PR #68, opened 2016-11-03
Remove the Subscription type from the Public API model

raveeram:
- It is used internally by the AzureContainerService and not by the acs-engine itself


anhowe:
@raveeram can you please close or bring up to date so no conflicts?

anhowe:
Closing pull request since it is obsolete.  Please re-open when you update this again.

PR #72, opened 2016-11-03
Video walkthrough of creating an ARM template using the acs-engine and deploying a k8s cluster on Azure

sauryadas:


rjtsdl:
@sauryadas need permit to checkout as well :)

anhowe:
@sauryadas please update or close.

PR #87, opened 2016-11-10
Updated docs with native installation steps for Linux and OS X

chzbrgr71:
Updated the base README to offer both Docker and native (but Docker is recommended). Instructions updated with steps for Linux and fixed a couple items in the OS X steps. Tested with Ubuntu Server 16.04.

Did not change or test anything for the Windows portion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,25 +44,7 @@ same environment used in our Continuous Integration system).
 
 ## Development (Native)
 
-### Requirements
-- PowerShell (Windows)
-- `bash` + `make` (Linux)
-- `git`
-- `go` (with a properly configured GOPATH)
-
-### Building (Linux)
-
-```shell
-make build
-```
-
-### Building (Windows, PowerShell)
-
-```shell
-cd ${env:GOPATH}/github.com/Azure/acs-engine
-go get .
-go build .
-```
+Follow the steps in the User Guide: [ACS Engine](docs/acsengine.md)
 
 
 ## Contributing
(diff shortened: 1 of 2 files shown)


colemickens:
Can you add a (slash move the existing) Docker section to the dev guide? The changes here look great other than that bit of info that is being lost. It's still nice to have documented for those that don't want to have to setup GOPATH or install tooling into their local dev box.

chzbrgr71:
By dev guide, you mean 'acsengine.md' right? 

Totally agree about the lost info. To me, it looked like there are 3 experiences here. One is running in Docker. Two is installing/running natively. And third is for the developer that wants to update acs-engine code itself in a developer environment. I imagine that third area could easily be added if it isn't clearly covered with the above.

colemickens:
You don't need to close it, you can force push another commit and we can get this merged.

I don't understand the third scenario there. Isn't that just "working with Git(Hub) 101"? I don't really want to document every single thing under the sun for Git, Go, etc. There's plenty of other resources out there; we can link to them if appropriate.

chzbrgr71:
Agreed. Update coming shortly.

README.md
```diff
@@ -26,45 +26,6 @@ The cluster definition file enables the following customizations to your Docker
 * [Managed Disks](examples/disks-managed) (under private preview) - shows how to use managed disks 
 * [Large Clusters](examples/largeclusters) - shows how to create cluster sizes of up to 1200 nodes
 
-## Development (Docker)
```
anhowe: Please ensure you update all documents that point to this section.  I would recommend instead leave "Development (Docker)" here and point to the new location.

anhowe: I thought this was linked in a few more places, but I think you found the only link in you change below so this is good.

colemickens:
Thanks for the contribution @chzbrgr71 !

PR #89, opened 2016-11-14
Update README.md

VincentLiu028:
untitled

There is no --name option for Azure XPlat CLI 0.10.0 just remove it.


colemickens:
@anhowe They're both correct. This PR is for `azure-xplat-cli`, the other PR is for `azure-cli`.

Issue #52, opened 2016-10-26
Think about vendoring


colemickens:
We now have external dependencies in this project.

We should strongly considering vendoring them into the repo.

This re-presents the problem of this repo being used as both a package and a standalone CLI.

Generally you want to vendor for binaries, but not vendor for packages. We may want to revisit splitting the CLI part out of this into a separate tool.

anhowe:
Let's keep learning best practice.

PR #91, opened 2016-11-14
Vendor dependencies

ahmetalpbalkan:


colemickens:
I have reservations about this since we are vendoring this package elsewhere. Common guidance seems to be "vendor deps in binaries, don't vendor deps in a library".

There almost needs to be another project for the CLI on top of the core acs-engine that can live a plain library.

Related: #52

ahmetalpbalkan:
You have a main() at the root, though. This is already a binary-producing repo. Does adding vendor.json at the root is going to impact you when you're vendoring subpackages?

colemickens:
Hm. Maybe not, probably it's fine. Which tool did you use for this?

ahmetalpbalkan:
https://github.com/kardianos/govendor

colemickens:
Before I go try to read up, do you know if govendor and gb-vendor both respect the "standard" vendor.json format?

ahmetalpbalkan:
@colemickens govendor claims to do so, however I'm not familiar with the format myself.

colemickens:
This is the problem I'm worried about: https://github.com/mattfarina/golang-broken-vendor (This would manifest in our RP, not in this project likely)

ahmetalpbalkan:
@colemickens that looks like the diamond dependency problem. Is the problem in this case the RP also uses the yaml library? Also it's unlikely that you'll end up assigning types from the yaml package to places, b/c it's just used for parsing things in 1-2 lines of code. So I'm assuming you won't hit this as much.

colemickens:
I think we should go ahead and do this. We can re-evaluate if this actually winds up being a problem.

But, I think we want to do the vendor with `gb-vendor` to be consistent with the RP. Or else evaluate why we chose gb upstream and then revisit this PR.

anhowe:
Thanks for raising this.  However, closing since we need to decide which vendoring strategy we should use.

PR #98, opened 2016-11-15
Fix name -> subscription in account set command

timfpark:
I believe the argument with the subscription should be --subscription and not --name.


anhowe:
Thanks Tim for this update.  I've just merged a similar pull request https://github.com/Azure/acs-engine/pull/90 from @jcorioland.

PR #101, opened 2016-11-16
set config mode as arm for xplatform cli before deploying acs

VincentLiu028:
default mode is asm after installation of xplatform cli. Need to run command to change mode to arm, or will get error run following command.


colemickens:
Thanks for the PR!

PR #106, opened 2016-11-17
add operation status definition

rjtsdl:
Define the operation status
The fields conform with ARM's microsoft internal api definition. I will not share anything more.


raveeram:
Agree with @colemickens . Probably see what @anhowe has to say.

amanohar:
As discussed in person this data structure belong in the service and not ACS Engine.

rjtsdl:
Talked offline, i will close this.

PR #110, opened 2016-11-18
Fix az cli usage docs

colemickens:


colemickens:
Equal signs were removed because of this bug: https://github.com/Azure/azure-cli/issues/1379 (and then on the other flags just to be consistent.

README.md
```diff
@@ -90,14 +90,14 @@ $ az login
 $ az account set --name "<SUBSCRIPTION NAME OR ID>"
 
 $ az group create \
```
kmala: this should be `az resource group create` instead of `az group create`

kmala: I thought of making a PR but since you already have a PR for the docs change i felt this change also can go along with other changes.

colemickens: Thanks, happy to roll it into this one. Good catch.

PR #121, opened 2016-11-24
Fixes #116. Naming container when starting

### LINKED ISSUE #116: [kubernetes] systemctl stop kubelet.service
Unable to stop kubelet.service via systemctl on k8s master node.

```
> systemctl stop kubelet.service
> systemctl status kubelet.service
Nov 22 01:29:19 k8s-master-1479775666-0 systemd[1]: Stopping Kubelet...
Nov 22 01:29:19 k8s-master-1479775666-0 docker[26894]: Error response from daemon: No such container: kubelet

```

Should be a fairly easy fix if you name the docker image when starting:
https://github.com/Azure/acs-engine/blob/master/parts/kubernetesmastercustomdata.yml#L323

jpoon:
Fixes https://github.com/Azure/acs-engine/issues/116 by naming container to `kubelet` on start such that it can be stopped through `systemctl`

### DIFF
--- a/parts/kubernetesmastercustomdata.yml
+++ b/parts/kubernetesmastercustomdata.yml
@@ -294,6 +294,7 @@ write_files:
     ExecStartPre=/bin/mount --bind /var/lib/kubelet /var/lib/kubelet
     ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
     ExecStart=/usr/bin/docker run \
+      --name=kubelet
       --net=host \
       --pid=host \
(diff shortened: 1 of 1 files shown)


colemickens:
Was this tested? Because it broke Kubernetes.

You're missing a `\` at the end of the line...

colemickens:
Not only that, it should have been done for the agents as well...

jpoon:
@colemickens fuuu. my bad.

PR #134, opened 2016-12-07
pki: remove wildcard from subject alt names

gabrtv:
This PR is designed to remove the wildcard FQDN from the end of the subjectAltNames in the APIServer certificate.  I was tempted to change some of the calling code to make the intent more clear, but opted for the 1-line change instead.

This PR resolves the following `pyopenssl` error:

```console
WARNING:requests.packages.urllib3.contrib.pyopenssl:A problem was encountered with the certificate that prevented urllib3 from finding the SubjectAlternativeName field. This can affect certificate validation. The error was Codepoint U+002A at position 1 of '*' not allowed
/usr/local/lib/python3.5/dist-packages/requests/packages/urllib3/connection.py:337: SubjectAltNameWarning: Certificate for 10.0.0.1 has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)
  SubjectAltNameWarning
ERROR:requests.packages.urllib3.connection:Certificate did not match expected hostname: 10.0.0.1. Certificate: {'subject': ((('commonName', 'apiserver'),),), 'subjectAltName': []}
ERROR There was a problem retrieving data from the Kubernetes API server. URL: https://10.0.0.1:443/api/v1/namespaces/test-288243432, params: {}
ERROR:scheduler:There was a problem retrieving data from the Kubernetes API server. URL: https://10.0.0.1:443/api/v1/namespaces/test-288243432, params: {}
```

### DIFF
--- a/pkg/acsengine/pki.go
+++ b/pkg/acsengine/pki.go
@@ -24,7 +24,7 @@ type PkiKeyCertPair struct {
 	PrivateKeyPem  string
 }
 
-func CreatePki(masterFQDN string, extraFQDNs []string, extraIPs []net.IP, clusterDomain string) (*PkiKeyCertPair, *PkiKeyCertPair, *PkiKeyCertPair, *PkiKeyCertPair, error) {
+func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string) (*PkiKeyCertPair, *PkiKeyCertPair, *PkiKeyCertPair, *PkiKeyCertPair, error) {
 	start := time.Now()
 	defer func(s time.Time) {
 		fmt.Fprintf(os.Stderr, "cert creation took %s\n", time.Since(s))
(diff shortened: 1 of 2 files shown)


pkg/acsengine/pki.go
```diff
@@ -110,7 +110,6 @@ func createCertificate(commonName string, caCertificate *x509.Certificate, caPri
 		template.KeyUsage |= x509.KeyUsageCertSign
 		template.IsCA = isCA
 	} else if isServer {
-		extraFQDNs = append(extraFQDNs, FQDN)
```
anhowe: Implement either of two recommendations:
1. keep your change, but then remove FQDN through full call stack, or
2. undo your change, and in setDefaultCerts remove the "*", and send in element 0 as the FQDN and [1:] as the FQDNs

gabrtv:
@anhowe I went with option 1 since FQDN is not used anymore. PTAL. Thanks!

PR #135, opened 2016-12-08
Adding capability to have CRP deploy keyvault certs during deployment

JackQuincy:
Updating types
Adding unit tests and making it work in the templates
Adding validation


colemickens:
@anhowe should review it too, but LGTM.

parts/dcosagentresourcesvmas.t
```diff
@@ -211,6 +211,31 @@
                 ]
               }
             }
+            {{if .HasSecrets}}
+            ,
+            "secrets":[
+            {{range  $vIndex, $vault := .Secrets}}
+              {{if $vIndex}}
```
colemickens: super nit, but these comma constructs could be a single line?

pkg/api/converterfromapi.go
```diff
@@ -191,6 +208,12 @@ func convertAgentPoolProfileToVLabs(api *AgentPoolProfile, p *vlabs.AgentPoolPro
 	p.VnetSubnetID = api.VnetSubnetID
 	p.SetSubnet(api.Subnet)
 	p.FQDN = api.FQDN
+	p.Secrets = []vlabs.KeyVaultSecrets{}
```
anhowe: we only want as part of VLabs, and probably not 20160330, since it is sealed.

pkg/api/convertertoapi.go
```diff
@@ -191,6 +197,23 @@ func convertVLabsAgentPoolProfile(vlabs *vlabs.AgentPoolProfile, api *AgentPoolP
 	api.VnetSubnetID = vlabs.VnetSubnetID
 	api.Subnet = vlabs.GetSubnet()
 	api.FQDN = vlabs.FQDN
+	api.Secrets = []KeyVaultSecrets{}
+	for _, s := range vlabs.Secrets {
```
anhowe: only VLAbs

pkg/api/converterfromapi.go
JackQuincy: I'm confused, isn't this only in vlabs in my CR.  That is what I meant to do. I didn't update the 20160330 types.

JackQuincy: the diff is doing weird things. this is only changing in the vlabs. the 20160330 is trailing lines on the previous change

pkg/api/types.go
```diff
@@ -134,12 +134,13 @@ type KubernetesConfig struct {
...
-	VnetSubnetID             string `json:"vnetSubnetID,omitempty"`
-	FirstConsecutiveStaticIP string `json:"firstConsecutiveStaticIP,omitempty"`
-	Subnet                   string `json:"subnet"`
+	Count                    int               `json:"count"`
+	DNSPrefix                string            `json:"dnsPrefix"`
+	VMSize                   string            `json:"vmSize"`
+	VnetSubnetID             string            `json:"vnetSubnetID,omitempty"`
+	FirstConsecutiveStaticIP string            `json:"firstConsecutiveStaticIP,omitempty"`
+	Subnet                   string            `json:"subnet"`
+	Secrets                  []KeyVaultSecrets `json:"secrets,omitempty"`
```
JackQuincy: should I be adding these on the Linux/windows profiles instead? That way I could check more easily for if we should have a certificate store specified.  It would also mirror how we are handling ssh keys and admin passwords. We would lose a little control on where we are storing certs. thoughts?

examples/keyvaultcerts/README.md
```diff
@@ -0,0 +1,17 @@
...
+ACS-Engine enables you to create customized Docker enabled cluster on Microsoft Azure with certs installed from key vault during deployment.
+
+The examples show you how to configure installing a cert from keyvault. These certs are assumed to be in the secrets portion of your keyvault:
+
+1. **dcos.json** - deploying and using [DC/OS](../../docs/dcos.md)
+2. **kubernetes.json** - deploying and using [Kubernetes](../../docs/kubernetes.md)
+3. **swarm.json** - deploying and using [Swarm](../../docs/swarm.md)
+4. **swarm-windows.json** - deploying and using [Swarm](../../docs/swarm.md)
+
+On windows machines certificates will be installed under the machine in the specified store
```
anhowe: period at end

pkg/api/vlabs/validate.go
```diff
@@ -95,6 +96,26 @@ func (a *AgentPoolProfile) Validate() error {
 	return nil
 }
 
+func validateKeyVaultSecrets(secrets []KeyVaultSecrets, requireCertificateStore bool) error {
+	for _, s := range secrets {
+		if len(s.VaultCertificates) == 0 {
+			return fmt.Errorf("Invalid KeyVaultSecrets must have none empty VaultCertificates")
```
anhowe: "have no"

JackQuincy:
Thanks for reviews

PR #142, opened 2016-12-12
Add "environment" flag in cluster definition to support AzureChinaCloud

wangtt03:
Due to Chinese GreatFire Wall, the default DC/OS packages cannot be downloaded from Azure CDN. This commit add Chinese local mirror for these packages and user can specify an "environment" variable in the cluster definition file to target the deployment environment to AzurePublicCloud/AzureChinaCloud.


wangtt03:
update kubernetes configuration to support AzureChinaCloud

PR #151, opened 2016-12-20
Feature optional calico policy daemon

simonswine:
This Pull Request adds a `networkPolicy` parameter with default value `none`. It won't change the cluster setup otherwise.

If you want to enable Calcio's policy enforcement using the hosted etcd-less install. Then you can set `networkPolicy=calico` and it will run kublets in CNI mode and run calico on the nodes.

This feature required some changes to the templates:
- [x] Use general templates for kubelet.service for both agents and masters
- [x] Use gzip compression for kubelet.service to fit everything into the custom-data
- [x] Adds role label to all nodes
- [x] Have separate files for kubernetes definitiions for calcio's daemonset and config map (including a label to roll it out using addons-manager
- [x] Upgrade kube-addons-manager to support config maps
- [x] Fix go test output. How to do that easily? I think the tests are hardly finding a real problem the way they are right now
- [x] Create a PR against calico to have the separate files in their main repo

These todos are left:
- [x] Use calicos repository for downloading manifests as soon as the PR merges
@mattbates @lxpollitt @gunjan5

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/151)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

lxpollitt:
cc @brendandburns

gunjan5:
Hi, just wanted to follow up to see what is the process to get this PR reviewed/discussed @rgardler @brendandburns 
Thanks!

simonswine:
@rgardler @brendandburns @gunjan5 @lxpollitt just rebased the changes for k8s v1.5.1

anhowe:
@simonswine - Hi Simon, thank you for this changes.  We are prioritizing the CR as a P0.

parts/kubernetesmastercustomscript.sh
```diff
@@ -92,6 +93,43 @@ function ensureKubectl() {
...
+function configNetworkPolicy() {
+    if [[ ! -z "${APISERVER_PRIVATE_KEY}" ]]; then
+        # on masters
+        ADDONS="calico-configmap.yaml calico-daemonset.yaml"
+        ADDONS_PATH=/etc/kubernetes/addons
+        CALICO_URL="https://github.com/simonswine/calico/raw/master/v2.0/getting-started/kubernetes/installation/hosted/k8s-backend"
+        if [[ "${NETWORK_POLICY}" = "calico" ]]; then
+            # download calico yamls
+            for addon in ${ADDONS}; do
+                curl -sSL --retry 12 --retry-delay 10 "${CALICO_URL}/${addon}" > "${ADDONS_PATH}/${addon}"
```
anhowe: How are you going to fail if this never downloads after 12 retries?

parts/kubernetesparams.t
```diff
@@ -53,7 +53,14 @@
         "description": "The kubernetes version."
       }, 
       "type": "string"
-    }, 
+    },
+    "networkPolicy": {
+      "defaultValue": "{{.OrchestratorProfile.KubernetesConfig.NetworkPolicy}}",
+      "metadata": {
+        "description": "The network policy enforcement to use (none|calico)"
+      },
```
anhowe: add:
"allowedValues": [
  "none",
  "calico"
],

pkg/acsengine/engine.go
```diff
@@ -379,9 +386,11 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 			if e != nil {
 				return ""
 			}
-			// add the master provisioning script
-			masterProvisionB64GzipStr := getBase64CustomScript(kubernetesMasterCustomScript)
-			str = strings.Replace(str, "MASTER_PROVISION_B64_GZIP_STR", masterProvisionB64GzipStr, -1)
+			// add artifcats
```
raveeram: spelling.

ahrkrak:
@simonswine please take a look at the feedback and update/resubmit w/ fixes?

parts/kubernetesmastercustomscript.sh
```diff
@@ -92,6 +93,43 @@ function ensureKubectl() {
...
+function setDockerOpts () {
+    sed -i "s#^DOCKER_OPTS=.*#DOCKER_OPTS=${1}#" /etc/default/kubelet
+}
+
+function configNetworkPolicy() {
+    if [[ ! -z "${APISERVER_PRIVATE_KEY}" ]]; then
+        # on masters
+        ADDONS="calico-configmap.yaml calico-daemonset.yaml"
+        ADDONS_PATH=/etc/kubernetes/addons
+        CALICO_URL="https://github.com/simonswine/calico/raw/master/v2.0/getting-started/kubernetes/installation/hosted/k8s-backend"
```
gunjan5: Seems like the manifests at this URL are using Calico v2.0.0, we have v2.0.2 out now and it's recommended to use that. See: http://docs.projectcalico.org/v2.0/releases/

mattbates:
Thanks for the review @anhowe.

@simonswine will be returning to this next week.

parts/kubernetesmastercustomscript.sh
simonswine: I guess I am not failing in that case. I have just copied that from: https://github.com/Azure/acs-engine/blob/bc344773b5da0d2afe775e6ef1a1e9d21a217726/parts/kubernetesmastercustomdata.yml#L20

parts/kubernetesparams.t
simonswine: Done

simonswine:
Hi @anhowe, thanks for your review!

I updated this PR according to the feedback. I added basic docs how to use the feature. (Linking mainly to Calico's simple policy example). 

The go tests are now passing (I am not too sure if that is the right thing to do, but I just made sure that the kubernetes expected JSON files contain the same like the ones already generated). I basically did this:

```bash
find pkg/acsengine/testdata -name "*.json.err" -exec echo mv {} {}XX \; | sed 's/\.errXX//g' 
```

I think we should wait for this PR (https://github.com/projectcalico/calico/pull/413) to merge and then I can update the URL for the manifests to calico's repository. (You can review it meanwhile)

caseydavenport:
>  Use calicos repository for downloading manifests as soon as the PR merges

@simonswine I'm wondering if this is the best option actually.  In most other integrations we've opted to host the manifests with the integration itself so that upstream changes can be pulled over in a more deliberate way.  Means fewer surprises, though at the cost of automatically pulling in updates.  WDYT?

simonswine:
@caseydavenport You are totally right, that it's a good idea to fix the version. I have done that by using a commit hash.

From my point of view this is ready to merge. Please review @amanohar @colemickens @raveeram @anhowe @gunjan5

lxpollitt:
Regarding question 1, for a mixed Linux / Windows cluster the network policy should just apply to the Linux boxes.  It was a deliberate spec choice for the Kubernetes policy API that it only gets enforced to the extent the underlying implementation can do so, and is not rejected if it can't be enforced.

simonswine:
@anhowe: The latest push should address both 1+2: I have added the validations. Please check if that is what you would expect from me, as I am not too sure how to test that. I have added some `go test` for my validation, just too make sure it works as expected.

I also make sure that you can only set calico for clusters that don't contain Windows nodes. I think it's the safe default, as it wouldn't leave users in the impression that the network policy is enforced, if it's actually not.

simonswine:
Another rebase, would be good to get a review rather sooner than later..

@amanohar @colemickens @raveeram @anhowe @gunjan5

docs/kubernetes.md
```diff
@@ -32,6 +32,34 @@ In the image above, you can see the following parts:
...
+
+```
+kubectl get pods --namespace kube-system -l k8s-app=calico-node -o wide
+NAME                READY     STATUS    RESTARTS   AGE       IP             NODE
+calico-node-034zh   2/2       Running   0          2h        10.240.255.5   k8s-master-30179930-0
+calico-node-qmr7n   2/2       Running   0          2h        10.240.0.4     k8s-agentpool1-30179930-1
+calico-node-z3p02   2/2       Running   0          2h        10.240.0.5     k8s-agentpool1-30179930-0
+```
+
+Per default Calico still allows all communication within the cluster. To enable a stricter policy follow this guide: [Simple Policy Demo](http://docs.projectcalico.org/v2.0/getting-started/kubernetes/tutorials/simple-policy)
```
lxpollitt: I think this should reference the official k8s docs as the primary reference: https://kubernetes.io/docs/user-guide/networkpolicies/
Probably also have the link to the Calico doc as a secondary reference example walkthrough.

docs/kubernetes.md
```diff
@@ -32,6 +32,34 @@ In the image above, you can see the following parts:
 
 All VMs are in the same private VNET and are fully accessible to each other.
 
+
+## Optional: Enable network policy enforcement using Calico
```
caseydavenport: I see an issue with the location of this section.  At this point in the walkthrough, the Kubernetes cluster is already deployed (i.e the user has followed the Deployment section above), which means if they get to this point and decide they want Calico they need to tear down the cluster they've deployed and create a new one.

Could we include this as an optional step in the `## Deployment` section above to make it more clear to users that this is an option?

lxpollitt: I think it should be split in two.  Most of this stays here as the "how to use it" part of the walk through.  But the `azuredeploy.parameters.json` oriented bit moves up to be just another numbered step in the deployment section near the top of the doc.

simonswine: I agree that it's not optimal that you are not aware of the parameters at creation time of a cluster. I guess that's a general problem for all possible parameters. I added a note and cross reference to the calico section. Let me know, if you think that's appropriate

simonswine:
Another rebase. Would be happy to get a review. (This time rebasing was quite complicated)
@amanohar @colemickens @raveeram @anhowe @gunjan5 @lxpollitt

lxpollitt:
@brendandburns we have rebased this a few times more.  If we rebased it again can we get a commitment to merge before it all changes again?

parts/kubernetesmastercustomdata.yml
```diff
@@ -298,68 +296,28 @@ write_files:
...
-    Requires=docker.service
-    After=docker.service
-
-    [Service]
-    Restart=always
-    ExecStartPre=/bin/mkdir -p /var/lib/kubelet
+    KUBELET_CLUSTER_DNS={{{kubeDnsServiceIP}}}
+    KUBELET_API_SERVERS=<<<concat('https://', variables('masterPrivateIpAddrs')[copyIndex()], ':443')>>>
+    KUBELET_IMAGE={{{kubernetesHyperkubeSpec}}}
+    KUBELET_NETWORK_PLUGIN=
```
colemickens: What implications does this have for pods scheduled on master? Will they be able to start if they are not on the host network?

docs/kubernetes.md
colemickens: Noted, the docs need some attention, I'll keep this in mind. The cross reference is a good solution.

parts/kubernetesmastercustomscript.sh
```diff
@@ -92,6 +93,43 @@ function ensureKubectl() {
...
+            done
+        else
+            # download calico yamls
+            for addon in ${ADDONS}; do
+                rm -f "${ADDONS_PATH}/${addon}"
+            done
+        fi
+    else
+        # on agents
+        if [[ "${NETWORK_POLICY}" = "calico" ]]; then
```
colemickens: This may be confusion on my part due to unfamiliarity with Calico, but is `NETWORK_POLICY` the same as `CNI_PLUGIN` in a sense?

For example, if I knew of another branch somewhere that is adding another CNI plugin, would they simply be able to add their plugin as a new option for `NETWORK_POLICY`, here and in the apimodel? It seems like it, but I wanted to check.

pkg/api/vlabs/types.go
```diff
@@ -118,6 +118,15 @@ const (
...
 type OrchestratorProfile struct {
 	OrchestratorType OrchestratorType `json:"orchestratorType"`
+	KubernetesConfig KubernetesConfig `json:"kubernetesConfig,omitempty"`
+}
+
+// KubernetesConfig contains the Kubernetes config structure, containing
+// Kubernetes specific configuration
+type KubernetesConfig struct {
+	KubernetesHyperkubeSpec string `json:"kubernetesHyperkubeSpec,omitempty"`
+	KubectlVersion          string `json:"kubectlVersion,omitempty"`
```
colemickens: I think this should be gone when you rebase.

pkg/api/vlabs/validate.go
```diff
@@ -247,6 +250,24 @@ func (a *Properties) Validate() error {
...
+		return nil
+	}
+
+	networkPolicy := a.OrchestratorProfile.KubernetesConfig.NetworkPolicy
+	if networkPolicy != "" && networkPolicy != "none" && networkPolicy != "calico" {
+		return fmt.Errorf("unknown networkPolicy '%s' specified", networkPolicy)
+	}
+
+	if networkPolicy == "calico" && a.HasWindows() {
+		return fmt.Errorf("unknown networkPolicy '%s' specified", networkPolicy)
```
colemickens: Can you make that error message more specific to the error condition.

colemickens:
I think we can get this merged in the next day or two, mostly pending a rebase. I will hold off on merging any of my PRs that will cause rebases here.

I just have a few small comments/questions that are very minor.

parts/kubernetesmastercustomscript.sh
```diff
@@ -92,6 +93,43 @@ function ensureKubectl() {
...
+}
+
+function setDockerOpts () {
+    sed -i "s#^DOCKER_OPTS=.*#DOCKER_OPTS=${1}#" /etc/default/kubelet
+}
+
+function configNetworkPolicy() {
+    if [[ ! -z "${APISERVER_PRIVATE_KEY}" ]]; then
+        # on masters
+        ADDONS="calico-configmap.yaml calico-daemonset.yaml"
```
colemickens: move these variables into the `if [[ "${NETWORK_POLICY}" = "calico" ]]; then` block, as they're only relevant there

parts/kubernetesmastercustomscript.sh
```diff
@@ -92,6 +93,43 @@ function ensureKubectl() {
...
+        CALICO_URL="https://raw.githubusercontent.com/projectcalico/calico/a4ebfbad55ab1b7f10fdf3b39585471f8012e898/v2.0/getting-started/kubernetes/installation/hosted/k8s-backend-addon-manager"
+        if [[ "${NETWORK_POLICY}" = "calico" ]]; then
+            # download calico yamls
+            for addon in ${ADDONS}; do
+                curl -sSL --retry 12 --retry-delay 10 "${CALICO_URL}/${addon}" > "${ADDONS_PATH}/${addon}"
+            done
+        else
+            # download calico yamls
+            for addon in ${ADDONS}; do
+                rm -f "${ADDONS_PATH}/${addon}"
```
colemickens: Is this logic needed? Meant for idempotency? I guess it shouldn't hurt but can the comment be updated?

pkg/acsengine/engine.go
```diff
@@ -72,6 +72,11 @@ const (
 	windowsParams                = "windowsparams.t"
 )
 
+var kubernetesAritfacts = map[string]string{
+	"MASTER_PROVISION_B64_GZIP_STR": kubernetesMasterCustomScript,
+	"KUBELET_SERVICE_B64_GZIP_STR":  "kuberneteskubelet.service",
```
colemickens: Can you make it a constant string for consistency

pkg/acsengine/engine.go
```diff
@@ -481,6 +487,11 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 				return ""
 			}
 
+			// add artifacts and addons
```
colemickens: I don't think this is supposed to be here? Duplicate of what's below and `str` isn't used...

simonswine:
@colemickens thanks for your review. Basically spent the whole day rebasing. But it looks like the current master is not that stable. Various dependency/timing issues with kubectl-extract. Not too sure but as soon as I solve these issues, it shouldn't take too long

brendandburns:

parts/kubernetesmastercustomdata.yml
```diff
@@ -168,7 +166,7 @@ write_files:
               - "--cloud-config=/etc/kubernetes/azure.json"
               - "--root-ca-file=/etc/kubernetes/certs/ca.crt"
               - "--service-account-private-key-file=/etc/kubernetes/certs/apiserver.key"
-              - "--leader-elect=true"
+              - "--leader-elect=true"
```
simonswine: Done

parts/kubernetesmastercustomdata.yml
simonswine: They won't be able to be started, before the CNI is initialized. This applies only to pods without hostNetwork: true and networkPolicy=calico

parts/kubernetesmastercustomscript.sh
simonswine: If I want to keep the delete then I need it there. (I could move the URL though)

parts/kubernetesmastercustomscript.sh
simonswine: I wanted to be able to remove on a running node with calling the configureCalcio function. Updated the comment

parts/kubernetesmastercustomscript.sh
simonswine: Yes that should be possible. We are not using calcio for overlay though, we are just enforcing the policy, so I thought this would be the better name

pkg/acsengine/engine.go
simonswine: Done

pkg/acsengine/engine.go
simonswine: No it's needed here, but I thought you were right at first till expanded the code in Github. It's once used for the Master and once for the Agent.

simonswine:
@colemickens Ok looks good from my point of view. Sometimes the kubectl-extract still fails

parts/kubernetesmastercustomscript.sh
colemickens: Hm, that opens up another set of questions that all stem from me not knowing Calico well enough or understand the distinction between using it as a CNI plugin but not using it for overlay.

So does this still rely on Azure networking and route tables being set up? I may need to have a quick Hangouts (or Zoom or w/e) call to understand the implications of this fully.

parts/kubernetesmastercustomdata.yml
colemickens: "before CNI is initialized". But how is CNI initialized if the NetworkingPlugin is set to nothing?

parts/kubernetesmastercustomscript.sh
colemickens: okay, it's fine as is then.

pkg/acsengine/engine.go
colemickens: Yup, I see it now too. Thanks.

parts/kubernetesmastercustomscript.sh
lxpollitt: Yes, this PR is inserting Calico for policy enforcement only.  It doesn't change the existing Azure networking or host-to-host route tables being set up.  CNI is the mechanism Calico (and a number of other network policy plugins) use to achieve this.

colemickens: @lxpollitt thank you, I assumed as much but I don't understand how these things compose. My understanding is that the existing CNI plugin being used, `kubenet`, is essential for making networking + cloudprovider work. But I guess somehow removing `kubenet` and replacing with `calico` just works still somehow? Maybe calico knows how to do the bridge reconfiguration that `kubenet` does?

There's a further extension to this question, but I don't really know if I can talk about it here yet. Are you or @simonswine available for a call anytime today? Can even just be voice? I just want to understand a tiny bit better before signing off.

colemickens: If so, shoot me mail colemick@microsoft.com. Thanks. Excited to get this in.

lxpollitt: Sorry I didn't manage to make time for a call today (and have meetings for the rest of the day/evening).  But hopefully I can help explain a little more.

When you use Calico as a CNI plugin for policy, it effectively replaces`kubenet`.  Calico will do the local networking (i.e. the networking within the host) instead of `kubenet` doing it.  The host-to-host networking remains untouched - so interacts with Azure SDN in exactly the same way as if you are using `kubenet`.  

This approach is fairly mainline and well tested in general on other public clouds.  In particular it is what most people who are using Calico with Kubernetes on GCE are doing.

@simonswine has been testing this throughout with Azure to confirm it is working as expected in this specific case too.

simonswine: Sorry @colemickens I am UK based, so it was a bit too late for me

Basically the way this works even though the `kubelet`s don't have `kubenet` set, is that the routes for the individual node's Pod IP range are setup in Azure (and GCE as @lxpollitt points out) by the `controller-manager` ([configured here](https://github.com/Azure/acs-engine/blob/master/parts/kubernetesmaster-kube-controller-manager.yaml#L19)) as long as cloud-provider is set. No special config on the `kubelet` is needed (IMHO).  

This will continue to work and allocate from the cluster wide Pod network. Calico then just takes the IP addresses out every nodes LocalPodRange. (this [setting](https://github.com/projectcalico/calico/blob/a4ebfbad55ab1b7f10fdf3b39585471f8012e898/v2.0/getting-started/kubernetes/installation/hosted/k8s-backend-addon-manager/calico-configmap.yaml#L20))

If you want to have a chat on Monday, we can set something up in your morning.

pkg/api/vlabs/types.go
```diff
@@ -125,6 +125,7 @@ type OrchestratorProfile struct {
 // Kubernetes specific configuration
 type KubernetesConfig struct {
 	KubernetesImageBase string `json:"kubernetesImageBase,omitempty"`
+	NetworkPolicy       string `json:"networkPolicy,omitempty"`
```
amanohar: Based on discussion in this PR looks like Calico is being used only for in host networking for now. But I think it could be expanded later to do host to host networking etc. Also, there might be a need for users to choose another plugin. 

With that in mind its best if the API is extensible for future network policy configurations. I recommend the  following approach:

```
type KubernetesConfig struct {
	KubernetesImageBase string 'json:"kubernetesImageBase,omitempty"'
	NetworkConfig NetworkConfig 'json:"networkConfig,omitempty"'
}

type NetworkConfig struct {
	NetworkPolicy       string 'json:"networkPolicy,omitempty"'
}
```

docs/kubernetes.md
```diff
@@ -12,7 +12,10 @@ Here are the steps to deploy a simple Kubernetes cluster:
 4. edit the [Kubernetes example](../examples/kubernetes.json) and fill in the blank strings
 5. [generate the template](acsengine.md#generating-a-template)
 6. [deploy the output azuredeploy.json and azuredeploy.parameters.json](../README.md#deployment-usage)
-7. Temporary workaround when deploying a cluster in a custom VNET with Kubernetes 1.5.3:
+  * To enable the optional network policy enforcement using calico, you have to
```
simonswine: The latest push contains the section in `clusterdefinition.md`: https://github.com/Azure/acs-engine/pull/151/files#diff-86820358fc5925785509e38c21c747f1R27

pkg/api/vlabs/types.go
simonswine: Calico is not really used for any networking. When `networkPolicy=calico` it is still using the existing `kubenet` Pod network. The only thing that calico does is managing iptables rules to prevent unspecified network communication.

`NetworkConfig` IMHO should not be limited to `KubernetesConfig`, as other orchestrators potentially have similar integrations with the same overlay networks. I would suggest to change this as soon as is becomes necessary.

colemickens:
Thanks @simonswine 

@amanohar I'm okay with changing this as needed whenever we get there. Can you sign off so we can merge before there's any more conflicts?

colemickens:
@acs-bot test this

colemickens:
@acs-bot test this please

colemickens:
This is going to break CI. I suspect the templates were either not updated, or were not updated from the development docker container. Thus they likely used an older version of Go. And without the basic test passing, we can't get a signal on if the rest of the change is safe... @simonswine Can you please update the templates using the docker development environment as documented?

Otherwise I can pull your branch, update it accordingly and run tests against it...

colemickens:
This is still blocked by rebase and fixing the testdata from the development docker container. At the very least it needs to be rebased. I'm still holding off on larger Kubernetes PRs in order to get this in, but I can't wait any much longer as we have other work that needs to be done. Please let me know.

The second this passes our CI and is rebased, I will merge it.

pkg/api/vlabs/types.go
amanohar: NetworkConfig will not be limited to KubernetesConfig. That's the point of adding a type (so it can be reused by other orchestrator's config when the need arises).

>>Calico is not really used for any networking.
But its still used for managing iptables so it is a network config. It makes sense for all network policies and other network settings to be inside network config so that the object structure aka API model is setup correctly from the get go and won't require revisiting.

colemickens:
Alternatively, we can finish this PR if you're okay with that.

simonswine:
Ok I have rebased this again. I had to run `gofmt -s -w` for `make ci` to pass. Sorry about my late responses, a lot things to do before Kubecon. :laughing: 

I can see your point @amanohar and if I would have a bit more capacity I would change it right away

simonswine:
I guess I can't do that:
@acs-bot test this please

simonswine:
@acs-bot test this please

colemickens:
Thank you for this @simonswine !

PR #152, opened 2016-12-21
Make acs engine deployable in AzureChinaCloud

wangtt03:
Due to Chinese firewall issue, google and azure CDN is blocked which will cause the acs engine deployment failed. In this PR, I extract some hard coded urls and configurations in parameters so that the users can specify them using Chinese local mirror.
- Common changes:
  - Add "environment" flag in cluster definition with default value "AzurePublicCloud". 
  - Update testdata to avoid test break.
- DC/OS changes:
   - Change dcosprovision.sh by adding Chinese local urls.
- Kubernetes changes:
  - Extract hard code docker download url, kubectl, kubernetes key images urls.
  - Put the default urls into const.go
  - Move and gzip hard coded configurations (such as kube-apiserver.yaml...) in kubernetesmastercustomdata.yml to seperate yaml files.
  - Add "cloud" parameter in azure.json so that when kubernetes start, it can point to the right endpoint.
  -  Change the hard code AzureProdFQDN by adding Chinese custom endpoint "cloudapp.chinacloudapi.cn"
  -  Change the provision.sh writekubconfig function so when user specify the "AzureChinaCloud" the correct FQDN suffix "cloudapp.chinacloudapi.cn" can be written.


anhowe:
@wangtt03 - thank you for these changes, we are prioritizing this for CR.

pkg/acsengine/testdata/disks-managed/swarm-vmas_expected.json
```diff
@@ -507,6 +507,13 @@
         "description": "SSH public key used for auth to all Linux machines.  Not Required.  If not set, you must provide a password key."
       },
       "type": "string"
+    },
+    "targetEnvironment": {
+      "defaultValue": "AzurePublicCloud",
+      "metadata": {
+        "description": "The azure deploy environment. eg. AzurePublicCloud, AzureChinaCloud, AzureGermanCloud, AzureUSGovermentCloud"
```
weinong: for the sake of consistency with azure powershell, these environments are
AzureCloud, AzureChinaCloud, AzureUSGovernment, and AzureGermanCloud.

parts/masterparams.t
```diff
@@ -52,6 +52,13 @@
         "description": "A string hash of the master DNS name to uniquely identify the cluster."
       },
       "type": "string"
+    },
+    "targetEnvironment": {
+      "defaultValue": "AzurePublicCloud",
```
weinong: same as my other comment, these should be AzureCloud, AzureChinaCloud, AzureUSGovernment, and AzureGermanCloud.

pkg/acsengine/testdata/disks-managed/swarm-vmss_expected.json
```diff
@@ -507,6 +507,13 @@
         "description": "SSH public key used for auth to all Linux machines.  Not Required.  If not set, you must provide a password key."
       },
       "type": "string"
+    },
+    "targetEnvironment": {
```
weinong: same as my other comment: these should be AzureCloud, AzureChinaCloud, AzureUSGovernment, and AzureGermanCloud

pkg/api/vlabs/types.go
```diff
@@ -38,6 +38,7 @@ type Properties struct {
 	MasterProfile           MasterProfile           `json:"masterProfile"`
 	AgentPoolProfiles       []AgentPoolProfile      `json:"agentPoolProfiles"`
 	LinuxProfile            LinuxProfile            `json:"linuxProfile"`
+	Environment				Environment				`json:"environment"`
```
amanohar: Environment can be derived by 'Location' field in ContainerService struct. I don't think a separate field 'Environment' is needed in properties for this.

pkg/acsengine/testdata/disks-managed/swarm-vmss_expected.json
wangtt03: Thanks for reviewing it. I am on vacation for Chinese new year and I will send the update once I get back to the office.

pkg/api/vlabs/types.go
wangtt03: Thanks for reviewing it. I am on vacation for Chinese new year and I will send the update once I get back to the office.

anhowe: @wangtt03 - are you able to update this week?

wangtt03: @anhowe yes.

wangtt03:
@amanohar Thanks for reviewing. The changes made in this commit:
* Remove 'Environment' flag from the cluster definition json file by reusing the 'Location' inherit from ContainerService (The value of the Location field should be one of the 'AzureCloud/AzureChinaCloud/AzureUSGovernment/AzureGermanCloud').
* Update defaults.go, change the SetOrchestratorDefaults function by adding location parameter, so the orchestrator properties can be set according to the location.
* Update DCOS related code by extracting hard coded urls.
* Update testcases to make go test pass.

pkg/acsengine/const.go
```diff
@@ -33,3 +30,24 @@ const (
 	// PublicAgent represents the public agent node type
 	DCOSPublicAgent DCOSNodeType = "DCOSPublicAgent"
 )
+
+const (
+	AzureCloud int = iota
+	AzureChinaCloud
+	AzureUSGovernment
```
amanohar: To be consistent can you update this to : "AzureUSGovernmentCloud"

pkg/api/v20160330/types.go
```diff
@@ -98,6 +98,31 @@ const (
...
+	KubernetesAddonResizerSpec      string `json:"kubernetesAddonResizerSpec,omitempty"`
+	KubernetesKubeDNSSpec           string `json:"kubernetesKubeDNSSpec,omitempty"`
+	KubernetesExecHealthzSpec       string `json:"kubernetesExecHealthzSpec,omitempty"`
+	KubernetesDashboardSpec         string `json:"kubernetesDashboardSpec,omitempty"`
+	KubernetesAddonManagerSpec      string `json:"kubernetesAddonManagerSpec,omitempty"`
+	KubernetesDNSMasqSpec           string `json:"kubernetesDNSMasqSpec,omitempty"`
+	KubernetesPodInfraContainerSpec string `json:"kubernetesPodInfraContainerSpec,omitempty"`
+	KubectlDownloadURL              string `json:"kubectlDownloadURL,omitempty"`
+	DockerInstallScriptURL          string `json:"dockerInstallScriptURL,omitempty"`
 }
```
amanohar: Is this something that really needs to be set in the API model? 

- Once ACS Engine knows the location/environment it can keep a list of download URLs for each location? Or is the expectation that call customer provides this info? If yes, why? How does this work in the K8s scenario today for Azure Public Cloud?

- If there is a need for such a provision providing flexibility to specify a different URL for each spec makes this complicated. Caller should provide a storage account/container/locations and files in that container should be expected to be named in a format that ACS Engine expects. This way we won't required 12 new fields.

amanohar:
cc: @colemickens

pkg/acsengine/const.go
weinong: @amanohar Azure uses AzureUSGovernment in powershell, cli, etc...

wangtt03: OK. I will use AzureUSGovernment.

pkg/api/v20160330/types.go
```diff
@@ -98,6 +98,31 @@ const (
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
 	OrchestratorType OrchestratorType `json:"orchestratorType"`
+	KubernetesConfig KubernetesConfig `json:"kubernetesConfig,omitempty"`
+	DCOSConfig       DCOSConfig       `json:"dcosConfig,omitempty"`
+}
+
+type DCOSConfig struct {
+	DCOS173_BootstrapDownloadURL string
```
wangtt03: OK. I will change it.

pkg/api/v20160330/types.go
wangtt03: I think I can remove these fields from API model.

wangtt03:
@amanohar Thanks for reviewing the code. This commit include changes:
* Remove redundant fields from API model.
* Merge.
* Update testcases.

pkg/acsengine/types.go
```diff
@@ -24,3 +24,36 @@ type V20160330ContainerService struct {
...
+}
+
+//DockerSpecConfig is default docker install script URL.
+type DockerSpecConfig struct {
+	DefaultDockerInstallScriptURL string
+}
+
+//KubernetesSpecConfig is the kubernetes container images used.
+type KubernetesSpecConfig struct {
+	DefaultKubernetesHyperkubeSpec         string
```
amanohar: Thanks for incorporating feedback comments. I still see KubernetesSpecConfig with 10 properties. It seems to me that for every new cloud these 10 fields will need to be added and someone can selectively override some of these fields. I am not sure if we want to give this much configurability to the caller. 

Also, this could become unmanageable/will need to versioned if a new spec config needs to be introduced.

In my earlier feedback comment what I meant was: can we expect the caller to give us the name of the base container like: "gcr.io/google_containers" or "mirror.azure.cn:5000/google_containers". And we also expect the name of spec to be a const.

Then for e.g. kubernetesDNSMasqSpec can be constructed by concatenating the base URL (provided by the caller): "mirror.azure.cn:5000/google_containers" and a constant string: "/kube-dnsmasq-amd64:1.3"  

Let me know if this makes sense or I missed something.

wangtt03: @amanohar I agree with you that 10 extra configurable fields are too much. Based on your comments, I remove these extra fields and use base url to concat the whole image url. Thanks!

pkg/acsengine/const.go
```diff
@@ -33,3 +30,21 @@ const (
 	// PublicAgent represents the public agent node type
 	DCOSPublicAgent DCOSNodeType = "DCOSPublicAgent"
 )
+
+const (
+	KubernetesHyperkubeImageName         = "hyperkube-amd64:v1.5.1"
+	KubernetesDashboardImageName         = "kubernetes-dashboard-amd64:v1.5.1"
+	KubernetesExechealthzImageName       = "exechealthz-amd64:1.2"
```
amanohar: This looks good for now but might change in the future where ACS Engine will need to load/install different versions of each component based on overall Kubernetes orchestrator version. Just fyi.

pkg/api/v20160330/types.go
```diff
@@ -98,6 +98,14 @@ const (
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
 	OrchestratorType OrchestratorType `json:"orchestratorType"`
+	KubernetesConfig KubernetesConfig `json:"kubernetesConfig,omitempty"`
+}
+
+// KubernetesConfig contains the Kubernetes config structure, containing
```
wangtt03: ooops, my bad. forget to delete it:(

pkg/acsengine/const.go
wangtt03: @amanohar OK! Thanks!

wangtt03:
@amanohar Thanks for reviewing! I updated the code:
* Remove redundant `KubernetesConfig ` from `v20160330/types.go`,`v20160930/types.go`, `v20170131/types.go`
* Update `KubernetesHyperkubeSpec` to `KubernetesImageBase` in `vlab/types.go`
* Update testcases.
* Merge from upstream.

wangtt03:
@amanohar Thanks for reviewing! I updated the code:
* Remove redundant `KubernetesConfig ` from `v20160330/types.go`,`v20160930/types.go`, `v20170131/types.go`
* Update `KubernetesHyperkubeSpec` to `KubernetesImageBase` in `vlab/types.go`
* Update testcases.
* Merge from upstream.

wangtt03:
@amanohar I made a new commit based on your comments. Could you please review it? Thanks!

wangtt03:
@amanohar @colemickens  Thanks for reviewing. I merge the code and also did some small changes to DCOS1.8.8 config file by extracting the hard code download url. Could you please have a look at it? Thanks!

amanohar:
@wangtt03, your changes LGTM. I will wait for @colemickens to take a look as well before signing off.

wangtt03:
@amanohar OK. Thanks!

amanohar:
Approved. Please rebase or merge and run tests before integrating. Thanks!

pkg/api/vlabs/types.go
```diff
@@ -118,6 +118,14 @@ const (
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
 	OrchestratorType OrchestratorType `json:"orchestratorType"`
+	KubernetesConfig KubernetesConfig `json:"kubernetesConfig,omitempty"`
+}
+
+// KubernetesConfig contains the Kubernetes config structure, containing
+// Kubernetes specific configuration
+type KubernetesConfig struct {
```
amanohar: In another discussion on ACS Talk someone had suggested adding an array of orchestrator options instead of typed property which would make this model flexible. E.g.:
“orchestratorOptions”: [
         “KubernetesImageBase”: "xyz",
         “KubectlVersion”: “1.5.2”
         ]

I think this is a good proposal. @wangtt03, @colemickens, @anhowe: what do you think?

colemickens: I'm not 100% tracking what each option looks like and that json is not valid, it would need to be an object of key/value pairs, not an array.

I like the wiring up of the internal model's KubernetesConfig to the one in vlabs to expose these options. I think this is exactly what Onur is doing for letting users enable CNI integration for the Kubernetes orchestrator.

I think it's okay to duplicate fields in various Kube/DCOS/SwarmConfig objects... otherwise we will end up with this giant list of options, some of which only apply to certain orchestrators and requires us to write a bunch of validation code for which permutations of flags are allowed based on the selected orchestrator.

parts/dcosagentvars.t
```diff
@@ -26,4 +26,5 @@
     "{{.Name}}LbIPConfigID": "[concat(variables('{{.Name}}LbID'),'/frontendIPConfigurations/', variables('{{.Name}}LbIPConfigName'))]", 
     "{{.Name}}LbIPConfigName": "[concat(variables('orchestratorName'), '-{{.Name}}-', variables('nameSuffix'))]", 
     "{{.Name}}LbName": "[concat(variables('orchestratorName'), '-{{.Name}}-', variables('nameSuffix'))]",
-{{end}}
\ No newline at end of file
+{{end}}
+    "dcosBootstrapURL": "[parameters('dcosBootstrapURL')]",
```
colemickens: Why put this in the agentvars instead of with the master vars or in dcosbase?

parts/dcoscustomdata173.t
```diff
@@ -1,342 +1,348 @@
...
-  path: /etc/mesosphere/setup-packages/dcos-provider-azure--setup/etc/ui-config.json
-- content: '{}'
-  path: /etc/mesosphere/setup-packages/dcos-provider-azure--setup/pkginfo.json
-- content: ''
-  path: /etc/mesosphere/roles/azure
-- content: 'PROVISION_STR'
-  path: "/opt/azure/containers/provision.sh"
-  permissions: "0744"
-  owner: "root"
+bootcmd:
```
colemickens: Do you know why this entire file changed?

What actually changed in this file?

parts/dcosagentvars.t
colemickens: It looks like it's in both places? How does that work out?

pkg/api/vlabs/types.go
colemickens: For example, I think this struct already exists in the internal model like this, right? And now it's just being wired up to the vlabs so users can input it.

pkg/api/types.go
```diff
@@ -123,8 +123,8 @@ type OrchestratorProfile struct {
 // KubernetesConfig contains the Kubernetes config structure, containing
 // Kubernetes specific configuration
 type KubernetesConfig struct {
-	KubernetesHyperkubeSpec string `json:"kubernetesHyperkubeSpec,omitempty"`
-	KubectlVersion          string `json:"kubectlVersion,omitempty"`
+	KubernetesImageBase string `json:"kubernetesImageBase,omitempty"`
+	KubectlVersion      string `json:"kubectlVersion,omitempty"`
```
colemickens: What if we got rid of `kubectlVersion` entirely, and I add a line to the customData that extracts it from the hyperkube image?

pkg/acsengine/azureconst.go
```diff
@@ -6,7 +6,9 @@ import "fmt"
 
 const (
 	// AzureProdFQDNFormat specifies the format for a prod dns name
-	AzureProdFQDNFormat = "%s.%s.cloudapp.azure.com"
+	AzurePublicProdFQDNFormat = "%s.%s.cloudapp.azure.com"
```
colemickens: It looks like these changes were done manually.

Instead, the PowerShell script needs to be updated so that it can output these things. We really need to make that thing not a PowerShell script...

parts/kubernetesmastercustomscript.sh
```diff
@@ -165,6 +167,11 @@ function writeKubeConfig() {
     chmod 700 $KUBECONFIGDIR
     chmod 600 $KUBECONFIGFILE
 
+    FQDNSuffix="cloudapp.azure.com"
+    if [ "$TARGET_ENVIRONMENT" = "AzureChinaCloud" ]
```
colemickens: Is this the only place the FQDN Suffix is used/needed? Otherwise we might want to compute it higher up rather than having this if/else down inside this script.

parts/kubernetesmastercustomdata.yml
```diff
@@ -317,11 +185,19 @@ write_files:
     [Service]
     Restart=always
     ExecStartPre=/bin/mkdir -p /var/lib/kubelet
+    ExecStartPre=/bin/mount --bind /var/lib/kubelet /var/lib/kubelet
```
colemickens: Can you elaborate on this change? Not that I would be surprised to know we need it... but just want to understand it.

parts/kubernetesmastercustomdata.yml
```diff
@@ -317,11 +185,19 @@ write_files:
...
     # Azure does not support two LoadBalancers(LB) sharing the same nic and backend port.
     # As a workaround, the Internal LB(ILB) listens for apiserver traffic on port 4443 and the External LB(ELB) on port 443
     # This IPTable rule then redirects ILB traffic to port 443 in the prerouting chain
     ExecStartPre=/bin/bash -c "iptables -t nat -A PREROUTING -p tcp --dport 4443 -j REDIRECT --to-port 443"
     ExecStartPre=/bin/sed -i "s|<kubernetesHyperkubeSpec>|{{{kubernetesHyperkubeSpec}}}|g" "/etc/kubernetes/addons/kube-proxy-daemonset.yaml"
+    ExecStartPre=/bin/sed -i "s|<kubernetesKubeDNSSpec>|{{{kubernetesKubeDNSSpec}}}|g; s|<kubernetesDNSMasqSpec>|{{{kubernetesDNSMasqSpec}}}|g; s|<kubernetesExecHealthzSpec>|{{{kubernetesExecHealthzSpec}}}|g" "/etc/kubernetes/addons/kube-dns-deployment.yaml"
+    ExecStartPre=/bin/sed -i "s|<kubernetesHeapsterSpec>|{{{kubernetesHeapsterSpec}}}|g; s|<kubernetesAddonResizerSpec>|{{{kubernetesAddonResizerSpec}}}|g" "/etc/kubernetes/addons/kube-heapster-deployment.yaml"
+    ExecStartPre=/bin/sed -i "s|<kubernetesDashboardSpec>|{{{kubernetesDashboardSpec}}}|g" "/etc/kubernetes/addons/kubernetes-dashboard-deployment.yaml"
```
colemickens: We might just want to use `find` and run the same sed expression on all of the files in /etc/kubernetes/manifests`. Any thoughts about doing that instead of having 7 sed lines?

Pro: less verbose here, any new manifests added in future will be automatically working
Con: One reason maybe not to is that it makes it explicit which values are "used" by each file...

parts/dcosprovision.sh
```diff
@@ -4,11 +4,31 @@ MESOSDIR=/var/lib/mesos/dl
...
+DOCKER_ENGINE_DOWNLOAD_URL=https://az837203.vo.msecnd.net/dcos-deps/docker-engine_1.11.2-0~xenial_amd64.deb
+LIBIPSET_DOWNLOAD_URL=https://az837203.vo.msecnd.net/dcos-deps/libipset3_6.29-1_amd64.deb
+IPSET_DOWNLOAD_URL=https://az837203.vo.msecnd.net/dcos-deps/ipset_6.29-1_amd64.deb
+UNZIP_DOWNLOAD_URL=https://az837203.vo.msecnd.net/dcos-deps/unzip_6.0-20ubuntu1_amd64.deb
+LIBLTDL_DOWNLOAD_URL=https://az837203.vo.msecnd.net/dcos-deps/libltdl7_2.4.6-0.1_amd64.deb
+
+case $DCOS_ENVIRONMENT in
+    # because of Chinese GreatWall Firewall, the default packages on Azure CDN is blocked. So the following Chinese local mirror url should be used instead.
+    AzureChinaCloud)
+        DOCKER_ENGINE_DOWNLOAD_URL=http://acsengine.blob.core.chinacloudapi.cn/dcos/docker-engine_1.11.2-0~xenial_amd64.deb
```
colemickens: Don't you have the URLs as part of the internal model? Should you use that here instead of repeating the if/else logic to determine the download urls?

pkg/acsengine/testdata/windows/kubernetes_expected_params.json
```diff
@@ -32,11 +32,38 @@
...
+    "value": "gcr.io/google_containers/kubernetes-dashboard-amd64:v1.5.1"
+  },
+  "kubernetesExecHealthzSpec": {
+    "value": "gcr.io/google_containers/exechealthz-amd64:1.2"
+  },
+  "kubernetesHeapsterSpec": {
+    "value": "gcr.io/google_containers/heapster:v1.2.0"
+  },
   "kubernetesHyperkubeSpec": {
-    "value": "gcr.io/google_containers/hyperkube-amd64:v1.5.3"
```
colemickens: Why is this change here? Is this rebased? Or was there a rebase problem that reverted from 1.5.3 to 1.5.1?

pkg/acsengine/engine.go
```diff
@@ -283,10 +291,26 @@ func prepareTemplateFiles(properties *api.Properties) ([]string, string, error)
 	return files, baseFile, nil
 }
 
-func getParameters(properties *api.Properties, isClassicMode bool) (map[string]interface{}, error) {
+//GetCloudSpecConfig returns the kubenernetes container images url configurations based on the deploy target environment
+//for example: if the target is the public azure, then the default container image url should be gcr.io/google_container/...
+//if the target is azure china, then the default container image should be mirror.azure.cn:5000/google_container/...
+func GetCloudSpecConfig(location string) AzureEnvironmentSpecConfig {
+	cloudSpecConfig := AzureCloudSpec
+	switch location {
```
colemickens: Is this going to silently ignore someone who tries to use `AzureGermanCloud`?

Seems like they're silently going to get PublicCloud? Do we need a validation check?

colemickens: Actually, you added the other clouds, can you add them to the switch statement? Or else remove the explicit ones you added for Germany/USGov.

parts/kubernetesmastercustomscript.sh
colemickens: For example, for any of this to work for Windows, we will need the FQDN Suffix in the windows files as well. This is something that should probably be in the internal model.

pkg/api/vlabs/types.go
amanohar: **Option 1: OrchestratorOptions  will have both Sovereign cloud config and other options.**

 	OrchestratorType OrchestratorType `json:"orchestratorType"` 
	orchestratorOption[] OrchestratorOptions  `json:"orchestratorOptions,omitempty"` 
} 

type OrchestratorOption struct {
    Key   string
    Value string
}

**Option 2: This option will have separate Sovereign cloud property and other config options will be a separate field. I am leaning towards this option. 

I recommend renaming KubernetesConfig to KubernetesSovereignConfig or something similar so its clear what the differences between both fields would be.**

 	OrchestratorType OrchestratorType `json:"orchestratorType"` 
	KubernetesConfig KubernetesConfig `json:"kubernetesConfig,omitempty"` 
	orchestratorOption[] OrchestratorOptions  `json:"orchestratorOptions,omitempty"` 
} 

type OrchestratorOption struct {
    Key   string
    Value string
}

parts/dcosagentvars.t
wangtt03: This parameter is in both agentvar and mastervar. I also created a new dcosparam.t to set the default value of the parameter. The value of this parameter is determined by the version of the DCOS specified in the cluster definition json. For example, if DCOS1.8.7 is specified, download url will be generated and set to the parameter based on the location of the cloud.

parts/dcosprovision.sh
wangtt03: For bootstrap download url I put it in the internal model. As for these urls I followed current design of this shell script to determine the url based on the env. Put these urls into internal model and dynamically generate them using parameters may introduce more configurations. I think it may be better to put them in current location. @colemickens do you think it make sense?

parts/kubernetesmastercustomdata.yml
wangtt03: I found that some of the yaml contents which will be extracted to /etc/kubernetes/manifest, and /etc/kubernetes/add-on are gziped and some are not. I just put all these kind of yamls as gzip content to make it clean. And another problem is that if there are too many content in this file, it will be an error when deploying it to Azure due to the limitation of the customData.

parts/kubernetesmastercustomdata.yml
wangtt03: I prefer the sed instead of find. As you said, it is easy to find that which file use which parameter. :-)

parts/kubernetesmastercustomscript.sh
wangtt03: AFAIK, this is the only 1 fqdn need to updated. There is also a fqdn config in azureconst.go which is used to generate kube config files based on the locations. However, I found these generated kubeconfigs are not used. The kubeconfig in the destination env is generated by this sh script.

wangtt03: Please correct me if I miss something.

pkg/acsengine/azureconst.go
wangtt03: Sure, I will do that!

pkg/acsengine/engine.go
wangtt03: Actually I don't know if the USGov/Germany have the same firewall issues that block the download of dcos packages and google. By default I suppose these two clouds should be work as the public cloud, but I am not sure. If the assumption is correct, I can remove the ones.

pkg/acsengine/testdata/windows/kubernetes_expected_params.json
wangtt03: It should be my bad. I will fix it.

pkg/api/types.go
wangtt03: KubectlVersion is also used to concat the kubectl download url. Personally I prefer keep the kubectlversion because we can simplify the logic in customData. Also what if the image url is: hyperkube.amd64:latest? In that case we can never determine the version of kubectl.

pkg/api/vlabs/types.go
wangtt03: I prefer the flat one. Introduce more configurations or concepts like "sovereignconfig" may cause more learning effort for users. @amanohar @colemickens what do you think about it?

parts/dcoscustomdata173.t
wangtt03: It is weird... Actually I changed two places:
1. Add an env variable
`
- content: 'DCOS_ENVIRONMENT={{{targetEnvironment}}}

    '
  owner: root
  path: /etc/mesosphere/setup-flags/dcos-deploy-environment
  permissions: '0644'
`
2. change the hard coded download url to configurable.
`
ExecStartPre=/usr/bin/curl -fLsSv --retry 20 -Y 100000 -y 60
    -o /var/lib/mesos/dl/bootstrap.tar.xz {{{dcosBootstrapURL}}}
`

pkg/acsengine/engine.go
```diff
@@ -179,7 +187,7 @@ func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerServ
 
 	properties := &containerService.Properties
 
-	if certsGenerated, err = SetPropertiesDefaults(properties); err != nil {
+	if certsGenerated, err = SetPropertiesDefaults(properties, containerService.Location); err != nil {
```
anhowe: should be SetPropertiesDefaults(containerService)

pkg/acsengine/engine.go
```diff
@@ -218,7 +226,7 @@ func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerServ
 	templateRaw = b.String()
 
 	var parametersMap map[string]interface{}
-	if parametersMap, err = getParameters(properties, t.ClassicMode); err != nil {
+	if parametersMap, err = getParameters(properties, t.ClassicMode, containerService.Location); err != nil {
```
anhowe: should be getParameters(containerService, t.ClassicMode)

pkg/acsengine/engine.go
wangtt03: @anhowe Great. Will update it soon.

pkg/acsengine/engine.go
wangtt03: @anhowe Sure!

pkg/api/vlabs/types.go
amanohar: @wangtt03 your change can stay as is. We will look into introducing OrchestratorOptions at some point in the future but the current setup is extensible so we should be good.

parts/dcosagentvars.t
colemickens: I'm confused how this works though. ARM allows you to have two variables with the same name in a template? Or am I still missing a detail?

parts/dcoscustomdata173.t
colemickens: In the diff for this file, I see this:

```
+- content: 'DCOS_ENVIRONMENT={{{targetEnvironment}}}
 +
 +    '
 +  owner: root
 +  path: /etc/mesosphere/setup-flags/dcos-deploy-environment
 +  permissions: '0644'
```

Which doesn't seem like quite what you wanted? Otherwise, makes sense. thanks.

pkg/api/types.go
colemickens: We only use the version to download `kubectl` though. If we just extract `kubectl` on the node from the `hyperkube` image, we don't need to construct the kubectl url at all, or download it at all.

It also has the benefit of meaning that if you use a custom hyperkube build with... say a new Disk type... you will automatically get the correct `kubectl`. Otherwise, you'd wind up with the wrong `kubectl` that would prevent you from testing the disk change properly... etc...

colemickens: This is just an idea. I'm fine if you want to punt it, I can play with it and send another PR some other time if you're not interested in this.

parts/kubernetesmastercustomscript.sh
colemickens: I think you're right. It looks like in Windows we only ever points at an IP address or localhost, never the external DNS name... Not sure why I was worried about this. I think you're right, nothing to worry about here.

parts/kubernetesmastercustomdata.yml
colemickens: Sure, I'm fine with that. I was more specifically asking about... why do you bind mount `/var/lib/kubelet`? Is that needed? Do other setups do this?

parts/kubernetesmastercustomdata.yml
colemickens: 👍

parts/dcosprovision.sh
colemickens: 👍 Actually, after thinking about it more, it is maybe even better you keep it this way. It means that if the user later tries to change the location from a China location to a public cloud location... without regenerating.... there's still a chance it would work out. At least for DC/OS. For Kubernetes, the hyperkube prefix is part of the model, so the user is kinda stuck with whatever they chose originally... unless you were to make that prefix a parameter that they can change in the generate parameters file... but I'll defer to you on whether or not you want to do it. (Frankly, I kinda think we're going to wind up in a model soon where you're required to regenerated and discouraged from manually adjusting the parameters file... so it's probably not worth the effort to make it an adjustable parameter).

pkg/acsengine/engine.go
colemickens: For now, let's assume that they can use the public mirrors. I'd still prefer that we either include the other clouds in the switch, or remove them from being explicit clouds. See what I mean? Otherwise someone is going to come along, change the AzureGermanCloud definition, and they may think they're done. But because AzureGermanCloud isn't part of the switch... they'll never actually get the updated values. So we should either go ahead and add it to the switch, or remove those environment definitions so that if  someone comes along and adds it later, they're more likely to look for the switch statement and add the new cloud.

colemickens:
@wangtt03 I've just replied to your responses, and am still waiting for a new push with some of the things we've discussed so far. Thanks again for all the work and patience on this!

parts/dcosagentvars.t
wangtt03: Sorry, I made a mistake. Put the variable to the master vars will be OK. I will change it. Thanks!

parts/dcoscustomdata173.t
wangtt03: It is a format issue. The correct content should be the one you specified.

pkg/api/types.go
wangtt03: I will take this issue to extract version from hyperkube image url. Let me do it:)

parts/kubernetesmastercustomdata.yml
wangtt03: Oh, it should be a merge or rebase issue. Actually I didn't change that line. What I did is just use `sed` to replace strings.

parts/dcosprovision.sh
wangtt03: Sounds good. I will make the prefix configurable.

pkg/acsengine/engine.go
wangtt03: Sure, I will remove the German and USGov definitions.

wangtt03:
@colemickens I did some code changes based on your and Anthony's comments:
1. Remove the redundant dcosBootstrapURL in DC/OS parameters.
2. Change the powershell script to generate azureconst.go by adding china east/north location
3. Remove the explicit defined AzureGermanCloud and AzureUSGovCloud.
4. Remove KubectlVersion parameter. Extract it from hyperkube image instead.
5. Interface changes requested by Anthony (properties -> containerService).
6. Update test cases.
7. Some minor changes.

@colemickens @anhowe Thanks for your reviewing😊

wangtt03:
@colemickens Could you please review this PR? Thanks😊!

parts/dcoscustomdata173.t
colemickens: Does this need to be fixed? It stills appears multi-line for me...

colemickens:
@wangtt03 Looks great. One remaining question: https://github.com/Azure/acs-engine/pull/152#discussion_r102846242

Does that need to be fixed? I assume you're testing it and it works, so I guess the extra whitespace must not hurt? Let me know and I'll merge this tomorrow.

wangtt03:
@colemickens Thanks for your reviewing. The extra white space doesn't hurt. Actually I just followed the existing coding convention in this file:)

colemickens:
I thought I'd seen it before... thanks for confirming.

PR #153, opened 2016-12-21
Fix for loop value

lastcoolnameleft:
See http://stackoverflow.com/questions/4176347/using-bash-c-with-loops-and-standard-output-redirection

```
# /bin/bash -c "for i in {1..5}; do echo MyFirstApp \$i;  done"
MyFirstApp 1
MyFirstApp 2
MyFirstApp 3
MyFirstApp 4
MyFirstApp 5
# /bin/bash -c "for i in {1..5}; do echo MyFirstApp $i;  done"
MyFirstApp 5
MyFirstApp 5
MyFirstApp 5
MyFirstApp 5
MyFirstApp 5
```

### DIFF
--- a/docs/dcos.md
+++ b/docs/dcos.md
@@ -70,7 +70,7 @@ This walk through is inspired by the wonderful digital ocean tutorial: https://w
    3. start a long running service
     1. click "Deploy Service"
     2. type "myfirstapp" for the id
-    3. type `/bin/bash -c "for i in {1..5}; do echo MyFirstApp \$i; sleep 1; done"` for the command
+    3. type `/bin/bash -c 'for i in {1..5}; do echo MyFirstApp $i; sleep 1; done'` for the command
     4. scroll to bottom and click Deploy
 
     ![Image of Deploy New Service dialog](images/deployfirstapp.png)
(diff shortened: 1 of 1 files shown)


rutsky:
LGTM.

IMO it's better to wrap whole `bash -c` argument with single quotes:

```
/bin/bash -c 'for i in {1..5}; do echo MyFirstApp $i; done'
```

lastcoolnameleft:
@rutsky For strings that don't require interpolation, I wholeheartedly agree.  However, in this case, we want the number, not the literal $i:

```
# /bin/bash -c 'for i in {1..5}; do echo MyFirstApp \$i; sleep 1; done'
MyFirstApp $i
MyFirstApp $i
MyFirstApp $i
MyFirstApp $i
MyFirstApp $i
# /bin/bash -c "for i in {1..5}; do echo MyFirstApp \$i; sleep 1; done"
MyFirstApp 1
MyFirstApp 2
MyFirstApp 3
MyFirstApp 4
MyFirstApp 5
```

lastcoolnameleft:
@rutsky Fair point.  I've changed it to use single quotes.

raveeram:
Thank you for your contribution!

PR #158, opened 2017-01-05
Add classic profile to make vlabs compatible with older versions.

raveeram:
With this change, templates created with older versions like 2016-03-30 and 2016-09-30 can be migrated to the newer "vlabs" version.


pkg/api/vlabs/validate.go
```diff
@@ -264,6 +274,17 @@ func validateUniqueProfileNames(profiles []AgentPoolProfile) error {
 	return nil
 }
 
+func validateClassicProfileType(profileType ClassicAgentPoolProfileType) error {
+	switch profileType {
+	case SwarmPublic:
+	case DCOSPublic:
+	case DCOSPrivate:
+	default:
+		return fmt.Errorf("ClassicAgentPoolProfile has unknown profile type: %s", profileType)
```
anhowe: can you just add case "", since case "" is a valid value meaning not classic.  It may be useful to declare a const for "" and call it "NotClassic" or similar name

pkg/api/vlabs/validate.go
```diff
@@ -93,6 +98,11 @@ func (a *AgentPoolProfile) Validate() error {
 	if len(a.Ports) == 0 && len(a.DNSPrefix) > 0 {
 		return fmt.Errorf("AgentPoolProfile.Ports must be non empty when AgentPoolProfile.DNSPrefix is specified")
 	}
+	if a.ClassicProfile != "" {
```
anhowe: validate should allow "", and then you don't need this guard.

pkg/api/vlabs/validate.go
raveeram: good idea, will add that.

PR #161, opened 2017-01-07
Add a simple cluster test script.

brendandburns:


test/cluster-tests/kubernetes/test.sh
```diff
@@ -0,0 +1,71 @@
...
+namespace="namespace-${RANDOM}"
+echo "Running test in namespace: ${namespace}"
+trap teardown EXIT
+
+function teardown {
+  kubectl delete namespaces ${namespace}
+}
+
+
+echo "Testing number of nodes is ${EXPECTED_NODE_COUNT}"
```
anhowe: Host OS Upgrades or Service healing may happen from time to time so we may not have all nodes.  Should this retry for up to 10 minutes if not found?

test/cluster-tests/kubernetes/test.sh
```diff
@@ -0,0 +1,71 @@
...
+
+EXPECTED_NODE_COUNT=4
+EXPECTED_DNS=2
+
+namespace="namespace-${RANDOM}"
+echo "Running test in namespace: ${namespace}"
+trap teardown EXIT
+
+function teardown {
+  kubectl delete namespaces ${namespace}
```
anhowe: are there any retries needed here?

anhowe:
@brendandburns - I've merged this change.  @weinong - please look at my two minor comments as you work on Jenkins.

PR #183, opened 2017-01-13
Make the deploy command generic.

brendandburns:
This is a pre-requisite for further continuous testing.

@colemickens

### DIFF
--- a/test/common.sh
+++ b/test/common.sh
@@ -22,14 +22,18 @@ function jqi() {
 }
 
 function deploy() {
+        cluster_type=${ORCHESTRATOR:-"kubernetes"}
 	# Check pre-requisites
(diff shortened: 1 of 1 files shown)


colemickens:
LGTM.

FYI, I've made [other (non-conflicting) changes to `common.sh` in my MSI branch](https://github.com/colemickens/acs-engine/commit/09ae65137c600b98e6d9f5f2b7d822fc26185409#diff-daaa154e22be44798ffcddf12a70e03e). I'll pull those out soon and send them in a separate PR. Specifically, when I was running this in the last two weeks, the SSH pubkey conversion wasn't working properly (and standard tools should be usable if the docker container has a proper user inside, which will be the case when I get the jobs running as Pods since the base Jenkins docker container contains a proper jenkins user for running the slave.jar).

Additionally, we should bump the CLI in the Dockerfile and then change `az resource group` -> `az group` but I'll do that as part of setting up MSI jobs which is next on my MSI list, so should be soon.

PR #187, opened 2017-01-17
Update Azure CLI 2.0 commands

Radek44:
Looks like az command line changed and requires `resource` to be included in front of `group`, updating where needed.


colemickens:
Please update your CLI. The `resource` part was removed from the commands. For reference, we merged the opposite of your PR about 7 days ago: https://github.com/Azure/acs-engine/pull/174

Issue #180, opened 2017-01-12
k8s: Make kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp configurable for custom VNet


alxlchnr:
In case you want to deploy a kubernetes cluster in an existing VNet which has a different address space than the default (10.0.0.0/8) you also want to adjust the hardcoded values for kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp. Currently the only way to achieve this is to deploy the templates and change the relevant files on the created master and agent vms. But you may still face problems.
It would be better to specify the values before template creation and get templates which fit the predefined VNet.

https://github.com/Azure/acs-engine/blob/2c4f3168751be5d3b2add8327eef8818c88daceb/parts/kubernetesmasteraddons-kube-dns-service.yaml#L11

https://github.com/Azure/acs-engine/blob/2c4f3168751be5d3b2add8327eef8818c88daceb/parts/kubernetesmastercustomdata.yml#L147

https://github.com/Azure/acs-engine/blob/2c4f3168751be5d3b2add8327eef8818c88daceb/parts/kubernetesmastervars.t#L58-L59

https://github.com/Azure/acs-engine/blob/2c4f3168751be5d3b2add8327eef8818c88daceb/pkg/acsengine/pki.go#L113

anhowe:
Thanks for filing this.  Marking this as P2 and we will continue to watch for customer interest.  Also, please consider submitting a PR on this item.

dcieslak19973:
I'm interested in this functionality

siamak-haschemi:
We also need to do that, since the 10.0.0.0/16 network is already blocked by our productive environment. I also did not understand the workaround above. Maybe @alxlchnr could comment that further. I am happy with a manual solution for now ...

siamak-haschemi:
In #303 I wrote the detailed steps to run a cluster within a custom VNET and with a different IP-subnet. Just in case someone needs that.

alxlchnr:
#191

alxlchnr:
Sorry for the late answer @siamak-haschemi . It seems that you already found my pull request #191 in order to find a manual solution for your problem. The Pull Request may have saved a lot of time if it was merged to master. 
With the PR it is possible to konfigure kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp in the template and all changes you have done manually will be done by acs-engine.

PR #191, opened 2017-01-18
#180 k8s in custom Vnet - configurable kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp

alxlchnr:
# Problem
In #180 I adressed some harcoded IP adresses which go with the default vnet 10.0.0.0/8. But several users have reported that they have already an existing vnet which often has a different address space than the default vnet. In this case the hard coded values would not fit and cause a non working kubernetes cluster. 
# Solution
The idea of this PR is as follows: all hardcoded IP-Adresses should be changed to more appropriate values. In order to find the correct values it is necessary to hand over these values to acs-engine before the arm template are generated. Therefore I added three new kubernetes parameters (kubeDnsServiceIp, kubeServiceCidr, kubeClusterCidr). With these parameters I am able to find appropriate values for all previously hardcoded IP adresses.
The introduction of new kubernetes parameters may break old templates because they are not present. To overcome this issue I check for presence of the newly added parameters and if they are not present I work with default values which are the hardcoded IP adresses from the master branch
# Why?
@siamak-haschemi described in #303 the manual steps which are necessary to deploy a kubernetes cluster in a custom vnet which has a different address space. In my opinion this is too much manual effort in order to deal with an apparantly frequent use case. The steps @siamak-haschemi described are the same I implemented in this PR. The PR will save future users time and manual adjustments on the generated arm templates.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/191)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

anhowe:
@alxlchnr - thanks for your changes.  We are CR'ing this PR this week.

pkg/acsengine/testdata/vnet/kubernetesvnet.json
```diff
@@ -49,6 +49,11 @@
       "clientPrivateKey": "clientPrivateKey",
       "kubeConfigCertificate": "kubeConfigCertificate",
       "kubeConfigPrivateKey": "kubeConfigPrivateKey"
+    },
+    "kubeNetworkConfig": {
+      "kubeDnsServiceIp": "172.17.1.10",
```
rjtsdl: How is the backward compatibility handled here ? I am not saying we need backward compatibility or not. Just wondering in your change, how this get treated ?

colemickens:
@alxlchnr Is this ready for review? Can you explain the tactic you took with this, and then also rebase it?

pkg/acsengine/testdata/vnet/kubernetesvnet.json
alxlchnr: Well I think old files won't have kubeNetworkConfig section. In this case I usethe old hardcoded values as defaults in pkg/acsengine/defaults.go
So users with old template files are still able to create the deployment templates.

alxlchnr:
@colemickens just answered Jing Tao's question. Also rebased

alxlchnr:
I will a last test, because after rebasing i get a deployment validation error. 
`error:   InvalidTemplate : Deployment template validation failed: 'The template resource 'k8s-master-39411720-0' at line
 '1' and column '31824' is not valid: The language expression length limit exceeded. Limit: '16384' and actual: '16673'.
. Please see https://aka.ms/arm-template-expressions for usage details.'.`

I will investigate and come back when it's fixed.

colemickens:
@alxlchnr Unfortunately this is caused by the customData being too large. One way to workaround is to remove whitespace from the customData. We need to address this soon though, here's an issue: https://github.com/Azure/acs-engine/issues/216

alxlchnr:
@colemickens so this is nothing I need to worry about?

colemickens:
@alxlchnr Depends. If we do nothing we block merging for a non-trivial amount of time. I will attempt to create some sort of workaround today so that we can unblock you and others.

colemickens:
That having been said, this is not properly rebased - there are 48 commits. I can't really review this in this state.

alxlchnr:
You were right, I'm sorry. May be it's better now.

anhowe:
@alxlchnr  - are you able to rebase per Coles request. We want to review and merge.

alxlchnr:
@anhowe I just rebased. Please have a look.

pkg/acsengine/testdata/vnet/kubernetesvnet.json
alxlchnr: Did my reply answer your question @rjtsdl or do you miss something?

alxlchnr:
@colemickens : I updated the description of this PR. I hope it adresses your question for the tactic I took with this PR.

pkg/acsengine/testdata/vnet/kubernetesvnet.json
rjtsdl: Thx, you answered my ask. And the approach looks good.

rjtsdl: Sounds good to me

colemickens:
We're currently figuring out how we want to enable optional features and extension points like this into the apimodel. We will come to a conclusion shortly and revisit this PR. Thanks.

brendandburns:

ashb:
Hi! TI was after this feature and master seemed to have moved on quite a lot (to do with now the .yaml contents are included in the userdata/gzipping) so I have rebased this on master here:  https://github.com/ashb/acs-engine/tree/customVnet -- I have maintained @alxlchnr's authorship on the commit.

I haven't opened a new pull request for this but I can do if that would be easier.

ashb:
Okay so my conclusion is I can't quite get this working as is, at least not using the default networkPolicy - the IPs alone aren't enough as we don't have control over things like the agent_n_Subnet params.

colemickens:
@ashb No network policy should work fine. If the route table was empty, it points to you not having correct service principal credentials. WIth the "azure" vnet network policy, this problem should actually be even easier since the pod ips will automatically come from the subnet range, so you'd actually need to adjust less flags.

I'd suggest focusing on the "none" network policy for now. Check your SP credentials and see if you get any further. If not, make sure your branch is pushed and ping us and we can see if we spot what's wrong.

ashb:
@colemickens Pushed as ashb/customVnet. I'm fairly sure I've got the SP right now as the nodes are coming up and `docker ps` shows the usual components, `kubectl get node` shows the right number and a status of Ready, but I've had a day of various failure modes and I'm not quite sure where I am with... anything... any more.

A second pair of eyes would be greatly appreciated.

(The goal we are trying to achieve is to have each of our three clusters in a separate IP range. Even though they won't be speaking to each other, they will be speaking to other IaaS components in Azure and our networking team wants them to be distinct.)

colemickens:
@ashb Are you on the Kubernetes Slack? I can offer some interactive assistance in ~90 minutes. Otherwise...

The cluster you have now with Ready nodes, is the networking working (route table populated)? Or do you still get issues when trying to `curl kubernetes` for example?

ashb:
I am, but timezones (Europe/London here) mean I'll be done in about 60 mins. I'm trying again with a clean RG and this as the (part) input to acs-engine:

```json
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "kubernetesConfig": {
        "dnsServiceIp": "192.168.1.255",
        "serviceCidr": "192.168.1.0/24"
      }
    },
```

I'll report how I get on. I'm hopeful this one at least should work as the service Cidr range isn't routed outside and is just iptables rules on each node.

(With apologies to @alxlchnr for hijacking his issue and spamming his inbox.)

ashb:
Ah progress! I have found an error message from the kube-proxy logs:

```
E0405 16:06:26.838902       1 proxier.go:1314] Failed to execute iptables-restore: exit status 2 (iptables-restore v1.4.21: host/network `<kubernetesDnsServiceIp>' not found
```

I missed a substitution somewhere!

colemickens:
@ashb That is related to a comment I was going to make -- I think you need to specify all of the custom cidrs, even if you're okay with the defaults. (We should also set defaults and then apply custom, but for the purpose of making progress now, I'd specify them all).

colemickens:
Per Slack, it sounds like @ashb has this working. I'm not sure if his branch is up to date or if there are other fixes that haven't been pushed.

Hopefully tomorrow @ashb can go ahead and send a new PR. It will be great to have this!

alxlchnr:
I'm glad you got it working and the discussion on the issue is revived.
Thanks @ashb for your effort.

anhowe:
this was completed in k8s in custom Vnet - configurable kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp #473

PR #197, opened 2017-01-20
Xtoph dockerfilefix

xtophs:
Missing dependency in the container. Edited Dockerfile to add github.com/ghodss/yaml


pkg/api/v20160330/types.go
```diff
@@ -126,7 +126,8 @@ type AgentPoolProfile struct {
 	FQDN      string `json:"fqdn,omitempty"`
 	OSType    OSType `json:"osType"` // TODO: This field is versioned to "2016-03-30"
 	// subnet is internal
-	subnet string
+	subnet     string
+	Attributes map[string]string `json:"attributes,omitempty"`
```
JackQuincy: this api version is closed and shouldn't have changes made to it. please revert this change

pkg/acsengine/engine.go
```diff
@@ -243,6 +243,7 @@ func getParameters(properties *api.Properties) (map[string]interface{}, error) {
 		if len(agentProfile.Ports) > 0 {
 			addValue(parametersMap, fmt.Sprintf("%sEndpointDNSNamePrefix", agentProfile.Name), agentProfile.DNSPrefix)
 		}
+
```
JackQuincy: nit: remove this extra newline

pkg/acsengine/engine.go
```diff
@@ -301,13 +302,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 		},
 		"GetDCOSMasterCustomData": func() string {
 			masterProvisionScript := getDCOSMasterProvisionScript()
-			str := getSingleLineDCOSCustomData(properties.OrchestratorProfile.OrchestratorType, properties.MasterProfile.Count, masterProvisionScript)
+			str := getSingleLineDCOSCustomData(properties.OrchestratorProfile.OrchestratorType, properties.MasterProfile.Count, masterProvisionScript, "")
```
JackQuincy: I'm not a fan of passing in an empty string like this. I'd prefer something where we didn't need to clear out the string in the masters case.

examples/dcos.184.attributes.json
```diff
@@ -0,0 +1,45 @@
...
+      "count": 1,
+      "dnsPrefix": "xtophmaster",
+      "vmSize": "Standard_D2_v2"
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "xtentprivate",
+        "count": 3,
+        "vmSize": "Standard_D2_v2",
+        "attributes" : {
```
anhowe: we are reviewing this API model change with the team.

pkg/acsengine/engine.go
anhowe: Can masters have attributes?  It would be great to add this capability to masters as well.

pkg/api/convertertoapi.go
```diff
@@ -174,6 +174,10 @@ func convertV20160330AgentPoolProfile(v20160330 *v20160330.AgentPoolProfile, api
 	api.FQDN = v20160330.FQDN
 	api.OSType = OSType(v20160330.OSType)
 	api.Subnet = v20160330.GetSubnet()
+	api.Attributes = map[string]string{}
+	for k, v := range v20160330.Attributes {
```
anhowe: again, this is only needed for vlabs.

xtophs: already backed out the vlabs change in master. will back out of this branch shortly

Dockerfile
```diff
@@ -22,4 +22,5 @@ RUN curl "https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_V
 ENV GOPATH /gopath
 ENV PATH "${PATH}:${GOPATH}/bin"
 RUN go get -u github.com/golang/lint/golint
+RUN go get -u github.com/ghodss/yaml
```
JackQuincy: this feels like all this review is about. If so could you rebase onto master and make this the only change? or if it is easier just start a new branch/PR with this being the only change?

xtophs:
Yep ... I'm pretty sure this branch was created in the browser from Azure/master since it was such a tiny change, but somehow (most likely user error) it branched of xtophs/master which had changes from other branches ..

I *think* I'll have everything in a good state shortly. If not I'll back it out and start over.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

colemickens:
This should be squashed both for reviewing and merging purposes.

JackQuincy:
Seems like your having trouble squashing the commits. here is a page talking about how. http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html 
Please note that afterwards you need to do a "git push --force" to avoid merging your squashed commits with the none squashed commits.

colemickens:
This is unnecessary. The `Makefile` executes `go get .` which already pulls in these deps.

Issue #185, opened 2017-01-16
Kubernetes/Agents: Create /etc/kubernetes/manifests directory


khenidak:
There are too many logs like this one:

Unable to read config path "/etc/kubernetes/manifests": path does not exist, ignoring

can be solved if /etc/kubernetes/manifests was created by default on all agents.

anhowe:
Thanks @khenidak for this bug report.  I have marked it as a P0.

anhowe:
This has now been fixed.  Closing issue.  Thanks again for the report.

PR #207, opened 2017-01-25
fix issue #185, pre-create manifests directory

anhowe:
This fixes issue #185.  This pre-creates manifests directory on manifest and agent in case there is ever a time where they do not have manifest files.


parts/kubernetesagentcustomdata.yml
```diff
@@ -13,6 +13,7 @@ runcmd:
 - systemctl restart docker
 - /usr/bin/curl -sSL --retry 12 --retry-delay 10 https://storage.googleapis.com/kubernetes-release/release/{{{kubectlVersion}}}/bin/linux/amd64/kubectl > /usr/local/bin/kubectl
 - chmod +x /usr/local/bin/kubectl
+- /bin/mkdir -p /etc/kubernetes/manifests
```
colemickens: Please use `mkdir`. If `/usr/bin` or `/bin` are never not-on-PATH, the system won't boot and we'll be so hosed it won't matter.

anhowe: fixed

anhowe: and deployed and confirmed working

colemickens:
Go for it. I approved already.

colemickens:
One more suggestion @anhowe. If your commit message or pull request text
says "Fixes #0" then the associated Issue is closed automatically with a
link to the PR.

As is now, you're closing without manually or automatically linking to the
PR so it's hard for people to discover where or how their issue was fixed.

PR #210, opened 2017-01-25
checkin two missing files

anhowe:
these two files didn't get checked in


colemickens:
(nit: PR title and commit title could be more descriptive, even though I figured out what's going on)

PR #211, opened 2017-01-25
Ensure Master NSG is added on Master NIC (Fixes #143)

### LINKED ISSUE #143: Master Network Security Group goes unused in DC/OS
In the output DC/OS template, the master network security group goes unused, as it is never actually associated with the master subnet. The following is the code snippet (in context) that attaches the master NSG to the master subnet:

```
...
    }, {
        "name": "[variables('virtualNetworkName')]",
        "apiVersion": "[variables('apiVersionDefault')]",
        "location": "[resourceGroup().location]",
        "dependsOn": ["[concat('Microsoft.Network/networkSecurityGroups/', variables('agentworkerNSGName'))]", "[concat('Microsoft.Network/networkSecurityGroups/', variables('agentcontrolNSGName'))]", "[concat('Microsoft.Network/networkSecurityGroups/', variables('agentproxyNSGName'))]"],
        "type": "Microsoft.Network/virtualNetworks",
        "properties": {
            "subnets": [{
                "name": "[variables('masterSubnetName')]",
                "properties": {
                    "addressPrefix": "[variables('masterSubnet')]",
                    "networkSecurityGroup": {
                        "id": "[resourceId('Microsoft.Network/networkSecurityGroups', variables('masterNSGName'))]"
                    }
                }
            }, {
                "name": "[variables('agentprivateSubnetName')]",
                "properties": {
...
```

Specifically, this section:
```
"networkSecurityGroup": {
         "id": "[resourceId('Microsoft.Network/networkSecurityGroups', variables('masterNSGName'))]"
}
```

This has been verified in the current template output of DC/OS.

anhowe:
Ensure Master NSG is added on Master NIC (Fixes #143).  This has been tested and deployed.


parts/dcosmasterresources.t
```diff
@@ -141,9 +141,8 @@
         "name": "nicLoopNode"
       },
       "dependsOn": [
-{{if .MasterProfile.IsCustomVNET}}
         "[variables('masterNSGID')]",
-{{else}}
+{{if not .MasterProfile.IsCustomVNET}}
```
colemickens: what is the purpose of listing variables in the dependsOn sections. It's not as if they will make a difference, right? If the variable isn't there, aren't you still going to get an upfront error when ARM parses the template?

anhowe: ARM uses the depends on to block creation of the resource that the variable represents.  For example, if variables('vnetID'), has not been created yet, then it will wait before creating the NIC.  However, in the bring your own VNET case, we are not creating a VNET, so we don't need to depends on clause in this case.

Interestingly as I was testing this, I forgot to add this in the depends and I actually got the following error from ARM, because it tried to create the NIC before the VNET:

        "message": "Resource /subscriptions/MYSUBIDREMOVED/resourceGroups/AHWE0124F/providers/M
icrosoft.Network/virtualNetworks/DCOS-VNET-35433090 not found."

colemickens: Right. In my templates I normally have `[resource('RP/', variable('resourceName')]` which is the exact thing you're doing. Don't know why I was confused. Thanks.

PR #212, opened 2017-01-25
Update DockerCE/Swarm Mode documentation

amanohar:


amanohar:
@anhowe, can I get one more quick CR? This is doc change only. Thanks.

PR #214, opened 2017-01-25
Add ability to specify CA Certificate and Key

### LINKED ISSUE #203: Feature ask: able to reuse generated CA cert in api model
it would be really great if we can reuse CA cert / KubeCtl cert specified in apimodel.js to generate new cluster. This allows user to run multiple clusters with same CA.

anhowe:
Add ability to specify CA Certificate and Key, Fixes #203

### DIFF
--- a/acs-engine.go
+++ b/acs-engine.go
@@ -108,46 +108,6 @@ func saveFile(dir string, file string, data []byte) error {
 	return nil
 }
 
-// injectCACertificateAndKey reads the ca cert and key files if they exist and inserts them into the container service model
-func injectCACertificateAndKey(containerService *api.ContainerService, caCertificatePath string, caKeyPath string) {
-	// confirm the caCerficate file exists
-	var err error
-	if len(caCertificatePath) > 0 {
-		if _, err = os.Stat(caCertificatePath); os.IsNotExist(err) {
-			usage(fmt.Errorf("caCertificatePath file %s does not exist", caCertificatePath))
-			os.Exit(1)
-		}
-		if len(caKeyPath) == 0 {
-			usage(errors.New("caKeyPath must be specified when providing caCertificatePath"))
-			os.Exit(1)
-		}
-		var caCertificateBytes []byte
-		if caCertificateBytes, err = ioutil.ReadFile(caCertificatePath); err != nil {
-			usage(err)
-			os.Exit(1)
-		}
-		containerService.Properties.CertificateProfile.CaCertificate = string(caCertificateBytes)
-	}
-
-	// confirm the caKey file exists
-	if len(caKeyPath) > 0 {
-		if _, err = os.Stat(caKeyPath); os.IsNotExist(err) {
-			usage(fmt.Errorf("caKeyPath file %s does not exist", caKeyPath))
-			os.Exit(1)
-		}
-		if len(caCertificatePath) == 0 {
-			usage(errors.New("caCertificatePath must be specified when providing caKeyPath"))
-			os.Exit(1)
-		}
-		var caKeyBytes []byte
-		if caKeyBytes, err = ioutil.ReadFile(caKeyPath); err != nil {
-			usage(err)
-			os.Exit(1)
-		}
-		containerService.Properties.CertificateProfile.SetCAPrivateKey(string(caKeyBytes))
-	}
-}
-
 func usage(errs ...error) {
 	for _, err := range errs {
 		fmt.Fprintf(os.Stderr, "error: %s\n\n", err.Error())
@@ -163,6 +123,9 @@ var noPrettyPrint = flag.Bool("noPrettyPrint", false, "do not pretty print outpu
 var artifactsDir = flag.String("artifacts", "", "directory where artifacts will be written")
 var classicMode = flag.Bool("classicMode", false, "enable classic parameters and outputs")
 var parametersOnly = flag.Bool("parametersOnly", false, "only output the parameters")
+
+// acs-engine takes the caKey and caCert as args, since the caKey is stored separately
+// from the api model since this cannot be easily revoked like the server and client key
 var caCertificatePath = flag.String("caCertificatePath", "", "the path to the CA Certificate file")
 var caKeyPath = flag.String("caKeyPath", "", "the path to the CA key file")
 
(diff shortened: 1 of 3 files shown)


acs-engine.go
```diff
@@ -122,6 +163,8 @@ var noPrettyPrint = flag.Bool("noPrettyPrint", false, "do not pretty print outpu
 var artifactsDir = flag.String("artifacts", "", "directory where artifacts will be written")
 var classicMode = flag.Bool("classicMode", false, "enable classic parameters and outputs")
 var parametersOnly = flag.Bool("parametersOnly", false, "only output the parameters")
+var caCertificatePath = flag.String("caCertificatePath", "", "the path to the CA Certificate file")
```
colemickens: Why do we plumb this into CLI but not other features in the apimodel.json? I guess maybe this is for convenience? Instead we could:

1. document it -- openssl and jq could inject them into the apimodel)
2. create a separate tool that would help spit out apimodel files based on flags (`acs-engine create-apimodel --orchestrator=k8s --ca-cert=blah.crt --ca-key=blah.key` which outputs the `apimodel.json`).

I worry a bit every apimodel knob is going to appear as a CLI flag if this starts a trend.

If you do leave this in, maybe make the flag a `StringPtr`. Then you can stop checking len = 0, and check against nil which seems more appropriate.

acs-engine.go
```diff
@@ -107,6 +108,46 @@ func saveFile(dir string, file string, data []byte) error {
 	return nil
 }
 
+// injectCACertificateAndKey reads the ca cert and key files if they exist and inserts them into the container service model
+func injectCACertificateAndKey(containerService *api.ContainerService, caCertificatePath string, caKeyPath string) {
```
colemickens: This seems overly complicated:

1. Why do you length check here instead of at the call site? I think this function should only be called if the args are specified. It seems like it's abusing `len(string)==0` to represent a nil value or a case where the function shouldn't be called. See the other comment below about the flag.

2. `ioutil.Readfile` will handle the error if the file is missing.

3. You could pass a `*string` into the function instead of the entire containerService model and have it set the string value. That way you can un-duplicate the body of this function and just call it twice.

pkg/acsengine/defaults.go
```diff
@@ -102,15 +102,17 @@ func setDefaultCerts(a *api.Properties) (bool, error) {
 		ips = append(ips, net.IP{firstMasterIP[12], firstMasterIP[13], firstMasterIP[14], firstMasterIP[15] + byte(i)})
 	}
```
colemickens: I think this file and the one below could be made more straightforward.

2. TryParse the ca cert and private key here.

3. Pass it as a `*pkiKeyCertPair` into `CreatePki` (much more clear than strings that might be empty) and then `CreatePki` has simpler logic for determining whether to add it's own or not.

This has the nice benefit of removing this code, a little bit lower, which was confusing for a bit:

```diff
+ if len(a.CertificateProfile.CaCertificate) == 0 || len(a.CertificateProfile.GetCAPrivateKey()) == 0 {
+		a.CertificateProfile.CaCertificate = caPair.CertificatePem
+		a.CertificateProfile.SetCAPrivateKey(caPair.PrivateKeyPem)
+	}
```

acs-engine.go
anhowe: The CA secret is not to be stored in the model.  Early on there was sensitivity around this, and that is why we made it not a public property in the API model and it is not deserialized or serialized to the file.  We could revoke the other certs/key, but not the root CA.

colemickens: So what are we doing instead? Always treating and storing apimodel + caCertificate separately?

acs-engine.go
rjtsdl: for 2. i was trying to leave a comment to abstract a helper function for file path check, and retrieve data. Seems ioutil.Reafile already handled it.

pkg/acsengine/defaults.go
anhowe: fixed

anhowe: fixed

acs-engine.go
anhowe: fixed

acs-engine.go
anhowe: apimodel + caKey is stored separately.  I have added comments to clarify this.

pkg/acsengine/defaults.go
```diff
@@ -102,15 +102,27 @@ func setDefaultCerts(a *api.Properties) (bool, error) {
...
 
-	caPair, apiServerPair, clientPair, kubeConfigPair, err := CreatePki(masterExtraFQDNs, ips, DefaultKubernetesClusterDomain)
+	// use the specified Certificate Authority pair, or generate a new pair
+	var caPair *PkiKeyCertPair
+	if len(a.CertificateProfile.CaCertificate) != 0 && len(a.CertificateProfile.GetCAPrivateKey()) != 0 {
+		caPair = &PkiKeyCertPair{CertificatePem: a.CertificateProfile.CaCertificate, PrivateKeyPem: a.CertificateProfile.GetCAPrivateKey()}
+	} else {
+		caCertificate, caPrivateKey, err := createCertificate("ca", nil, nil, false, nil, nil)
```
colemickens: I like it. thanks for making this change.

acs-engine.go
```diff
@@ -147,6 +155,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	if (len(*caCertificatePath) > 0 && len(*caKeyPath) == 0) ||
```
colemickens: Doing this validation here means that it only applies when they are user-supplied flags.

If the user puts in the CA private key, but not the CA certificate, then they will not hit this logic, and instead will wind up in a situation where their private key is ignored because of your logic below: `len(a.CertificateProfile.CaCertificate) != 0 && len(a.CertificateProfile.GetCAPrivateKey()) != 0`.

anhowe: I don't quite understand.  The API model doesn't allow the specification of the CA private key.  The only way to supply the CA private key is through the flags.  How do you mean?

colemickens: Doh. Never mind.

PR #218, opened 2017-01-26
ensure recovery from any panics

anhowe:
A panic may happen during template generation from inner functions.  This change ensures that a panic never kills the process

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -114,15 +114,19 @@ func InitializeTemplateGenerator(classicMode bool) (*TemplateGenerator, error) {
 }
 
 // GenerateTemplate generates the template from the API Model
-func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerService) (string, string, bool, error) {
-	var err error
+func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerService) (templateRaw string, parametersRaw string, certsGenerated bool, err error) {
+	// named return values are used in order to set err in case of a panic
+	templateRaw = ""
+	parametersRaw = ""
+	certsGenerated = false
+	err = nil
+
 	var templ *template.Template
-	certsGenerated := false
 
 	properties := &containerService.Properties
 
 	if certsGenerated, err = SetPropertiesDefaults(properties); err != nil {
-		return "", "", certsGenerated, err
+		return templateRaw, parametersRaw, certsGenerated, err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/acsengine/engine.go
```diff
@@ -135,26 +139,49 @@ func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerServ
...
+			return templateRaw, parametersRaw, certsGenerated, err
 		}
 		if _, err = templ.New(file).Parse(string(bytes)); err != nil {
-			return "", "", certsGenerated, err
+			return templateRaw, parametersRaw, certsGenerated, err
 		}
 	}
+	// template generation may have panics in the called functions.  This catches those panics
+	// and ensures the panic is returned as an error
+	defer func() {
```
colemickens: What exactly is `panic`ing?

pkg/acsengine/engine.go
```diff
@@ -135,26 +139,49 @@ func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerServ
...
+		if r := recover(); r != nil {
+			// find out exactly what the error was and set err
+			switch x := r.(type) {
+			case string:
+				err = errors.New(x)
+			case error:
+				err = x
+			default:
+				err = errors.New("Unknown panic")
```
colemickens: You could replace the entire switch with `fmt.Error("%v", r)` I believe.

pkg/acsengine/engine.go
anhowe: There are a few callback functions that panic in case of bugs.  It is more useful to catch those errors, than hiding them.

pkg/acsengine/engine.go
anhowe: thanks, this simplifies things nicely.

fixed

pkg/acsengine/engine.go
colemickens: Do we have bugs filed for those as well?

I can't totally visualize the effects of this PR without running it in a crashy scenario and observing the difference, so please humor me:

* Before: the panic would happen, we blow up, get a stack trace (and the error?)
* After: panic is suppressed, we still end up fatally exitting, but we see just the err string instead?

Is this right? Does this add extra information, or is it just to hide the scary looking panic stack trace.

anhowe: They are not bugs yet, because we have never hit those panics.

This is added since panics are undesirable in a running service.

colemickens: Got it.

PR #225, opened 2017-01-30
add kubernetes 1.5.1 for windows (without kube-proxy)

anhowe:
This adds Kubernetes 1.5.1 for Windows.

The following will be added in additional iterations:
1. Readme.md - for building windows binaries
2. Readme.md - for example
3. Working Kube-proxy to implement the service network.


parts/kubernetesagentresourcesvmas.t
```diff
@@ -112,7 +112,7 @@
         "osProfile": {
           "adminUsername": "[variables('username')]",
           "computername": "[concat(variables('{{.Name}}VMNamePrefix'), copyIndex())]",
-          {{GetKubernetesAgentCustomData .}}
```
colemickens: Significances of this change?

parts/kubernetesagentvars.t
```diff
@@ -4,6 +4,16 @@
     "{{.Name}}StorageAccountsCount": "[add(div(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')), mod(add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),2), add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),1)))]",
     "{{.Name}}VMNamePrefix": "[concat(variables('orchestratorName'), '-{{.Name}}-', variables('nameSuffix'), '-')]", 
     "{{.Name}}VMSize": "[parameters('{{.Name}}VMSize')]",
+{{if .IsWindows}}
+    "{{.Name}}IPAddressName": "[concat(variables('orchestratorName'), '-agent-ip-', variables('{{.Name}}Index'), '-', variables('nameSuffix'))]",
+    "{{.Name}}LbBackendPoolName": "[concat(variables('orchestratorName'), '-{{.Name}}-', variables('nameSuffix'))]", 
+    "{{.Name}}LbID": "[resourceId('Microsoft.Network/loadBalancers',variables('{{.Name}}LbName'))]",
```
colemickens: Why are we doing any of this? I'd prefer:

1. NAT to the RDP port for the Windows nodes (but if you do this, expose the normal agents over NAT as well...)

2. We continue to not expose agents to the Internet and instead tell people to SSH port forward to reach the Windows nodes.

parts/kuberneteswindowssetup.ps1
```diff
@@ -0,0 +1,299 @@
...
+    [ValidateNotNullOrEmpty()]
+    $MasterIP = "10.240.255.5",
+
+    [parameter()]
+    [ValidateNotNullOrEmpty()]
+    $KubeDnsServiceIp = "10.0.0.10",
+
+    [parameter()]
+    [ValidateNotNullOrEmpty()]
+    $CACertificate = "<<<caCertificate>>>",
```
colemickens: Why this dual parameterization? Either:

1. The script takes these as args, we should just pass them as args.

2. The script is only used by us and we interpolate directly in. In this case, don't make them parameters, and make the other variables work the same.

parts/kuberneteswindowssetup.ps1
```diff
@@ -0,0 +1,299 @@
...
+
+    [parameter(Mandatory=$true)]
+    [ValidateNotNullOrEmpty()]
+    $AzureHostname
+)
+
+$global:DockerServiceName = "Docker"
+$global:RRASServiceName = "RemoteAccess"
+$global:KubeDir = "c:\k"
+$global:KubeBinariesSASURL = "https://acsengine.blob.core.windows.net/v1-5-1/k.zip?st=2017-01-30T15%3A47%3A00Z&se=2020-01-31T15%3A47%3A00Z&sp=rl&sv=2015-12-11&sr=b&sig=C9ho0Vl%2F1ute7d2mAs%2BiVM43%2BQfhEITk%2Fg2oZv5r66Y%3D"
```
colemickens: Why? Maybe they should be versioned but no reason to require SAS.

parts/kuberneteswindowssetup.ps1
```diff
@@ -0,0 +1,299 @@
...
+New-InfraContainer()
+{
+    cd $global:KubeDir
+    docker build -t kubletwin/pause . 
+}
+
+function
+Get-PodCIDR
+{
+    $argList = @("--hostname-override=$AzureHostname","--pod-infra-container-image=kubletwin/pause","--resolv-conf=""""","--api-servers=https://${MasterIP}:443","--kubeconfig=c:\k\config")
```
colemickens: Is `--hostname-override` really needed? Is this because the Windows hostname != the actual name of the VM in Azure?

parts/kuberneteswindowssetup.ps1
```diff
@@ -0,0 +1,299 @@
...
+    $process = Start-Process -FilePath c:\k\kubelet.exe -PassThru -ArgumentList $argList
+
+    $podCidrDiscovered=$false
+    $podCIDR=""
+    # run kubelet until podCidr is discovered
+    while (-not $podCidrDiscovered)
```
colemickens: Log a message here so the user knows that we're waiting on cidr. Otherwise looks like it will just silently loop/hang here.

parts/kuberneteswindowssetup.ps1
```diff
@@ -0,0 +1,299 @@
...
+    }
+    else 
+    {
+        # keep for debugging purposes
+        Write-Log ".\CustomDataSetupScript.ps1 -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp -MasterFQDNPrefix $MasterFQDNPrefix -Location $Location -CACertificate $CACertificate -AgentCertificate $AgentCertificate -AgentKey $AgentKey -AzureHostname $AzureHostname"
+    }
+}
+catch
+{
+    Write-Error $_
```
colemickens: Do you want this script to pass as successful in this case? Might want to rethrow the exception and/or fail explicitly?

pkg/acsengine/engine.go
```diff
@@ -431,6 +433,18 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
...
 			return fmt.Sprintf("\"customData\": \"%s\"", str)
 		},
+		"GetKubernetesWindowsAgentCustomData": func() string {
+			b, err := Asset(kubernetesWindowsAgentCustomDataPS1)
+			if err != nil {
+				// this should never happen and this is a bug
+				panic(err.Error())
+			}
+			str := string(b)
+			str = strings.Replace(str, "<<<caCertificate>>>", base64.StdEncoding.EncodeToString([]byte(properties.CertificateProfile.CaCertificate)), -1)
```
colemickens: Why? Why are we adding these in customdata when we normally use customscript for this?

Where is the private key?

In general, this is a good example of a place where a new Go template function is really not needed. Whatever we handle the template here should have enough data that it can just splat out the CA Certificate, instead of needing to call a custom Go template function. In some sense... you're basically doing another layer of templating here when you fill in those `<<<placeholders>>>`.

pkg/acsengine/testdata/windows/kubernetes_expected_params.json
```diff
@@ -0,0 +1,68 @@
...
+  "clientPrivateKey": {
+    "value": "Y2xpZW50UHJpdmF0ZUtleQ=="
+  },
+  "firstConsecutiveStaticIP": {
+    "value": "10.240.255.5"
+  },
+  "kubeConfigCertificate": {
+    "value": "a3ViZUNvbmZpZ0NlcnRpZmljYXRl"
+  },
+  "kubeConfigPrivateKey": {
```
colemickens: These are not real, right?

anhowe: fixed

PR #226, opened 2017-01-30
Improve testing basics; add simple deploy script

colemickens:
This PR does a few things:

1. Cleanup `test/common.sh`. `common.sh` contains the guts of the deployment process and a couple helpers.

2. Create `test/deploy.sh` (or rather re-write `deploy-k8s.sh` to be more generic).

   `deploy.sh` can now be used in three ways (Which are better documented in code, a README-modification will be added later):

   1. create a specific named cluster
   2. create a timestamp-suffixed cluster quickly
   3. used from Prow (or some other CI system) - via environment variables

3. Add a script that can be used from a CI system to test a PR ref.

This PR is needed for subsequent PRs that add testing scripts for dcos/kubernetes/swarmmode.


test/common.sh
```diff
@@ -13,78 +13,79 @@ DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
...
 	[[ ! -z "${LOCATION:-}" ]] || (echo "Must specify LOCATION" && exit -1)
 	[[ ! -z "${CLUSTER_DEFINITION:-}" ]] || (echo "Must specify CLUSTER_DEFINITION" && exit -1)
-        which az || (echo "az must be on PATH" && exit -1)
-
-	if [[ "${cluster_type}" == "kubernetes" ]]; then
-            [[ ! -z "${SERVICE_PRINCIPAL_CLIENT_ID:-}" ]] || (echo "Must specify SERVICE_PRINCIPAL_CLIENT_ID" && exit -1)
-	    [[ ! -z "${SERVICE_PRINCIPAL_CLIENT_SECRET:-}" ]] || (echo "Must specify SERVICE_PRINCIPAL_CLIENT_SECRET" && exit -1)
-	    which kubectl || (echo "kubectl must be on PATH" && exit -1)
-	fi
+	[[ ! -z "${SUBSCRIPTION_ID:-}" ]] || (echo "Must specify SUBSCRIPTION_ID" && exit -1)
```
anhowe: turn off logging (set +x), so that secrets and other mbi data like sub id is not printed to traces.  And keep that as a practice for other secret and mbi like data throughout the scripts.

test/user.env.example
```diff
@@ -0,0 +1,10 @@
+export INSTANCE_NAME_PREFIX=colemick
+export LOCATION=westus2
+export SUBSCRIPTION_ID=23d59db5-7c0e-4825-b738-5859cec6b1a1
```
anhowe: are these real at all?

colemickens: Nope.

test/common.sh
colemickens: They're printed regardless as part of the outputs of various commands - I had originally done what you suggest, but it isn't sufficient.

For now the CI system must be locked down. I have a follow-up to find a better longer-term solution. The better solution will be to wrap these scripts and manually `sed` out things that we know are sensitive. I plan to do this soon, as I want the output to be something we can have users send us, etc for debugging.

PR #227, opened 2017-01-30
Add basic Swarm Mode validation script

colemickens:
Note, this stacks on top of #226. Please only review the latest commit here.

This adds a basic Docker Swarm Mode validation script.

cc: @amanohar

### DIFF
--- a/test/cluster-tests/swarmmode/test.sh
+++ b/test/cluster-tests/swarmmode/test.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -x
+set -e
+set -u
+
+remote_exec="ssh -i ${SSH_KEY} -o StrictHostKeyChecking=no -p2200 azureuser@${INSTANCE_NAME}.${LOCATION}.cloudapp.azure.com"
+
+function teardown {
+  ${remote_exec} docker service rm nginx || true
+  sleep 10
+  ${remote_exec} docker network rm network || true
+}
+
+trap teardown EXIT
+
+# TODO: investigate if this is really needed - replace with deterministic wait or retry loop
+sleep 60
+
+${remote_exec} docker network create \
+	--driver overlay \
+	--subnet 10.0.9.0/24 \
+	--opt encrypted \
+	network
+
+${remote_exec} docker service create \
+	--replicas 3 \
+	--name nginx \
+	--network network \
+	--publish 80:80 \
+	nginx
+
+sleep 10
+
+wait=5
+count=12
+success="n"
+while (( $count > 0 )); do
+  curl --fail "http://${INSTANCE_NAME}0.${LOCATION}.cloudapp.azure.com:80/"
+  if [[ $? == 0 ]]; then
+    success="y"
+    break
+  fi
+done
+if [[ "${success}" != "y" ]]; then
+  echo "gave up waiting for service to be externally reachable"; exit -1
+fi
--- a/examples/swarm.json
+++ b/examples/swarm.json
@@ -11,7 +11,7 @@
     },
     "agentPoolProfiles": [
       {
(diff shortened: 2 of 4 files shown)


test/cluster-tests/swarmmode/test.sh
```diff
@@ -0,0 +1,34 @@
...
+	--replicas 3 \
+	--name nginx \
+	--network network \
+	--publish 80:80 \
+	nginx
+
+sleep 10
+
+# only publicagent pool is exposed to internet
+wget http://${INSTANCE_NAME}0.${LOCATION}.cloudapp.azure.com:80/
```
anhowe: add retries over at least 60 seconds?

test/bootstrap/checkout-pr.sh
```diff
@@ -0,0 +1,27 @@
...
+set -eu -o pipefail
+set -x
+
+git init .
+git clean -dfx
+git reset --hard
+git config --local user.name 'ACS Bot'
+git config --local user.email 'acs-bot@microsoft.com'
+git fetch --tags https://github.com/${REPO_OWNER}/${REPO_NAME} master +refs/pull/${PULL_NUMBER}/head:refs/pr/${PULL_NUMBER}
+git checkout -B test "${PULL_BASE_SHA}"
```
amanohar: Would this be an issue if the bot tries to run tests for 2 PRs at the same time if the branch name is always test?

test/cluster-tests/swarmmode/test.sh
```diff
@@ -0,0 +1,34 @@
...
+	network
+
+${remote_exec} docker service create \
+	--replicas 3 \
+	--name nginx \
+	--network network \
+	--publish 80:80 \
+	nginx
+
+sleep 10
```
amanohar: Sometimes it takes longer for the service to come up. I recommend either sleeping longer or adding retries.

test/user.env.example
```diff
@@ -0,0 +1,10 @@
+export INSTANCE_NAME_PREFIX=colemick
+export LOCATION=westus2
```
amanohar: Is it better to use region instead of location since that's what Azure calls it?

test/deploy.sh
```diff
@@ -0,0 +1,89 @@
...
+###############################################################################
+
+set -e
+set -u
+set -o pipefail
+
+ROOT="${DIR}/.."
+source "${ROOT}/test/common.sh"
+
+function cleanup {
```
amanohar: Should there be some kind of retry in here in case first cleanup attempt fails for some unrelated reason?

test/user.env.example
colemickens: Maybe, but it's sort of become standard, I see it used in a loooot of places - I'd prefer to keep with LOCATION for consistency's sake.

test/cluster-tests/swarmmode/test.sh
colemickens: Ack. Will make this loop, same per Anthony's comment.

test/bootstrap/checkout-pr.sh
colemickens: They get separate workspaces. (In fact, in the system I showed you, each build job happens in a new Kubernetes Pod - clean slate everytime, etc).

test/cluster-tests/swarmmode/test.sh
colemickens: Will do.

test/deploy.sh
colemickens: Maybe, going to punt for now as I have enough of these bulky retry loops in here. If I add a better `retry` helper function to common, I could use it here as well and will do so.

PR #230, opened 2017-01-30
Configuring docker live restore mode

kim0:
As per https://docs.docker.com/engine/admin/live-restore/ this enables live-restore mode for docker, ensuring pods don't die when a docker upgrade hits the wire

### DIFF
--- a/parts/kubernetesagentcustomdata.yml
+++ b/parts/kubernetesagentcustomdata.yml
@@ -31,6 +31,14 @@ write_files:
     ExecStart=
     ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay
 
+- path: "/etc/docker/daemon.json"
+  permissions: "0644"
+  owner: "root"
+  content: |
+    {
+      "live-restore": true
+    }
+
 - path: "/etc/kubernetes/certs/ca.crt"
   permissions: "0644"
   encoding: "base64"
--- a/parts/kubernetesmastercustomdata.yml
+++ b/parts/kubernetesmastercustomdata.yml
@@ -37,6 +37,14 @@ write_files:
     ExecStart=
     ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay
 
+- path: "/etc/docker/daemon.json"
+  permissions: "0644"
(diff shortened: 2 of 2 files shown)


kim0:
@msftclas I already signed the cla ? How can I convince you to remove the tag :)

colemickens:
Did you just recently sign in? I thought it used to have a way to re-trigger without needing to close/reopen...

colemickens:
What cases does this help? Are you saying that unattended upgrades will actually restart the docker daemon rather than waiting for a reboot?

kim0:
@colemickens .. Yes I just signed the CLA after pushing this PR .. I guess it didn't feel it

This helps in the following way. Assuming a docker update (security?) comes out, and you apply it. I believe apt will restart docker daemon for you. If it doesn't I would still restart it to really apply the update. Currently restarting the docker daemon kills all running containers/pods (sigh!). With this live-restore option, running pods are not killed. Instead they are swiftly handed over to the new docker daemon version.

colemickens:
@kim0 Does this option actually change the behavior in the reboot case - I would assume they're not checkpointing/restoring on reboot?

I don't want to take this unless someone can point me to docs that say `apt` is going to auto-restart docker. That sounds a bit unexpected to me.

kim0:
@colemickens .. This mode changes nothing in the reboot case. A reboot is a reboot, and everything comes up fresh as usual. This mode does not do persistent checkpointing/restoring of containers, instead it only refers to passing of control of a running container from old dockerd to newer dockerd without killing the running container. 

From my memory, I remembered upgrading docker-engine did kill running containers (which is why I always use that option on my servers, and which is why docker folks created the live-restore to begin with) .. However since you asked, I spun up a VM to verify. I found that it does kill running containers upon docker-engine upgrade. A full log is attached

[dockerupgrade.log.txt](https://github.com/Azure/acs-engine/files/740793/dockerupgrade.log.txt)

colemickens:
@kim0 Thank you very much for digging in and showing my assumption to be wrong.

I'm no longer opposed to this and am okay with the change. I will Approve this now and merge tomorrow if there are no objections.

Thanks!

PR #231, opened 2017-01-31
Remove deprecated --reconcile-cidr, add explicit --storage-backend, bump addons-manager.

brendandburns:
@colemickens

### DIFF
--- a/parts/kubernetesagentcustomdata.yml
+++ b/parts/kubernetesagentcustomdata.yml
@@ -116,7 +116,6 @@ write_files:
             --cloud-config=/etc/kubernetes/azure.json \
             --hairpin-mode=promiscuous-bridge \
             --network-plugin=kubenet \
-            --reconcile-cidr \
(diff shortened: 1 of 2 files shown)


colemickens:
When do you think we need to move to `etcd3`, @brendandburns ?

brendandburns:
@anhowe what is the best way to automatically update those files?

@colemickens we should bump to etcd3 when we roll to 1.6 (where it is the default)

anhowe:
Run "go test" under the pkg/acsengine directory.  From there diff the expected vs. err files, and then diff each file, and mv .err file to expected file.

Also, ping me, and I can run this quickly.

brendandburns:
@anhowe sure, go ahead and update the tests, thanks!

(I tried running, but I see lots of non-k8s diffs, so I must have some problem in my setup...)

anhowe:
@brendandburns - test files are updated and merged

PR #233, opened 2017-02-01
etcd clustering template changes

raveeram:


parts/kubernetesmastercustomdata.yml
```diff
@@ -117,7 +117,7 @@ write_files:
               - "--cloud-provider=azure"
               - "--cloud-config=/etc/kubernetes/azure.json"
               - "--service-cluster-ip-range={{{kubeServiceCidr}}}"
-              - "--etcd-servers=http://127.0.0.1:4001"
+              - "--etcd-servers=http://127.0.0.1:{{{masterEtcdClientPort}}}"
```
colemickens: Does `apiserver` not also need to `leader-elect=true`?

parts/kubernetesmastercustomdata.yml
```diff
@@ -8,7 +8,7 @@ packages:
  - traceroute
 
 runcmd:
-- /bin/echo DAEMON_ARGS=--advertise-client-urls "http://127.0.0.1:2379,http://{{{masterPrivateIp}}}:2379" --listen-client-urls "http://0.0.0.0:2379,http://0.0.0.0:4001" | tee -a /etc/default/etcd
+- /bin/bash -c "/bin/echo DAEMON_ARGS=--name ""<<<variables('masterVMNames')[copyIndex()]>>>"" --initial-advertise-peer-urls ""<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>"" --listen-peer-urls ""<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>"" --advertise-client-urls ""<<<variables('masterEtcdClientURLs')[copyIndex()]>>>"" --listen-client-urls ""<<<concat(variables('masterEtcdClientURLs')[copyIndex()], ',http://127.0.0.1:', variables('masterEtcdClientPort'))>>>"" --initial-cluster-token ""k8s-etcd-cluster"" --initial-cluster ""<<<variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]>>>"" --initial-cluster-state ""new"" | tee -a /etc/default/etcd"
```
colemickens: Where did we end up on the whole double quote thing here? At the very least, I'd prefer they were all removed if they end up removing in the final file. Even better would be escaping them normally:

```
runcmd:
- /bin/bash -c "/bin/echo DAEMON_ARGS=--name \"<<<variables('masterVMNames')[copyIndex()]>>>\" --initial-advertise-peer-urls \"<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>\" --listen-peer-urls \"<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>\" --advertise-client-urls \"<<<variables('masterEtcdClientURLs')[copyIndex()]>>>\" --listen-client-urls \"<<<concat(variables('masterEtcdClientURLs')[copyIndex()], ',http://127.0.0.1:', variables('masterEtcdClientPort'))>>>\" --initial-cluster-token \"k8s-etcd-cluster\" --initial-cluster \"<<<variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]>>>\" --initial-cluster-state \"new\" | tee -a /etc/default/etcd"
```

raveeram: Ah, I"ll validate that next and see if I can get rid of these. Will update.

parts/kubernetesmastercustomdata.yml
raveeram: api-server talks to the local etcd instance which in-turn takes of replication.

parts/kubernetesmastervars.t
```diff
@@ -71,6 +71,47 @@
     "masterFirstAddrOctet4": "[variables('masterFirstAddrOctets')[3]]",
     "masterFirstAddrPrefix": "[concat(variables('masterFirstAddrOctets')[0],'.',variables('masterFirstAddrOctets')[1],'.',variables('masterFirstAddrOctets')[2],'.')]",
     "masterVMNamePrefix": "[concat(variables('orchestratorName'), '-master-', variables('nameSuffix'), '-')]",
+    "masterVMNames": [
+      "[concat(variables('masterVMNamePrefix'), '0')]",
+      "[concat(variables('masterVMNamePrefix'), '1')]",
+      "[concat(variables('masterVMNamePrefix'), '2')]",
+      "[concat(variables('masterVMNamePrefix'), '3')]",
```
anhowe: let's keep it to 5 machines, and then only allow 1,3,5 combinations to match the other orchestrators.

colemickens:
Do we need conditional logic in here to support single masters? Thinking of the etcd args in particular.

parts/kubernetesmastervars.t
raveeram: Thanks, will update.

raveeram:
Cole and I chatted offline regarding his comment above. No action needed.

PR #234, opened 2017-02-01
Add instructions to build 'acs-engine' when using Docker

dluc:


colemickens:
Thanks for this!

Would you be willing to take your new section and instead replace the Linux steps, and then point the Docker guide to the Linux steps. I don't want to have those steps duplicating/diverging since the build process once inside the container is identical to on Linux.

Sound reasonable?

PR #240, opened 2017-02-03
fix expression length issue

anhowe:


pkg/acsengine/engine.go
```diff
@@ -486,12 +483,11 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 			if e != nil {
 				return ""
 			}
-			// add the master provisioning script
-			masterProvisionB64GzipStr := getBase64CustomScript(kubernetesMasterCustomScript)
-			str = strings.Replace(str, "MASTER_PROVISION_B64_GZIP_STR", masterProvisionB64GzipStr, -1)
-
 			return fmt.Sprintf("\"customData\": \"[base64(concat('%s'))]\",", str)
 		},
+		"GetKubernetesB64Provision": func() string {
```
colemickens: Why was a new custom template function added instead of using something like ` {{ readFile masterProvision | b64 | gzip }}`. We really should cut back on adding these custom functions when builtins will work.

karataliu:
`MASTER_PROVISION_B64_GZIP_STR` has no reference now, consider removing?

https://github.com/Azure/acs-engine/blob/0330677c0d5c6d6109377c353e89a96b4b0ade43/pkg/acsengine/engine.go#L93

seanknox:
@karataliu Refactoring/code cleanup is always welcome—feel free to open a PR!

PR #241, opened 2017-02-03
switch 'app' to 'k8s-app' to follow convention.

brendandburns:
@colemickens @anhowe


colemickens:
~~Dashboard is still using `app`: https://github.com/kubernetes/dashboard/blob/master/src/deploy/kubernetes-dashboard.yaml~~ but I guess in the kubernetes tree it's using `k8s-app`: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-service.yaml

(In this same light, I thought there was an issue or PR about changing how `kubernetes.io/cluster-service` works with addon-manager but I can't find it anywhere.)

brendandburns:
Yeah, there is actually an e2e test for this now:

https://github.com/kubernetes/kubernetes/blob/61b7b3fb660d09917af0b843c1b151279ea574ef/test/e2e/dashboard.go#L51

PR #243, opened 2017-02-04
(experimental) Enable Swarm Mode to support Windows agents

amanohar:


amanohar:
cc: @anhowe

parts/Join-SwarmMode-cluster.ps1
```diff
@@ -0,0 +1,335 @@
...
+    $currentRetry = 0;
+    $success = $false;
+    $getTokenCommand = "docker -H $($SwarmMasterIP):2375 swarm join-token -q worker"
+    $swarmmodetoken;
+
+    do {
+        try
+        {
+            Write-Log "Executing [$getTokenCommand] command...."
+            <#& $swarmmodetoken#>
```
anhowe: what does this do?

amanohar: This command gets the token for the worker from the master. Then in the last command at the end of this function uses that token join the Swarm.

PR #247, opened 2017-02-07
K8s: Add an internal load balancer for Agent->Master communication

raveeram:
- Also enable 3 and 5 master clusters


parts/kubernetesmastercustomdata.yml
```diff
@@ -308,6 +308,7 @@ write_files:
     [Service]
     Restart=always
     ExecStartPre=/bin/mkdir -p /var/lib/kubelet
+    ExecStartPre=/bin/bash -c "iptables -t nat -A PREROUTING -p tcp --dport 4443 -j REDIRECT --to-port 443"
```
anhowe: add a comment above to what this is doing and why

parts/kubernetesagentcustomdata.yml
```diff
@@ -70,7 +70,7 @@ write_files:
     - name: localcluster
       cluster:
         certificate-authority: /etc/kubernetes/certs/ca.crt
-        server: https://{{{masterPrivateIp}}}:443
+        server: https://{{{masterInternalLbIp}}}:443
```
anhowe: did you search for all occurences of "masterPrivateIP" in the Yaml and templates file?  I don't see a change to the windows agents anywhere in here.

parts/kubernetesmasterresources.t
```diff
@@ -171,6 +171,63 @@
     },
     {
       "apiVersion": "[variables('apiVersionDefault')]",
+      "dependsOn": [
+       "[variables('vnetID')]"
```
anhowe: This will not work for custom VNET, because "vnetID" does not exist.

anhowe: You will need something like:
{{if .MasterProfile.IsCustomVNET}}
         "[variables('nsgID')]",		         "[variables('nsgID')]",
 {{else}}		 {{else}}
          "[variables('vnetID')]",		          "[variables('vnetID')]",
  {{end}}		  {{end}}

colemickens:
Looks okay other than the things Anthony mentioned. Ping me for another review when you push the rest of the changes.

parts/kubernetesmastercustomdata.yml
raveeram: Updated.

parts/kubernetesmasterresources.t
raveeram: Thanks, updated.

parts/kubernetesagentcustomdata.yml
raveeram: Updated winagent as well.

parts/kubernetesmastercustomdata.yml
```diff
@@ -308,6 +310,10 @@ write_files:
     [Service]
     Restart=always
     ExecStartPre=/bin/mkdir -p /var/lib/kubelet
+    # Azure does not support two LoadBalancers(LB) sharing the same nic and backend port.
+    # As a workaround, the Internal LB forwards apiserver traffic from port 4443 to 443 while the
```
colemickens: I don't think the comment is accurate ? (or else I'm reading it wrong...)

I thought it went like this: `ILB:443 -> Node:4443 -> iptables -> internal:443`

pkg/acsengine/defaults.go
```diff
@@ -98,8 +98,10 @@ func setDefaultCerts(a *api.Properties) (bool, error) {
 
 	ips := []net.IP{firstMasterIP}
 
-	for i := 1; i < a.MasterProfile.Count; i++ {
-		ips = append(ips, net.IP{firstMasterIP[12], firstMasterIP[13], firstMasterIP[14], firstMasterIP[15] + byte(i)})
+	// Include the Internal load balancer as well
+	for i := 1; i < (a.MasterProfile.Count + 1); i++ {
+		ip := net.IP{firstMasterIP[12], firstMasterIP[13], firstMasterIP[14], firstMasterIP[15] + byte(i)}
```
colemickens: Can you remind me what's going on here? `firstMasterIP[12] == first octet` ? When named `firstMasterIP` and why the indexes are `12,13,14,15+i` ?

I realize you're just adapting the existing code, but can you tell me just for my knowledge? Thanks.

parts/kuberneteswinagentresourcesvmas.t
```diff
@@ -227,7 +227,7 @@
...
         "protectedSettings": {
-          "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('masterPrivateIp'),' -KubeDnsServiceIp ',variables('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -AgentKey ',variables('clientPrivateKey'),' -AzureHostname ',variables('{{.Name}}VMNamePrefix'),copyIndex(),variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]"
+          "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('masterInternalLbIp'),' -KubeDnsServiceIp ',variables('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -AgentKey ',variables('clientPrivateKey'),' -AzureHostname ',variables('{{.Name}}VMNamePrefix'),copyIndex(),variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]"
```
colemickens: (I really dislike that GitHub doesn't do the darker sub-line highlight for the actual diff like VSTS.) What changed here?

raveeram: Ah, variables('masterPrivateIp') -> variables('masterInternalLbIp')

parts/kubernetesmastercustomdata.yml
raveeram: You're right. The second line could be clearer.

parts/kuberneteswinagentresourcesvmas.t
colemickens: 👍

pkg/acsengine/defaults.go
raveeram: Ah no, this is the fourth octet. IP is a v4/v6 agnostic 16-byte slice.

colemickens: I'm still stuck on what I presume is my own confusion...  `net.IP` is a type alias for `[]byte`, yeah?

It SEEMS like the code here would be doing the following. At least, from eye-balling it, since it seems to work:

```
ip[0] = firstMasterIP[12]
ip[1] = firstMasterIP[13]
ip[2] = firstMasterIP[14]
ip[3] = firstMasterIP[15] + byte(i)
```

Since you're using the array literal initializer syntax.

colemickens: So, it looks like the stdlib tries to be smart here and all of the functions internally check for len(ip)==4. But then why is it `firstMasterIP[12]` and not `firstMaster[0]`? Still stuck there...

colemickens: Okay. So, my guess is... the code that constructs `firstMasterIP` is using `IPv4()` constructor which puts the ipv4 octets in 12,13,14,15.

So, that's what we should do here. Can you change it to use the IPv4 constructor? (I still have no idea how this is working now I guess...)

colemickens: Okay, one last comment from me... here's how I'd recommend writing it:

```
masterIP4, err := firstMasterIP.To4()
if err != nil { whatever }
ip := net.IPv4(masterIP[0], masterIP4[1], masterIP4[2], masterIP4[3] + byte(i))
```

This no longer requires understanding the `net.IP` internals.

raveeram: Right, we're bumping up the 4th octet. Yeah, considering the way ParseIP is written: https://golang.org/src/net/ip.go?h=ParseIP#L648 (ParseIPv4 isn't exposed), net/http doesn't want to distinguish between the two at the type level i guess. We could explicitly use the IPv4, i don't feel strongly one way or the other.

PR #250, opened 2017-02-07
added multi-regional acs-engine tests

dmitsh:


test/deploy.sh
```diff
@@ -40,7 +39,7 @@ function cleanup {
 #
 # Prow:
 #   export PULL_NUMBER=...
-#   export VALIDATE=true
+#   export VALIDATE=<script path>
 #   export CLUSTER_DEFIITION=examples/kubernetes.json
 #   export CLUSTER_TYPE=kubernetes
```
colemickens: remove this, it's not used.

test/deploy.sh
```diff
@@ -69,16 +68,25 @@ else
...
-deploy
+deploy_template
 
 if [[ -z "${VALIDATE:-}" ]]; then
 	exit 0
 fi
 
-export SSH_KEY="${ROOT}/_output/${INSTANCE_NAME}/id_rsa"
-export KUBECONFIG="${ROOT}/_output/${INSTANCE_NAME}/kubeconfig/kubeconfig.${LOCATION}.json"
+export SSH_KEY="${OUTPUT}/id_rsa"
```
colemickens: It feels weird that we export these here and then also export them in `./test/step.sh`. If we want to split this script into smaller bites, then let's go ahead and yank out the call to validate here. (FWIW, in my opinion, you can make breaking changes to this -- I'm the only one who seems to be using it)

test/common.sh
```diff
@@ -72,20 +76,28 @@ function deploy() {
...
 		--template-file "${OUTPUT}/azuredeploy.json" \
 		--parameters "@${OUTPUT}/azuredeploy.parameters.json"
-
-	echo "${INSTANCE_NAME} files -> ${OUTPUT}"
 }
 
 function cleanup() {
-	az group delete --no-wait --name="${INSTANCE_NAME}" || true
+	echo "Cleanup"
+	az group delete --no-wait --name="${RESOURCE_GROUP}" --force || true
```
colemickens: Is `--force` actually live yet? Oh, maybe it's in `b3` and I haven't noticed because I'm using a tagged old version? Can you confirm the `--force` works and isn't being silently missed due to the `|| true`?

test/common.sh
```diff
@@ -72,20 +76,28 @@ function deploy() {
 		--tenant "${TENANT_ID}" &>/dev/null
 
 	az account set --subscription "${SUBSCRIPTION_ID}"
+}
+
+function deploy_template() {
+	# Check pre-requisites
+	[[ ! -z "${DEPLOYMENT_NAME:-}" ]] || (echo "Must specify DEPLOYMENT_NAME" && exit -1)
```
colemickens: Can we consolidate on `INSTANCE_NAME` or `DEPLOYMENT_NAME`? For 99.999% of cases, they're likely to be the same, no?

test/common.sh
```diff
@@ -18,20 +18,14 @@ ROOT="${DIR}/.."
...
 	[[ ! -z "${INSTANCE_NAME:-}" ]] || (echo "Must specify INSTANCE_NAME" && exit -1)
-	[[ ! -z "${LOCATION:-}" ]] || (echo "Must specify LOCATION" && exit -1)
 	[[ ! -z "${CLUSTER_DEFINITION:-}" ]] || (echo "Must specify CLUSTER_DEFINITION" && exit -1)
-	[[ ! -z "${SUBSCRIPTION_ID:-}" ]] || (echo "Must specify SUBSCRIPTION_ID" && exit -1)
-	[[ ! -z "${TENANT_ID:-}" ]] || (echo "Must specify TENANT_ID" && exit -1)
-	[[ ! -z "${SERVICE_PRINCIPAL_CLIENT_ID:-}" ]] || (echo "Must specify SERVICE_PRINCIPAL_CLIENT_ID" && exit -1)
-	[[ ! -z "${SERVICE_PRINCIPAL_CLIENT_SECRET:-}" ]] || (echo "Must specify SERVICE_PRINCIPAL_CLIENT_SECRET" && exit -1)
-	which kubectl || (echo "kubectl must be on PATH" && exit -1)
-	which az || (echo "az must be on PATH" && exit -1)
+#	[[ ! -z "${SERVICE_PRINCIPAL_CLIENT_ID:-}" ]] || (echo "Must specify SERVICE_PRINCIPAL_CLIENT_ID" && exit -1)
```
colemickens: Where did this end up? Commenting it out breaks k8s. Leaving it in does not hurt dcos/swarm...

test/acsengine.deploy
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env groovy
```
colemickens: Why is this a new file rather than modifying the existing pipeline? Or, why not remove the old one if this is the way forward?

test/acsengine.deploy
```diff
@@ -0,0 +1,101 @@
...
+
+node {
+  withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'AZURE_CLI_SPN_ACS_TEST',
```
colemickens: Might want to re-use the existing SP credentials ID rather than adding a new one. If not, we should cleanup the old one when we remove the old job.

test/acsengine.deploy
```diff
@@ -0,0 +1,101 @@
...
+            // Build and Test
+            stage('Build') {
+              sh('make ci')
+            }
+
+            stage('Create templates') {
+              sh("echo \"export INSTANCE_NAME=acs-test\$(printf \"%x\" \$(date '+%s'))\" > test/user.env")
+              sh("echo \"export CLUSTER_DEFINITION=examples/${ORCHESTRATOR}.json\" >> test/user.env")
+              sh("echo \"export CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID=${SERVICE_PRINCIPAL_CLIENT_ID}\" >> test/user.env")
+              sh("echo \"export CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET}\" >> test/user.env")
```
colemickens: Hm, you should be able to just export these as env vars, rather than explicitly writing out user.env. Maybe that's desirable for others reasons though?

(Historical context: I wrote the `user.env` path simply with the intention of it being used for local wip dev scenarios.)

test/step.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
colemickens: This script seems to just call `${ROOT}/test/$@.sh`. Any reason for it being an explicit script instead of callers just invoking `./test/step.sh <stepname>`?

test/acsengine.deploy
```diff
@@ -0,0 +1,101 @@
...
+          locations_str = "\
+australiaeast australiasoutheast \
+brazilsouth \
+canadacentral canadaeast \
+centralindia southindia westindia\
+centralus eastus2 eastus northcentralus southcentralus westcentralus westus2 westus \
+eastasia southeastasia \
+japaneast japanwest \
+northeurope westeurope \
+uksouth ukwest"
```
amanohar: Make to run tests against at least all these regions. Technically these are ACS regions but its good run ACS Engine tests in these regions. Especially US and Europe regions as they tend to get a lot more traffic.

australiasoutheast
northeurope       
brazilsouth       
australiaeast     
japaneast         
northcentralus    
westus            
eastasia          
eastus2           
southcentralus    
southeastasia     
eastus            
westeurope        
Centralus
japanwest

test/acsengine.deploy
amanohar: Does the test runner provide credentials at the beginning of a test run? If not how are the credentials retrieved?

test/step.sh
```diff
@@ -0,0 +1,48 @@
...
+;;
+
+deploy_template)
+#  trap cleanup EXIT
+  deploy_template
+;;
+
+verify)
+  export SSH_KEY="${OUTPUT}/id_rsa"
+  export KUBECONFIG="${OUTPUT}/kubeconfig/kubeconfig.${LOCATION}.json"
```
amanohar: I might be missing something but isn't this step K8s specific. If yes then what happens in case of other orchestrators at this step?

test/common.sh
JackQuincy: I've got it when i updated my docker image today

dmitsh: Without "--force" it expects user's confirmation on removing the resource group.

test/common.sh
dmitsh: Certainly. The reason I didn't do that is I wanted to preserve deploy.sh

colemickens: I figured that might be the case. I very much appreciate it, but I'm more than happy to adjust stuff on my end if you're going to cleanup how this is organized 😄.

test/common.sh
colemickens: Yup, I'm now seeing this as well. My comment here can be ignored.

test/step.sh
```diff
@@ -32,17 +27,19 @@ generate_template)
...
 
 deploy_template)
-#  trap cleanup EXIT
   deploy_template
 ;;
 
 verify)
-  export SSH_KEY="${OUTPUT}/id_rsa"
-  export KUBECONFIG="${OUTPUT}/kubeconfig/kubeconfig.${LOCATION}.json"
+  if [ ${ORCHESTRATOR} = "kubernetes" ]; then
```
colemickens: Pretty sure you need SSH_KEY for the other orchestrators.

PR #252, opened 2017-02-09
Enable ACR + test for ACR integration

colemickens:
This PR:
* fixes a small bug from the test refactoring regarding AZURE_CONFIG_DIR (cc: @dmitsh)
* adds a test for ACR integration - now we do a smoke test for external load balancer provisioning, and we also create an ACR registry and ensure that an image pushed to it is able to start in the cluster without manually injecting any credentials.
* adds support for ACR integration in Kubernetes by enabling the required flags on `kubelet`.

Note, this PR needs to wait for `v1.5.3` but I needed to do the work anyway to validate  https://github.com/kubernetes/kubernetes/pull/41208 so I decided to go ahead and automate it to prevent regressions.

cc: @brendandburns


colemickens:
Note to self: this needs to be modified to only do ACR-verification in regions where ACR is live.

colemickens:
CC: @dmitsh Now that we are on 1.5.3, I'd like to get this turned on... but there are implications for testing.

We only want to run the ACR verification steps on regions where ACR is enabled. For now, I will hack this and just check against a hard-coded list of regions where I know that ACR is available, but we probably want to think about how to handle and organize these "scenario" tests in ACS-Engine.

colemickens:
@anhowe @dmitsh I dismissed your reviews so that I can get a new one.

PTAL at the second to last commit "test: k8s: only test ACR in regions where it is available".

I've tested it with `westus` and `westus2` - works as expected.

parts/kubernetesagentcustomdata.yml
```diff
@@ -123,6 +123,7 @@ write_files:
             --cluster-domain=cluster.local \
             --cloud-provider=azure \
             --cloud-config=/etc/kubernetes/azure.json \
+            --azure-container-registry-config=/etc/kubernetes/azure.json \
```
anhowe: should this be added to the windows powershell (function Write-KubernetesStartFiles)

colemickens: Absolutely it should. New commits pushed with this fixed.

colemickens:
Good thing I tested this... `/etc/kubernetes/azure.json` sure enough doesn't exist on a Windows box.

Worse, there appears to be no azure config file dropped on the Windows boxes...

I may back out the Windows change and merge this in... we'll see if I get to it tonight before we have a chance to sync... especially since I want to understand implications of using / not using cloudprovider with the Windows kubelets...

parts/kubernetesagentcustomdata.yml
colemickens: This was pulled back out, we need other changes first, will do in a separate PR.

PR #245, opened 2017-02-05
Update build.sh

dwaiba:
* https://github.com/jteeuwen/go-bindata is not maintained anymore
* Actively maintained on fork https://github.com/lestrrat/go-bindata

PR #253, opened 2017-02-09
Lessen development pain (rebasing, etc)

colemickens:
Rebasing is rough right now for a few reasons, this attempts to address them, at least partially:

1. ~~Use the `lestrrat` fork of `go-bindata` since it is being maintained (similar to #245)~~

2. Remove all the (broken, not working) build remnants from a long time ago

3. Make `go-bindata` ignore timestamp updates. This cuts down on a huge amount of needlessly dirty workspaces, warnings when switching branches, and often the need for manual merge of `templates.go`.

4. In the case that there ARE changes to parts that require merging, or result in `templates.go` needing a merge... you can now run `./scripts/rebase-helper.sh` and it will attempt to auto-resolve the currently rebasing commit and advance the rebase.

5. Bump us up to Go 1.8.

6. Bumps us up to `0.1.2rc1` of `azure-cli`.

Note number 4 is not tested yet, but I wanted to get this out for a review.

(This would fully subsume #245.)


scripts/rebase-helper.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# If anything fails, we bail out and leave the user to repair.
+
+# This will fail to push the rebase forward if there are other conflicts to be resolved.
+# It's recommended you re-run this script again after resolving them to ensure that the
+# generated file and expected test output reflects your merge resolutions.
+
+go generate -v ./...
+cd pkg/acsengine/testdata
+mv **/*.json{.err,}
```
anhowe: what does this move command do?

colemickens: It's "globbing".

It will (or rather, should - testing is in progress) expand to:
```
$ mv foo/bar.json.err foo/bar.json
$ mv whatever/something/else/cool.json.err whatever/something/else/cool.json
etc...
```

ams0:
I'm facing build errors when I ```make build```:

```
<snip>
pkg/acsengine/fileloader.go
flag provided but not defined: -nometadata
Usage: go-bindata [options] <input directories>

  -debug=false: Do not embed the assets, but provide the embedding API. Contents will still be loaded from disk.
  -nocompress=false: Assets will *not* be GZIP compressed when this flag is specified.
  -nomemcopy=false: Use a .rodata hack to get rid of unnecessary memcopies. Refer to the documentation to see what implications this carries.
  -o="./bindata.go": Optional name of the output file to be generated.
  -pkg="main": Package name to use in the generated code.
  -prefix="": Optional path prefix to strip off asset names.
  -tags="": Optional set of uild tags to include.
  -version=false: Displays version information.
pkg/acsengine/fileloader.go:3: running "go-bindata": exit status 2
<snip>
```

Perhaps go-bindata changed? I'm on go1.8, Ubuntu16.04 and
```
$ go-bindata -version
go-bindata 3.1.0 (Go runtime go1.2.1).
```

Removing "-nometadata" (effectively reverting commit 4b20f1a) make the build succeed.

PR #255, opened 2017-02-09
k8s: add default storage class

colemickens:


colemickens:
testing in progress, adding 'DO-NOT-MERGE' for now.

colemickens:
Tested. Works as expected. A PV is dynamically provisioned. I didn't walk through a full attachment demo on 1.5.1 builds (too slow from other bugs). Works well in 1.6. This is safe to add anyway (and will provide value in 1.5.3) so I'm going to merge.

After I get a review I suppose. 😃

colemickens:
(cc: @brendandburns since I'm just copying this from your PR: https://github.com/kubernetes/kubernetes/pull/41182)

lostintangent:
@colemickens Awesome! Out of curiosity: what is the v1.5.3 value you're referring to?

colemickens:
@lostintangent `1.5.3` in my comment there refers to `v1.5.3` of Kubernetes which is due soon. `v1.5.2` has a regression regarding load balancers so we're sort of ignoring it. `v1.5.3` will also include a number of improvements to the disk attach/formatting code thanks to user 'codablock' on GitHub (don't want to send them spam by tagging)

Just a note... you can use all of this today with a 1.5.x cluster, you just need to add the `DefaultStorageClass` yourself, or else add your own `StorageClass` and select it in your PVC. Just in case you're anxious to use it!)

lostintangent:
@colemicken Very cool! Thanks a lot for that explanation.

PR #256, opened 2017-02-10
added email notification

dmitsh:


test/acsengine.deploy
```diff
@@ -101,6 +105,12 @@ uksouth ukwest"
             // Generate reports
             try {
               junit("${junit_dir}/**/junit.xml")
+              if(sendReport && (sendTo != "")) {
```
colemickens: I think there's a way to query the current build status - as in "failed" or "succeeding". You might consider that instead of the `sendReport` sentinel.

test/acsengine.deploy
```diff
@@ -101,6 +105,12 @@ uksouth ukwest"
...
+                emailext(
+                  to: "${sendTo}",
+                  subject: "[ACS Engine Jenkins Failure] ${env.JOB_NAME} #${env.BUILD_NUMBER}",
```
colemickens: *nit:* Maybe add something that tells what kind of test from a quick glance since I assume other types of jobs will be extended to send mail as well... Is it a smoke failure, a BVT failure?

dmitsh: The name of the test is in the subject: env.JOB_NAME

colemickens: Oh of course. Never mind.

test/acsengine.deploy
dmitsh: You are right! Implemented.

PR #257, opened 2017-02-10
Update an old deprecated kubelet flag, removed in 1.6.x

brendandburns:
Ref: https://github.com/kubernetes/kubernetes/pull/40048

It's been present in kubelet since: https://github.com/kubernetes/kubernetes/commit/e43ccdbf2c0b69bfa19108a4d82894962c108016 so it is safe for 1.5.x clusters too

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/257)


brendandburns:
Updated tests...  ptal.

colemickens:
Sorry to be a nag, but is there any chance you are using Go 1.7? There's a bit of diff in there that makes me think so. Unfortunately, the CI is now on Go 1.8 and bindata output changes based on Go version.

(I'm guessing this based on the templates.go file having far more changes than it should based on your change).

Alternatively, I have this change incorporated in my branch that I'll be sending a PR from in the next 18 hours or so...

brendandburns:

JackQuincy:
@brendandburns @colemickens is this something we should be trying to merge or can I close this PR?

colemickens:
I'm pretty sure it's needed for 1.6 but I was going to try to take a pass at that on Friday or the weekend and see if it's only this flag. We may need to do other stuff as part of 1.6, the other PR I have open is relevant, we may want to switch to etcd3 too...

(Plus the test data / templates.go isn't updated in this PR)'

So on the one hand, it may be obsoleted, but on the other, it's more than I have to offer at the moment....

shrutir25:
@colemickens do we want to merge this PR ?

shrutir25:
@brendanburns @anhowe how do we want to proceed with this PR ?

anhowe:
closing since this change is captured here: https://github.com/Azure/acs-engine/blob/master/parts/kuberneteskubelet.service

PR #259, opened 2017-02-10
someone didn't run go generate running it

JackQuincy:


JackQuincy:
I could be wrong actually I thought your change to ignore timestamp was already in.

colemickens:
@dmitsh was seeing the same thing in Jenkins though.

Looks like diff contains more than just date changes.

If you clone master and run `make ci` it fails... so I think we still want this.

PR #264, opened 2017-02-13
add kubernetes windows classic

anhowe:


pkg/api/vlabs/validate.go
```diff
@@ -275,7 +275,7 @@ func validatePoolName(poolName string) error {
 }
 
 func validateDNSName(dnsName string) error {
-	dnsNameRegex := `^[a-z][a-z0-9-]{3,45}[a-z0-9]$`
+	dnsNameRegex := `^[a-z][a-z0-9-]{1,43}[a-z0-9]$`
```
raveeram: Is this an intentional change?

raveeram: discussed offline

Issue #266, opened 2017-02-13
[k8s] Allow customizing flags for kubelet


wbuchwalter:
There are a number of scenarios where someone would want to customize the flags passed to kubelet on the agents (i.e. enabling alpha features).

Any idea on how that could be done?

@colemickens initial input:
>We would need to come to agreement on what this would look like in the apimodel - do we support a number of specific toggles and support checking for invalid combinations? Do we just allow the user to wholesale override all of the attributes passed to the various components, etc?

jackfrancis:
@wbuchwalter I think the most practical step forward is to continue to add configuration of kubelet (and other k8s core components) flags on an as-needed basis. You can see in the current kubelet service definition that we ship to clusters there's a fair amount of ad hoc customization going on:

https://github.com/Azure/acs-engine/blob/master/parts/kuberneteskubelet.service

Generalizing this would be a decent amount of work, and not entirely practical given that the set of kubelet key/vals from one version of kubernetes to the next is subject to change (and would have to be managed appropriately at template generation time to ensure we produce a usable kubelet spec).

I would welcome any improvement to this surface area as a PR, up to and including a tasteful general model! But I'd prefer to keep this out of the backlog as it's not likely something we'd tackle anytime soon in a comprehensive way.

PR #267, opened 2017-02-13
[k8s] Add experimental-nvidia-gpus flag when VM has GPU

wbuchwalter:
Mostly for discussion, as I don't think we want custom code like that for every k8s experimental feature.

In order to allow Kubernetes to schedule workloads requiring GPU resource against a node, the 
`--experimental-nvidia-gpus` needs to be set to `1`.
This PR checks the `VMSize` of the agent profile, and set the flag to 1 for any `Standard_N*` size.

Broader related issue: https://github.com/Azure/acs-engine/issues/266

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/267)

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -480,8 +480,8 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 			// return the custom data
 			return fmt.Sprintf("\"customData\": \"[base64(concat('%s'))]\",", str)
 		},
-		"GetKubernetesAgentCustomData": func() string {
-			str, e := getSingleLineForTemplate(kubernetesAgentCustomDataYaml)
+		"GetKubernetesAgentCustomData": func(profile *api.AgentPoolProfile) string {
+			str, e := getSingleLineKubernetesAgentCustomData(kubernetesAgentCustomDataYaml, profile.Name)
 			if e != nil {
 				return ""
 			}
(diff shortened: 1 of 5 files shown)


pkg/acsengine/engine.go
```diff
@@ -564,6 +564,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
 		"HasWindowsSecrets": func() bool {
 			return properties.WindowsProfile.HasSecrets()
 		},
+		"ExperimentalGpuFlag": func() int {
+			//Only test a single profile, since all agents should be the same size
+			isGpuVM := strings.HasPrefix(properties.AgentPoolProfiles[0].VMSize, "Standard_N")
```
anhowe: what if pool 1 has Standard_N, but pool 0 has Standard_D2_v2?  Will this still enable?

pkg/acsengine/engine.go
```diff
@@ -564,6 +564,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(properties *api.Properties) map[s
...
+			if isGpuVM {
```
colemickens: Just return the value of `HasPrefix`.

anhowe:
I've unapproved to make sure you turn it on with a GPU capable pool.

to pkg/api/types.go add something like the following (I just confirmed with Drew, the following code will be stable for the next 6-12 months):

func (a *AgentPoolProfile) HasGPU() bool {
	return strings.HasPrefix(a.VMSize, "Standard_N")
}

Then in the template just call "{{.HasGPU}}" instead of "{{ExperimentalGpuFlag}}"

wbuchwalter:
@anhowe Thanks for the feedback, totally missed that.
I may have missed something about how acs-engine works, first time in this code for me, but I also had to modify `engine.go` to allow the variable name to be different for each agent pool.
I didn't find anywhere else where this was done, so I had to add a new syntax `[[VMName]]`.
I'm guessing this might need more discussion to get approved, feel free to point me in another direction or request further changes.

brendandburns:

wbuchwalter:
I will close this PR unless someone sees a reason not to.
A new PR will be submitted soon with kubernetes 1.6 support instead.
That will allow us to get rid of the ugly `[[VMName]]` bit, and make the code simpler anyway.

wbuchwalter:
Closing in favor of #385.

PR #270, opened 2017-02-14
cleanup resource group after test

dmitsh:


test/deploy.sh
```diff
@@ -84,7 +84,3 @@ export KUBECONFIG="${OUTPUT}/kubeconfig/kubeconfig.${LOCATION}.json"
 "${ROOT}/${VALIDATE}"
 
 echo "post-test..."
-
```
colemickens: I assume this means the trap works now that the inner script doesn't redefine the function?

dmitsh: correct :)

PR #278, opened 2017-02-15
DCOS 1.8.8 Support

colemickens:
~~This is a WIP, still in testing, but getting it out for an initial review.~~

Tested with walkthrough, looks good.


pkg/acsengine/testdata/simple/dcos1.8.8.json
```diff
@@ -0,0 +1,41 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "DCOS187"
```
JackQuincy: 188?

colemickens:
Thanks @JackQuincy @raveeram for the catches. Looks like it's still not working, but probably closer. Check latest commits for the fixes. (Will squash eventually before merging).

pkg/acsengine/engine.go
```diff
@@ -635,7 +635,8 @@ func getPackageGUID(orchestratorType api.OrchestratorType, masterCount int) stri
 func getDCOSCustomDataPublicIPStr(orchestratorType api.OrchestratorType, masterCount int) string {
 	if orchestratorType == api.DCOS173 ||
 		orchestratorType == api.DCOS184 ||
-		orchestratorType == api.DCOS187 {
+		orchestratorType == api.DCOS187 ||
+		orchestratorType == api.DCOS188 {
```
amanohar: Are there any template code changes expected with this new DCOS version?

colemickens: Not as far as we can tell from diffing the 1.8.7 and 1.8.8 templates from Mesosphere.

amanohar: I was trying to find dcoscustomdata188.t but couldn't for some reason. I refreshed, now I see it.

JackQuincy:
Something I'm not seeing that I would expect is a change in the bootstrap id for 188. in 187 it is defined in custom data on line like so
- content: 'BOOTSTRAP_ID=5b4aa43610c57ee1d60b4aa0751a1fb75824c083

colemickens:
@JackQuincy Thanks for the catch, testing now. Any reason that isn't parameterized out like the other "guid" values?

JackQuincy:
it wasn't as it didn't need to be. The other ones change based on number of masters. This one is consistent 1,3, or 5 masters

colemickens:
@JackQuincy hm okay. Thanks for the help here. Looks like everything is healthy.

Ran through the entire DCOS walkthrough... all works as expected.

ping @anhowe for another review

PR #287, opened 2017-02-16
regenerate templates... again.

colemickens:
Shame on me.


colemickens:
Never mind, Anthony got it anyway.

PR #291, opened 2017-02-17
Adding VM SKU Sizes

uday31in:
Currently we are missing some common sizes e.g. Standard_DS2_v2


colemickens:
Are those really recently added? Otherwise why would you have not already got them from your automatic ps script @anhowe?

anhowe:
This doesn't look autogenerated, note the timestamp 2017-01-24 09:25:08Z.  Closing as we shouldn't update the autogenerated file without running the generator.  I've re-run as #293.  This adds the L sizes.  Note that sizes with < 2 cores and small disks are excluded from generation.

PR #292, opened 2017-02-17
Update IP layout for Kubernetes Windows agents to be Dynamic & allow multiple Windows agent pools

amanohar:


colemickens:
Thanks again for catching the route table regression and documenting it!

PR #293, opened 2017-02-17
refresh Azure sizes

anhowe:


uday31in:
@anhowe @colemickens  -  Any specific reason why "Standard_DS2_v2" is not in the list of allowed sizes? 

FYI - Sizes of  "Standard_D2",  "Standard_D2_v2" - is allowed.

Uday

anhowe:
@uday31in - DCOS requires ephemeral disks of at least 100GB for the masters and that is why DS2_v2 is excluded, since it only has a 14GB ephemeral disk, whereas standard_D2 and Standard_D2_v2 have disks >= 100GB.

However, I have further relaxed this requirement for Kubernetes and Swarm masters.  Check out the latest changes which will now allow DS2* for Kubernetes and Swarm masters.

PR #296, opened 2017-02-17
fixed shunittest validation script

dmitsh:


test/shunit/validate_deployment.sh
```diff
@@ -12,6 +10,11 @@ function shunittest_validate_deployment {
   fi
 
   script="${HOME}/test/cluster-tests/${ORCHESTRATOR}/test.sh"
-  [ -x "$script" ] || echo "$script: No such file or directory"; exit 2
-  "$script"
+
+  if [ -x "$script" ]; then
```
colemickens: Can we be consistent `${script}`? Also, if `script` isn't guaranteed to be set... do you want `${script:-}` ?

PR #297, opened 2017-02-17
Regenerate templates to pass ci test

JiangtianLi:


colemickens:
All of these changes are non-material and can't be the cause of the problem.

Can you tell me what `go version` says? I strongly suspect you're on Go 1.8...

JiangtianLi:
go version go1.7.5 windows/amd64.

Which version should I use?

colemickens:
Hm, okay... Any chance you can do the generate again from the devenv/docker container and push it up?

It could even just be 1.7.4 vs 1.7.5 since I think the container is still on 1.7.4 (I have a PR for that too)

(Basically, if you can run `make ci` locally, on HEAD cleanly, then something else is up - like Go version, or version of Go-bindata or something. )

JiangtianLi:
@colemickens  I did "make build" && "make ci" in the devenv/docker container. It seems the changes are file mode. I used to build on Windows box. Maybe that is the root cause?

colemickens:
We should not suppress the output of `validate-generated` so that we can see the problem in the CI job... (cc: @dmitsh)

Maybe it's the newline at the bottom actually... no idea why that would be new though...

https://github.com/Azure/acs-engine/blob/master/scripts/validate-generated.sh#L22 This should prevent any lines with metadata from triggering a validation failure... hence why I don't think the filemode or timestamps matter.

If you think this will help, I'll go ahead and approve. It certainly shouldn't hurt.

JiangtianLi:
Thanks, Cole! I'll give it a try.

PR #304, opened 2017-02-18
fix k8s windows provisioning that gets blocked on upper case vm name

anhowe:


parts/kuberneteswindowssetup.ps1
```diff
@@ -161,7 +161,7 @@ Get-PodCIDR
     Write-Host "waiting to discover pod CIDR"
     while (-not $podCidrDiscovered)
     {
-        $podCIDR=c:\k\kubectl.exe --kubeconfig=c:\k\config get nodes/$AzureHostname -o custom-columns=podCidr:.spec.podCIDR --no-headers
```
raveeram: Just for my information, this is not required anymore after Dmitry's change or is it that Cole added the cloud-provider flag?

raveeram: Ah looks like this was updated.

PR #305, opened 2017-02-18
remove cloud provider on windows k8s agents

anhowe:
When cloud provider runs on windows, the routing table and pod cidrs go out of sync.


colemickens:
I think you need to update the testdata expected files.

Otherwise LGTM.

PR #313, opened 2017-02-21
Fixes the issue I just introduced.

colemickens:
There's an ordering problem.
`make ci` calls for validation. Validation requires `go-bindata` but `go-bindata` isn't installed until the build command.

This moves the prereqs into an explicit build target and then make sures dependent targets require it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,19 +2,21 @@
 
 .PHONY: build test validate-generated lint ci devenv
 
-build:
+prereqs:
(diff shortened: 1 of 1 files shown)


colemickens:
(This regressed because I took it out of the Dockerfile so that there was only a single source of truth for how to acquire this dependency).

Normally, I'd just install `go-bindata` by `go get .` but that doesn't work if the project doesn't build locally, etc, etc.

PR #317, opened 2017-02-22
enabled 'classic', 'windows' and 'dcos-versions' tests

dmitsh:

PR #319, opened 2017-02-23
fixed templates for acs-engine tests

dmitsh:


colemickens:
Is this due to  https://github.com/Azure/acs-engine/pull/317/files ?

PR #327, opened 2017-02-28
Kubelet + Docker Argument refresh + bump docker to 1.13.x

colemickens:
1. Fix kubelet flags. `--masters` was deprecated. `--kubeconfig` alone doesn't appear to be sufficient, so also added `--require-kubeconfig`. Also, `--pod-manifest-path` instead of `--config`.

2. Fix `docker` flags so that it will leave iptables/ipmasq alone, also add optional `ExecStartPre`'s to undo anything `docker` may have done at install time.

3. Pin `docker` to `1.13.0` now that we are in full control of our iptables destiny.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/327)


brendandburns:

parts/kubernetesagentcustomdata.yml
```diff
@@ -30,7 +30,10 @@ write_files:
   content: |
     [Service]
     ExecStart=
-    ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay
```
rjtsdl: add the comment here as well ?

# Note, the preceeding dash tells systemd to ignore non-zero exit code

parts/kubernetesagentcustomdata.yml
```diff
@@ -113,13 +116,13 @@ write_files:
       --volume=/srv/kubernetes/:/srv/kubernetes/:ro \
         {{{kubernetesHyperkubeSpec}}} \
           /hyperkube kubelet \
-            --api-servers="https://{{{masterInternalLbIp}}}:443" \
```
rjtsdl: Reason to remove this ?

--api-servers

colemickens: It's a deprecated flag. The `kubeconfig` file already has the masters location inside, so this flag was deprecated. The `require-kubeconfig` flag is required for a reason I don't 100% fully understand, but I think it's a temporary thing and will go away after 1.6.

parts/kubernetesagentcustomdata.yml
colemickens: will do!

parts/kubernetesagentcustomdata.yml
rjtsdl: I see

colemickens:
Heads up. I'm about to:

* rebase
* revert back to 1.12.x (1.13 is not supported officially by Kubernetes)
* merge

PR #328, opened 2017-02-28
added retry logic for testing proxy via dashboard

dmitsh:


colemickens:
@dmitsh I know you weren't able to manually trigger it, but do you think it was possible that this problem was introduced by a recent PR?

PR #330, opened 2017-02-28
regenerate templates

colemickens:
cc: @wangtt03 

The build was broke by merging https://github.com/Azure/acs-engine/pull/152

This should hopefully fix it.


colemickens:
cc: @dmitsh

PR #343, opened 2017-03-02
set correct 'targetEnvironment' value

dmitsh:


brendandburns:

pkg/acsengine/engine.go
```diff
@@ -295,15 +295,25 @@ func prepareTemplateFiles(properties *api.Properties) ([]string, string, error)
...
+		return AzureChinaCloudSpec
+	default:
+		return AzureCloudSpec
 	}
+}
 
-	return cloudSpecConfig
+func GetCloudTargetEnv(location string) string {
+	switch location {
+	case "chinaeast":
```
colemickens: You can do this too for the switches...

```
switch {
case location=="chinaeast" || location=="chinanorth":
    return AzureChinaCloud
default:
    return AzurePublicCloud
}
```

I'll let you decide if it's better or worse 😄 .

dmitsh:
I deployed and verified all 4 orchestrators with these changes. Does it count?

Thanks,
-Dmitry

pkg/acsengine/engine.go
dmitsh: @weinong thinks we should add strings like "China East" and "China North". Should we?

colemickens:
@dmitsh that's your PR builder commenting, not Brendan :P

pkg/acsengine/engine.go
colemickens: Can we canonicalize them first? Is it true that for any "Location Name" that "locationname" will always be valid?

@weinong is probably right even though I wish he weren't... if that makes sense...

pkg/acsengine/engine.go
```diff
@@ -295,15 +300,27 @@ func prepareTemplateFiles(properties *api.Properties) ([]string, string, error)
...
+	switch GetCloudTargetEnv(location) {
+	case azureCloudChina:
+		return AzureChinaCloudSpec
+	default:
+		return AzureCloudSpec
 	}
+}
 
-	return cloudSpecConfig
+func GetCloudTargetEnv(location string) string {
```
weinong: there is a potential bug here. "The rest of the locations" are not public azure. there are USGov, Germany, etc... :(

dmitsh: addressed, thanks!

wangtt03:
@dmitsh Thanks for your commit.😊👍

PR #347, opened 2017-03-02
Fix kubectl extract bug.

wangtt03:
@colemickens I guess this should be a timing issue. I made a deployment in westus2 and found kubectl is not extracted correctly. I went through the syslog and found that when kubectl extract service is running, the docker service is stopped. I guess this is caused by the `systemctl restart docker` command in `kubernetesmastercustomscript.sh`. So I moved the kubectl-extract service to this script right after the `ensureKubelet` and deploy again, then the kubectl successfully extracted.


brendandburns:

wangtt03:
@colemickens could you please review this PR? Thanks!

parts/kubernetesmastercustomscript.sh
```diff
@@ -127,6 +127,11 @@ function ensureKubelet() {
     systemctl restart kubelet
 }
 
+function extractKubectl(){
+    systemctl enable kubectl-extract
```
colemickens: Is there an advantage to keeping it as a systemd unit rather than just executing the `docker run` and `mv` here? Does this all work out okay even when the master is rebooted (since this is `enable`d as well as started?)?

wangtt03: I don't have prefer. I think both of the solution will be fine. DC/OS use systemd to do extract and setup job that's why I followed the convention. For the reboot, I think it is OK because there is a condition check before the service running.

wangtt03: I did serval deployments and it seems that it works:)

wangtt03:
@colemickens I made another commit to update the testcases. I upgrade the go to 1.8 and hope it doesn't break the CI. 😊

wangtt03:
@colemickens Thanks for your quick response!

colemickens:
@wangtt03 

I pulled your branch and ran this:

```
$ make devenv
# make ci
```

and it failed with a number of diffs. Can you please run this locally;
```
$ make devenv
# ./scripts/rebase.sh
# exit
$ git push origin HEAD
```

that should update your PR with properly generated templates and testdata.

colemickens:
It looks like the testdata was right, but `templates.go` wasn't, which is weird since the testdata actual data is created from `templates.go`. I'm wondering if you missed staging it in that last commit?

colemickens:
Hi @wangtt03. Just saw you comment on another issue. Can you address this one ASAP? There are a number of people encountering it today and creating noise. Thanks.

wangtt03:
@colemickens i am sure that the template.go is added because i use vscode to add the all, i remember in one of your post you mentioned two different go-bindata, is that the problem?

wangtt03:
I use the jtween/go-bindata

colemickens:
@wangtt03 It could be, but it won't happen if you use the development docker container. Plus, if you used the wrong `go-bindata`, then the testdata would've been wrong too.... but it wasn't 😄 . Maybe we should chat on IM to unblock this.

wangtt03:
@colemickens Could you please review the PR? Hopefully it should fix the CI issue.😊

colemickens:
Thanks @wangtt03. It looked good when I ran it, so I've squashed and merged!

mfaizanse:
@wangtt03 @colemickens 
I pulled the latest commit d9373bb8b32ce1eab8676b1791d4e5de694540cf 
and generated and deployed a win+k8s cluster but when I SSH to the master node and run kubectl, it says "Command not found". I looked into the systemd failed units and found:

systemctl list-units --state=failed
```
  UNIT                    LOAD   ACTIVE SUB    DESCRIPTION
● hv-fcopy-daemon.service loaded failed failed Hyper-V File Copy Protocol Daemon
● kubectl-extract.service loaded failed failed Kubectl extraction
```
Is the issue resolved or is it still in progress? I am deploying to East US.

colemickens:
I tried several times last night and they were all successful. Can you provide output of: `sudo journalctl -u kubectl-extract`

colemickens:
Another user is reporting that extraction is flaky. Let me look at it, or @wangtt03 if you can...

wangtt03:
@colemickens Let me fix it.

wangtt03:
@mfaizanse Could you please provide more info about `sudo journalctl -u kubectl-extract`? Or could you please share me a copy of `/var/log/syslog` and `/var/log/azure/clusterProvision.log` on which the kubectl is missing?

mfaizanse:
@wangtt03 I pulled the latest commit 36bfce7fec50733ecfffbf31e65e8108674c9100 and delpoyed, now the kubectl extraction is successful.

wangtt03:
@mfaizanse That is great!😊

mfaizanse:
@wangtt03 see my comment here #348

PR #351, opened 2017-03-02
Managed disks are now GA - update readme

rgardler:


brendandburns:

PR #364, opened 2017-03-05
Update cluster denition.md by adding national azure cloud descriptions.

wangtt03:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/364)


wangtt03:
@amanohar Could you please review this PR? Thanks!

docs/clusterdefinition.md
```diff
@@ -10,12 +10,21 @@ Here are the cluster definitions for apiVersion "vlabs"
 |---|---|---|
 |apiVersion|yes|The version of the template.  For "vlabs" the value is "vlabs".|
 
+### location
+`location` describe the deployment target of the cluster. Current available:
```
amanohar: describes

amanohar: target deployment region

amanohar: Available regions:

docs/clusterdefinition.md
```diff
@@ -10,12 +10,21 @@ Here are the cluster definitions for apiVersion "vlabs"
 |---|---|---|
 |apiVersion|yes|The version of the template.  For "vlabs" the value is "vlabs".|
 
+### location
+`location` describe the deployment target of the cluster. Current available: 
+```    
+ "australiaeast", "australiasoutheast", "brazilsouth", "canadacentral", "canadaeast", "centralindia", "centralus", "eastasia", "eastus", "eastus2", "japaneast", "japanwest", "koreacentral", "koreasouth", "northcentralus", "northeurope", "southcentralus", "southeastasia", "southindia", "uksouth", "ukwest", "westcentralus", "westeurope", "westindia", "westus", "westus2", "chinaeast", "chinanorth"
+```
+
+This value can be ignored by default if the deployment is go to public azure. However, if the deployment is go to non public azure such `azure china`, the correct value must be specified. ACS engine will based on this value to determine the download URLs of the required installation binaries such Docker, DC/OS, Kubernetes Docker images. Because different location may have different policy such as China Great Firewall which blocks Azure CDN and GCR.
```
amanohar: Value of which field? I am assuming you don't mean location field.

amanohar: does not need to be specified if the deployment location is a Public Azure region. It would be useful if Public Azure regions are specified separately or even better would be refer to Azure's official region list then this list is not be out of date if new regions are added.

docs/clusterdefinition.md
```diff
@@ -24,6 +33,8 @@ Here are the valid values for the orchestrator types:
 3. `Swarm` - this represents the [Swarm orchestrator](swarm.md).
 4. `Swarm Mode` - this represents the [Swarm Mode orchestrator](swarmmode.md).
 
+For `KubernetesConfig.KubernetesImageBase` the default value is GCR (gcr.io/google_containers/), ACS engine will use this base URL and the default Kubernetes Docker image name to concat the full URL of the Docker image URL such as `gcr.io/google_containers/hyperkube-amd64:v1.5.3`. In some the location GCR may be blocked or unstable to access, by specifying this value can redirect to custom Docker image download URL.
```
amanohar: .

amanohar: names

amanohar: some locations

amanohar: .

amanohar: If KubernetesImageBase is specified then generated templates will use custom Docker image download URL.

wangtt03:
@amanohar Thanks for your comments. I updated the contents based on your feedback. 😊

colemickens:
@wangtt03 

1. the `kubectl-extract` needs to be in a separate PR
2. the history for this PR now looks incorrect and needs to be remedied (or we can squash it)
3. remove the `restart docker` is not an okay fix. This restart is essential because there are a variety of cases where `docker` is not successfully running after the package is installed


Let's keep the discussion about the fix in the Issue or in a new PR.

PR #367, opened 2017-03-07
Fix kubectl extraction bug.

wangtt03:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/367)


wangtt03:
@colemickens Hi Cole, I closed the previous PR and create this new one. Could you please review it? BTW, I am a little bit stucked by squashing this merge commit and normal commits into one, do you have ideas? Thanks!

colemickens:
@wangtt03 Best thing to do is to create two branches off of master. One that includes your documentation changes. The other than contains the kubectl-extract fix. Each with their own PR. Now we've lost all the review feedback for the documentation change, unfortunately.

wangtt03:
😲Sorry about that. https://github.com/Azure/acs-engine/pull/364 is the previous feedbacks about the document. hope it helps.

wangtt03:
@colemickens I tried several deployments using this PR and all of them success. 
@mfaizanse Could you please try this PR? It should fix the kubectl extraction bug.

wangtt03:
Ping @colemickens

parts/kubernetesmastercustomscript.sh
```diff
@@ -130,6 +130,23 @@ function ensureKubelet() {
 function extractKubectl(){
     systemctl enable kubectl-extract
     systemctl restart kubectl-extract
+    kubectlExtracted=1
```
colemickens: This fix still feels brittle. It doesn't feel right to create a systemd service to be scheduled by systemd, and then have this manual loop to keep kicking it until it does the right thing.

Can we just remove the systemd unit? Or get systemd to restart it and drive it to success instead of us having to do it manually?

wangtt03: Thanks for your feedback! I think I can move the systemd unit after the kubelet.service. If the kubelet service is started correctly, then the kubectl extraction should be fine. Does it make sense?

colemickens: Yes, I was thinking something similar. I still think the service should be restartable so that systemd restarts it in the case that it does happen to fail. And then we can remove the retry code you just recently added? Yeah? If that works out, I think it'd be okay.

wangtt03: cool. I will send the PR soon.

wangtt03:
@colemickens could you please review this PR? Thanks!

wangtt03:
Ping @cloemickens

colemickens:
@acs-bot test this please

colemickens:
@wangtt03 can you confirm that you tested this multiple times in public cloud? The CI looks green, but we don't check for `kubectl` being present on the master currently.

wangtt03:
I deployed 5 k8s clusters on west europe, all of them successful. So i think this pr should solve the problem.

colemickens:
Thanks @wangtt03. I'm going to hold off for a short bit to see if we can get the Calico one in first. Thanks for fixing this, I'll get it merged in the next day at least.

wangtt03:
@colemickens thank you😁

PR #372, opened 2017-03-07
added go utility to run concurrent deployment/validation tests

dmitsh:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/372)

### DIFF
--- a/test/acs-engine-test.sh
+++ b/test/acs-engine-test.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+###############################################################################
+
+set -e
+set -u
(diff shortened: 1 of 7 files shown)


test/acs-engine-test/acs-engine-test.json
```diff
@@ -0,0 +1,20 @@
+{"deployments":
```
colemickens: Can you run this through `jq` for consistency's sake?

test/acs-engine-test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+###############################################################################
+
+set -e
+set -u
+set -o pipefail
+
+ROOT="${DIR}/.."
+
+# Check pre-requisites
```
colemickens: Why not do these checks for env vars in Go?

test/acs-engine-test/config.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return nil
+}
+
+func getTestConfig(fname string) (*TestConfig, error) {
+	data, err := ioutil.ReadFile(fname)
+	if err != nil {
+		return nil, err
+	}
+	config := new(TestConfig)
```
weinong: as I was reading the some go material, it's suggested that it's idiomatic to use var config TestConfig to initialize the variable with default values. or config := TestConfig{}.

test/acs-engine-test/acs-engine-test.json
dmitsh: done

test/acs-engine-test.sh
dmitsh: done

test/acs-engine-test/config.go
dmitsh: in this case it should be config := &TestConfig{}, as  I return pointer. Done.

PR #385, opened 2017-03-09
[K8s] Add multi-gpu support

### LINKED ISSUE #679: GPU support in Kubernetes
Kubernetes v1.6 introduced GPU support that is useful in a variety of workloads, including data science and machine learning.

ritazh:
This PR allows Kubernetes to schedule workloads requiring multiple GPU resources against a node. This feature was recently added to K8 v1.6.beta.1 in this [commit](https://github.com/kubernetes/kubernetes/pull/42116). To turn on this feature add the `--feature-gates=Accelerators=true` flag. Once v1.6 is officially released this month, I will update the `hyperkube-amd64` value so that this PR can be merged.

I have also included [Kubernetes Multi-GPU support Walkthrough](https://github.com/ritazh/acs-engine/blob/enable-k8v1.6-multiplegpu/docs/kubernetes.gpu.md) for reference. 

This has been tested against ACS in eastus. 

> NOTE: kubelet's `--config` flag has been deprecated, using `--pod-manifest-path` instead [Changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#v160-beta1)

cc: @wbuchwalter

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/385)

Fixes #679


colemickens:
cc: @amanohar @anhowe PTAL, we should discuss.

parts/kubernetesagentcustomdata.yml
```diff
@@ -135,7 +135,8 @@ write_files:
             --allow-privileged=true \
             --enable-server \
             --enable-debugging-handlers \
-            --config=/etc/kubernetes/manifests \
+            --feature-gates=Accelerators=true \
```
anhowe: If the machine doesn't have GPU, does this flag have any effect?  We want to avoid affecting existing customers who do not want GPU support.

docs/kubernetes.gpu.md
```diff
@@ -0,0 +1,63 @@
+# Microsoft Azure Container Service Engine - Kubernetes Multi-GPU support Walkthrough
```
anhowe: Can you add a link to this from the main page, or from the Kubernetes page?

pkg/acsengine/const.go
```diff
@@ -32,7 +32,7 @@ const (
 )
 
 const (
-	KubernetesHyperkubeImageName         = "hyperkube-amd64:v1.5.3"
+	KubernetesHyperkubeImageName         = "hyperkube-amd64:v1.6.0-beta.1"
```
anhowe: This will have to wait for the official 1.6 stable release.  Also you will need to adjust all 1.5.* references everywhere including the dashboard, documenation, kubectl download links, windows binaries.  It may actually be best to submit as 2 PRs, one from v1.6, and one for GPU support.

examples/kubernetesgpu.json
```diff
@@ -0,0 +1,35 @@
+{
```
anhowe: this should go under example/gpu/kubernetesgpu.json

pkg/acsengine/const.go
ritazh: Absolutely agree with all the 1.6 updates that need to happen. One thing is that this feature is only introduced in 1.6 so 1.6 is a prerequisite.

docs/kubernetes.gpu.md
ritazh: Done.

parts/kubernetesagentcustomdata.yml
wbuchwalter: This flag is only used for GPU discovery currently (might be used for other type of hardware discovery in the future).
If a VM has this flag enabled, but no GPU is present, nothing really happens. The list of available GPU will be empty and no GPU workloads will be scheduled on it as you see [here](https://github.com/kubernetes/kubernetes/blob/ed479163fa3772895be6a837d0b645ec62bfd383/pkg/kubelet/gpu/nvidia/nvidia_gpu_manager.go#L212).
No error will be logged either as far as I can tell, so that should be transparent for non-GPU customers.  
  
The only thing is that if the container engine used is not docker, then an error message will get displayed and the accelerator flag will be disabled, but anyway I don't think we support anything else than docker? (see [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L804))

docs/kubernetes.gpu.md
```diff
@@ -0,0 +1,63 @@
...
+* You need to specify `alpha.kubernetes.io/nvidia-gpu: 2` as a limit
+* You need to expose the drivers to the container as a volume. If you are using TF original docker image, it is based on ubuntu 16.04, just like your cluster's VM, so you can just mount `/usr/bin` and `/usr/lib/x86_64-linux-gnu`, it's a bit dirty but it works. Ideally, improve the previous script to install the driver in a specific directory and only expose this one.
+
+``` yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gpu-test
+spec:
+  volumes:
```
macat: I tried to use the pod config from here, but the nvidia devices were not available in the pod. Shouldn't this config contain volumes exposing `/dev/nvidia*` devices for the container?

wbuchwalter: Did you correctly install the drivers on the host machine and made sure `/dev/nvidia-uvm` was online?  
You don't need to expose the devices as kubernetes already does that for you as you can see [here](https://github.com/kubernetes/kubernetes/blob/f53ba5581b97c4993205b5ab4468abd26041f153/pkg/kubelet/gpu/nvidia/nvidia_gpu_manager.go#L193)

ritazh: @macat Please take a look at [Kubernetes Multi-GPU support Walkthrough](https://github.com/ritazh/acs-engine/blob/enable-k8v1.6-multiplegpu/docs/kubernetes.gpu.md) for reference. Specifically, you have to install the drivers by running `install-nvidia-driver.sh`.

macat: Wow, it really works. This is a great improvement! 

After running this container, do you not have a problem with linking libraries attached from the host machine?

After importing `tensorflow` in python I'm getting this (using the `:tensorflow/tensorflow:1.0.1-gpu-py3` docker image):

```
$ python3 
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tensorflow as tf
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcublas.so.8.0 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcudnn.so.5 locally
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcufft.so.8.0 locally
I tensorflow/stream_executor/dso_loader.cc:126] Couldn't open CUDA library libcuda.so.1. LD_LIBRARY_PATH: /usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib/x86_64-linux-gnu
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:165] hostname: tpod2
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:189] libcuda reported version is: Not found: was unable to find libcuda.so DSO loaded into this program
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:363] driver version file contents: """NVRM version: NVIDIA UNIX x86_64 Kernel Module  375.26  Thu Dec  8 18:36:43 PST 2016
GCC version:  gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) 
"""
I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:193] kernel reported version is: 375.26.0
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1065] LD_LIBRARY_PATH: /usr/local/nvidia/lib:/usr/local/nvidia/lib64:/usr/lib/x86_64-linux-gnu
I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:1066] failed to find libcuda.so on this system: Failed precondition: could not dlopen DSO: libcuda.so.1; dlerror: libnvidia-fatbinaryloader.so.375.26: cannot open shared object file: No such file or directory
I tensorflow/stream_executor/dso_loader.cc:135] successfully opened CUDA library libcurand.so.8.0 locally
```

macat: Got the same result with the `gcr.io/tensorflow/tensorflow:latest-gpu` image as well.

macat: `nvidia-docker` creates a new `ldcache` for these libraries and [mounts them as a volume](https://github.com/NVIDIA/nvidia-docker/blob/master/src/nvidia/volumes.go).  I wonder if the best option is mounting that `nvidia-docker` volume into the kubernetes pod.

ritazh: @macat you should not be seeing those errors. are you mounting the volumes like this in your yaml file?
```
apiVersion: v1
kind: Pod
metadata:
  name: gpu-test
spec:
  volumes:
  - name: binaries
    hostPath:
      path: /usr/bin/
  - name: libraries
    hostPath:
      path: /usr/lib/x86_64-linux-gnu
  containers:
  - name: tensorflow
    image: gcr.io/tensorflow/tensorflow:latest-gpu
    ports:
    - containerPort: 8000
    resources:
      limits:
        alpha.kubernetes.io/nvidia-gpu: 2
    volumeMounts:
    - mountPath: /usr/bin/
      name: binaries
    - mountPath: /usr/lib/x86_64-linux-gnu
      name: libraries
```

wbuchwalter: @macat This is not an issue with this PR or `acs-engine`, so I suggest opening an issue on the kubernetes repo to discuss this. The volumes you need to mount are different depending on how you install the drivers, which host OS you are using, which OS in the container etc...
Hopefully soon k8s will do this for you just like DCOS and nvidia-docker.

macat: @ritazh Yes I mounted them exactly like this.

@wbuchwalter You are right. I just wanted to give feedback and share my experience with the guide. I know that gpu + k8s combo is quite fragile right now, so my issue could be related to some mismatch in lib versions. 

Thanks for writing up this BTW. This is the first usable documentation on 1.6's gpu support.

pkg/acsengine/const.go
ritazh: @anhowe I started working on porting to the official 1.6 release. Please let me know if you guys are working on it as well to avoid duplicate effort.

docs/kubernetes.gpu.md
seanknox: Great guide! I was able to get GPU workloads using this PR branch + @wbuchwalter's blog post.

docs/kubernetes.gpu.md
```diff
@@ -0,0 +1,63 @@
...
+    hostPath:
+      path: /usr/bin/
+  - name: libraries
+    hostPath:
+      path: /usr/lib/x86_64-linux-gnu
+  containers:
+  - name: tensorflow
+    image: gcr.io/tensorflow/tensorflow:latest-gpu
+    ports:
+    - containerPort: 8000
```
seanknox: Jupyter is running on port 8888 by default.

colemickens: Hm. What does that mean? How does that relate to tensorflow on 8000?

seanknox: Maybe it's irrelevant? Saw this when tailing the logs on the pod:

```
$ k logs -f gpu-test
[I 02:41:17.155 NotebookApp] Writing notebook server cookie secret to /root/.local/share/jupyter/runtime/notebook_cookie_secret
[W 02:41:17.165 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using encryption. This is not recommended.
[I 02:41:17.170 NotebookApp] Serving notebooks from local directory: /notebooks
[I 02:41:17.170 NotebookApp] 0 active kernels
[I 02:41:17.170 NotebookApp] The Jupyter Notebook is running at: http://[all ip addresses on your system]:8888/?token=617d39ac4c374a75ad0f1efbb3189f0932e64643e991f26c
[I 02:41:17.170 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 02:41:17.170 NotebookApp]

    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8888/?token=617d39ac4c374a75ad0f1efbb3189f0932e64643e991f26c
```

colemickens: Ah. Yeah, I probably don't know enough about how `tensorflow` is used. I think if the user is connecting via localhost like suggested in the output, they'll probably be using `kubectl port-forward` to get the port locally. If the user wanted to expose Jupyter via a Service though, I think they'd need to add the port here.

ritazh: @seanknox you are right. If you want to run Jupyter, it should be port 8888. If you want to run TensorBoard, it should be port 6006 if using the default tensorflow image. I will update this before merging it.

examples/gpu/kubernetesgpu.json
```diff
@@ -0,0 +1,35 @@
+{
+  "apiVersion": "vlabs",
```
amanohar: Since GPU is enabled by default and ACS Engine API doesn't provide a way to disable GPU, is a separate example needed? Since this example doesn't do anything new compared to other examples. It's better to update documentation with the info that GPU is enabled by default and will work on GPU enabled skus.

shrutir25:
@ritazh We have made changes to the acs-engine repo, where we have removed test comparison with the expected files. Please pull the latest changes and resubmit.

examples/gpu/kubernetesgpu.json
ritazh: This example has been removed.

ritazh:
Rebased, ready for another round of review.

ritazh:
_bump_

parts/kuberneteskubelet.service
```diff
@@ -45,6 +45,7 @@ ExecStart=/usr/bin/docker run \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
         --hairpin-mode=promiscuous-bridge \
         --network-plugin=${KUBELET_NETWORK_PLUGIN} \
+        --feature-gates=Accelerators=true \
```
JackQuincy: This feature is only on 1.6.x. We need to be able to not set it for 1.5.x clusters.

seanknox: For my own edification, does the templating engine support wrapping manifests like this one with golang conditionals, e.g. 

```
{{if ge .OrchestratorProfile.OrchestratorVersion 1.6.0}}
--feature-gates=Accelerators=true \
{{end}}
```

anhowe: the *.service is not going through the go template engine.  We did a similar update to customdata.

shrutir25:
@ritazh we will be closing this PR as there has been no response from you

ritazh:
@shrutir25 apologies for the delay on this. Have been busy with few other things. I will try to get this done this week. Can you provide some pointers for how to use `{{if ge .OrchestratorProfile.OrchestratorVersion 1.6.0}}` in kubernetesagentcustomdata.yml? Thanks!!!

wbuchwalter:
@JackQuincy @anhowe Can you re-review this PR? It should be in a mergeable state now hopefully.
Thanks!

JackQuincy:
we need to drop the line ending changes in all the vendored packages. easiest way would be to git reset --soft to a commit before they happened and then run git checkout vendor, that will ignore the changes we could commit again, as is, this is hard to read, I'll try when I have time. Currently I'm on live site issues

seanknox:
@jackfrancis @ritazh @wbuchwalter I have a PR to fix line endings here: https://github.com/Azure/acs-engine/pull/663

If we can merge that in that will make this PR easier to review (and diff churn in general).

seanknox:
Oops, that was supposed to be @JackQuincy. Sorry Jack's.

seanknox:
@ritazh 

> Can you provide some pointers for how to use {{if ge .OrchestratorProfile.OrchestratorVersion 1.6.0}} in kubernetesagentcustomdata.yml? Thanks!!

Here's an example of using a conditional in the custom data YAML: https://github.com/Azure/acs-engine/blob/master/parts/kubernetesmastercustomdata.yml#L237

```go
{{if gt .MasterProfile.Count 1}}
    # Azure does not support two LoadBalancers(LB) sharing the same nic and backend port.
    # As a workaround, the Internal LB(ILB) listens for apiserver traffic on port 4443 and the External LB(ELB) on port 443
    # This IPTable rule then redirects ILB traffic to port 443 in the prerouting chain
    iptables -t nat -A PREROUTING -p tcp --dport 4443 -j REDIRECT --to-port 443
{{end}}
```

So if you're gating a feature based on Kubernetes version, I think it would be pretty close to what you used:
```golang
{{if ge .OrchestratorProfile.OrchestratorVersion 1.6.0}}
<your logic here>
{{end}}
```
(I saw that you created a `IsKubernetesGte16()` function which is a great idea too. If you go that route might be nice to pass the version as an argument so as to work with any version of Kubernetes)

While the `parts/kubernetesmastercustomdata.yml` and `parts/kubernetesagentcustomdata.yml` are interpolated through the template engine (and thus can use golang sprinkled through them) I'm not sure if the same is true about `parts/kuberneteskubelet.service`. @JackQuincy would know!

JackQuincy:
To my knowledge they aren't today. We make it look like they might be but those are actually just done with string.replace in code. The right answer might be to start interpolating it though. @amhowe would you agree. However I don't like how this would make us go from template to choose to template. Not sure there is a better way in our current patterns

wbuchwalter:
@seanknox Are you sure we can access `OrchestratorProfile` from within `parts/kubernetesagentcustomdata.yml` ? @ritazh and I couldn't make it work.  
It seems that the context is only `AgentPoolProfile` as we can see here: https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/engine.go#L623  
For `kubernetesmastercustomdata.yml` we are providing the full context (`profile *api.Properties`) allowing access to `OrchestratorProfile`.  
That's why I went with the custom function using the `ContainerService` object instead to access the `OrchestratorProfile`.  
Let me know if I am missing something.

wbuchwalter:
@seanknox added parameters for k8s version checking function.

parts/kuberneteskubelet.service
```diff
@@ -45,7 +45,8 @@ ExecStart=/usr/bin/docker run \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
         --hairpin-mode=promiscuous-bridge \
         --network-plugin=${KUBELET_NETWORK_PLUGIN} \
-        --v=2
+        --feature-gates=Accelerators=true \
```
JackQuincy: feels like we need to drop this line. Am I missing something?

JackQuincy:
Also if you could rebase on top of master we should be able to get 1000 or so files out of this review.

parts/kuberneteskubelet.service
seanknox: I believe this is needed for GPU support. Since the templating system doesn't currently support these files (we can't put a conditional in here), we'll need another way to gate.

wbuchwalter: @JackQuincy @seanknox Jack is right, stupid mistake on my part. The flag is passed from agentcustomdata, I'll fix that in a few moments.

wbuchwalter:
@JackQuincy Fixed.

parts/kuberneteskubelet.service
```diff
@@ -45,7 +45,7 @@ ExecStart=/usr/bin/docker run \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
         --hairpin-mode=promiscuous-bridge \
         --network-plugin=${KUBELET_NETWORK_PLUGIN} \
-        --v=2
+        --v=2 ${KUBELET_FEATURE_GATES}
```
seanknox: Nice approach!

pkg/acsengine/engine.go
```diff
@@ -500,6 +530,13 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) map[str
 			return cs.Properties.OrchestratorProfile.OrchestratorType == api.DCOS &&
 				cs.Properties.OrchestratorProfile.OrchestratorVersion == api.DCOS190
 		},
+		"IsKubernetesVersionGe": func(version string) bool {
+			targetVersion := api.OrchestratorVersion(version)
+			targetVersionOrdinal := VersionOrdinal(targetVersion)
+			orchestratorVersionOrdinal := VersionOrdinal(cs.Properties.OrchestratorProfile.OrchestratorVersion)
+			return cs.Properties.OrchestratorProfile.OrchestratorType == api.Kubernetes &&
+				orchestratorVersionOrdinal >= targetVersionOrdinal
+		},
```
seanknox: Overall this looks great, excited to get this in. Would love a unit test for this new function. Is that possible?

wbuchwalter: Since `getTemplateFuncMap` returns `map[string]interface{}` I had to do quite a lot of reflection to test this function, making my test pretty much unreadable, unless I missed something (I am not a Go expert).    
So I tested `VersionOrdinal` instead, which is the important part.

seanknox: Thanks and no worries @wbuchwalter—this is helpful.

seanknox:
@acs-bot test this please

seanknox:
@ritazh @wbuchwalter can you describe how you tested this?

wbuchwalter:
@seanknox I have a fork of this branch, where the only difference is the addition of a script that downloads and install the NVIDIA drivers on agent nodes (you can see the [diff here](https://github.com/ritazh/acs-engine/compare/enable-k8v1.6-multiplegpu...wbuchwalter:k8s-gpu?expand=1)).
I create 2 clusters, one with k8s 1.5.3 and one with k8s 1.6.2.
The cluster is composed of two pools:  
* pool1: DS2_v2
* pool2: NC_6  

First, if we look at the generated `customData` for the agent pools VMs, we can see that `KUBELET_FEATURE_GATES=--feature-gates=Accelerators=true` only appears in the `azuredeploy.json` for  the `1.6.2` cluster.

When the clusters finish creating, if I describe the nodes:
* On both clusters, for pool1, no GPU is detected (` alpha.kubernetes.io/nvidia-gpu: 0`).
* On the cluster with 1.5.3, the pool2 agent also has ` alpha.kubernetes.io/nvidia-gpu: 0`
* On the cluster with 1.6.2, the pool2 agent has `alpha.kubernetes.io/nvidia-gpu: 1`

I also scheduled some GPU workloads on the 1.6.2 cluster and everything works as expected.

wbuchwalter:
@seanknox Is there anything else we need to look at? Or is this PR good to be merged?

wbuchwalter:
@seanknox @JackQuincy @anhowe Can we get this merged if you have no further comments? Thanks!

JackQuincy:
@acs-bot  test this please

JackQuincy:
please update the branch from. Sorry it is taking awhile to merge I haven't seen it up to date and with a passing CI run to be able to merge.

wbuchwalter:
@JackQuincy Done

JackQuincy:
@acs-bot test this please

seanknox:
alright folks, we're getting this in today! :) The [first release of acs-engine](https://github.com/Azure/acs-engine/releases/tag/v0.1.0) has been cut so we can advise teams to use releases while HEAD continues to move. As an aside, our testing infrastructure is tightening up due to great work by @shrutir25 @dmitsh and @jchauncey. 

@wbuchwalter @ritazh can you do another rebase? I'll personally make sure this gets in today. I'll also add you to a whitelist so you can kick off CI yourselves for the future.

wbuchwalter:
@seanknox Thanks, merge done.

JackQuincy:
@acs-bot test this please

PR #395, opened 2017-03-13
acsengine-features allows setting CUSTOM_HYPERKUBE_SPEC as a build param

colemickens:
cc: @dmitsh / @weinong 

We need this for investigations and to setup jobs for pre-release builds of Kubernetes.

Please let me know thoughts, I know these are not right:
~~1. convention for naming build parameters?~~
2. does this need to be changed in other groovy scripts?

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/395)

### DIFF
--- a/test/acsengine-features.groovy
+++ b/test/acsengine-features.groovy
@@ -40,7 +40,7 @@ node {
 
                 // First check to see if var exists in context, then check for true-ness
                 // In Groovy, null and empty strings are false...
-                if(getBinding().hasVariable("myparameter") && CUSTOM_HYPERKUBE_SPEC) {
+                if(getBinding().hasVariable("CUSTOM_HYPERKUBE_SPEC") && CUSTOM_HYPERKUBE_SPEC) {
                     env.CUSTOM_HYPERKUBE_SPEC="${CUSTOM_HYPERKUBE_SPEC}"
                 }
 
--- a/test/acsengine-features.groovy
+++ b/test/acsengine-features.groovy
@@ -37,7 +37,10 @@ node {
                 env.SUBSCRIPTION_ID="${SUBSCRIPTION_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET}"
-                if(CUSTOM_HYPERKUBE_SPEC) {
+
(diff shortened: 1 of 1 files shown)


test/acsengine-features.groovy
```diff
@@ -37,6 +37,9 @@ node {
                 env.SUBSCRIPTION_ID="${SUBSCRIPTION_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET}"
+                if(CUSTOM_HYPERKUBE_SPEC) {
```
weinong: have you tried this on a Jenkins job without this parameter? IIRC, Jenkins loves to throw undefined property exception of sort.....

weinong: also, PR bot will not be able to catch it cuz it doesn't use the groovy pipeline.

weinong: I just tried, it did throw MissingPropertyException. So you can add a try-catch...

weinong:
either update the existing jobs to add this parameter, or add a try-catch at the beginning of the pipeline to swallow exception.. If you want to do the former, let me know. wont block the PR. :)

test/acsengine-features.groovy
colemickens: Thanks. Fixed with something better than `try/catch`. I've tested it as well.

test/acsengine-features.groovy
```diff
@@ -37,7 +37,10 @@ node {
                 env.SUBSCRIPTION_ID="${SUBSCRIPTION_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_ID}"
                 env.CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET="${CLUSTER_SERVICE_PRINCIPAL_CLIENT_SECRET}"
-                if(CUSTOM_HYPERKUBE_SPEC) {
+
+                // First check to see if var exists in context, then check for true-ness
+                // In Groovy, null and empty strings are false...
+                if(getBinding().hasVariable("myparameter") && CUSTOM_HYPERKUBE_SPEC) {
```
weinong: myparameter??

colemickens: Oh my... was doing something else at the same time... I only checked the negative case. That was bad, thanks for catching...

Updated PR has been pushed.

PR #406, opened 2017-03-15
Enable data disks encryption

clementnero:
Encryption option enabled by default on data disks' storage account.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/406)


anhowe:
Thank you @clementnero for suggesting this change.  However, currently we do not have this in scope for acs-engine.

Also, we will need some discussion around this change, including the following questions, to gain better understanding?

1. How should the API model change to accommodate this request?
1a. How should managed disks be modeled with subscriptions
1b. Do we only want to consider storage account encryption (where keys are managed by storage)? or do we want to consider the encryption extension where customers provide the keys, and how does that look in the model?
2. Why do we want encryption on by default?
3. What is the boot performance impact of encryption on cluster boot for K8S, DCOS, and Swarm?  For 100 agents?
4. What is the post boot performance impact of encryption on cluster boot for K8S, DCOS, and Swarm where disks are in use, for example a postgres database? (Post boot is interesting since this is where encryption starts happening).
5. Are ephemeral disks encrypted and how does this impact where we use ephemeral disks?

We don't necessarily have to implement the above questions, but it would be good to get a better understanding of the issues involved.

I'm going to close this PR, but please start an issue to track, and once a plan has been designed, feel free to re-open.

PR #411, opened 2017-03-15
Kubernetes VNET integration

ofiliz:
Enable complete Azure SDN support for containers:
- [x] Add "azure" to network policy options: Feature is currently disabled by default. Can be enabled by specifying networkPolicy: "azure" in orchestrator properties.
- [x] Allocate IP addresses for containers: Pre-provision IP addresses for containers from Azure VNET CA space. These addresses are allocated to containers on demand.
- [x] Disable route table, address spoofing and node CIDR allocation: These are no longer needed because SDN stack will own routing packets directly to containers.
- [x] Add network policy example files: Apimodel.json file for testing and to serve as an example.
- [x] Install Azure VNET CNI plugins: Install VNET CNI plugins when feature is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/411)


pkg/acsengine/const.go
```diff
@@ -1,55 +1,59 @@
...
-
-const (
-	//MsecndDCOSBootstrapDownloadURL Azure CDN to download DCOS1.7.3
-	MsecndDCOSBootstrapDownloadURL = "https://az837203.vo.msecnd.net/dcos/%s/bootstrap/%s.bootstrap.tar.xz"
-	//AzureEdgeDCOSBootstrapDownloadURL is the azure edge CDN download url
-	AzureEdgeDCOSBootstrapDownloadURL = "https://dcosio.azureedge.net/dcos/%s/bootstrap/%s.bootstrap.tar.xz"
-	//AzureChinaCloudDCOSBootstrapDownloadURL is the China specific DCOS package download url.
-	AzureChinaCloudDCOSBootstrapDownloadURL = "https://acsengine.blob.core.chinacloudapi.cn/dcos/%s.bootstrap.tar.xz"
-)
+package acsengine
```
colemickens: Looks like line endings changed here. This has happened before, I don't know how to prevent it because the `.gitattributes` should be forcing everything to LF anyway...

Anyway, can you tell me what the change is here to review?

pkg/acsengine/defaults.go
```diff
@@ -73,7 +73,9 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 	cloudSpecConfig := GetCloudSpecConfig(location)
 	if a.OrchestratorProfile.OrchestratorType == api.Kubernetes {
 		a.OrchestratorProfile.KubernetesConfig.KubernetesImageBase = cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
-		a.OrchestratorProfile.KubernetesConfig.NetworkPolicy = DefaultNetworkPolicy
```
colemickens: I am also confused as to how Calico could've worked... cc: @simonswine

pkg/api/vlabs/const.go
```diff
@@ -47,6 +47,10 @@ const (
 	MinDiskSizeGB = 1
 	// MaxDiskSizeGB specifies the maximum attached disk size
 	MaxDiskSizeGB = 1024
+	// MinIPAddressCount specifies the minimum number of IP addresses per network interface
+	MinIPAddressCount = 1
+	// MaxIPAddressCount specifies the maximum number of IP addresses per network interface
+	MaxIPAddressCount = 256
```
colemickens: Is it 256? I thought someone told me 250 at one point?

pkg/api/vlabs/const.go
```diff
@@ -67,3 +71,8 @@ const (
 	// ManagedDisks means that the nodes use managed disks for their os and attached volumes
 	ManagedDisks = "ManagedDisks"
 )
+
+// Network policy
+var (
+	NetworkPolicyValues = [...]string{"", "none", "azure", "calico"}
```
colemickens: Huh. I didn't even know that was valid syntax... `[]string{"literal", "literal"}` is more idiomatic Go.

pkg/acsengine/const.go
colemickens: Looks like it's probably just adding the `DefaultAgentIPAddressCount` and `DefaultAgentMultiIPAddressCount` ?

colemickens: Ignore, you pushed an update and this went away.

parts/kubernetesagentresourcesvmas.t
```diff
@@ -15,20 +15,26 @@
...
+            "name": "ipconfig{{$seq}}",
             "properties": {
+              {{if eq $seq 1}}
+              "primary": true,
+              {{end}}
               "privateIPAllocationMethod": "Dynamic",
               "subnet": {
-                "id": "[variables('{{.Name}}VnetSubnetID')]"
-             }
+                "id": "[variables('{{$.Name}}VnetSubnetID')]"
```
colemickens: Does this have different semantic meaning? `$.Name` vs `.Name`? If not, can you revert the change to this line for consistency's sake?

ofiliz: Yes. Inside the new "range" block, ".Name" now refers to the Name field of the inner iterator value (so $seq.Name) instead of AgentPool (AgentPool.Name). So the "$" is required to refer to AgentPool.Name by reaching it through the global variable "$".

pkg/api/vlabs/const.go
ofiliz: I don't think there is an actual upper limit enforced by Azure SDN stack. I thought putting some reasonable upper limit is a good idea so that customers don't consume all addresses and resources inside the VM by mistake. 256 seemed large enough to fit all cases, and small enough to not cause any out-of-resource issues.

256, not 250, because coders do it in hex :)

pkg/api/vlabs/const.go
ofiliz: I think [...] defines an array, where as [] defines a slice. They are equivalent, except my instincts tell me [...] should be slightly more performant. Anyway, I can convert it to [].

pkg/api/vlabs/const.go
ofiliz: You were right that there is a limit, but it is 256. It turns out NRP enforces 256 total IP addresses. I tried creating an agent pool with 256, it worked.

parts/kubernetesagentresourcesvmas.t
colemickens: 👍

pkg/api/vlabs/const.go
colemickens: Interesting. Never seen it like that anywhere before. I'll defer to you, seems like what you did was appropriate.

pkg/api/vlabs/const.go
colemickens: 👍

ofiliz:
CC @sharmasushant

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +103,70 @@ function setDockerOpts () {
...
-        if [[ "${NETWORK_POLICY}" = "calico" ]]; then
-            # download calico yamls
-            for addon in ${ADDONS}; do
-                curl -o "${ADDONS_PATH}/${addon}" -sSL --retry 12 --retry-delay 10 "${CALICO_URL}/${addon}"
-            done
-        else
-            # make sure calico yaml are removed
-            for addon in ${ADDONS}; do
-                rm -f "${ADDONS_PATH}/${addon}"
-            done
```
ofiliz: Note I have removed this for loop that removes yaml files. They seemed like dead code to me, because this provision script is run only once during master node first boot, and at that point we can't possibly have the yaml's in the add-on path anyway. It would also be harder to keep this loop in the new restructured function, so I removed it. Let me know if I am mistaken.

colemickens: I'm fine with it. I'd questioned the same thing when they sent the Calico PR.

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +103,70 @@ function setDockerOpts () {
...
 }
 
-function configNetworkPolicy() {
+function configAzureNetworkPolicy() {
+    # Download netconfig.
+    NETCONFIGDIR=/etc/cni/net.d
+    NETCONFIGFILE=$NETCONFIGDIR/10-azure.conf
+    mkdir -p $NETCONFIGDIR
+
+    echo "{
```
colemickens: nit: Consider using a heredoc: http://stackoverflow.com/questions/2953081/how-can-i-write-a-here-doc-to-a-file-in-bash-script

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +103,70 @@ function setDockerOpts () {
...
+" > $NETCONFIGFILE
+
+    chown -R root:root $NETCONFIGDIR
+    chmod 755 $NETCONFIGDIR
+    chmod 644 $NETCONFIGFILE
+
+    # Download Azure VNET CNI plugins.
+    CNIBINDIR=/opt/cni/bin
+    mkdir -p $CNIBINDIR
+    curl -fsSL --retry 12 --retry-delay 10 https://github.com/Azure/azure-container-networking/releases/download/v0.7/azure-cni-linux-amd64-v0.7.tgz | tar -xz -C $CNIBINDIR
```
colemickens: Can you do the retry externally, around curl? I know it sounds silly, but curl only retries on some subset, and has some weird behavior. It also doesn't really timeout on connection issues for a really long time, can you add a flag to curl for timeout as well?

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +103,70 @@ function setDockerOpts () {
...
+    curl -fsSL --retry 12 --retry-delay 10 https://github.com/containernetworking/cni/releases/download/v0.4.0/cni-amd64-v0.4.0.tgz | tar -xz -C $CNIBINDIR ./loopback
```
colemickens: Should we "externalize" these versions, at least to variables at the top of this script? `$CNI_VERSION` or something?

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +103,70 @@ function setDockerOpts () {
...
-            setNetworkPlugin kubenet
-            setDockerOpts ""
-        fi
+        setNetworkPlugin cni
+        setDockerOpts " --volume=/etc/cni/:/etc/cni:ro --volume=/opt/cni/:/opt/cni:ro"
+    fi
+}
+
+function configNetworkPolicy() {
+    if [[ "${NETWORK_POLICY}" = "azure" ]]; then
```
colemickens: yes, I like this better than where they were doing this `if` previously. thanks.

parts/kubernetesmastercustomscript.sh
ofiliz: Sure, I'll put the version numbers in variables in script. If you don't mind, I don't want to do the plumbing all the way from go templates just yet. I want to complete the Windows CNI part first, see what that looks like, and then come up with a unified solution that works for both.

parts/kubernetesmastercustomscript.sh
ofiliz: Thanks. Done.

parts/kubernetesmastercustomscript.sh
ofiliz: I am assuming you are referring to how you've done it in customdata yaml files? Sure, I'll add a utility wrapper for curl to do that.

parts/kubernetesmastercustomscript.sh
colemickens: Yup, that's all the more I was thinking for now anyway.

parts/kubernetesmastervars.t
```diff
@@ -58,6 +58,11 @@
 {{if AnyAgentHasDisks}}
     "dataStorageAccountPrefixSeed": 97,
 {{end}}
+{{if IsVNETIntegrated}}
+    "allocateNodeCidrs": false,
+{{else}}
+    "allocateNodeCidrs": true,
+{{end}}
```
amanohar: Small nit: Does it need an additional bool or can IsVNETIntegrated be used directly? Or is it because function cannot be used in the yml?

pkg/api/vlabs/validate.go
```diff
@@ -274,14 +277,21 @@ func (a *Properties) Validate() error {
...
 
-func (a *Properties) ValidateNetworkPolicy() error {
-
+func (a *Properties) validateNetworkPolicy() error {
 	if a.OrchestratorProfile.OrchestratorType != Kubernetes {
 		return nil
 	}
 
 	networkPolicy := a.OrchestratorProfile.KubernetesConfig.NetworkPolicy
-	if networkPolicy != "" && networkPolicy != "none" && networkPolicy != "calico" {
```
amanohar: This might fall through if a new network policy is added. Can you change this to: if networkPolicy == "azure"

pkg/api/vlabs/const.go
amanohar: What is the different between "" and "none"?

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +109,75 @@ function setDockerOpts () {
...
+	AZURE_PLUGIN_VERSION=v0.7
+
+    # Create network config file.
+    CNI_CONFIG_DIR=/etc/cni/net.d
+    NET_CONFIG_FILE=$CNI_CONFIG_DIR/10-azure.conf
+    mkdir -p $CNI_CONFIG_DIR
+
+    cat <<- EOF > $NET_CONFIG_FILE
+	{
+	  "cniVersion": "0.2.0",
```
ofiliz: This (0.2.0) is the CNI spec version that the plugin implements. CNI_VERSION (0.4.0) is the version of the CNI release. They are separate.

parts/kubernetesmastervars.t
ofiliz: AFAIK, function cannot be used in the yml.

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +109,75 @@ function setDockerOpts () {
...
-        else
-            # make sure calico yaml are removed
-            for addon in ${ADDONS}; do
-                rm -f "${ADDONS_PATH}/${addon}"
-            done
-        fi
+
+        # download calico yamls
+        for addon in ${ADDONS}; do
+            curl -o "${ADDONS_PATH}/${addon}" -sSL --retry 12 --retry-delay 10 "${CALICO_URL}/${addon}"
```
ofiliz: Maybe, I guess that is up to Calico to decide. Cole had asked me to do it with the for loop. I don't know whether Calico ever had a problem with retries or whether they would want that. Not really related to this PR.

pkg/api/vlabs/const.go
ofiliz: I guess "" is when there are no values specified. "none" is what acs-engine internally uses. In any case, those values were not introduced nor changed by this PR. I agree it would be nice to clean that up later.

pkg/api/vlabs/validate.go
ofiliz: Sorry, I don't understand. I am removing this line. The values are now stored in an array (in const.go), instead of hardcoded inline.

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +109,75 @@ function setDockerOpts () {
     sed -i "s#^DOCKER_OPTS=.*#DOCKER_OPTS=${1}#" /etc/default/kubelet
 }
 
-function configNetworkPolicy() {
+function configAzureNetworkPolicy() {
+	CNI_VERSION=v0.4.0
```
ofiliz: This is the CNI_VERSION that AzureNetworkPolicy uses. Other network policy providers may choose to use other versions. Hence it would be confusing to put it at the top of the script.

pkg/api/vlabs/validate.go
amanohar: My bad!

parts/kubernetesmastercustomscript.sh
amanohar: Can one of the var be renamed to be more descriptive and for the anyone who is new and might not know the difference when they are reading the code.

ofiliz: I have no control over "cniVersion" in the netconfig file. This is a well-known field defined in CNI spec.

I can change CNI_VERSION. In retrospect, CNI_RELEASE_VERSION would probably have been a better name. I will move this function to a common shell script when adding support for DC/OS in the next week or so. Do you mind if I rename this at that time? Would save us the time to retest the change.

amanohar: Ok. Can you add a comment describing the two are different?

ofiliz: Sure. Thanks.

pkg/api/vlabs/validate.go
```diff
@@ -274,14 +277,21 @@ func (a *Properties) Validate() error {
...
 }
 
-func (a *Properties) ValidateNetworkPolicy() error {
-
+func (a *Properties) validateNetworkPolicy() error {
 	if a.OrchestratorProfile.OrchestratorType != Kubernetes {
 		return nil
 	}
 
 	networkPolicy := a.OrchestratorProfile.KubernetesConfig.NetworkPolicy
```
anhowe: Add validation to ensure this only is set when orchestratorType is Kubernetes, and it is a non-windows cluster.

parts/kubernetesagentcustomdata.yml
```diff
@@ -9,6 +9,7 @@ runcmd:
 - echo "deb {{{dockerEngineDownloadRepo}}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
 - "echo \"Package: docker-engine\nPin: version {{{dockerEngineVersion}}}\nPin-Priority: 550\n\" > /etc/apt/preferences.d/docker.pref"
 - apt-get update
+- apt-get install -y ebtables
```
anhowe: Is this needed if not using CNI?  should this be wrapped in an if statement

examples/networkpolicy/kubernetes-azure.json
```diff
@@ -0,0 +1,38 @@
+{
+  "apiVersion": "vlabs",
```
anhowe: Can you add and test an example with multiple agent pools.

parts/kubernetesagentresourcesvmas.t
```diff
@@ -15,22 +15,31 @@
...
+              }
             }
           }
-        ],
+          {{if lt $seq $.IPAddressCount}},{{end}}
+          {{end}}
+        ]
+{{if not IsVNETIntegrated}}
```
anhowe: Why IsVNETIntegrated, and not IsCNIEnabled or IsAzureNetworkPolicy?

parts/kuberneteskubelet.service
```diff
@@ -10,6 +10,7 @@ ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
 ExecStartPre=/bin/mkdir -p /var/lib/kubelet
 ExecStartPre=/bin/mount --bind /var/lib/kubelet /var/lib/kubelet
 ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
+ExecStartPre=-/sbin/ebtables -t nat --list
```
anhowe: Is this needed for non-azure network policy and can it be wrapped in an if block

parts/kubernetesmastercustomdata.yml
```diff
@@ -15,6 +15,7 @@ runcmd:
 - echo "deb {{{dockerEngineDownloadRepo}}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
 - "echo \"Package: docker-engine\nPin: version {{{dockerEngineVersion}}}\nPin-Priority: 550\n\" > /etc/apt/preferences.d/docker.pref"
 - apt-get update
+- apt-get install -y ebtables
```
anhowe: wrap in if statement

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +109,75 @@ function setDockerOpts () {
...
+	EOF
+
+    chown -R root:root $CNI_CONFIG_DIR
+    chmod 755 $CNI_CONFIG_DIR
+    chmod 644 $NET_CONFIG_FILE
+
+    # Download Azure VNET CNI plugins.
+    CNI_BIN_DIR=/opt/cni/bin
+    mkdir -p $CNI_BIN_DIR
+    downloadUrl https://github.com/Azure/azure-container-networking/releases/download/$AZURE_PLUGIN_VERSION/azure-cni-linux-amd64-$AZURE_PLUGIN_VERSION.tgz | tar -xz -C $CNI_BIN_DIR
```
anhowe: only download if needed.

parts/kubernetesmastercustomscript.sh
```diff
@@ -95,6 +95,12 @@ function ensureKubectl() {
     fi
 }
 
+function downloadUrl () {
```
anhowe: This is a great improvement.  Please adapt all places in this script to use download URL (I looked and there is only one other).

parts/kubernetesmastercustomscript.sh
```diff
@@ -103,32 +109,75 @@ function setDockerOpts () {
...
     else
         # on agents
-        if [[ "${NETWORK_POLICY}" = "calico" ]]; then
-            setNetworkPlugin cni
-            setDockerOpts " --volume=/etc/cni/:/etc/cni:ro --volume=/opt/cni/:/opt/cni:ro"
-        else
-            setNetworkPlugin kubenet
-            setDockerOpts ""
-        fi
+        setNetworkPlugin cni
```
anhowe: only call if required.

parts/kubernetesagentresourcesvmas.t
colemickens: CNI enabled is true when Calico is set, but that's not the case here. "AzureNetworkPolicy" is not an accurate descriptor either. That having been said, `$.NetworkPlugin == "Azure"` might be a more valid way of specifying this condition.

parts/kubernetesagentcustomdata.yml
colemickens: I'd suggest doing the conditional logic with bash... in fact I think you basically have to.

parts/kubernetesagentresourcesvmas.t
ofiliz: The check we are making here is whether we are using native VNET routing and addresses, or the RouteTable + AddressSpoofing that we used until now. This is really independent of whether you are using Azure CNI plugin or Azure network policy.

Consider these two scenarios:
1) CNI may also be used by other components, like Calico. You can have "CNIEnabled" and yet still be using RouteTable + AddressSpoofing. This is what Calico does today. Hence IsCNIEnabled is not suitable.

2) Similarly, IsAzureNetworkPolicy is not suitable, because you may not be using Azure network policy, yet use VNET routing. This is what I expect Calico (or other 3rd parties) will use in the future.

So the name should not be tied to CNI or AzureNetworkPolicy. This is an independent switch that just happens to be used only by Azure CNI plugin for now.

pkg/api/vlabs/validate.go
colemickens: (I think this is a new bit of validation that theoretically should've been added when Calico introduced the KubernetesConfig field.

Basically we need additional type checking of sorts to say `OrchestratorProfile` can only contain `KubernetesConfig` when `OrchestratorProfile.OrchestratorType == Kubernetes`.

pkg/acsengine/const.go
```diff
@@ -15,12 +15,16 @@ const (
 	DefaultFirstConsecutiveKubernetesStaticIP = "10.240.255.5"
 	// DefaultAgentSubnetTemplate specifies a default agent subnet
 	DefaultAgentSubnetTemplate = "10.%d.0.0/16"
```
anhowe: Now with CNI, we can max out our nodes.  There are 12 agent pools, and each can have 100 nodes.  When CNI is chosen, we should add these under a separate range from the master IPs, and make that range support 78,000, perhaps a /15 (12 pools * 100 machines * 65 addresses per host (machine + 64) == 76,800).

parts/kubernetesmastercustomscript.sh
ofiliz: Discussed with Anthony offline. Yes, that's only called when feature is enabled.

parts/kubernetesmastercustomscript.sh
ofiliz: Discussed with Anthony offline. Yes, that's only called when feature is enabled (if networkpolicy=="azure").

parts/kubernetesmastercustomdata.yml
ofiliz: We discussed and it doesn't look like there is a way to add a check here. The provision script can't contain apt commands due to race conditions, either.

parts/kuberneteskubelet.service
ofiliz: This isn't needed if the feature is disabled. The "-" in front ensures any errors will be ignored, so it has no side effects, even if ebtables is not even installed.

pkg/api/vlabs/validate.go
ofiliz: Sure, agreed. I'll add a check in OrchestratorProfile.Validate() to cover both me and Calico.

examples/networkpolicy/kubernetes-azure.json
ofiliz: Adding kubernetes-azure-hybrid example with two agent pools: one Windows, one Linux.

parts/kubernetesmastercustomscript.sh
ofiliz: Ok, @amanohar had asked for the same thing a few comments above. I'll add it, but again, there are no test cases for Calico. For regression testing, all I can do is to check if the addons are downloaded to that file path. I have no other means to test if something is broken because of this change.

parts/kubernetesmastercustomscript.sh
ofiliz: Because @anhowe has asked me to touch this file for some other change anyway, I'll do this now. No need to postpone.

pkg/acsengine/const.go
ofiliz: Updated the default IP address count per network interface to 50, so that even the largest Kubernetes cluster with 12 pools * 100 nodes * 50 IP addresses = 60000 still fits a single /16 subnet. Added comments to the code to document how we decided on this default value.

ofiliz:
@colemickens @anhowe @amanohar Thanks for the great feedback.

I grouped changes related to the extra validation and pre-existing issues together and made a separate commit. I'd appreciate it if you do NOT squash these commits, to clarify the reasoning for future reviewers.

Hopefully all issues are addressed now and we can merge this soon :)

examples/networkpolicy/kubernetes-azure-hybrid.json
```diff
@@ -0,0 +1,52 @@
...
+        "ipAddressCount": 20,
+        "osType": "Linux"
+      },
+      {
+        "name": "windowspool2",
+        "count": 2,
+        "vmSize": "Standard_DS3_v2",
+        "availabilityProfile": "AvailabilitySet",
+        "ipAddressCount": 30,
+        "osType": "Windows"
```
anhowe: Isn't this going to hit the validation error?

The intention of the request is to ensure this change works on > 1 agent pool.  So if you don't want to put a multi linux agent pool in the examples folder, put in the test folder.

parts/kubernetesmasterresources.t
```diff
@@ -316,7 +316,7 @@
             }
           }
 {{if IsVNETIntegrated}}
-          {{range $seq := loop 2 64}}
+          {{range $seq := loop 2 50}}
```
anhowe: define a function in engine.go, and pull this value from const.go.

colemickens: Why does it need to be a function? Can we please make it a part of the internal model rather than defining more small template functions?

examples/networkpolicy/kubernetes-azure-hybrid.json
ofiliz: It is going to hit an error only for a few days, until I send the Windows PR later this week. But sure, I'll convert the second pool to Linux for now.

parts/kubernetesmasterresources.t
ofiliz: Made master IPAddressCount part of internal model.

examples/networkpolicy/kubernetes-azure-hybrid.json
```diff
@@ -0,0 +1,52 @@
...
+      "dnsPrefix": "",
+      "vmSize": "Standard_D2_v2"
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "linuxpool1",
+        "count": 2,
+        "vmSize": "Standard_D2_v2",
+        "availabilityProfile": "AvailabilitySet",
+        "ipAddressCount": 20,
```
anhowe: during our scrum @colemickens commented that this will have to be configured as a pod restriction.  Otherwise, differing ip address counts will cause intermittent failures.  In this example, one agent from pool2 will be able to take a 21st node, and the second agent from pool1 will not, and it will be not be clear why.

colemickens: see here: https://kubernetes.io/docs/admin/kubelet/

`      --max-pods int32                                          Number of Pods that can run on this Kubelet. (default 110)`

Hopefully this is an easy change to make. Just plumbing the agent pool CA address count into a new flag on kubelet. It can just default to 110 if not in CNI if that makes it easier to add this flag.

ofiliz: As discussed, I am using a single large /12 subnet for all masters, agents and pods when VNET integration is enabled. This allows us to allocate 128 IP addresses per node, more than the default --max-pods value of 110, fixing the root cause. This also allows creation of huge Kubernetes clusters with 1200 nodes now and more in the future.

colemickens:
(I don't understand why you made the changes in this commit: a3e61bb007881e7491b1daaf816f71d6fdd018e7 I preferred it the other way. Those orchestrator specific options are eventually going to have lots of things inside. A quick check against nil seems clearer than checking for any of the fields being set.)

ofiliz:
@colemickens I agree, I also liked the nil check. However, that solution would cause a regression where acs-engine crashes when a KubernetesConfig is not specified. Basically I had to insert tens of "if Orchestrator.KubernetesConfig != nil" checks everywhere, even in some part files. Too much code churn. I agree the nil check is better, but we should do it in a separate PR.

colemickens:
Got it, I concur.

anhowe:
@acs-bot test this please

weinong:
@acs-bot test this please

anhowe:
@acs-bot test this please

anhowe:
@acs-bot test this please

anhowe:
@acs-bot test this please

colemickens:
Hitting merge... it looks green to me. 
screen shot 2017-03-23 at 7 56 18 pm

Issue #311, opened 2017-02-21
Stamp git information into the binary


colemickens:
We already prescribe that people should be building in Docker. I think we should go all-in on requiring this to be built under Linux, using the Makefile.

In the Makefile, we can stamp information about the current git state into the binary. Then `acs-engine --version` can return the git revision it was built with.

Further, once this PR is in (https://github.com/Azure/acs-engine/pull/253) git workspaces won't get dirty from just regenerating (unless there are real changes) so we can just trust the dirty/clean bit to know if they had local changes.

This is probably preferrable to having to guess: "What version are you on? Are you absolutely sure you rebuilt? Did you also re-generate?" etc...

colemickens:
See #356 for an example of the wrapper script. It might be time to introduce a `build/` or `hack/` directory and put such a script there. And then maybe the Makefile just shells out for everything, and we keep the Makefile as an easy entry point and for tab completion.

anhowe:
@raveeram fixed this in #412

PR #412, opened 2017-03-16
Use Git SHA-1 hash as version for acs-engine

raveeram:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/412)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,16 @@
 
 .PHONY: prereqs build test test_fmt validate-generated fmt lint ci devenv
 
+VERSION=`git describe --always --long --dirty`
+BUILD=`date +%FT%T%z`
+
 prereqs:
 	go get github.com/jteeuwen/go-bindata/...
 
 build: prereqs
 	go generate -v ./...
 	go get .
-	go build -v
+	go build -v -ldflags="-X main.AcsEngineBuildSHA=${VERSION} -X main.AcsEngineBuildTime=${BUILD}"
 	cd test/acs-engine-test; go build -v
 
 test: prereqs test_fmt
--- a/acs-engine.go
+++ b/acs-engine.go
@@ -125,6 +125,7 @@ func usage(errs ...error) {
 	fmt.Fprintf(os.Stderr, "\n")
 	fmt.Fprintf(os.Stderr, "options:\n")
 	flag.PrintDefaults()
+	fmt.Fprintf(os.Stderr, "\n%s --version or -v to get the build version \n", os.Args[0])
 }
 
 var noPrettyPrint = flag.Bool("noPrettyPrint", false, "do not pretty print output")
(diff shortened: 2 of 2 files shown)


raveeram:
#311

weinong:
@acs-bot test this please

Makefile
```diff
@@ -2,13 +2,15 @@
 
 .PHONY: prereqs build test test_fmt validate-generated fmt lint ci devenv
 
+VERSION=`git describe --always --long --dirty`
```
weinong: it doesn't print the full sha-1 hash. or is it on purpose? i like the full sha-1, though...

acs-engine.go
```diff
@@ -132,12 +132,25 @@ var artifactsDir = flag.String("artifacts", "", "directory where artifacts will
...
+// AcsEngineBuildSHA is the Git SHA-1 of the last commit
+var AcsEngineBuildSHA string
+
 // acs-engine takes the caKey and caCert as args, since the caKey is stored separately
 // from the api model since this cannot be easily revoked like the server and client key
 var caCertificatePath = flag.String("caCertificatePath", "", "the path to the CA Certificate file")
 var caKeyPath = flag.String("caKeyPath", "", "the path to the CA key file")
 
 func main() {
+	if (len(os.Args) == 2) &&
```
weinong: Please add --version to the usage flag.

acs-engine.go
```diff
@@ -132,12 +132,25 @@ var artifactsDir = flag.String("artifacts", "", "directory where artifacts will
 var classicMode = flag.Bool("classicMode", false, "enable classic parameters and outputs")
 var parametersOnly = flag.Bool("parametersOnly", false, "only output the parameters")
 
+// AcsEngineBuildSHA is the Git SHA-1 of the last commit
+var AcsEngineBuildSHA string
```
weinong: since we are adding commit hash, i highly suggest to add build time, too. It's a lot easier to see how dated the binary is.

acs-engine.go
colemickens: Hm, did you intentionally not use the `flag` package as we do for our other flags?

rjtsdl: +1 on this question

Makefile
rjtsdl: Use `git rev-parse HEAD`

acs-engine.go
raveeram: Ah yeah, it was intentional. I felt like the existing flags are tightly coupled with the apimodel.json side of things.

Makefile
raveeram: @weinong Yes it was on purpose. Cole and I discussed and felt like this should be sufficient. I don't feel strongly either way...

acs-engine.go
raveeram: Added.

acs-engine.go
raveeram: Added

weinong:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/412)*

PR #415, opened 2017-03-16
Add Extensions functionality

RobBagby:
- Allow 1st and 3rd party to write extensions to acs-engine.
- To add a hello world extension, user would add the following to their
template
- "extensionsProfile": [{ "name": "hello-world-k8s", "version": "v1" }],
- extensions are really just linked templates that call a custom script
extension
- the magic is in the script that is called - written by extension
author
- only enabled in k8s for this release - will add to others shortly
- examples/kubernetes.json illustrates hello world - to test - you will
have to add rootUrl
that points to "https://bagbyimages.blob.core.windows.net:443/".
rootUrl allows for
testing of extensions prior to the extension being accepted into
acs-engine
- docs/extensions.md describes how to consume, as well as author
extensions

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/415)


pkg/api/types.go
```diff
@@ -44,6 +44,7 @@ type Properties struct {
 	AgentPoolProfiles       []AgentPoolProfile      `json:"agentPoolProfiles"`
 	LinuxProfile            LinuxProfile            `json:"linuxProfile"`
 	WindowsProfile          WindowsProfile          `json:"windowsProfile"`
+	ExtensionsProfile       []ExtensionProfile      `json:"extensionsProfile"`
```
amanohar: @RobBagby, I am starting to review this PR and also read your mail (thanks for the detailed description). However, I am still not convinced that the extension profile should be at the root level without describing what its targeting. 

The current approach certainly provides flexibility but at the same time makes the model of the cluster very implicit & harder to read at a glance. Also, this modeling diverges from the principle with which ACS API was designed of being very explicit of the cluster architecture. 

E.g. the fact the extension will only run on the first master is a design detail of the extension (and dependent on extension author's design choice). However, the cluster creator can still choose to say that they want to run this extension only on the master pool. And the way to describe this in the model is to either have extension profile at the root level (like it is today) refer explicitly to its target pools or each agent/master pool supports extension profile. 

Having extension profile at the root level provides advantage of targeting more than one pool at a time.

Thoughts?

amanohar: If this is targeting only Kubernetes for now please add a check for this (and it can be removed later when other orchestrators are supported).

amanohar: However, I do recognize that the extension author can base their extension code on any parameter in the template and choose to noop on other pools. This certainly causes more confusion when looking that the API model.

RobBagby: Hi - we did add a check for this.  That is the role of supported-orchestrators.json.  That file contains a json array of supported orchestrators for the extension.  There is a function orchestratorSupportsExtension that validates that the configured orchestrator is in that array.

RobBagby: Also, I have created another branch and have DC/OS support completed.  After this PR is complete, I will issue another PR with DC/OS

shrutir25: @amanohar how do we want to proceed with this PR ?

shrutir25:
@RobBagby We have made changes to the acs-engine repo, where we have removed test comparison with the expected files. Please pull the latest changes and resubmit.

skasturi:
DC/OS as well please.

shrutir25:
@RobBagby  - we will be closing this PR since there has been no response from you

shrutir25:
@RobBagby We have made changes to the acs-engine repo, where we have removed test comparison with the expected files. Please pull the latest changes and resubmit.

RobBagby:
I am pulling the PR.  I will issue another that has taken into account all of the changes in acs-engine.

annegi:
Is this available for all acs-engine versions?

PR #417, opened 2017-03-20
Adding pre exec check for kubelet service, to prevent exponential mou…

rushabhnagda11:
There are a couple of issues related to this
https://github.com/kubernetes/kubernetes/issues/39508 -> specifically on azure
https://github.com/kubernetes/kubernetes/issues/29059 -> on gke, might not be entirely related

I think it is related with kubelet restarts on acs-engine.

**Steps to reproduce:**

run `systemctl restart kubelet` 32 times. After a certain no of times, it should give you an error "failed to mount /var/lib/kubelet no space left on device"

This is because /var/lib/kubelet is already mounted on the first start, now on further restarts, the existing mount is also remounted. Further restarts lead to an exponential increase in the number of mounts.

**Fix:**
The pre exec check for this service should also check if /var/lib/kubelet is already mounted. This PR should fix this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/417)


colemickens:
It's not clear to me that this is the cause of kubelet remounting the secret over and over in #29059.

To understand more, you're seeing this by SSHing to the node and issuing `systemctl restart kubelet` a significant number of times?

rushabhnagda11:
>> To understand more, you're seeing this by SSHing to the node and issuing systemctl restart kubelet a significant number of times?

Yes. One of the reasons I have to do this is because, whenever I edit the kube-apiserver.yaml the kubelet/apiserver don't restart. I have to manually do a systemctl restart kubelet.

colemickens:
Can you run `./scripts/rebase.sh` to fixup the templates, etc? I'd like to merge this soon. Thanks.

BTW, there's an upstream issue for kube-apiserver not restarting when editting the pod manifests. It's not got a lot of attention though. I filed it if you want to look for it.

colemickens:
:/ Did you actually try this? You can't just pass literal shell syntax to systemd like that... I'm fixing it in #486...

colemickens:
This was addressed as part of #486.

rushabhnagda11:
Hey sorry, was on vacation. Thanks for fixing this issue! :)

Issue #350, opened 2017-03-02
Should setup container log retention policy on master & agent node


yangl900:
## Problem
Master node stops working for everything and claim etcd failure or it is not configured correctly.

## Root Cause
Looking into logs we found that the master node container log has taken 27 GB and used all disk space in OS disk. The cluster has been running for ~2 months.

## Proposed Fix
Setup log retention policy so we never run out of disk space.

colemickens:
@yangl900 Ah, were the huge files: `/var/lib/docker/containers/[CONTAINER ID]/[CONTAINER_ID]-json.log`?

We should mitigate this short term with some sort of log rotations, long term presumably we'd want to run fluentd w/ logrotate and push cluster logs to some storage system...

yangl900:
That's correct. Note that the file is holding by the running container, we had to restart in order to remove the file completely.

@brendandburns @anhowe as FYI, I think we should take this as a high priority fix. we cannot run cluster in production without the rotation.

brendandburns:
Yes, we should add this ASAP.  The 'logrotate' tool is a good simple tool to add.

--brendan


________________________________

rjtsdl:
I am looking into this problem.

Currently, the reason that docker daemon created multiple `/var/lib/docker/containers/[CONTAINER ID]/[CONTAINER_ID]-json.log` files, is because, docker daemon is using the default log-driver : json-file. 

Now i have two options for the short term purpose:
1. Keep using json-file log-driver for docker, then we can leverage docker's option like `max-size` to limit the size.

2. Change log-driver to journald, in this way, we can leverage journald's log rotation strategy to manage docker along with other daemons. 

I personally prefer option 2. However, i may overlook some issues, such as backward compatibility.

colemickens:
I like option 2 as well, and I doubt we're going to end up supporting any distros that don't use systemd anyway.

tekmaven:
This may become a problem on Windows Agents I'm assuming, so that needs to be considered.

rjtsdl:
#373 

@anhowe @JiangtianLi , how do you guys think about Windows Agents concern?

colemickens:
It'd be one thing if if the Windows Service was harvesting the output of `kubelet` on windows, but instead we're writing out to a text file manually. Does NSSM take care of pumping stdout/err to Windows logs? If so, that might be easiest.

JiangtianLi:
[...]

PR #444, opened 2017-03-29
make kubernetes nodes logrotate by journald

rjtsdl:
I double checked dcos's logrotation strategry. It is very similar with what we have now for k8s. The difference is dcos's kern.log is not growing as fast as k8s. So rotate every week, with 4 weeks kern.log doesn't hit the disk limit. For k8s, i need to force restart systemd-journald.service to load my customized conf file, and it will not forward to syslog. All the logs are managed by journald.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/444)


rjtsdl:
@acs-bot test this please

rjtsdl:
Fix for #350 , see my description.

parts/kubernetesmastercustomscript.sh
```diff
@@ -298,6 +304,7 @@ ensureDocker
 configNetworkPolicy
 ensureKubelet
 extractKubectl
+ensureJournal
```
rjtsdl: I found, if i don't do reload and restart journal, it will not take my customized config.

rjtsdl:
@anhowe @colemickens 
Ping on review.

PR #455, opened 2017-03-30
switch to using pointers to structs in fields

JackQuincy:
This is so that we can omit fields in output in the managed azure service.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/455)

### DIFF
--- a/pkg/api/convertertoapi.go
+++ b/pkg/api/convertertoapi.go
@@ -20,13 +20,17 @@ func ConvertV20160930ContainerService(v20160930 *v20160930.ContainerService) *Co
 	c.ID = v20160930.ID
 	c.Location = v20160930.Location
 	c.Name = v20160930.Name
-	convertV20160930ResourcePurchasePlan(&v20160930.Plan, &c.Plan)
+	if v20160930.Plan != nil {
+		c.Plan = &ResourcePurchasePlan{}
+		convertV20160930ResourcePurchasePlan(v20160930.Plan, c.Plan)
+	}
 	c.Tags = map[string]string{}
 	for k, v := range v20160930.Tags {
 		c.Tags[k] = v
 	}
 	c.Type = v20160930.Type
-	convertV20160930Properties(&v20160930.Properties, &c.Properties)
+	c.Properties = &Properties{}
+	convertV20160930Properties(v20160930.Properties, c.Properties)
 	return c
 }
 
@@ -36,13 +40,17 @@ func ConvertV20160330ContainerService(v20160330 *v20160330.ContainerService) *Co
 	c.ID = v20160330.ID
 	c.Location = v20160330.Location
 	c.Name = v20160330.Name
-	convertV20160330ResourcePurchasePlan(&v20160330.Plan, &c.Plan)
+	if v20160330.Plan != nil {
+		c.Plan = &ResourcePurchasePlan{}
+		convertV20160330ResourcePurchasePlan(v20160330.Plan, c.Plan)
+	}
 	c.Tags = map[string]string{}
 	for k, v := range v20160330.Tags {
 		c.Tags[k] = v
 	}
 	c.Type = v20160330.Type
-	convertV20160330Properties(&v20160330.Properties, &c.Properties)
+	c.Properties = &Properties{}
+	convertV20160330Properties(v20160330.Properties, c.Properties)
 	return c
 }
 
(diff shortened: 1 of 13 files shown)


JackQuincy:
@acs-bot test this please

pkg/api/vlabs/validate.go
```diff
@@ -25,7 +25,7 @@ func (o *OrchestratorProfile) Validate() error {
 	}
 
 	if o.OrchestratorType != Kubernetes &&
-		(o.KubernetesConfig.KubernetesImageBase != "" || o.KubernetesConfig.NetworkPolicy != "") {
+		(o.KubernetesConfig != nil) {
```
amanohar: The case of KubernetesConfig being present and KubernetesImageBase/NetworkPolicy  should also be handled.

weinong:
Reviewed 6 of 13 files at r1.
Review status: 6 of 13 files reviewed at latest revision, 4 unresolved discussions.

---

*[pkg/api/convertertoapi.go, line 391 at r1](https://reviewable.io:443/reviews/azure/acs-engine/455#-KgWJNjSHkcy_1BvEQmX:-KgWJNjSHkcy_1BvEQmY:b-i17vjz) ([raw file](https://github.com/azure/acs-engine/blob/16421b9bb1cf73021eff33cf423ad6954d7531d9/pkg/api/convertertoapi.go#L391)):*
> ```Go
> func convertV20160330VMDiagnostics(v20160330 *v20160330.VMDiagnostics, api *VMDiagnostics) {
> 	api.Enabled = v20160330.Enabled
> 	api.StorageURL = v20160330.StorageURL
> ```

StorageURL needs to be a pointer, and copies when it's not nil.

---

*[pkg/api/types.go, line 182 at r1](https://reviewable.io:443/reviews/azure/acs-engine/455#-KgWIslv-PasjmOcHQKB:-KgWIslv-PasjmOcHQKC:b7bb0ew) ([raw file](https://github.com/azure/acs-engine/blob/16421b9bb1cf73021eff33cf423ad6954d7531d9/pkg/api/types.go#L182)):*
> ```Go
> 	// This field is readonly as ACS RP will create a storage account
> 	// for the customer.
> 	StorageURL neturl.URL `json:"storageUrl"`
> ```

I think it needs to be a pointer

---

*[pkg/api/v20160330/types.go, line 146 at r1](https://reviewable.io:443/reviews/azure/acs-engine/455#-KgWIn6CLF3UtC0TA2a-:-KgWIn6CLF3UtC0TA2a0:b7bb0ew) ([raw file](https://github.com/azure/acs-engine/blob/16421b9bb1cf73021eff33cf423ad6954d7531d9/pkg/api/v20160330/types.go#L146)):*
> ```Go
> 	// This field is readonly as ACS RP will create a storage account
> 	// for the customer.
> 	StorageURL neturl.URL `json:"storageUrl"`
> ```

I think it needs to be a pointer

---

*[pkg/api/v20170131/types.go, line 159 at r1](https://reviewable.io:443/reviews/azure/acs-engine/455#-KgWIRplyRaVKMmAOyuY:-KgWIRpmnTZQz9zhba5t:bdn56xi) ([raw file](https://github.com/azure/acs-engine/blob/16421b9bb1cf73021eff33cf423ad6954d7531d9/pkg/api/v20170131/types.go#L159)):*
> ```Go
> 	// This field is readonly as ACS RP will create a storage account
> 	// for the customer.
> 	StorageURL neturl.URL `json:"storageUrl"`
> ```

I think it needs to be a pointer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/455)*

pkg/api/vlabs/validate.go
```diff
@@ -263,6 +263,9 @@ func (a *Properties) validateNetworkPolicy() error {
 
 	switch a.OrchestratorProfile.OrchestratorType {
 	case Kubernetes:
+		if a.OrchestratorProfile.KubernetesConfig == nil {
+			return nil
+		}
```
amanohar: Should it return nil or empty since Go's default string value is empty?

pkg/api/vlabs/validate.go
JackQuincy: good catch

pkg/api/vlabs/validate.go
JackQuincy: this is returning an error as this is validation. so returning nil means this is a valid value. I'll skip the rest of this case which will leave networkPolicy as "" which is how this method behaves today.

JackQuincy:
Good idea on the more simple test, that doesn't depend on our test inputs as much. I ran classic/kubernetes.classic.json through and didn't see unexpected objects.

pkg/api/convertertoapi.go
```diff
@@ -307,7 +382,8 @@ func convertV20160930VMDiagnostics(v20160930 *v20160930.VMDiagnostics, api *VMDi
 }
 
 func convertV20160330DiagnosticsProfile(v20160330 *v20160330.DiagnosticsProfile, api *DiagnosticsProfile) {
-	convertV20160330VMDiagnostics(&v20160330.VMDiagnostics, &api.VMDiagnostics)
+	api.VMDiagnostics = &VMDiagnostics{}
+	convertV20160330VMDiagnostics(v20160330.VMDiagnostics, api.VMDiagnostics)
```
weinong: nil check on v20160330.VMDiagnostics?

JackQuincy: It doesn't makes sense to not have it if you have a diagnostics profile, it is the only field. So I turned off omit empty. But I guess we could be called without it.

weinong: stupid GitHub doesn't allow me to leave comments elsewhere. I meant nil check on v20160330.VMDiagnostics inside convertV20160330VMDiagnostics

JackQuincy:
@acs-bot test this please

PR #462, opened 2017-04-03
Making converting diagnostics profiles not throw nil pointer panics

JackQuincy:
I made a bad assumption about the state of the target object( thought we had serialized it from json).

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/462)


JackQuincy:
@acs-bot test this please

PR #466, opened 2017-04-04
Sizes issue 352

shrutir25:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/466)


shrutir25:
@anhowe - could you take a look ?

parts/dcosagentresourcesvmss.t
```diff
@@ -189,7 +189,7 @@
       "sku": {
         "capacity": "[variables('{{.Name}}Count')]",
         "name": "[variables('{{.Name}}VMSize')]",
-        "tier": "Standard"
+        "tier": "[variables('{{.Name}}VMSizeTier')]"
```
colemickens: Is this change still needed if we're not supporting 'Basic'?

colemickens: Or is this so that everything still works even if the customer manually tries to edit and use a 'Basic_X' vm size?

anhowe: Even though we don't use basic, there may be new sizes in the future, and the error produced is hard to troubleshoot, so we should fix it.

pkg/acsengine/Get-AzureConstants.ps1
```diff
@@ -27,7 +30,7 @@ Get-AllSizes() {
 		#Write-Output $location.Location
 		$sizes = Get-AzureRmVMSize -Location $location.Location
 		ForEach ($size in $sizes) {
-			if (!$sizeMap.ContainsKey($size.Name)) {
+			if (!$sizeMap.ContainsKey($size.Name) -and !($size.Name.split('_')[0] -eq 'BASIC')) {
```
anhowe: 1. Consider case insensitive comparison
2. add a comment to why basic is being filtered out (no LB support)

parts/dcosagentresourcesvmss.t
colemickens: 👍 thanks!

pkg/acsengine/Get-AzureConstants.ps1
shrutir25: As discussed, sizes are only in 'Basic' format, so updated the if condition accordingly.

shrutir25:
@anhowe  I have made the changes. Could you take a look ?

anhowe:
I just noticed one thing, @shrutir25 please scrub examples and tests to make sure no basic is listed.

JackQuincy:
most recent change looks good. I'll merge this.

dmitsh:
Did we run Jenkins PR validation job for this PR?

JackQuincy:
Apparently not my bad. I will send out a PR to get us back in a good state

dmitsh:
It is rather my bad :) I thought we configured GitHub not to allow PRs without building them first ...

PR #467, opened 2017-04-04
[Kubernetes] Move etcd to attached persistent volumes on master VMs

amanohar:
- Move etcd to attached persistent volume on master VMs in a Kubernetes cluster
- Add VM name hash as a tag to the VMs

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/467)


amanohar:
@acs-bot test this please

parts/kubernetesmasterresources.t
```diff
@@ -395,6 +396,7 @@
           {{end}}
         },
         "storageProfile": {
+          {{GetEtcdDisk}}
```
colemickens: Do we really need a new template function for this? :( Is this because it needs information that's not available in the model passed to the template here? Looks like maybe the OrchType or MasterProfile?

amanohar: @colemickens, valid point. I thought I was following/supposed to follow this pattern to get a disk but there is no need for this, I will add the JSON directly to the template file.

pkg/acsengine/engine.go
```diff
@@ -984,6 +987,27 @@ func getDataDisks(a *api.AgentPoolProfile) string {
...
+              "createOption": "Empty",
+              "diskSizeGB": "%d",
+              "lun": %d,
+              "name": "[concat(variables('masterVMNamePrefix'), copyIndex(),'-etcddisk')]",
+              "vhd": {
+                "uri": "[concat(reference(concat('Microsoft.Storage/storageAccounts/',variables('masterStorageAccountName')),variables('apiVersionStorage')).primaryEndpoints.blob,'vhds/',    variables('masterVMNamePrefix'),copyIndex(),'-etcddisk.vhd')]"
+               }
+            }`
+
+	buf.WriteString(fmt.Sprintf(etcdDisks, 1, 0))
```
anhowe: Do we know the size of etcd in a 5000 node cluster or what etcd will grow to?  Perhaps we can look at Dmitry's 1200 node cluster.

Let's determine what to do in each of the standard and premium cases.  Based on pricing (https://azure.microsoft.com/en-us/pricing/details/managed-disks/):
1-128GB premium disk is $19.71 /month
1GB standard will be about $0.43 / month

I think it will be safe to go with 128GB for both.

pkg/acsengine/engine.go
```diff
@@ -984,6 +987,27 @@ func getDataDisks(a *api.AgentPoolProfile) string {
...
+	etcdDisks := `            {
+              "createOption": "Empty",
+              "diskSizeGB": "%d",
+              "lun": %d,
+              "name": "[concat(variables('masterVMNamePrefix'), copyIndex(),'-etcddisk')]",
+              "vhd": {
+                "uri": "[concat(reference(concat('Microsoft.Storage/storageAccounts/',variables('masterStorageAccountName')),variables('apiVersionStorage')).primaryEndpoints.blob,'vhds/',    variables('masterVMNamePrefix'),copyIndex(),'-etcddisk.vhd')]"
+               }
+            }`
+
```
anhowe: For lun, add as constant, with comment on why it must be lun0 (https://github.com/anhowe/scratch/tree/master/simplelinux-attachblankdisk#lun-0-issues)

pkg/acsengine/engine.go
```diff
@@ -984,6 +987,27 @@ func getDataDisks(a *api.AgentPoolProfile) string {
...
+	buf.WriteString(fmt.Sprintf(etcdDisks, 1, 0))
+	buf.WriteString("\n          ],")
+	return buf.String()
```
anhowe: Another consideration is do we want to enable encryption at rest by default for etcd?  There have been a few questions around this on the issues.  I'm not sure what this does performance wise.

parts/kubernetesmastercustomdata.yml
```diff
@@ -7,8 +7,29 @@ packages:
...
+- /bin/echo DAEMON_ARGS=--name "<<<variables('masterVMNames')[copyIndex()]>>>" --initial-advertise-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --listen-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --advertise-client-urls "<<<variables('masterEtcdClientURLs')[copyIndex()]>>>" --listen-client-urls "<<<concat(variables('masterEtcdClientURLs')[copyIndex()], ',http://127.0.0.1:', variables('masterEtcdClientPort'))>>>" --initial-cluster-token "k8s-etcd-cluster" --initial-cluster "<<<variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]>>> --data-dir "/var/lib/etcddisk"" --initial-cluster-state "new" | tee -a /etc/default/etcd
+- sudo /bin/chmod -R 0777  /var/lib/etcddisk
+- systemctl stop etcd
+- sudo rm -rf /var/lib/etcd/default
+- sudo rm -rf /var/lib/etcddisk/member
+- systemctl restart etcd
+- until curl http://127.0.0.1:2379/v2/machines; do sleep 2; done
```
amanohar: Sure, there is only one other curl pattern in the code. I can switch to using that.

parts/kubernetesmastercustomdata.yml
```diff
@@ -7,8 +7,29 @@ packages:
...
+- /bin/echo DAEMON_ARGS=--name "<<<variables('masterVMNames')[copyIndex()]>>>" --initial-advertise-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --listen-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --advertise-client-urls "<<<variables('masterEtcdClientURLs')[copyIndex()]>>>" --listen-client-urls "<<<concat(variables('masterEtcdClientURLs')[copyIndex()], ',http://127.0.0.1:', variables('masterEtcdClientPort'))>>>" --initial-cluster-token "k8s-etcd-cluster" --initial-cluster "<<<variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]>>> --data-dir "/var/lib/etcddisk"" --initial-cluster-state "new" | tee -a /etc/default/etcd
+- sudo /bin/chmod -R 0777  /var/lib/etcddisk
+- systemctl stop etcd
+- sudo rm -rf /var/lib/etcd/default
```
amanohar: No its not. I was starting the basic PR of replacing the current etcd dir with this and the current script does a remove so I was matching that behavior. I was then going the change this in subsequent PRs as upgrade script begins to take form to see how this will need to evolve since I didn't want to do all the upgrade design up front. But I am happy to remove this code for now and revisit later.

raveeram: IIRC, we were doing this earlier to cleanup the default data dir created by the systemd instance before cloud-init runs.

parts/kubernetesmasterresources.t
anhowe: Correct, functions are only needed where it is not possible to get information from the model, or the formatting is not easy to express in the templates (bz, managed disks vs, std disk formate, etc).

pkg/acsengine/engine.go
amanohar: Updated. 

I don't know the size of etcd in a 5000 cluster. But for a 3 master, 2 agent cluster its 3MB. I anticipate this will grow to some degree over time as well. Even if the etcd size grows linearly with each new node added we should be within the 128 GB limit. Though it would be interesting to find out what 1200 node cluster shows about etcd size change..

parts/kubernetesmastercustomdata.yml
colemickens: Oh, I didn't realize etcd was activating itself by default after install, so weird that Ubuntu does that.

I'm fine with it either way, just wanted to understand. If this mirrors current behavior and you're iterating on it next, it seems fine to leave it as is.

parts/kubernetesmastercustomdata.yml
```diff
@@ -7,8 +7,29 @@ packages:
...
+- /bin/echo DAEMON_ARGS=--name "<<<variables('masterVMNames')[copyIndex()]>>>" --initial-advertise-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --listen-peer-urls "<<<variables('masterEtcdPeerURLs')[copyIndex()]>>>" --advertise-client-urls "<<<variables('masterEtcdClientURLs')[copyIndex()]>>>" --listen-client-urls "<<<concat(variables('masterEtcdClientURLs')[copyIndex()], ',http://127.0.0.1:', variables('masterEtcdClientPort'))>>>" --initial-cluster-token "k8s-etcd-cluster" --initial-cluster "<<<variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]>>> --data-dir "/var/lib/etcddisk"" --initial-cluster-state "new" | tee -a /etc/default/etcd
+- sudo /bin/chmod -R 0777  /var/lib/etcddisk
```
brendandburns: +1, we should `chown` this to the `etcd` user...  (assuming that such a user exists, and if it doesn't we should create/use an `etcd` user first :)

parts/kubernetesmastercustomdata.yml
brendandburns: What is the path back? e.g. what happens if this breaks? Shouldn't we move the directory to the side rather than just blowing it away?

parts/kubernetesmastercustomdata.yml
brendandburns: this makes me slightly nervous since it goes for forever. Perhaps a "really long" (e.g. 1 hour?) timeout?

brendandburns:
Meta-comment: This is a lot of commits. We should squash it down to 1-2 before merging.

parts/kubernetesmastercustomdata.yml
amanohar: I have updated this to add a long timeout.

amanohar:
@brendandburns, yes I will squash all commits when I merge.

parts/kubernetesmastercustomdata.yml
amanohar: Peeking into the logs it looks like it does:
__Adding system user `etcd' (UID 113) ...
Adding new group `etcd' (GID 117) ...
Adding new user `etcd' (UID 113) with group `etcd' ...
Creating home directory `/var/lib/etcd/' ..._
Adding system user `etcd' (UID 113) ...
Adding new group `etcd' (GID 117) ...
Adding new user `etcd' (UID 113) with group `etcd' ...
Creating home directory `/var/lib/etcd/' ..._

I will update this.

pkg/acsengine/engine.go
amanohar: Did you mean add a variable to the template and pass a value by setting the parameter with the const? Is that really necessary?

pkg/acsengine/engine.go
amanohar: I will open a GitHub issue to track this as I will need investigate the performance hit.

amanohar:
@acs-bot test this please

pkg/acsengine/engine.go
amanohar: Update: Ran into errors when attaching a 128 GB disk. Related issue: https://github.com/Azure/WALinuxAgent/issues/651

I will update once I have more information from WALinuxAgent team.

amanohar:
@acs-bot test this please

amanohar:
@anhowe, @colemickens  please review when you get a chance.

JackQuincy:
on the squashing. Do we want to do that in github ui, squash and commit, or in git command line with "git rebase -i"  in the general case?

colemickens:
Not sure it matters a ton. The end result should be the same either way.

parts/kubernetesmastercustomscript.sh
```diff
@@ -169,12 +169,6 @@ function configNetworkPolicy() {
     fi
 }
 
-function ensureEtcd() {
```
amanohar: I have verified by creating a file at the end of a successful etcd setup and then reading it during CSE run. Synchronization is not an issue here since waagent logs still point to the same issue and seem to fail before: https://github.com/Azure/WALinuxAgent/issues/651. 

In the meantime I think this would be a good check to add anyways to CSE run to make sure etcd is up so I am adding it and will update this PR.

colemickens: I thought the VM wasn't finished provisioning until CustomData was finished executing... don't we have this problem in other places?

amanohar: Yeah that's my impression as well on discussing this further with the Linux team that CSE does not run until cloud init finishes. 

Plus regarding:

> now that this is removed from the script, and since CSE is started in parallel, what is the synchronization point between the etc setup finishing, and this script starting?

My understanding is that this was never mean for synchronization in the first place. But just to be sure I will add the check.

amanohar:
@acs-bot test this please

amanohar:
@anhowe, I have updated the PR to enable 128 GB etcd disk. Please review.

parts/kubernetesmastercustomdata.yml
```diff
@@ -7,8 +7,30 @@ packages:
...
 
+disk_setup:
+  /dev/sdc:
+      table_type: gpt
+      layout: true
+      overwrite: true
+
+fs_setup:
+- special:
+  cmd: /sbin/mkfs.ext4 /dev/sdc1 -F -E lazy_itable_init=1,lazy_journal_init=1
```
colemickens: Is this the ultimate fix for the overall issue? How fragile is this? Is there a longer term fix for there being some sort of race condition here?

colemickens: It's interesting that this is the fix... this is what we had to do in k8s too because mkfs.ext4 was being awfully slow in certain conditions.

amanohar: I don't think this is fragile based on my understanding and reading of this. Any reason why it would be?

This is lazy formatting where enough blocks are written but complete disk is not formatted with the rest happening in the background when its first mounted. It was taking about 8 mins for the formatting to finish for a 128 GB disk that caused Linux Agent to time out.

Another way to fix this would be to not do time consuming operations like setting up etcd and apt-get during cloud init. Instead kick of another script during cloud init that runs in the background and read the status during CSE run. 

However, at this time I don't see this as a blocking issue or we could go back to using a smaller etcd disk.

amanohar:
@colemickens, I am merging as Anthony signed off on the latest iteration. Let me know if you have any other feedback.

parts/kubernetesmastercustomdata.yml
colemickens: Discussed offline. Sounds good to me.

PR #469, opened 2017-04-04
update 2016-03-30 conversion to allow public agent pools

JackQuincy:
Also added test json data to make sure we run these. Need to think about
what is the right way to unit test these changes

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/469)


JackQuincy:
@acs-bot test this please

PR #471, opened 2017-04-05
Proposal: Docker for mac setup

tripdubroot:
When using docker for mac I get an error:

docker: Error response from daemon: error while creating mount source path '/usr/local/bin/docker': chown /usr/local/bin/docker: no such file or directory.

I've added a Dockerfile that installs docker.io into the acs-engine container. I've also added a new devenv build script that builds the new Dockerfile and removes mounting the docker cli into the container.

Docs updated to reflect changes.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/471)


JackQuincy:
@colemickens @rjtsdl have you had to do something similar to work on mac?

colemickens:
I don't understand why this is needed at all. Why is Docker needed inside the dev container?

colemickens:
Sorry, I grok it now. And even if Docker was able to see the full FS properly, the binary for Mac obviously won't be usable inside the container... whoops...

Can you go ahead and make these changes to the main Dockerfile/devenv.sh script? I'd prefer to not have to maintain two sets.

tripdubroot:
@colemickens I will update now.

colemickens:
One last request, I've been burnt by Ubuntu's docker package too many times. Do you mind changing it to do an official install? I don't mind if you do the `curl | bash` or if you add the gpg key + apt repo and then use apt directly.

colemickens:
Thank you for fixing this by the way, I really want everyone to be on the same page and use the Docker container, so I appreciate having this fixed for Mac-based contributors!

tripdubroot:
@colemickens sure I'll update the file to curl|Bash.

tripdubroot:
@colemickens updated Dockerfile to use official install. Tested OK after rmi relevant images and new build executed.

Image build is pretty lengthy.

colemickens:
@acs-bot test this please

Issue #437, opened 2017-03-28
Doc update: distinguish between custom VNET *within* 10.0.0.0/8 and otherwise


rocketraman:
Based on the docs, I was under the impression that one had to go through a lot of manual work or use code in out-of-date pull request [1] to deploy Kubernetes to *any* VNET other than 10.0.0.0/8. However, after trying this for myself, I now realize deploying with a custom VNET that uses some subset of the 10.0.0.0/8 address space e.g. 10.1.0.0/16 works quite well already, by following the "Custom VNET" instructions (and tempory workaround [2]) to preconfigure the VNET and associated address space.

It would be useful to clearly distinguish in the docs between the second simple case using a subnet of 10.0.0.0/8, and the first currently not-so-simple case of using an entirely different address space.

[1] https://github.com/Azure/acs-engine/issues/180, https://github.com/Azure/acs-engine/issues/303, https://github.com/Azure/acs-engine/pull/191

[2] Step 7 @ https://github.com/Azure/acs-engine/blob/master/docs/kubernetes.md

colemickens:
Can you show your changes? We hardcode the master ips to something that I don't believe falls in 10.1.0.0/16, so I'm not sure how it would be working.

I'm not really sure I understand how/why the docs are confusing, could you possibly suggest the change with a PR?

rocketraman:
> Can you show your changes?

I followed the instructions on ACS-Engine for a custom VNET. 

1. Create a vnet in a resource group -- I used 10.2.0.0/16 for this install.

2. Update `kubernetes.json` as per the instructions in the CustomVNET directory of ACS-Engine:
2a. Include the path to the vnet created above in `properties.masterProfile.vnetSubnetId`
2b. Set `properties.masterProfile.firstConsecutiveStaticIP` to `10.2.255.239` (an IP inside the 10.2.0.0/16 subnet -- this is probably the key step to avoid the master IP hard-coding)
2c. Set `agentPoolProfiles[*].vnetSubnetId`

3. Deploy the templates generated by ACS-Engine to the same resource group as the custom vnet was originally created in

4. Finally, update the created routetable with the subnet information as instructed in temporary step 7.

[...]

PR #473, opened 2017-04-06
k8s in custom Vnet - configurable kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp

### LINKED ISSUE #180: k8s: Make kubeClusterCidr, kubeServiceCidr, kubeDnsServiceIp configurable for custom VNet

ashb:
This is a rebased version of #191.

The one change from the original pull request is where the config lives: rather than having a "kubeNetworkConfig" at the top level, I have addeded 3 keys to the existing orchestratorProfile.kubernetesConfig structure.

Closed #180, #191, #437

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/473)


pkg/api/converterfromapi.go
```diff
@@ -285,6 +285,17 @@ func convertOrchestratorProfileToV20160330(api *OrchestratorProfile, o *v2016033
...
 	o.OrchestratorType = vlabs.OrchestratorType(api.OrchestratorType)
+
+	if api.KubernetesConfig != nil {
+		o.KubernetesConfig = &vlabs.KubernetesConfig{}
+		convertKubernetesConfigToVLabs(api.KubernetesConfig, o.KubernetesConfig)
+	}
+}
+
+func convertKubernetesConfigToVLabs(api *KubernetesConfig, vlabs *vlabs.KubernetesConfig) {
+	vlabs.DnsServiceIp = api.DnsServiceIP
```
JackQuincy: seems like this is missing the inverse to these lines from convertertoapi.go

api.KubernetesImageBase = vlabs.KubernetesImageBase		  	
api.NetworkPolicy = vlabs.NetworkPolicy

could you add them please

pkg/acsengine/defaults.go
```diff
@@ -209,6 +220,24 @@ func setDefaultCerts(a *api.Properties) (bool, error) {
...
+	_, serviceCidr, err := net.ParseCIDR(a.OrchestratorProfile.KubernetesConfig.ServiceCIDR)
+	if err != nil {
+		return false, err
+	}
+	// Based off https://play.golang.org/p/m8TNTtygK0
+	// Increment the last octet if it's zero.
+	firstIP := func(ip net.IP) net.IP {
+		for j := len(ip) - 1; j >= 0; j-- {
+			if ip[j] == 0 {
+				ip[j]++
```
JackQuincy: Can you explain the need for this first ip? We didn't have this idea before. Was it covered from the first master ip?

JackQuincy:
Also huge thanks for picking this work up.

JackQuincy:
@acs-bot test this please

pkg/acsengine/defaults.go
ashb: I can now having played with it a bit more.

There is a service called `kubernetes` created in every cluster, and it has as it's IP the first IP in the service CIDR range. By default this is 10.0.0.1 (the default range is 10.0.0.0/16). Previously this was handled in CreatePki (and removed in this diff in the only changes in `pkg/acsengine/pki.go`), but rather than passing another argument in for it to then add to the IPs I handle it here.

The reason I do the ParseCIDR and then add the net is to be more lenient in the input - for instance this would cope with:

- 192.168.1.128/25 (192.168.1.128 would be the serviceIp)
- 192.168.1.1/25 or 192.168.1.0/25, both yielding 192.168.1.1 as the service IP etc.

ashb:
I'll make the review feedback as requested in individual fixup! commits so it's hopefully easier to check the fixes, and then will squash to cleanliness before a final merge.

pkg/acsengine/const.go
```diff
@@ -29,6 +29,15 @@ const (
 	DefaultInternalLbStaticIPOffset = 10
 	// DefaultNetworkPolicy is disabling network policy enforcement
 	DefaultNetworkPolicy = "none"
+	// DefaultKubernetsDnsServiceIP specifies the IP address that kube-dns
+	// listens on by default. must by in the default Service CIDR range.
+	DefaultKubernetsDnsServiceIP = "10.0.0.10"
```
colemickens: There's a lot of `Kubernets` in here. (Missing the last `e`: `kubernets vs kubernetes`).

pkg/acsengine/defaults.go
```diff
@@ -80,6 +80,17 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 		if a.OrchestratorProfile.KubernetesConfig.NetworkPolicy == "" {
 			a.OrchestratorProfile.KubernetesConfig.NetworkPolicy = DefaultNetworkPolicy
 		}
+
+		if a.OrchestratorProfile.KubernetesConfig.DnsServiceIP == "" {
+			a.OrchestratorProfile.KubernetesConfig.DnsServiceIP = DefaultKubernetsDnsServiceIP
+		}
+		if a.OrchestratorProfile.KubernetesConfig.ServiceCIDR == "" {
+			a.OrchestratorProfile.KubernetesConfig.ServiceCIDR = DefaultKubernetsServiceCIDR
```
colemickens: Can you validate that the DnsServiceIP falls into the ServiceCIDR? (and isn't the .255 since I think you got tripped up on that)

ashb: I can do that, and yes good call :)

ashb: To make the validation a code nicer I'm going to mandate that dnsServiceIp and serviceCIDR have to both be specified if either is. 

(Otherwise the validation code would have to know about the default values - validation is currently done on the versioned API model, which hasn't yet had the defaults applied.)

colemickens:
@acs-bot test this please

ashb:
@colemickens @JackQuincy Thanks! All squashed to clean commits and conflicts resolved.

JackQuincy:
@acs-bot test this please

pkg/acsengine/testdata/vnet/kubernetesvnet.json
```diff
@@ -2,7 +2,12 @@
   "apiVersion": "vlabs",
   "properties": {
     "orchestratorProfile": {
-      "orchestratorType": "Kubernetes"
+      "orchestratorType": "Kubernetes",
+      "kubernetesConfig": {
+        "dnsServiceIp": "172.17.1.10",
```
amanohar: To make the naming convention consistent can you update this to: dnsService**IP** instead of dnsService**Ip** like in  firstConsecutiveStaticIP

ashb: Good call, have done.

ashb:
@colemickens Any chance of a merge today?

JackQuincy:
@acs-bot test this please

JackQuincy:
We can merge once we pass ci/cd tests from @acs-bot

pkg/acsengine/pki.go
```diff
@@ -117,8 +117,6 @@ func createCertificate(commonName string, caCertificate *x509.Certificate, caPri
 		template.KeyUsage |= x509.KeyUsageCertSign
 		template.IsCA = isCA
 	} else if isServer {
-		extraIPs = append(extraIPs, net.ParseIP("10.0.0.1"))
```
anhowe: Why was this removed?  Kubeproxy still writes an ip table rule, and kubedns relies on this.

anhowe: @ashb Why was this removed? Kubeproxy still writes an ip table rule, and kubedns relies on this.

colemickens: because it's done here: https://github.com/Azure/acs-engine/pull/473/files#diff-8edb2653431a5da9ad1f4286097702a7R224

Again, otherwise this wouldn't work in the first place.

anhowe: Thanks, I've read through the changes and they look good.

lshahar:
I added to kubernetes.json example the following configuration under orchestratorProfile section
                "kubernetesConfig": {
                  "dnsServiceIP": "172.17.1.10",
                  "serviceCidr": "172.16.0.0/14",
                  "clusterCidr": "172.40.0.0/16"
 And try to generate the configuration by ./acs-engine example/kubernetes.json but under _output I still got configuration with 10.240.x, there is anymore configuration that should be done?

ashb:
This will not change the network space the k8s creates. If you want to spin it up on a completely different subnet you will have to create the VNet and subnet yourself and use the "external VNet" example. Be careful that the subnet you choose doesn't overlap with the Docker subnet otherwise things might get.... interesting? (It might not, but the default docker subnet is 172.17.0.0/16 so....)

colemickens:
To clarify,

@Ishahar, those options let you change:

* the ClusterIP used for KubeDNS. This must be determined ahead of time so that it can be configured on all `kubelet`s for the `--cluster-dns` flag.
* the service CIDR - a fake range used for ClusterIPs (the former must fall in this range)
* the cluster CIDR - the pod cidr used for the cluster

None of those change the actual IPs of the nodes. In that case, you'd want to do as @ashb suggests and use the existing vnet scenario examples (though now I'm wondering how that works, given that the PKI assets need to have the master IP known ahead of time... how does that even work @ashb ?)

colemickens:
I guess in the existing vnet scenario, the assumption is still that k8s nodes can have free range on 10.x? I think this PR is not quite enough then, and we need to go further to allow flexibility of the node IP ranges as well.

Or is this PR enough to let me peer (or deploy to an existing vnet) with 10.x.x.x. stuff already in it, @ashb?

ashb:
I'm just testing spinning up a k8s cluster in a VNet with an address space of 10.105.248/21, and I think, one bug with routetable association (debugging and will open an issue if so) it works and I can spin up a small cluster in a IP space fully under my control.

(The other downside right now is that the nodes are claiming a /24 for pods each where as I would really like it if I could make it /25 or /26. Again, different issue I think)

ashb:
The PKI needs the master IP, but the master IP is predictable within the subnet namespace -- it is statically assigned based on `masterProfile.firstConsecutiveStaticIP` (which needs to be 14 IPs from the last IP in the subnet in practice for the internalLbIp to be valid). 

It's probably a bit error prone and must be configured exactly right, and I wouldn't want to gerenate the certs for it myself, but it is doable.

PR #475, opened 2017-04-06
☁️  mounting .azure

sozercan:
Mounting .azure so user is logged in to az inside the container

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/475)

### DIFF
--- a/scripts/devenv.sh
+++ b/scripts/devenv.sh
@@ -3,6 +3,8 @@
 set -eu -o pipefail
 set -x
 
+sudo -u $(logname) mkdir -p ~/.azure
+
 docker build --pull -t acs-engine .
 
 docker run -it \
--- a/scripts/devenv.sh
+++ b/scripts/devenv.sh
@@ -3,8 +3,6 @@
 set -eu -o pipefail
 set -x
 
-sudo -u $(logname) mkdir -p ~/.azure
-
 docker build --pull -t acs-engine .
 
 docker run -it \
@@ -15,3 +13,4 @@ docker run -it \
 	-w /gopath/src/github.com/Azure/acs-engine \
 		acs-engine /bin/bash
 
+chown -R $(logname):$(id -gn $(logname)) . ~/.azure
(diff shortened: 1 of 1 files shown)


scripts/devenv.sh
```diff
@@ -9,6 +9,7 @@ docker run -it \
 	--privileged \
 	-v /var/run/docker.sock:/var/run/docker.sock \
 	-v `pwd`:/gopath/src/github.com/Azure/acs-engine \
+	-v ~/.azure:/root/.azure \
```
rjtsdl: This is the devenv.sh in acs-engine.

The line you added implies that 
1. ~/.azure file exist
2. user need az cli

For users want to run this script first, and then login to az cli. It will break, because ~/.azure never exists.

colemickens:
My only concern is:

1. user has never logged into CLI
2. user enters container
3. user logs in
4. user exits container
5. `az` tries to refresh an access token or something, and now it can't write to the files in ~/.azure because they're owned by root since they were created in the container.

Not sure if it's enough to block this change, but wanted to mention it.

scripts/devenv.sh
rjtsdl: Noticed if file missing, docker run -v will create it for u.
Could you add something before running this docker run, something like `touch ~/.azure`. It will also help to resolve the issue which Cole pointed out, docker created the file with root permission.

sozercan: Interesting, looks like this happens when running `devenv.sh` with `sudo` and not added to docker group.

Just pushed a change to make sure it's running as user, let me know if this works.

colemickens:
This can still wind up with users who have `~/.azure/accessTokens.json` owned by root, which will be awkward, but this is already true for any files created in container, like `_output/*` etc... so I'm going to go ahead and give my approval since this is probably a win for acs-engine usability.

colemickens:
I guess maybe adding a `sudo -u $(loguser) chown -R $(loguser) .` or whatnotafter the `docker run` might help a tiny bit?

sozercan:
@colemickens `chown` after `docker run` seems to be a better solution, let me know if this works

PR #478, opened 2017-04-08
QoS: Change how provisioning blocking works

colemickens:
Three things:

1. Fixes `AZURE_CONFIG_DIR` that has been sorta broken since I added it probably. Only noticeable in some rare cases.

2. Moves us to the 2.0 version of CustomScript for Linux to avoid this issue: https://github.com/Azure/azure-linux-extensions/issues/367

3. Fixes the QoS impacting issue so that SPN failures will propagate up as deployment failures:

   i. makes the master CSE run in blocking mode so that failure is propagated up
   ii. makes it so that Windows agent provision in the background via use of `Start-Job` with PowerShell

4. Only wait 5 minutes for apiserver instead of 10.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/478)


colemickens:
@acs-bot test this please

colemickens:
This was tested with a bad SP. The deployment fails fast, meaning that the windows agents didn't block the deployment, and it shows that the master CSE failed properly as a result of apiserver not coming up. These were the stated objectives for this particular QoS bug.

colemickens:
cc: @anhowe

colemickens:
The deployment failure manifests like this. There doesn't appear to be a way to get CSE to propagate a proper error message all the way up out to the CLI, unfortunately, so this is the best we can do from a UX perspective.

```
+ az group deployment create --name colemick-58e830a2 --resource-group colemick-58e830a2 --template-file /mnt/persistentdisk/home/cole/code/colemickens/acs-engine_gopath/src/github.com/Azure/acs-engine/test/../_output/colemick-58e830a2/azuredeploy.json --parameters @/mnt/persistentdisk/home/cole/code/colemickens/acs-engine_gopath/src/github.com/Azure/acs-engine/test/../_output/colemick-58e830a2/azuredeploy.parameters.json
At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details. {
  "status": "Failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "VMExtensionProvisioningError",
        "message": "VM has reported a failure when processing extension 'cse0'. Error message: \"Enable failed: failed to execute command: command terminated with exit status=1\"."
      }
    ]
  }
}  Correlation ID: 15445a1b-3eed-4dbe-91ed-28e5d49bec66
```

colemickens:
@acs-bot test this please

colemickens:
cc: @dmitsh 

So, when the deploy script was cut up, the AZURE_CONFIG_DIR stopped actually working. When I fix it here, it ends up breaking Jenkins... I suspect that Jenkins is calling these scripts in such a way that the AZURE_CONFIG_DIR is set for the login, but not set when running the actual deployment. 

I can't really trace through all of the steps that are going on in this job anymore. There's a shell script in Jenkins which runs Docker which runs a shell script, which runs Go code, which later runs the bash scripts in `test/`. Can you take a look and see where the `export AZURE_CONFIG_DIR...` should go such that it will be in effect for the login and the actual deployment step(s)?

Thanks.

colemickens:
@acs-bot test this please

colemickens:
Okay, well, I just moved it so that AZURE_CONFIG_DIR is only set/used for the all-in-one deploy.sh script. Jenkins is proceeding now at least...

colemickens:
@acs-bot test this please

parts/kubernetesmastervars.t
```diff
@@ -178,7 +178,7 @@
     "agentWindowsSku": "2016-Datacenter-with-Containers",
     "agentWindowsVersion": "2016.0.20170127",
     "singleQuote": "'",
-    "windowsCustomScriptSuffix": " $inputFile = '%SYSTEMDRIVE%\\AzureData\\CustomData.bin' ; $outputFile = '%SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.ps1' ; Copy-Item $inputFile $outputFile ; Invoke-Expression('{0} {1}' -f $outputFile, $arguments) ; "
+    "windowsCustomScriptSuffix": " $inputFile = '%SYSTEMDRIVE%\\AzureData\\CustomData.bin' ; $outputFile = '%SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.ps1' ; Copy-Item $inputFile $outputFile ; Start-Job -command { Invoke-Expression('{0} {1}' -f $outputFile, $arguments) } ; "
```
amanohar: With the update to running this in the background with Start-Job how does the failure propagate back up and fail the operation?

colemickens: It's intentional - that was the discussed desired behavior:
* master blocks deployment if apiserver fails to come up
* all agent nodes run in the background and don't fail the deployment

Running in non-blocking mode and being able to send a signal back up for the deployment are basically mutually exclusive.

amanohar: Yes, I agree that they are mutually exclusive. I wanted to understand the intent of the change. Although I do think that there should be some kind of validation that at least one agent node is up but that can be discussed outside of this PR.

parts/kubernetesmastercustomscript.sh
```diff
@@ -215,7 +215,7 @@ function ensureJournal(){
 
 function ensureApiserver() {
     kubernetesStarted=1
-    for i in {1..600}; do
+    for i in {1..300}; do
```
anhowe: I would prefer to keep this at 10 minutes, for any slowness in container download or read/write latencies from disk.

colemickens:
@acs-bot test this please

colemickens:
@acs-bot test this please

colemickens:
@acs-bot test this please

PR #483, opened 2017-04-11
eliminate throttling issue

dmitsh:
eliminate throttling issue by
not generating /etc/kubernetes/azure.json on the worker nodes

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/483)


parts/kuberneteskubelet.service
```diff
@@ -37,8 +37,6 @@ ExecStart=/usr/bin/docker run \
         --cluster-domain=cluster.local \
         --register-schedulable=${KUBELET_REGISTER_SCHEDULABLE} \
         --node-labels=${KUBELET_NODE_LABELS} \
-        --cloud-provider=azure \
```
JackQuincy: @colemickens seemed pretty set that this should be an option and not the default. That it would cause issues that a user would need to opt into.

dmitsh:
@acs-bot test this please

colemickens:
I do have reservations about this. This completely breaks FD-based scheduling in the kube-scheduler. I think this is something we should document, rather than merge. Or make it an option of the agentPoolProfile

rjtsdl:
Is this the recommended way to solve throttling issue? And I don't see the issue referenced. We may want to first create the issue.

amanohar:
@anhowe, I have updated the PR to enable 128 GB etcd disk. Please review.

PR #477, opened 2017-04-07
fix kubelet.service template

vmakhaev:
Missing [Install] section prevents creating symlink while `systemctl enable kubelet`. Service does not start after node restart.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/477)

PR #486, opened 2017-04-12
Various fixes - kubelet restart, /var/lib/kubelet re-bind-mount, k8s flag updates, docker to leave iptables alone

colemickens:
Lots of things here, trying to tidy up:

1. Pull in #477 (fixes kubelet service file to restart, looks like a regression from Calico PR) and regenerate templates. (Thanks @vmakhaev!)
2. Pull in #417 (fixes a problem that can occur when restarting apiserver a lot) and regenerate templates. (Thanks @rushabhnagda11 / @Prokure!)
3. Fixes docker so that it will leave iptables + ip masq alone. (I did NOT up us to a newer version of Docker as they've not been tested with K8s)
4. Updates some of the Kubernetes flags to be more modern and avoid deprecation warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/486)


colemickens:
@acs-bot test this please

colemickens:
@acs-bot test this please

colemickens:
@acs-bot test this please

parts/kubernetesagentcustomdata.yml
```diff
@@ -29,7 +29,10 @@ write_files:
   content: |
     [Service]
     ExecStart=
-    ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay --log-driver=journald
+    ExecStartPre=-/usr/bin/iptables -t nat -F
```
anhowe: how does this interact with kube-proxy

colemickens:
@anhowe Really great question. I'm actually going to change this around. We should be dropping the docker/kubernetes override files ahead of actually installing the various packages. In this way, we will prevent Docker from ever touching the various tables, so we don't have to worry about kubelet restarting and clobbering existing nat table entries.

Will do this, test and update the PR very shortly.

colemickens:
@anhowe I reset your approval, please take a look at the last commit which does what I mentioned in my previous comment.

colemickens:
Alright, Build #142 was for the current latest commit. I had to fix one additional bug unfortunately, around kubelet stopping... it looks like it was probably never properly fixed. Wasted more time than I should've on this and wound up just marking exit status 143 as a success. I don't know if docker is doing the wrong thing, or it kubelet is not doing the right thing. I suspect docker. Whatever, the kubelet stop is fine, and this prevents the unit from going into a failed state and alarming users.

Aaaand going to merge this.

PR #494, opened 2017-04-12
fix DNSName regex

rjtsdl:
UpperCase in DNSPrefix is allowed in production.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/494)


colemickens:
I pulled that regex from the azure error message, plus the portal disagrees: 
untitled


Or is this like a really recent development or?

rjtsdl:
I think i miss read the behavior. My DNSPrefix is somehow having UpperCases now, which shouldn't. I assume production accepting that before. Looks it is not the case. I will close this PR.

rjtsdl:
I need to do more to let this match ACS current behavior. I will send updates shortly. And i want to reopen this.

rjtsdl:
In our parts/, the generated template will do tolower for DNSNamePrefix. And we do allow user to specify DNSPrefix with uppercases in their api model. So when we want to validate, we also need to allow uppercases.

parts/kubernetesmastervars.t
```diff
@@ -21,7 +21,7 @@
     "servicePrincipalClientId": "[parameters('servicePrincipalClientId')]",
     "servicePrincipalClientSecret": "[parameters('servicePrincipalClientSecret')]",
     "username": "[parameters('linuxAdminUsername')]",
-    "masterFqdnPrefix": "[toLower(parameters('masterEndpointDNSNamePrefix'))]",
```
rjtsdl: other places using tolower instead of toLower. I just want to keep it consistent.

pkg/api/v20160330/validate.go
```diff
@@ -122,7 +122,7 @@ func validatePoolName(poolName string) error {
 }
 
 func validateDNSName(dnsName string) error {
-	dnsNameRegex := `^([a-z][a-z0-9-]{1,45}[a-z0-9])$`
```
colemickens: Is there a validate function for vlabs too?

rjtsdl: yes, i overlooked vlab version. 
Just updated, and also correct a little bit to match the error message.

colemickens: >also correct a little bit to match the error message

How's that?

rjtsdl: the change is, {1,45} -> {1,43}

As the message is saying "The DNS name must contain between 3 and 45 characters", we need to count the first and last ones.

colemickens: 👍

Issue #493, opened 2017-04-12
Regression to azure network plugin introduced by latest CIDR changes


anhowe:
@ashb @ofiliz - the latest CIDR changes have broken the azure network plugin.  Deploying a standard template repros crash looping of the dns and dashboard processes.

The logs of DNS show the following error:

E0412 04:27:57.862786       1 reflector.go:214] pkg/dns/dns.go:156: Failed to list *api.Service: Get https://10.0.0.1:443/api/v1/services?resourceVersion=0: dial tcp 10.0.0.1:443: i/o timeout

ashb:
That normally appears at the start of the log for a few times. Just to confirm: that just keeps happening?

ofiliz:
If anyone actively investigating this? If you can't figure it out, please undo this change.

ofiliz:
We've discovered that the azure-vnet CNI plugin suffers from a problem similar to #439. I'll fix that first, then check if it fixes this problem as well.

ashb:
Is there anything I can do to help this along? It's not affecting me but I'm keen to not see my previous merge backed out :)

rocketraman:
I'm having the same issue. I don't see the error message shown by @anhowe on my setup, but the behavior is exactly the same -- the DNS and dashboard pods are in a crash loop (several hours now).

@ashb Here is what I am doing -- perhaps you can repro in your environment:

1. Create a custom vnet using [this template](https://github.com/Azure/acs-engine/blob/master/examples/vnet/vnetarmtemplate/azuredeploy.kubernetes.json) -- modify the `addressPrefix` in two places to 10.10.0.0/16.

2. Create an engine template with the following `orchestratorProfile`:
```
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "kubernetesConfig": {
        "clusterCidr": "10.10.0.0/16",
        "serviceCidr": "10.11.0.0/16",
        "dnsServiceIP": "10.11.255.1"
      }
    },
```
and `"firstConsecutiveStaticIP": "10.10.255.239"`. Update servicePrincipal info as necessary.

3. Generate the deployment templates and deploy them to the same RG as the custom vnet.

[...]

PR #507, opened 2017-04-19
Fix k8s custom VNet example so that the pods work

ashb:
Without this fix the agent nodes will be assigned IPs by DHCP at the start of this range (10.239.0.4, .5, .6 etc is common for three agents) and when the RouteTable is manually associated the kube cloud provider will add routes - including one of 10.239.0.0/24 to one the nodes which will immediately make the other two nodes unreachable. The fix is to make sure the nodes are assigned IP address from the end of the network space (or alternatively from outside the clusterCidr all together but still in the VNet space)

Addresses one of the problems mentioned in #493 (but perhaps not the original reported problem.)

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/507)

### DIFF
--- a/examples/vnet/kubernetesvnet.json
+++ b/examples/vnet/kubernetesvnet.json
@@ -2,7 +2,8 @@
   "apiVersion": "vlabs",
   "properties": {
     "orchestratorProfile": {
-      "orchestratorType": "Kubernetes"
+      "orchestratorType": "Kubernetes",
+      "clusterCidr": "10.239.0.0/16"
     },
     "masterProfile": {
       "count": 1,
--- a/examples/vnet/vnetarmtemplate/azuredeploy.kubernetes.json
+++ b/examples/vnet/vnetarmtemplate/azuredeploy.kubernetes.json
@@ -18,7 +18,7 @@
           {
             "name": "KubernetesSubnet",
             "properties": {
-              "addressPrefix": "10.239.0.0/16"
+              "addressPrefix": "10.239.255.0/24"
             }
           }
         ]


colemickens:
@ashb I think I understand how this could contribute to the problem, but I have a number of questions still:

1. Why is this a problem, but only when the AzureVNET CNI plugin is activated?
2. Why is this a problem with the default values? I thought the defaults you used were the same defaults used previously (meaning your change should have been a no-op for cases where users didn't customize the CIDRs)?
3. Is this the best way to solve it? In the interest of driving people to the "pit of success", we should see if we can fix the defaults to avoid this overlap (after I understand question 1 better...)

Please feel free to ping me on Slack if you're available. We basically have to either fix this today or back out the change. The instability and worry that there is something wrong with ACS-Engine at the moment is blocking some other (non-GitHub) work.

ashb:
1. It was also a problem in my testing with "None". It might be related or un-related to the Azure cni plugin - I didn't test that.
2. The defaults (as in when acs-engine can create it's own network.) are 10.0.0.0/8 for the Vnet and 10.240.0.0/16 for the cluster CIDR. I think the defaults are fine. This only comes up when you deploy k8s into a custom VNet.
3. It would be nice to do some validation, but acs-engine doesn't speak to the Azure API so can't find out information about the previously created VNet+subnet. Nothing immediately springs to mind.

I'll grab you on slack now too.

ashb:
Right okay, I've confused matters a bit with this PR.

To be clear. This is _not needed_ to get working pods.

With the example as it was you will have master and agent VM IPs in the 10.239.0.0/16 IP space, and Pod IPs in the default 10.244.0.0/16 IP space.

A change like this is only needed if you want the Pod IPs to lie within the 10.239.0.0/16 IP space _as well_. I'm going to close this PR and a take a look at just updating the docs somewhere (maybe in this example folder, maybe in the docs/clusterdefinition.md file) tomorrow/Friday.

PR #529, opened 2017-04-25
Make OrchestratorType case insensitive

JiangtianLi:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/529)

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -283,20 +283,20 @@ func GenerateKubeConfig(properties *api.Properties, location string) (string, er
 func prepareTemplateFiles(properties *api.Properties) ([]string, string, error) {
 	var files []string
 	var baseFile string
-	if properties.OrchestratorProfile.OrchestratorType == api.DCOS188 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS190 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS187 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS184 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS173 {
+	if properties.OrchestratorProfile.OrchestratorType.Equal(api.DCOS188) ||
+		properties.OrchestratorProfile.OrchestratorType.Equal(api.DCOS190) ||
+		properties.OrchestratorProfile.OrchestratorType.Equal(api.DCOS187) ||
+		properties.OrchestratorProfile.OrchestratorType.Equal(api.DCOS184) ||
+		properties.OrchestratorProfile.OrchestratorType.Equal(api.DCOS173) {
 		files = append(commonTemplateFiles, dcosTemplateFiles...)
 		baseFile = dcosBaseFile
-	} else if properties.OrchestratorProfile.OrchestratorType == api.Swarm {
+	} else if properties.OrchestratorProfile.OrchestratorType.Equal(api.Swarm) {
 		files = append(commonTemplateFiles, swarmTemplateFiles...)
 		baseFile = swarmBaseFile
-	} else if properties.OrchestratorProfile.OrchestratorType == api.Kubernetes {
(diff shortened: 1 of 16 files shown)


weinong:
@acs-bot please test this

weinong:
@amanohar

amanohar:
Reviewing...

pkg/api/types.go
```diff
@@ -186,6 +187,17 @@ type VMDiagnostics struct {
 // OrchestratorType defines orchestrators supported by ACS
 type OrchestratorType string
 
+// OrchestratorTypeEqualer compares OrchestratorType
+type OrchestratorTypeEqualer interface {
+	Equal(OrchestratorTypeEqualer) bool
```
amanohar: 'Equals' is usually the function name in all languages.

dmitsh:
Wouldn't it be easier to lowercase the string at the time of assignment, and then use regular comparison?

pkg/api/types.go
```diff
@@ -186,6 +187,17 @@ type VMDiagnostics struct {
 // OrchestratorType defines orchestrators supported by ACS
 type OrchestratorType string
 
+// OrchestratorTypeEqualer compares OrchestratorType
+type OrchestratorTypeEqualer interface {
```
amanohar: Any reason to simply not use: https://golang.org/pkg/strings/#EqualFold ?

pkg/api/vlabs/types.go
```diff
@@ -197,6 +199,17 @@ type KeyVaultCertificate struct {
 // OrchestratorType defines orchestrators supported by ACS
 type OrchestratorType string
 
+// OrchestratorTypeEqualer compares OrchestratorType
+type OrchestratorTypeEqualer interface {
```
amanohar: StringComparer so it can reused.

pkg/api/types.go
amanohar: Discussed offline. Adding a helper function makes the code concise. My suggestion is to convert the name of a generic StringComparer so it can be reused.

pkg/api/types.go
JiangtianLi: Changed.

pkg/api/vlabs/types.go
JiangtianLi: Changed.

pkg/api/v20170131/types.go
```diff
@@ -159,9 +160,20 @@ type VMDiagnostics struct {
...
+type StringComparer interface {
+	Equals(StringComparer) bool
+}
+
 // OrchestratorType defines orchestrators supported by ACS
 type OrchestratorType string
 
+// Equals returns true if two OrchestratorType are compared to be the same.
+// Here it is string case insensitive comparison
+func (o OrchestratorType) Equals(u StringComparer) bool {
```
amanohar: Talked offline. My recommendation would be to look into the option of generalizing this function so any code in ACS Engine can use it instead of each type having to duplicate this function.

pkg/acsengine/engine.go
```diff
@@ -283,20 +283,20 @@ func GenerateKubeConfig(properties *api.Properties, location string) (string, er
 func prepareTemplateFiles(properties *api.Properties) ([]string, string, error) {
 	var files []string
 	var baseFile string
-	if properties.OrchestratorProfile.OrchestratorType == api.DCOS188 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS190 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS187 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS184 ||
-		properties.OrchestratorProfile.OrchestratorType == api.DCOS173 {
+	if properties.OrchestratorProfile.OrchestratorType.Equals(api.DCOS188) ||
+		properties.OrchestratorProfile.OrchestratorType.Equals(api.DCOS190) ||
```
rjtsdl: I think, this will be a perfect place to use switch.

JiangtianLi:
Closed per a new #535

PR #534, opened 2017-04-26
add validation for k8s agentpool dnsprefix

rjtsdl:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/534)

### DIFF
--- a/pkg/api/v20160930/validate.go
+++ b/pkg/api/v20160930/validate.go
@@ -35,7 +35,7 @@ func (m *MasterProfile) Validate() error {
 }
 
 // Validate implements APIObject
-func (a *AgentPoolProfile) Validate() error {
+func (a *AgentPoolProfile) Validate(orchestratorType OrchestratorType) error {
 	if e := validateName(a.Name, "AgentPoolProfile.Name"); e != nil {
 		return e
 	}
@@ -47,6 +47,17 @@ func (a *AgentPoolProfile) Validate() error {
 	}
 	if e := validateName(a.VMSize, "AgentPoolProfile.VMSize"); e != nil {
 		return e
+	}
+	// Kubernetes don't allow agent DNSPrefix
+	if orchestratorType == Kubernetes {
+		if e := validateNameEmpty(a.DNSPrefix, "AgentPoolProfile.DNSPrefix"); e != nil {
+			return e
+		}
+	}
+	if a.DNSPrefix != "" {
+		if e := validateDNSName(a.DNSPrefix); e != nil {
+			return e
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


rjtsdl:
Here, i only do validation fix. The compatible change need to be made somewhere else.

pkg/api/v20160930/validate.go
```diff
@@ -48,6 +48,11 @@ func (a *AgentPoolProfile) Validate() error {
 	if e := validateName(a.VMSize, "AgentPoolProfile.VMSize"); e != nil {
 		return e
 	}
+	if orchestratorType == Kubernetes {
```
JiangtianLi: We need to do case insensitive comparison. See #529. You can check in first and I'll change.

weinong:
why is it different than vlab validation?

rjtsdl:
vlab use len(ports) to determine whether the DNSPrefix could be empty or not.
In version 2017-01-31 and 2016-09-30, ports is not in the api model. We have to do things in another way. We use DNSPrefix to determine ports should be set or not. 

However, I don't see why we cannot keep the principle consistent in vlab. I can also do validate change in vlab, let it also use DNSPrefix to determine ports.

@amanohar , how do you think?

pkg/api/v20160930/validate.go
rjtsdl: I approved your pr, if you get checked in earlier. I will modify upon it.

JiangtianLi: I am working on a solution that doesn't need to change string comparison. So feel free to check in.

amanohar:
@rjtsdl, yes I think validation should be consistent through all the API versions. Let's discuss tomorrow (and I agree with what you mentioned about validation being the same throughout).

rjtsdl:
@acs-bot test this please

rjtsdl:
@amanohar @weinong , I intentionally make all the version's DNSPrefix validation follow the same philosophy. They have subtle difference, mainly because of the version's difference.

rjtsdl:
@acs-bot test this please

pkg/api/vlabs/validate.go
```diff
@@ -290,6 +303,13 @@ func (a *Properties) validateNetworkPolicy() error {
 	return nil
 }
 
+func validateNameEmpty(name string, label string) error {
```
amanohar: validateFieldEmpty or validateStringEmpty is better IMO.

rjtsdl: Saw this, after i merged it. We can do another one to correct the name. There is validateName as well. It needs a broader change.

PR #535, opened 2017-04-26
Make OrchestratorType case insensitive. After json is unmarshal, do a…

JiangtianLi:
… case insensitive comparison of OrchestratorType in Validate and save OrchestratorType as the defined constant string for future comparison.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/535)

### DIFF
--- a/pkg/api/v20160330/validate.go
+++ b/pkg/api/v20160330/validate.go
@@ -4,18 +4,14 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-	"strings"
 )
 
 // Validate implements APIObject
 func (o *OrchestratorProfile) Validate() error {
-	switch {
-	case strings.EqualFold(string(o.OrchestratorType), string(DCOS)):
-		o.OrchestratorType = DCOS
-	case strings.EqualFold(string(o.OrchestratorType), string(Mesos)):
-		o.OrchestratorType = Mesos
-	case strings.EqualFold(string(o.OrchestratorType), string(Swarm)):
-		o.OrchestratorType = Swarm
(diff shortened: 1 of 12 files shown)


pkg/api/v20160330/validate.go
```diff
@@ -4,14 +4,18 @@ import (
...
+	"strings"
 )
 
 // Validate implements APIObject
 func (o *OrchestratorProfile) Validate() error {
-	switch o.OrchestratorType {
-	case DCOS:
-	case Mesos:
-	case Swarm:
+	switch {
```
rjtsdl: So, validation actually change the orchestratorProfile?

JackQuincy:
Don't block on this right now, but I don't like how this makes validate change the api model contents. I'd like this to be part of the unmarshalling, but I understand that is much more work. Not sure if there is a different better spot. possibly the call to defaults as that updates the model currently.

JiangtianLi:
@JackQuincy Correct. It is better to do in unmarshalling. Right now we rely on validate right after unmarshal.

pkg/api/v20160330/validate.go
JiangtianLi: Yes, it changes the case of OrchestratorType

JackQuincy:
If you haven't found it yet, look at this https://golang.org/pkg/encoding/#TextUnmarshaler . If our types implement it json.Unmarshall should be able to corrcet casing for us

JiangtianLi:
@JackQuincy Thanks for the pointer. That's very helpful! I've made the change.

PR #552, opened 2017-04-28
use pointer for []*AgentPoolProfile

rjtsdl:
change AgentPoolProfiles's type to []*AgentPoolProfile. It conforms with other properties fields. And also by making this change, it avoids complicated reference syntax, which is not necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/552)


rjtsdl:
Btw, I noticed, we didn't do nil pointer check before we call validate. Such as 
```
if e := containerService.Properties.Validate();
```
Properties's type is *Properites. It could be nil. So i guess, there should be more work to make this more robust. It should be done in a seperate PR though.

dmitsh:
@acs-bot test this please

rjtsdl:
#553 , just filed an issue. We will need to address it ASAP. maybe right after April

PR #571, opened 2017-05-04
Update Kubernetes cloudinit syntax for creating file system to get around cmd cloudinit bug 

amanohar:
- Update Kubernetes cloudinit syntax for creating file system to get around cmd cloudinit bug 
- Add etcd validation in CSE script

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/571)


parts/kubernetesmastercustomscript.sh
```diff
@@ -249,6 +249,23 @@ function ensureEtcd() {
     done
 }
 
+function ensureEtcdDataDir() {
+    etcdStarted=1
+    mount | grep /dev/sdc1 | grep /var/lib/etcddisk
+    if [ "$?" = "0" ]
+    then
```
colemickens: nit: I assume you copy/pasted one that was looping? Since there's no loop here, there's no need for the second if. You can just early return if the exit code was zero and "exit 1" if you make it out of the `if`. (Or inversed)

parts/kubernetesmastercustomscript.sh
```diff
@@ -249,6 +249,23 @@ function ensureEtcd() {
...
+        echo "Etcd running with data dir at: /var/lib/etcddisk"
+        etcdStarted=0
+        break
+    fi
+
+    if [ $etcdStarted -ne 0 ]
+    then
+        echo "Etcd data dir was not found at: /var/lib/etcddisk"
```
anhowe: are there any mitigations we can do before failing like re-mounting?

amanohar: I consider mounting a basic operation and should not have to write code to retry it IMO. DC/OS also creates fs and mounts it in cloud init (ephemeral).

I don't think there are any retries built into cloud init's syntactical way of specifying mount and fs setup. If you want a retry then etcd disk setup needs to be move to CSE which is not bad by itself but I am not convinced that retries need to build for something as fundamental as mounting.

Running tests in a loop for a couple of days will give me better idea about mount reliability.

amanohar:
@acs-bot test this please

amanohar:
@acs-bot test this please

parts/kubernetesmastercustomscript.sh
anhowe: The retry I was imagining was something similar to what we have in ensureApiServer and ensureDocker with a sleep and call to "mount -a", that will just kick anything unmounted in fstab.  We can wait until you can determine reliability and decide then.  I'll approve now so you can get this in.

amanohar: I have add the check and retry.

amanohar:
@acs-bot test this please

amanohar:
@acs-bot test this please

amanohar:
@acs-bot test this please

amanohar:
@acs-bot test this please

PR #576, opened 2017-05-04
Fix POD deletion on Windows node in 1.6.0 due to cgroups-per-qos intr…

JiangtianLi:
…oduced

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/576)


parts/kuberneteswindowssetup.ps1
```diff
@@ -178,7 +178,7 @@ c:\k\kubelet.exe --hostname-override=`$global:AzureHostname --pod-infra-containe
     if ($global:KubeBinariesVersion -ne "1.5.3")
     {
         $KubeletArgList += "--enable-cri=false"
-        $KubeletCommandLine += " --enable-cri=false --image-pull-progress-deadline=20m"
+        $KubeletCommandLine += " --enable-cri=false --image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`""
```
rjtsdl: Not sure what the additional parameter do?

JiangtianLi: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node-allocatable.md

JiangtianLi: https://github.com/kubernetes/kubernetes/blob/release-1.6/cmd/kubelet/app/options/options.go also has comment for the option. Basically --cgroups-per-qos is not supported on Windows and needs to be set to false. With that, --enforce-node-allocatable has to be set to empty string instead of default "pods".

rjtsdl: thx

JiangtianLi:
@rjtsdl Thanks.

Issue #575, opened 2017-05-04
Protect docker daemon using TLS for Swarm and Swarmmode.


iphilpot:
This change would allow customers to connect from client to ACS instance of swarm/swarmmode using TLS, which is a published best practice. Currently customers have to create an SSH tunnel to execute Docker commands on the remote host. The following are proposed changes to ACS-Engine to enable this functionality.

Current Customer/Partner example: [CodeShip CI/CD](https://github.com/codeship-library/azure-deployment/blob/acs-swarm/sshtunnel/ssh-tunnel.sh)

I may have missed a couple of things during my investigation, but this should be the highlights. 

### Proposed Changes to engine:
| Pkg	| File	| Method/Struct	| Change |
|-------|-------|---------------|--------|
|Acsengine	|Defaults.go	|certGenerationRequired	|Swarm/SwarmMode return true|
|Api/vlabs	|Types.go	|CertificateProfile	|Change cert struct to allow for both K8 and swarm certs|
|Acsengine	|Pki.go	|CreatePki	|extraFQDNs if k8|
|Acsengine	|Pki.go	|CreatePki	|If K8 gen certs per struct/if swarm gen certs per struct|

### Proposed Changes to provision scripts (swarm/swarmmode):
Script locations: /opt/azure/containers/

| Template name | Script name | Changes |
|------------------|---------------|----------|
| agentCustomScript | configure-swarm-cluster.sh & configure-swarmmode-cluster.sh | Save cert/keys & configure TLS startup |

### Proposed Changes to example ClusterDefinitionFiles for Swarm/Swarmmode:
https://github.com/Azure/acs-engine/tree/master/examples

  - Add TLS port: 2376

### Proposed changes to documentation: 
These docs will need to be updated:
  1. https://github.com/Azure/acs-engine/blob/master/docs/swarm.md
  2. https://github.com/Azure/acs-engine/blob/master/docs/swarmmode.md

  - Update Architecture image to reflect TLS port
  - Step 2 remove ssh tunnel 
  - Add client TLS setup:
		$ mkdir -pv ~/.docker
        $ cp -v {ca,cert,key}.pem ~/.docker
        $ export DOCKER_HOST=tcp://$HOST:2376 DOCKER_TLS_VERIFY=1

amanohar:
Ack @tripdubroot. I will review this soon.

mboersma:
[...]

PR #580, opened 2017-05-08
Updated cert type for swarm cert/key

tripdubroot:
PR for Issue #575 

Signed-off-by: Ian Philpot <ianphil@microsoft.com>

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/580)

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -76,6 +76,14 @@ type CertificateProfile struct {
 	KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
 	// caPrivateKey is an internal field only set if generation required
 	caPrivateKey string
+	// SwarmTlsServerCertificate is the Docker server daemon certificate and signed by the CA
+	SwarmTlsServerCertificate string `json:"swarmTlsSeverCertificate,omitempty"`
+	// SwarmTlsClientCertificate is the Docker client cli certificate and signed by the CA
+	SwarmTlsClientCertificate string `json:"swarmTlsClientCertificate,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -76,6 +76,14 @@ type CertificateProfile struct {
 	KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
 	// caPrivateKey is an internal field only set if generation required
 	caPrivateKey string
+	// SwarmTlsServerCertificate is the Docker server daemon certificate and signed by the CA
```
amanohar: CertificateProfile also contains the following which is K8s specific:

```
// KubeConfigCertificate is the client certificate used for kubectl cli and signed by the CA
KubeConfigCertificate string `json:"kubeConfigCertificate,omitempty"`
// KubeConfigPrivateKey is the client private key used for kubectl cli and signed by the CA
KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
```
SwarmMode cert and Kube config certs live side by side in one struct? Its confusing about what is the right combination. Should the CertificateProfile be renamed to KubernetesCertificateProfile and another struct created called "SwarmModeCertificateProfile". That way there can be validation check that one only type is set in properties based on orchestrator profile.

tripdubroot: @amanohar I'll make this change.

shrutir25:
@acs-bot test this please

shrutir25:
@amanohar - how do we want to proceed with this PR ?

tripdubroot:
closing PR... will open new

PR #581, opened 2017-05-08
mitigate ARM throttling issue

dmitsh:
- removing cloud provider service
 - retrieving FD from the wireserver
 - retrieving node's VM size from the template

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/581)


parts/kubernetesagentcustomdata.yml
```diff
@@ -113,7 +113,7 @@ write_files:
     DOCKER_OPTS=
     CUSTOM_CMD=/bin/true
     KUBELET_REGISTER_SCHEDULABLE=true
-    KUBELET_NODE_LABELS=role=agent
+    KUBELET_NODE_LABELS=role=agent,failure-domain.beta.kubernetes.io/region={{WrapAsVariable "location"}}
```
JiangtianLi: What is this label used for? Why some labels are appended in kuberneteskubelet.service? I am asking because I am looking into matching node labels for Windows node with Linux node.

parts/kubernetesagentcustomdata.yml
```diff
@@ -128,7 +128,8 @@ write_files:
   owner: "root"
   content: |
     #!/bin/bash
-    exit 0
+    echo $(curl -q http://169.254.169.254/metadata/v1/InstanceInfo 2>/dev/null | sed "s/.*\"FD\":\"\([0-9]*\).*/\1/") > /etc/kubernetes/fd
```
rjtsdl: 1. What the url (http://169.254.169.254/metadata/v1/InstanceInfo) for?
2. This url contain explicit IP address, and the link itself is not very meaningful. So is it possible to find an alternative link?
2. For curl, i think we should do retry.

parts/kubernetesagentcustomdata.yml
```diff
@@ -128,7 +128,8 @@ write_files:
...
+    echo "',variables('{{.Name}}VMSize'),'" > /etc/kubernetes/vmsize
```
rjtsdl: vmsize is a Azure concept. Not sure if this should be located at /etc/kubernetes/ hierarch.

parts/kubernetesagentcustomdata.yml
anhowe: do we need retry logic, or is imd always going to answer?

parts/kubernetesmastervars.t
```diff
@@ -35,7 +35,6 @@
 {{if .HasManagedDisks}}
     "apiVersionStorageManagedDisks": "2016-04-30-preview",
 {{end}}
-    "location": "[resourceGroup().location]",
```
anhowe: why was this deleted?

dmitsh: it is overridden by another "location" statement

parts/kubernetesagentcustomdata.yml
dmitsh: There are two labels associated with Fault Domain (FD): failure-domain.beta.kubernetes.io/region and failure-domain.beta.kubernetes.io/zone

parts/kubernetesagentcustomdata.yml
dmitsh: 1. This is the metadata server IP
2. This is a well-known non routable IP address that can be accessed only from within the Azure VM. 
3. Thanks, will add "--retry" flag to curl

parts/kubernetesagentcustomdata.yml
dmitsh: vmsize is consumed by kubelet

parts/kubernetesagentcustomdata.yml
dmitsh: adding retry flag

dmitsh:
@acs-bot test this please

khenidak:
Guys why did you remove cloud provider and cloud provider config?

JackQuincy:
We were getting throttling from ARM on subscriptions that had large clusters or lots of clusters. Cloud shell had to do this manually and the call was made to just rip it out. I know there was talks about making it an option in the api model to do one or the other.

dmitsh:
@khenidak We had throttling issue with ARM. Does this change introduce a problem for you?

dmitsh:
@dsemarketer Changing VM size wasn't supported before this change.

dmitsh:
reverting...

PR #592, opened 2017-05-09
changing default values in template to be valid. These aren't ever us…

JackQuincy:
…ed as they are set in the parameters file but it is a cleanliness thing/I had a parameters file that didn't have that parameter and it caused a failure

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/592)


JackQuincy:
correct I will regen them

JackQuincy:
@acs-bot test this please

PR #593, opened 2017-05-09
Restructure commands. Introduce cobra/logrus.

colemickens:
Starting to retrofit some stuff from `azkube` in here. This is a prerequisite for adding the `upgrade` command.

1. Introduce `logrus`. We're already using this internally.

2. Introduce `cobra`. Cobra is a library for creating structured CLIs. It has a huge, huge feature list. Notably I am using: support for subcommands, autogenerated usage, command/flag validation.

3. Switch to using master FQDN prefix instead of the cluster hash to name the output directory.

4. Change `./test/common.sh` accordingly for this change.

5. Change `Makefile` to always rather than sometimes nooping, and change the version stamping since I moved where those variables live.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/593)


cmd/generate.go
```diff
@@ -0,0 +1,128 @@
...
+		Use:   generateName,
+		Short: generateShortDescription,
+		Long:  generateLongDescription,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return gc.run(cmd, args)
+		},
+	}
+
+	f := generateCmd.Flags()
+	f.StringVar(&gc.clusterDefinitionPath, "cluster-definition", "", "")
```
amanohar: Update to APIModel

colemickens: Done.

PR #594, opened 2017-05-10
Add initial skeleton of an `upgrade` command

colemickens:
This gets us a valid `ServicePrincipalToken` (aka accessToken) so we can hand it, and other relevant information, to the upgrade engine.

This stacks on top of #593, so only look at the latest commit for only the upgrade command parts.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/594)

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -1,7 +1,11 @@
 package cmd
 
 import (
+	"os"
+	"path"
+
 	"github.com/Azure/acs-engine/pkg/acsengine"
+	"github.com/Azure/acs-engine/pkg/api"
 
 	"github.com/Azure/go-autorest/autorest/adal"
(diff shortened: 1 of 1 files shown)


cmd/upgrade.go
```diff
@@ -0,0 +1,135 @@
...
+	upgradeCmd := &cobra.Command{
+		Use:   upgradeName,
+		Short: upgradeShortDescription,
+		Long:  upgradeLongDescription,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return uc.run(cmd, args)
+		},
+	}
+
+	f := upgradeCmd.Flags()
```
amanohar: Doesn't it need to take the APIModel or directory to the generated template folder that has the APIModel file?

PR #603, opened 2017-05-11
Upgrade container service API

amanohar:
- Add upgrade container service API validation
- Hookup with upgrade command
- Query all to get a list of matching VM resources that need to be upgraded

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/603)


pkg/api/vlabs/upgradetypesandvalidate.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "fmt"
+
+// UpgradeContainerService API model
+type UpgradeContainerService struct {
+	OrchestratorProfile *OrchestratorProfile `json:"orchestratorProfile,omitempty"`
+}
+
+// Validate implements APIObject
+func (ucs *UpgradeContainerService) Validate() error {
```
colemickens: Does this need to take the existing containerService model to evaluate against? I'm cool with punting, just wanted to mention.

pkg/operations/upgradecluster.go
```diff
@@ -1,52 +1,79 @@
...
+}
 
 // UpgradeCluster runs the workflow to upgrade a Kubernetes cluster.
 // UpgradeContainerService contains target state of the cluster that
 // the operation will drive towards.
-func (uc *UpgradeCluster) UpgradeCluster(subscription uuid.UUID, rg string,
-	cs *vlabs.ContainerService, ucs *vlabs.UpgradeContainerService) {
-	// 	ListClusterResources(subscription uuid.UUID, rg string) error
-}
+func (uc *UpgradeCluster) UpgradeCluster(subscriptionID uuid.UUID, resourceGroup string,
```
colemickens: per other comment, please remove Token from here.

pkg/operations/armhelpers/azureclients.go
```diff
@@ -0,0 +1,32 @@
...
+	SubscriptionID   string
+	AzureEnvironment azure.Environment
+	TenantID         string
+
+	GroupsClient *resources.GroupsClient
+	VMClient     *compute.VirtualMachinesClient
+}
+
+// Create method creates various Azure clients
+func (ac *AzureClients) Create(token *adal.ServicePrincipalToken) (*AzureClients, error) {
```
colemickens: Generally, the ctor style in Golang would be to make this:

`func NewAzureClient() (*AzureClients, error) {` or maybe, but less likely:

`func (ac *AzureClients) Initiailize(token) error {`

as is, this is fairly un-idiotmatic since it returns itself and is a pointer receiver function at the same time.

I was expecting to see the `Create` called "higher up" and then attached to the upgradeCluster operation before the upgradeCluster operation is kicked off.

colemickens: I think we should try to get to the point where we're just passing around a fully self-sufficient `AzureClients` struct and not need to keep plumbing subId and token everywhere.

cmd/upgrade.go
```diff
@@ -63,6 +69,7 @@ func NewUpgradeCmd() *cobra.Command {
 	f.StringVar(&uc.rawSubscriptionID, "subscription-id", "", "the subscription ID where the cluster is deployed")
 	f.StringVar(&uc.resourceGroupName, "resource-group", "", "the resource group where the cluster is deployed")
 	f.StringVar(&uc.deploymentDirectory, "deployment-dir", "", "the location of the output from `generate`")
+	f.StringVar(&uc.upgradeModelFile, "upgrademodel-file", "", "file path to upgrade API model")
```
amanohar: I was trying to be consistent. Its confusing to have deployment-dir and upgrade-model, no?

pkg/api/vlabs/upgradetypesandvalidate.go
amanohar: That's the validation I was referring to earlier that I still need to do. Though I don't think this is the right place for it.

pkg/operations/armhelpers/azureclients.go
amanohar: Yes, I agree and the RP already does this (https://msazure.visualstudio.com/DefaultCollection/One/_git/Compute-containerService?path=%2Fsrc%2Fmicrosoft.com%2Fcontainerservice%2Ftemplates%2Fhelper.go&version=GBmaster&_a=contents). 

I tried writing something similar but reverted back to this because I wanted to see it working :). Do you want me to do same thing in this PR?

pkg/operations/upgradecluster.go
amanohar: Updated

pkg/operations/armhelpers/azureclients.go
amanohar: Updated as a stop gap until ACS Engine and RP code is merged.

dmitsh:
@acs-bot test this please

cmd/upgrade.go
amanohar: I am going to leave this as is for consistency with deployment-dir flag.

amanohar:
@acs-bot test this please

pkg/api/types.go
```diff
@@ -266,6 +266,19 @@ type V20170131ARMContainerService struct {
...
+// needed because the json that is sent to ARM and acs-engine
+// is different from the json that the ACS RP Api gets from ARM
+type VlabsUpgradeContainerService struct {
+	TypeMeta
+	*vlabs.UpgradeContainerService
+}
+
+// UpgradeContainerService API model
+type UpgradeContainerService struct {
+	OrchestratorProfile *OrchestratorProfile `json:"orchestratorProfile,omitempty"`
```
colemickens: What about having Properties in there? That's how the regular apimodel works, it has "apiVerson" and "properties" as the two top-level things.

(context: I was trying to test some integration stuff and wanted to cheat and generate with one version and then pass the other versioned apimodel to the upgrade call, but it nil derefs in the switch statement because it doesn't actually deserialize properly since there's no Properties level here)

Not that we are trying to accommodate the apimodel as the upgrade model input, but for consistency sake maybe?

pkg/operations/upgradecluster.go
```diff
@@ -1,52 +1,73 @@
...
-	"github.com/Azure/acs-engine/pkg/api/vlabs"
-	"github.com/satori/uuid"
+	"fmt"
+
+	"strings"
+
+	"github.com/Azure/acs-engine/pkg/api"
+	"github.com/Azure/acs-engine/pkg/operations/armhelpers"
+	"github.com/Azure/azure-sdk-for-go/arm/compute"
+	"github.com/prometheus/common/log"
```
colemickens: !! oh! I couldn't figure out how logrus had started showing me line numbers automagically!

colemickens: cool!

colemickens: Hm, it doesn't seem to easily expose Entry, or implement the functions on the pointer receiver making it hard to use everywhere though.

colemickens: I guess the long term is to pull what's already in the RP anyway, and I guess it's doing the same thing the prometheus code is doing.

pkg/api/types.go
amanohar: I thought about this. Properties are needed for PUT model to be ARM compliant. I couldn't find anything in ARM documentation (and also poked around in the CRP & Disk for e.gs) for other operations that follow this model.

We don't need fields like name, tag, location during non PUT operations. These are needed by ARM to distinguish between its fields vs RP's fields (everything inside properties that are just sent to the RPs without ARM parsing them).

E.g.:
https://docs.microsoft.com/en-us/rest/api/virtualmachinescalesets/manage-all-vms-in-a-set
https://docs.microsoft.com/en-us/rest/api/manageddisks/disks/disks-grant-access

So I couldn't think of a good reason to follow this format for the sake of it in upgrade. But happy to discuss if this is indeed needed..

pkg/operations/upgradecluster.go
amanohar: Yes, long terms we need to pull stuff from the RP but this is a good start, makes it functional at least.

PR #607, opened 2017-05-12
avoid using ILB for a single kubernetes master

anhowe:
Due to kubernetes issue kubernetes/kubernetes#41185, this is required in case ILB restarts.  For 3 and 5 master cases, we are using #608 for tracking.

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/607)


parts/kubernetesagentcustomdata.yml
```diff
@@ -66,7 +66,7 @@ write_files:
     - name: localcluster
       cluster:
         certificate-authority: /etc/kubernetes/certs/ca.crt
-        server: https://{{WrapAsVariable "masterInternalLbIp"}}:443
+        server: https://{{WrapAsVariable "masterIP"}}:443
```
rjtsdl: The way this PR handle this, is to replace everywhere using masterIP. 
To me, it is a little bit confusing, when i saw "masterIP", which actually is the "masterInternalLbIp". 

Wondering, is there a way to make the name more explicit?

rjtsdl:
Can you also refer to the issue 608? 

You can point out, this PR only partially alleviate the issue.

parts/kubernetesagentcustomdata.yml
anhowe: renamed to kubernetesAPIServerIP

anhowe:
@acs-bot test this please

anhowe:
updated comment with #608.

PR #609, opened 2017-05-12
Gated, experimental `upgrade` command + uber AzureClient + generate can --deploy

colemickens:
1. Import `AzureClient` from [`azkube`](https://github.com/colemickens/azkube). Refactor and clean it up further to handle OAuthConfig, automatic tenant ID and to stop stashing things on the client that shouldn't really be necessary to persist. Remove need for create of AzureClient to know the `tenantID` ahead of time. The client ctors will auto-discover it.

2. Add UberClient interface, TemplateDeployer interface, along with a VMClient interface. Now the `UberClient` can have a mock implementation, and the real code can use the `AzureClient` implementation that is backed by real Azure SDK clients.

3. Add `./test/upgrade.sh` script that wraps the CLI... this is really not necessary at all with the device auth support though, especially with the token caching. I might remove this anyway.

4. Add device auth support. It means you can `upgrade` or `generate --deploy` with only supplying the `--subscription-id`. TenantID is auto-discovered and the user can auth the same way they do with `az login`. The token is also cached, so they only need to interactive login once!

5. Along with that, move all of the auth flags / client creation to `root.go` so that `generate` and `upgrade` can share that code (and eventually `scale`, etc).

6. Improve UX for the `--upgrademodel-flag` file.

7. Change all instances of `logrus.Logger` to `logrus.Entry` as that's the more appropriate thing to pass around.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/609)


colemickens:
Closing to send smaller PRs.

PR #614, opened 2017-05-12
Add initial upgrade command behind env var flag + container service API validation, wiring up to the CLI

colemickens:
First set of changes. This is merging most of what's in `k8supgrade` branch now into `master`.

I've amended your "Add upgrade container..." commit, @amanohar, with the check for the `--upgrademode-file` flag check so that the commit history is still, in theory, buildable (it had been out of order based on my rebasing previously).

This, generally, should all be code that has been CR'd already.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/614)


cmd/generate.go
```diff
@@ -26,7 +26,7 @@ type generateCmd struct {
 	noPrettyPrint     bool
 	parametersOnly    bool
 
-	// Parsed from inputs
+	// derived
```
amanohar: Derived from?

PR #615, opened 2017-05-13
AzureClient improvements + ACSEngineClient interface

colemickens:
This stacks on top of #614.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/615)


pkg/acsengine/munge.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	templateMap[resourcesFieldName] = append(resources[:nsgIndex], resources[nsgIndex+1:]...)
+
+	return nil
+}
+
+// NormalizeMasterResourcesForScaling takes a template and removes elements that are unwanted in any scale up/down case
+func NormalizeMasterResourcesForScaling(logger *logrus.Entry, templateMap map[string]interface{}) error {
```
amanohar: This name and NormalizeForK8sVMASScalingUp (or NormalizeForVMSSScaling) are inconsistent in naming.
NormalizeForK8sVMASScalingUp : could be NormalizeK8sVirtualMachineAgentsForScaling
NormalizeForVMSSScaling : could be NormalizeForVMSSAgentsScaling

pkg/armhelpers/azureclient.go
```diff
@@ -0,0 +1,313 @@
...
+
+var (
+	// RequiredResourceProviders is the list of Azure Resource Providers needed for ACS-Engine to function
+	RequiredResourceProviders = []string{"Microsoft.Compute", "Microsoft.Storage", "Microsoft.Network"}
+)
+
+// AzureClient is the uber client
+// If done right, we really shouldn't need SubscriptionID or TenantID in the client anywhere else
+// they're only used for setting up the clients, we can add them back later if needed
+type AzureClient struct {
```
amanohar: Can you rename of "AzureClients"?

pkg/armhelpers/azureclient.go
```diff
@@ -0,0 +1,313 @@
...
+	DeploymentsClient     resources.DeploymentsClient
+	GroupsClient          resources.GroupsClient
+	RoleAssignmentsClient authorization.RoleAssignmentsClient
```
amanohar: This can be removed for now since its not being used.

pkg/armhelpers/interfaces.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// the master client interface
+// if this is mocked out, then the entire context of the target environment is mocked:
+// once one of the NewAzureClient... returns, it has returned a fully initialized UberClient
+// all other code that needs to do work in subscription/AAD tenant will use this
+// currently only one impl: AzureClient
+type UberClient interface {
```
amanohar: Rename to something more descriptive. ACSEngineClients?

pkg/armhelpers/deployer.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/prometheus/common/log"
+)
+
+type AzureTemplateDeployer struct{}
+
+func (az *AzureClient) Validate(resourceGroup, name string, d resources.Deployment) (resources.DeploymentValidateResult, error) {
```
amanohar: Why AzureClient and not AzureTemplateDeployer?

pkg/armhelpers/deployer.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (az *AzureClient) Delete(resourceGroup, name string, cancel chan struct{}) (<-chan autorest.Response, <-chan error) {
+	return az.DeploymentsClient.Delete(resourceGroup, name, cancel)
+}
+
+func (az *AzureClient) Get(resourceGroup, name string) (resources.DeploymentExtended, error) {
+	return az.DeploymentsClient.Get(resourceGroup, name)
+}
+
+func DeployTemplate(deployer TemplateDeployer, resourceGroupName, deploymentName string, template map[string]interface{}, parameters map[string]interface{}) (response *resources.DeploymentExtended, err error) {
```
amanohar: I might be missing the purpose of AzureTemplateDeployer. Shouldn't DeployTemplate be associated with AzureTemplateDeployer? Or it just supposed to take TemplateDeployer (which is fine as well) then shouldn't CreateOrUpdate be associated with AzureTemplateDeployer?

pkg/acsengine/munge.go
```diff
@@ -0,0 +1,194 @@
...
+
+		if !removeCustomData(logger, virtualMachineProfile) || !removeImageReference(logger, virtualMachineProfile) {
+			continue
+		}
+	}
+	return nil
+}
+
+// NormalizeForK8sVMASScalingUp takes a template and removes elements that are unwanted in a K8s VMAS scale up/down case
+func NormalizeForK8sVMASScalingUp(logger *logrus.Entry, templateMap map[string]interface{}) error {
```
colemickens: This was straight copied from RP. I think the 'S' is VMAS + Scaling up. I don't see an "extra" S.

pkg/armhelpers/azureclient.go
colemickens: I assume you mean undo the rename of AzureClients -> AzureClient? Sure.

pkg/armhelpers/deployer.go
colemickens: There's no need for the struct here. The struct was when I thought I'd have to wrap more, I'll remove the struct line

pkg/armhelpers/deployer.go
colemickens: The point is that it's a utility function that works with anything that implements TemplateDeployer. The struct is simply unnecessary. But I don't understand this question: `shouldn't CreateOrUpdate be associated with AzureTemplateDeployer?` I think the question just goes away when I remove the AzureTemplateDeployer struct that should've never been in the PR.

pkg/acsengine/munge.go
colemickens: Makes sense, I like your suggested names better as well.

pkg/acsengine/munge.go
```diff
@@ -0,0 +1,194 @@
+package acsengine
```
colemickens: To me, those sound like generic C#-y names. I've not really seen Go files names like that, especially given that they're more groupings of functionality than C# where it's usually 1:1 class<->file.

I'm open to something beyond "munge" but it seems more descriptive than "TemplateProcessor" to me. "TemplateProcessor" seems more like the stuff we have in `engine.go`.

pkg/armhelpers/azureclient.go
colemickens: The more I think about this, the more I don't want to do it. It's not common for struct types to be pluralized and seems a bit misleading. What would it mean to have an array of `AzureClients`, etc?

pkg/acsengine/munge.go
colemickens: VMAS = VM Availability Set, not VM Agent S. I think that's the confusion?

pkg/acsengine/munge.go
colemickens: Hm, given my comment here, do you still think they need to be renamed? https://github.com/Azure/acs-engine/pull/615#discussion_r116417929

I could see this maybe:

`NormalizeForK8sVMASScalingUp ->`
`NormalizeForK8sVMASScaling` to be more consistent with:
`NormalizeForVMSSScaling` and changing this to be ->
`NormalizeForK8sVMSSScaling`

What do you think? Otherwise I feel yours are possibly inconsistent given spelling out "VirualMachine" but leaving VMSS.

colemickens:
Pushed a new commit:

Implement some of Anubhuti's feedback.
  - remove unused clients
  - renamed UberClient interface -> ACSEngineClient
  - removed the unused AzureTemplateDeployer as part of the interfaces
  cleanup

Cleanup the usage of interfaces. More closely follows the pattern I'd
been trying to follow. The AzureClient winds up implementing the
ACSEngine interface, as well as the interfaces inside ACSEngine.

Also fixed some warnings here and there in the new code I've been adding

---

I'm not sure if the two layers of interfaces is needed. It might be more prudent to just collapse the "DeploymentsClient" and "VirtualMachinesClient" interface into the AzureClient, instead of having two layers.

JackQuincy:
*[pkg/acsengine/munge.go, line 120 at r4](https://reviewable.io:443/reviews/azure/acs-engine/615#-Kjz9dgQwjopb6gCqwxm-r4-120:-Kk9tQSaDY5NEZoynlG5:bto2ymm) ([raw file](https://github.com/azure/acs-engine/blob/84eeaccb0f6c1c4a4717d822ed237ae96a58fb17/pkg/acsengine/munge.go#L120)):*
<details><summary><i>Previously, colemickens (Cole Mickens) wrote…</i></summary><blockquote>

Hm, given my comment here (https://github.com/Azure/acs-engine/pull/615#discussion_r116417929), do you still think they need to be renamed? 

I could see this maybe:

`NormalizeForK8sVMASScalingUp ->`
`NormalizeForK8sVMASScaling` to be more consistent with:
`NormalizeForVMSSScaling` and changing this to be ->
`NormalizeForK8sVMSSScaling`

What do you think? Otherwise I feel yours are possibly inconsistent given spelling out "VirualMachine" but leaving VMSS.
</blockquote></details>

NomalizeK8sVMASScalingUp has the Up because it Cannot be used to scale down. While the VMSS one can be. If we want to add agents in the end to specify we are scaling agents that seems fine. The VMSS is how we refer to those agent pools else where in the code so I was just starting consistent 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/615)*

pkg/acsengine/munge.go
amanohar: I disagree with names being "C#-y". The name should represent what the file is doing. It can be something else but I don't think munge is more Go-like either.

pkg/acsengine/munge.go
amanohar: Yeah, I misread it.

pkg/acsengine/munge.go
amanohar: Sure sounds fine. I think we should keep the words 'Agents'  and 'Up' (in case of VMAS) in the function name though.

pkg/acsengine/transform.go
```diff
@@ -0,0 +1,194 @@
...
+
+		if !removeCustomData(logger, virtualMachineProfile) || !removeImageReference(logger, virtualMachineProfile) {
+			continue
+		}
+	}
+	return nil
+}
+
+// NormalizeForK8sVMASScalingUp takes a template and removes elements that are unwanted in a K8s VMAS scale up/down case
+func NormalizeForK8sVMASScalingUp(logger *logrus.Entry, templateMap map[string]interface{}) error {
```
rjtsdl: NormalizeForK8sVMASScalingUp -> NormalizeForK8sVMASScaling, to be consistent with other Normalize functions.

pkg/armhelpers/azureclient.go
```diff
@@ -0,0 +1,295 @@
...
+
+func getClient(env azure.Environment, subscriptionID string, armSpt *adal.ServicePrincipalToken, adSpt *adal.ServicePrincipalToken) (*AzureClient, error) {
+	c := &AzureClient{
+		deploymentsClient:     resources.NewDeploymentsClientWithBaseURI(env.ResourceManagerEndpoint, subscriptionID),
+		resourcesClient:       resources.NewGroupClientWithBaseURI(env.ResourceManagerEndpoint, subscriptionID),
+		providersClient:       resources.NewProvidersClientWithBaseURI(env.ResourceManagerEndpoint, subscriptionID),
+		virtualMachinesClient: compute.NewVirtualMachinesClientWithBaseURI(env.ResourceManagerEndpoint, subscriptionID),
+	}
+
+	authorizer := autorest.NewBearerAuthorizer(armSpt)
```
rjtsdl: I don't know there is a NewBearerAuthorizer. I will use it in my PR. Thx

pkg/acsengine/transform.go
colemickens: It only works for scale up though, and Anubhuti is going to be refactoring/renaming in this file in the next set of PRs.

PR #627, opened 2017-05-15
First round of i18n in acs-engine based on gettext. Subsequent change…

JiangtianLi:
…s will replace and extract resource strings in acs-engine and make acsengine/api package's i18n used by other go program.

This PR is to get preview and feedback. Subsequent change will do the real format in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/627)


JiangtianLi:
Note, please don't merge as there will be more commit.

Dockerfile
```diff
@@ -29,6 +29,10 @@ RUN git clone https://github.com/akesterson/cmdarg.git /tmp/cmdarg \
 RUN git clone https://github.com/akesterson/shunit.git /tmp/shunit \
     && cd /tmp/shunit && make install && rm -rf /tmp/shunit
 
+# Go tool for internationalization and localization
+RUN go get github.com/JiangtianLi/gettext/... \
```
amanohar: Should ACS Engine be getting stuff from private repos?

cmd/generate.go
```diff
@@ -18,17 +20,19 @@ const (
...
 
 type generateCmd struct {
-	apimodelPath      string
-	outputDirectory   string // can be auto-determined from clusterDefinition
-	caCertificatePath string
-	caPrivateKeyPath  string
-	classicMode       bool
-	noPrettyPrint     bool
-	parametersOnly    bool
+	translationsDirectory string
```
amanohar: Caller should not need to pass translation dir to any command ever. This decreases usability of ACS Engine. This should happen automatically. In fact do we even care of localize ACS Engine commands? Localization is primarily needed for the RP.

cmd/generate.go
```diff
@@ -18,17 +20,19 @@ const (
 )
 
 type generateCmd struct {
```
amanohar: What about deploy and update commands?

cmd/generate.go
```diff
@@ -59,6 +64,13 @@ func (gc *generateCmd) validate(cmd *cobra.Command, args []string) {
 	var caCertificateBytes []byte
 	var caKeyBytes []byte
 
+	locale, err := i18n.LoadTranslations(gc.translationsDirectory)
```
amanohar: It seems that a lot of setup is happening even before ACS Engine functions are called. Which means evey time someone needs to use the primary ACS Engine functions (in the commands or RP) these steps need to be followed. IMO this should be as simple as translation dir being passed to InitializeTemplateGenerator. And then everything is setup inside it.

Dockerfile
JiangtianLi: I tried not to but I spent some time and didn't find a tool. That package is basically used for extract strings out of source file for translation so it will only run every time we update translation strings. I forked https://github.com/gosexy/gettext to address the parsing we need. I can try to merge into upstream.

cmd/generate.go
JiangtianLi: You are right, localization of ACS Engine command has lower priority. The primary translation happens in acsengine and api package, for the purpose of RP. I added here for the e2e test and it is nice to have for ACS Engine. For the translationsDirectory  parameter, it will not be needed and default to the translation folder if acs-engine is run at the package root directory (I think that's the most used scenario?) It has the issue if go install is used without deploy translation file. Another approach is to used go-bindata to avoid such caveat. I'll look into it first.

cmd/generate.go
JiangtianLi: Thanks for the suggestion. The translation init routine includes set the locale, bind to translation directory, and select domain. I thought about the approach of have package level init function and have each exported function call it if it haven't called before. [GNU gettext doc](https://www.gnu.org/software/gettext/manual/gettext.html#Libraries) has the guidance of how to init in library. However, in our case, the ACS Engine package is called by RP, which serves requests with different language on multiple thread. For example, if InitializeTemplateGenerator is called by the same thread on different language, the locale needs to be updated. If InitializeTemplateGenerator is called by different thread, each thread needs to keep its own locale. If InitializeTemplateGenerator is called by the same thread with the same language, then we don't need to init again. I haven't found a simple way to handle that other than the caller needs to do some initialization and pass the locale in the context (adopting  [Go context pattern)](https://blog.golang.org/context).

cmd/generate.go
JiangtianLi: Thanks for reminding. I didn't notice that and will address it.

Dockerfile
JiangtianLi: I submitted a PR to that tool: https://github.com/gosexy/gettext/pull/16

JiangtianLi:
@acs-bot test this please

JiangtianLi:
@amanohar Can you please take a look? There is line ending issue caused in [commit](https://github.com/Azure/acs-engine/pull/627/commits/1867b0c3d665524342a6172c23fffe9e8ebcba66) so it appears to have a lot of changes. Although those are benign, I discussed with Cole and will address it later (Cole already has related #650). So please don't merge.

JiangtianLi:
@acs-bot test this please

JiangtianLi:
After merge and due to #663, the CRLF changes have been merged and there are much less changes now.

shrutir25:
@JiangtianLi how do we want to proceed with this PR ?

JiangtianLi:
@shrutir25 It is pending approval. It has been a while so merge is probably needed.

JiangtianLi:
@rjtsdl @weinong PTAL

pkg/api/apiloader.go
```diff
@@ -2,39 +2,44 @@ package api
...
+	Translator *i18n.Translator
+}
+
 // LoadContainerServiceFromFile loads an ACS Cluster API Model from a JSON file
-func LoadContainerServiceFromFile(jsonFile string) (*ContainerService, string, error) {
+func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string) (*ContainerService, string, error) {
 	contents, e := ioutil.ReadFile(jsonFile)
 	if e != nil {
-		return nil, "", fmt.Errorf("error reading file %s: %s", jsonFile, e.Error())
+		return nil, "", a.Translator.Errorf("error reading file %s: %s", jsonFile, e.Error())
```
rjtsdl: Do you need to replace more places with `Translater.Errorf`? At least i know i introduced quite a few of them ...

rjtsdl:
@acs-bot test this please

pkg/api/apiloader.go
JiangtianLi: @rjtsdl Yes, the change is incremental and we can to that in several iterations. The bottom line is to get the framework setup so that people can start use it.

weinong:
Reviewed 1 of 28 files at r2, 114 of 1420 files at r4, 9 of 16 files at r5, 11 of 33 files at r6, 15 of 18 files at r7.
Review status: 144 of 145 files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/i18n/i18n.go, line 117 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomnyiYMU3r69SaBxx5:-KomnyiYMU3r69SaBxx6:b-hwzm42) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/i18n.go#L117)):*
> ```Go
> }
> 
> // NT translates a text string into the appropriate plural form, based on GNU's gettext library.
> ```

any example?

---

*[pkg/i18n/i18n.go, line 127 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomnpCI40_HPeGT3mBO:-KomnpCJME7Jpp07pAIl:b-68mvx1) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/i18n.go#L127)):*
> ```Go
> }
> 
> // NErrorf produces an error with a translated error string in the appropriate plural form.
> ```

any example of using this?

---

*[test/i18n/i18ntestinput.go, line 8 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomqIWSm5XGofUJLxTL:-KomqIWSm5XGofUJLxTM:b-22so5x) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/test/i18n/i18ntestinput.go#L8)):*
> ```Go
> )
> 
> // This file is used to generate the test translation file
> ```

these instructions need to be in README.md perhaps under pkg/i18n

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

weinong:
LGTM
@seanknox as well

Dockerfile
seanknox: Just echoing others and calling out we can't include this change as is, the repo has to be publically accessible.

pkg/i18n/translations/default/LC_MESSAGES/acsengine.po
```diff
@@ -0,0 +1,45 @@
+# Test translations for acsengine package.
+# Copyright (C) 2017
+# Jiangtian Li <jiangtianli@hotmail.com>, 2017.
+#
```
seanknox: I don't think we can include code that isn't MIT/Apache v2/etc.

pkg/i18n/translations/en_US/LC_MESSAGES/acsengine.po
```diff
@@ -0,0 +1,45 @@
+# Test translations for acsengine package.
+# Copyright (C) 2017
+# Jiangtian Li <jiangtianli@hotmail.com>, 2017.
+#
```
seanknox: Copyright problem here too, and other spots. Won't call all of them out.

Dockerfile
seanknox: Looks like https://github.com/JiangtianLi/gettext is public now (might have been from the beginning), so we're OK to use the fork, as long as we've registered it at https://ossmsft.visualstudio.com/_oss.

JiangtianLi:
Review status: 144 of 145 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[pkg/i18n/i18n.go, line 117 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomnyiYMU3r69SaBxx5:-KpXeT8kYSxbbrYpe8f8:b6rl4r8) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/i18n.go#L117)):*
<details><summary><i>Previously, weinong (Weinong Wang) wrote…</i></summary><blockquote>

any example?
</blockquote></details>

Yes, example is in i18n_test.go

---

*[pkg/i18n/i18n.go, line 127 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomnpCI40_HPeGT3mBO:-KpXeea5JzHaXuFB8Giv:b-te3it2) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/i18n.go#L127)):*
<details><summary><i>Previously, weinong (Weinong Wang) wrote…</i></summary><blockquote>

any example of using this?
</blockquote></details>

Also in i18n_test.go

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
@seanknox @anhowe @amanohar PTAL

JiangtianLi:
@acs-bot test this please

JiangtianLi:
Review status: 123 of 160 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/i18n/translations/default/LC_MESSAGES/acsengine.po, line 4 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KkT_lpeH4t6W79i8o8o-r7-4:-Kq9CHU-fzVIyxyQfw9l:bwathnp) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/translations/default/LC_MESSAGES/acsengine.po#L4)):*
<details><summary><i>Previously, seanknox (Sean Knox) wrote…</i></summary><blockquote>

I don't think we can include code that isn't MIT/Apache v2/etc.
</blockquote></details>

Discussed with Sean offline. It is generated by poedit and will be modified by loc team. Currently it follows the same as Kubernetes translation files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
Review status: 123 of 160 files reviewed at latest revision, 10 unresolved discussions.

---

*[pkg/i18n/translations/en_US/LC_MESSAGES/acsengine.po, line 4 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KkT_lpeH4t6W79i8o8q-r7-4:-Kq9C_my2k0LWaVxlO6u:b-7l2cjg) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/pkg/i18n/translations/en_US/LC_MESSAGES/acsengine.po#L4)):*
<details><summary><i>Previously, seanknox (Sean Knox) wrote…</i></summary><blockquote>

Copyright problem here too, and other spots. Won't call all of them out.
</blockquote></details>

The same as other translation file copyright.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
Review status: 123 of 160 files reviewed at latest revision, 10 unresolved discussions.

---

*[test/i18n/i18ntestinput.go, line 8 at r7](https://reviewable.io:443/reviews/azure/acs-engine/627#-KomqIWSm5XGofUJLxTL:-Kq9ChkuAYN_4tbDBt5m:b-kmnnke) ([raw file](https://github.com/azure/acs-engine/blob/384787095b1a212eca7e815978b35160299e544c/test/i18n/i18ntestinput.go#L8)):*
<details><summary><i>Previously, weinong (Weinong Wang) wrote…</i></summary><blockquote>

these instructions need to be in README.md perhaps under pkg/i18n
</blockquote></details>

Added README.md

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
Review status: 123 of 160 files reviewed at latest revision, 10 unresolved discussions.

---

*[Dockerfile, line 33 at r2](https://reviewable.io:443/reviews/azure/acs-engine/627#-KkBzFq9kE63SztEvq5--r2-33:-Kq9Cm1_LwAST4rItb69:bhz1xg2) ([raw file](https://github.com/azure/acs-engine/blob/78be458bee62e08ec58a0415af79afb6ec1bfc9c/Dockerfile#L33)):*
<details><summary><i>Previously, seanknox (Sean Knox) wrote…</i></summary><blockquote>

Looks like https://github.com/JiangtianLi/gettext is public now (might have been from the beginning), so we're OK to use the fork, as long as we've registered it at https://ossmsft.visualstudio.com/_oss.
</blockquote></details>

Resolved offline. The dependence is all MIT licence.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
@acs-bot test this please

JiangtianLi:
Review status: 123 of 160 files reviewed at latest revision, 10 unresolved discussions.

---

*[Dockerfile, line 33 at r2](https://reviewable.io:443/reviews/azure/acs-engine/627#-KkBzFq9kE63SztEvq5--r2-33:-KqAbnVfY9YcSMPgBvxv:b-896fix) ([raw file](https://github.com/azure/acs-engine/blob/78be458bee62e08ec58a0415af79afb6ec1bfc9c/Dockerfile#L33)):*
<details><summary><i>Previously, JiangtianLi (Jiangtian Li) wrote…</i></summary><blockquote>

Resolved offline. The dependence is all MIT licence.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/627)*

JiangtianLi:
@acs-bot test this please

PR #632, opened 2017-05-16
Enable custom cluster subnets for Kubernetes

### LINKED ISSUE #493: Regression to azure network plugin introduced by latest CIDR changes

ofiliz:
This PR is complementary to #473, enabling the same "custom cluster CIDR" feature to run correctly when the azure-vnet CNI plugin is enabled. PR #473 and this PR can be applied in any order to re-enable that functionality. However it is probably easier to merge this PR first to avoid merge conflicts.

The regression caused by #473 was due to the flat subnet used by the azure-vnet plugin not being compatible the way that PR computed subnets. When azure-vnet plugin is enabled, all masters, agents and pods use a single large /12 subnet by default. When computing the value of "clusterCidr" parameter being passed to kubelet and kubeproxy, this must be taken into account as this PR does in the file default.go.

The combined behavior implemented by #473 and this PR can be summarized as follows:

When azure-vnet plugin is disabled (Default)

ClusterSubnet | Master#1 IP address | Agent subnet | Pod Subnet
--------------- | --------------- | --------------- | ---------------
Not specified | 10.240.255.5 | 10.240.0.0/16 | 10.244.0.0/16
10.230.0.0/16 | 10.240.255.5 | 10.240.0.0/16 | 10.230.0.0/16

When azure-vnet plugin is enabled

ClusterSubnet | Master#1 IP address | Agent subnet | Pod Subnet
--------------- | --------------- | --------------- | ---------------
Not specified | 10.240.255.5 | 10.240.0.0/12 | 10.240.0.0/12
10.230.0.0/16 | 10.230.255.5 | 10.230.0.0/16 | 10.230.0.0/16

In other words, KubernetesConfig.ClusterSubnet controls only the pod subnet when plugin is disabled. It controls all of masters, agent and pod subnets when plugin is enabled.

Fixes #493 

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/632)


lachie83:
Will review

pkg/acsengine/defaults.go
```diff
@@ -82,6 +82,14 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 		if a.OrchestratorProfile.KubernetesConfig.NetworkPolicy == "" {
 			a.OrchestratorProfile.KubernetesConfig.NetworkPolicy = DefaultNetworkPolicy
 		}
+		if a.OrchestratorProfile.KubernetesConfig.ClusterSubnet == "" {
+			if a.OrchestratorProfile.IsVNETIntegrated() {
+				// When VNET integration is enabled, all masters, agents and pods share the same large subnet.
```
seanknox: What do you think about either of the following to cut down on nested conditionals:

1. refactor into a function, e.g. something like `vnetIsIntegrated()`
1. multiple conditions
```golang
		if a.OrchestratorProfile.KubernetesConfig.ClusterSubnet == "" && a.OrchestratorProfile.IsVNETIntegrated() {
			a.OrchestratorProfile.KubernetesConfig.ClusterSubnet = DefaultKubernetesSubnet
		} else {
			a.OrchestratorProfile.KubernetesConfig.ClusterSubnet = DefaultKubernetesClusterSubnet
		}
```

pkg/acsengine/defaults.go
```diff
@@ -238,3 +247,27 @@ func certGenerationRequired(a *api.Properties) bool {
 		return false
 	}
 }
+
+// getFirstConsecutiveStaticIPAddress returns the first static IP address of the given subnet.
+func getFirstConsecutiveStaticIPAddress(subnetStr string) string {
+	_, subnet, err := net.ParseCIDR(subnetStr)
+	if err != nil {
+		return DefaultFirstConsecutiveKubernetesStaticIP
+	}
```
seanknox: You should handle the error case here.

pkg/api/converterfromapi.go
```diff
@@ -383,6 +383,17 @@ func convertOrchestratorProfileToVLabs(api *OrchestratorProfile, o *vlabs.Orches
 	if api.OrchestratorVersion != "" {
 		o.OrchestratorVersion = vlabs.OrchestratorVersion(api.OrchestratorVersion)
 	}
+
+	if api.KubernetesConfig != nil {
+		o.KubernetesConfig = &vlabs.KubernetesConfig{}
+		convertKubernetesConfigToVLabs(api.KubernetesConfig, o.KubernetesConfig)
```
seanknox: When would this be nil? Could this be written to be more declarative?

pkg/acsengine/defaults.go
ofiliz: (2) wouldn't work for the case where user-specified ClusterSubnet != "". We would override it with the default. This has to be nested because the inner if check must run iff KubernetesConfig.ClusterSubnet == "".

I thought this logic is simple enough and refactoring it to another function would just make the code less readable.

pkg/api/converterfromapi.go
ofiliz: This is from #473, I'm just re-adding the same code that was previously here.

KubernetesConfig could be nil if the input apimodel to acs-engine does not have a KubernetesConfig section. But I think it is also a good defensive practice to check for nil before accessing pointers.

Is there a better way to write this? What did you have in mind?

anhowe: Onur is correct, this is nil, when the customer hasn't specified KubernetesConfig in the model.

pkg/acsengine/defaults.go
ofiliz: Yep, maybe returning the default is not the right thing to do. Will update.

pkg/api/converterfromapi.go
seanknox: > KubernetesConfig could be nil if the input apimodel to acs-engine does not have a KubernetesConfig section. But I think it is also a good defensive practice to check for nil before accessing pointers.

Sure, I'm trying to understand what situations exist where that would be true.

ofiliz: I think when customer KubernetesConfig is nil, we still insert an empty KubernetesConfig struct during the initial parsing phase. This if check is just being defensive, which I think is the right thing to do.

seanknox: Got it, that makes sense. Thanks!

pkg/acsengine/defaults.go
ofiliz: On second thought, this is the "defaults" path, which is just supposed to set the defaults and therefore it is not supposed to fail. It fails only because acs-engine sets the defaults _after_ it parses the customer's input json file, so I was forced to put the input parsing here.

@anhowe Would you like me to update and return an error for user input in this set-default path, or do you think you guys are going to change the acs-engine to set the defaults first and then load the user input file?

I did implement the error case, either way is fine for me.

seanknox:
@ofiliz Thanks for answering my questions. I didn't see any tests included—what is the team policy around testing?

ofiliz:
@anhowe I created 4 clusters for the 4 variations you mentioned

- [x] Verified that the given subnet is actually used
- [x] Verified that the kube-services previously crashing (DNS and dashboard) have a "RESTARTS" counts of zero after 10 minutes.

@seanknox I think #473 has the test cases, this is just fixing the regression in it. Let me check.

seanknox:
> @seanknox I think #473 has the test cases, this is just fixing the regression in it. Let me check.
 
Right—if this fixes a regression, shouldn't the tests that missed the regression be updated?

pkg/acsengine/defaults.go
ofiliz: I updated the commit to do the subnet checks during validation phase, in validate.go. I also added corresponding test cases in validate_test.go.

I'm still returning the default here, just to cover the case when the hardcoded default in const.go is somehow not a valid subnet. I did not want to pollute the defaults logic here with validation code which also requires updating these functions to return error codes all the way up to main().

ofiliz:
Pushed new commit that adds test cases covering KubernetesConfig.ClusterSubnet validation.

lachie83:
If there's no urgency I would still like to take a pass at reviewing this.

ofiliz:
Renamed the PR because the changes have grown based on review feedback. We are now enabling this feature for both the VNET-integrated (CNI plugin) and not-VNET-integrated (UDR and NAT) case.

Pushed a commit to update cluster definition documentation.

docs/clusterdefinition.md
```diff
@@ -31,7 +31,8 @@ Here are the valid values for the orchestrator types:
 |Name|Required|Description|
 |---|---|---|
 |kubernetesImageBase|no|This specifies the image of kubernetes to use for the cluster.|
-|networkPolicy|no|Specifies the network policy tool for the cluster. Default is `none`, which won't enforce network policy. This can be set to `calico` for clusters with Linux agents only.|
+|networkPolicy|no|Specifies the network policy tool for the cluster. Valid values are:<br>`none` (default), which won't enforce any network policy,<br>`azure` for applying Azure VNET network policy,<br>`calico` for Calico network policy for clusters with Linux agents only.<br>See [network policy examples](../examples/networkpolicy) for more information.|
+|clusterSubnet|no|The IP subnet used for allocating IP addresses for pod network interfaces.|
```
lachie83: Can you please mention the default

lachie83: Are there any restrictions on what this subnet can be? For example does it have to fall under the VNet supernet? Must it be RFC1918 compliant?

dmitsh:
@acs-bot test this please

dmitsh:
@acs-bot test this please

pkg/acsengine/defaults.go
seanknox: 👍

dmitsh:
@acs-bot test this please

PR #634, opened 2017-05-16
operations: add CleanDeleteVirtualMachine

colemickens:
- add new ACSEngineClient functions as needed
- add new interface for the AzureStorageClient
  (it's a separate interface because there might be more
  than one storage account used in the context of a request)

`CleanDeleteVirtualMachine` deletes a VM and the associated NIC and OS Disk blob.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/634)


pkg/armhelpers/storage.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// DeleteBlob deletes the specified blob
+// TODO(colemick): why doesn't SDK give a way to just delete a blob by URI?
+// it's what it ends up doing internally anyway...
+func (as *AzureStorageClient) DeleteBlob(vhdContainer, vhdBlob string) error {
+	bs := as.client.GetBlobService()
+	containerRef := bs.GetContainerReference(vhdContainer)
+	blobRef := containerRef.GetBlobReference(vhdBlob)
+
```
JackQuincy: Does the skills specify page or block? or is that handled internally?

JackQuincy: SDK not skills... Autocorrect 😀

colemickens: Doesn't matter. When the call hits the wire it's just a generic HTTP DELETE against the blob URI. Same semantics for block/page blobs.

colemickens:
@JackQuincy Can you elaborate on this: *"Though I'd like a unit test validating the passing around of the os disk uri in the vm delete method."* I do have a unit test that checks the proper decomposition of the VHD URI... are you looking for something beyond that?

colemickens:
@acs-bot test this please

JackQuincy:
Yes and that was great I just wanted in level higher as well. Where you mock the responses from the clients and pass things between the clients. Maybe that is more important when we start doing more error handling. Understood if it is it out of scope.

colemickens:
@acs-bot test this please

amanohar:
https://github.com/Azure/acs-engine/pull/634

PR #641, opened 2017-05-17
force overwrite=false for disk_setup and fs_setup on etcd disk

colemickens:
We only want to set the partition table and format the first partition on initial VM boot. Regardless of if cloud-init is meant to re-execute those stages, we don't want them to overwrite. Specifically, consider an upgrade scenario with immutable replacement of VMs. We don't want the etcd data disk to be reattached and then wiped/formatted during the (new) VM's initial boot.

Not addressed in this PR: There is concern around whether `fs_setup` will respect `overwrite` if a custom command is set (though I'm not sure the `extra_opts` alone qualifies as a custom command)

cc: @amanohar

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/641)


colemickens:
@acs-bot test this please

colemickens:
This already was merged as part of one of Anubhuti's PRs. Closing.

Issue #591, opened 2017-05-09
K8s Agent Cannot Use Standard_F1


robbiezhang:
Standard_F1 is defined in GetKubernetesAgentAllowedSizes,
https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/azureconst.go#L324
but it doesn't exist in GetSizeMap:
[https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/azureconst.go#L560](url)

The ARM template is generated successfully, but it failed to process.

> The language expression property 'Standard_F1' doesn't exist. available properties are ...

JackQuincy:
@shrutir25 was this something we meant to do?

anhowe:
@robbiezhang Standard_F1 is supported for Agents but not for Masters.  Masters require >= 2 cores.

anhowe:
replace https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/Get-AzureConstants.ps1#L264-L276, with all sizes

anhowe:
@shrutir25 - please note the above

robbiezhang:
This issue is for agent pool. I use Standard_F1 in the agent pool, and generate the ARM template. The error happens when executing the template.

SaaldjorMike:
I believe this one can be closed.

mboersma:
I'm closing this old issue since acs-engine is deprecated in favor of [aks-engine](https://github.com/Azure/aks-engine).

Issue #353, opened 2017-03-02
Get-AzureConstants.ps1 needs to be converted to a bash script


colemickens:

anhowe:
@shrutir25 has this been completed?  I still see the .ps1 file.

jackfrancis:
Hopefully python fits the bill!

https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/Get-AzureConstants.py

PR #644, opened 2017-05-17
Python get azure constants file

shrutir25:
- Added a python file to generate azure constants
- addressed issue #591 by opening up Kubernetes sizes in GetSizeMap
- updated azureconst.go file

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/644)


shrutir25:
@anhowe - could you review this ?

dmitsh:
@acs-bot test this please

pkg/acsengine/Get-AzureConstants.py
```diff
@@ -0,0 +1,241 @@
...
+    if 'S' in capability or 's' in capability:
+        return "Premium_LRS"
+    else:
+        return "Standard_LRS"
+
+def getFileContents(dcosMasterMap, masterAgentMap, kubernetesAgentMap, sizeMap, locations):
+    text = 'package acsengine'
+    text += '\n\nimport "fmt"'
+
+    text += '\
```
colemickens: It should be possible to use a different type of string literal to avoid the manual newline characters through this.

Generally, I still wish that this only generated a single file with constants in it. I don't understand why we're spitting out static functions in a generated file (FormatAzureProdFQDNs, why are the FQDN formats in this file, etc)

pkg/acsengine/Get-AzureConstants.py
```diff
@@ -0,0 +1,241 @@
...
+        // AzurePublicProdFQDNFormat specifies the format for a prod dns name \n\
+        AzurePublicProdFQDNFormat = "%s.%s.cloudapp.azure.com"\n\
+        //AzureChinaProdFQDNFormat specify the endpoint of Azure China Cloud\n\
+        AzureChinaProdFQDNFormat = "%s.%s.cloudapp.chinacloudapi.cn"\n\
+)\n\n'
+
+    text += '// AzureLocations provides all azure regions in prod.\n\
+// Related powershell to refresh this list:\n\
+//   Get-AzureRmLocation | Select-Object -Property Location\n\
+var AzureLocations = []string{\n'
```
colemickens: Generally, I still wish that this only generated a single file with only constants in it. I don't understand why we're spitting out static functions in a generated file (FormatAzureProdFQDNs, why are the FQDN formats in this file, etc)

shrutir25: I agree. It should be a file with only constants in it. I replicated everything that was in the old azureconst.go. We could remove it.
@anhowe - should we remove the static functions from azureconst.go ?

anhowe: Let's focus this PR on #591 and #353.  Please file another issue to split this file in two, azureconst.go, and something like azure.go with the following functionality:
- azure.go - reduce constants to contain struct of sizes.
- azureconst.go - split out the functions into an azure.go, that explicitly formats and uses the data from this file

shrutir25:
@colemickens  I updated the string literals. Could you take a look ?

PR #646, opened 2017-05-17
bugfix `deploy` command

colemickens:
Some commits that missed my initial `deploy` PR apparently.

* Fixes some UX problems
* a nil deref if flags were missing
* auto-creates the RG if it's not there
* drops the polling duration for template deployments to 5 seconds

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/646)

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -110,7 +110,7 @@ func (gc *generateCmd) validate(cmd *cobra.Command, args []string) {
 }
 
 func (gc *generateCmd) run() error {
-	log.Infoln("Generating...")
+	log.Infoln("Generating assets...")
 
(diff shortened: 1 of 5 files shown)


cmd/generate.go
```diff
@@ -110,8 +110,6 @@ func (gc *generateCmd) validate(cmd *cobra.Command, args []string) {
 }
 
 func (gc *generateCmd) run() error {
-	log.Infoln("Generating...")
```
amanohar: Why not leave this comment in the command?

colemickens: I'll put it back, not sure why I'd removed.

PR #650, opened 2017-05-18
Fix .gitattributes to normalize EOL to LF. Normalize existing content.

colemickens:
This seems to do the right thing, based on reading and testing (natively from Linux, natively from macOS, natively from Windows [using Git Bash]. I'm not sure about from Docker in Windows yet, hopefully someone can try for me; I don't have a way of testing that).

When I first pulled this branch on Windows and ran `glide install`, it showed files in `git status` as being modified, but `git status` showed no changes after executing `git add -A .`. Further invocations of `glide install` had `git status` consistently showing no changes (good!).

TODO: Can someone who uses the docker container on Windows, pull this branch, run `glide install` and then `git add -A .` and then check to see if `git status` shows files as modified. thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/650)


amanohar:
@colemickens, I tried glide install and then git add -A . It doesn't seem to resolve the issue for me. Still shows files as modified.

colemickens:
This fix doesn't seem to be sufficient. Here's the problem. Git does different things depending on if it thinks it's running in Windows or Linux: 
problem-iter1

JackQuincy:
When you install git in windows you configure it to use lf or crlf. I'm not sure you can overwrite it from the repo. I wonder if we could make the windows Dev container do similar. Interesting to note. I haven't had this issue in gb with our windows Dev containers

colemickens:
It seems to depend on where I run the `glide` command from. If I normalize to LF, then only do `glide`, etc from within the container (without changing the .gitattributes from what is in master now) then it seems to handle everything okay.

But if I run it from Windows natively, it writes in native Windows line format, and then tries to commit like that even though I have `* -crlf` in `.gitattributes`.

.gitattributes
```diff
@@ -1 +1 @@
-* -crlf
+* eolf=lf
```
JiangtianLi: I didn't find eolf from https://git-scm.com/docs/gitattributes. The closest is eol=lf. But I am not familiar with gitattributes ...

colemickens: gah, thanks. I will have to retest now that I've fixed my typos.

also, fwiw, I'm changing this branch around a lot right now just trying to find a combination that maybe does the right thing...

JiangtianLi: Thanks for the fix.

seanknox:
FWIW I think https://github.com/Azure/acs-engine/pull/663 will fix the line encodings issue.

colemickens:
Yeah #663 is good. It still has the same problem if you do some actions inside the docker container on windows and then do the git actions inside Windows natively... but I guess Windows devs are going to just have to be a bit careful... I spent many hours on trying to get a Windows config that works consistently and couldn't seem to get it to do the right thing.

Thanks for #663. Closing this out since it's basically the bulk of #663.

seanknox:
@colemickens no worries. the bad news is folks have to do a bit of manual config on their end, but once they do, git should do the "write" (get it? sorry).

PR #651, opened 2017-05-18
Enable staged upgrade workflow for multiple master Kubernetes cluster

amanohar:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/651)


pkg/operations/kubernetes162upgrader.go
```diff
@@ -64,14 +66,21 @@ func (ku *Kubernetes162upgrader) RunUpgrade() error {
...
 	upgradeMasterNode.Client = ku.Client
 
-	// for _, vm := range *ku.ClusterTopology.MasterVMs {
-	// 1.	Shutdown and delete one master VM at a time while preserving the persistent disk backing etcd.
-	upgradeMasterNode.DeleteNode()
-	// 2.	Call CreateVMWithRetries
-	upgradeMasterNode.CreateNode(loopCount)
-	loopCount++
+	// Sort by VM Name (e.g.: k8s-master-22551669-0) offset no. in descending order
+	sort.Sort(sort.Reverse(armhelpers.ByVMNameOffset(*ku.ClusterTopology.MasterVMs)))
```
colemickens: why `sort.Sort(sort.Reverse(...))`?

I could reason about `sort.Reverse(sort.Sort(...))` maybe but not the other way, but I've not used the sort interface before so I'm probably missing it.

colemickens:
(oh and squash again please :) )

pkg/operations/kubernetes162upgrader.go
amanohar: Because if you see the sort implementation for Go (sort.Interface) it only has Less function that needs to be implemented. So basically you can only sort ascending and the only way I could find to sort descending is by reversing the sorted list.

amanohar: The only other way to 'mis-implement' the Less function and have it do descending instead of ascending.

PR #672, opened 2017-05-24
add DeleteResourceGroup function in ArmHelpers

rjtsdl:

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/672)

### DIFF
--- a/pkg/armhelpers/groupsclient.go
+++ b/pkg/armhelpers/groupsclient.go
@@ -2,6 +2,7 @@ package armhelpers
 
 import (
 	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/prometheus/common/log"
 )
 
@@ -17,4 +18,9 @@ func (az *AzureClient) EnsureResourceGroup(name, location string) (resourceGroup
 	}
 
 	return &response, nil
+}
+
+// DeleteResourceGroup delete the named resource group
+func (az *AzureClient) DeleteResourceGroup(name string, cancel chan struct{}) (<-chan autorest.Response, <-chan error) {
+	return az.groupsClient.Delete(name, cancel)
 }


pkg/armhelpers/groupsclient.go
```diff
@@ -18,3 +20,16 @@ func (az *AzureClient) EnsureResourceGroup(name, location string) (resourceGroup
 
 	return &response, nil
 }
+
+// DeleteResourceGroup delete the named resource group
+func (az *AzureClient) DeleteResourceGroup(name string, cancel chan struct{}) error {
+	resCh, errCh := az.groupsClient.Delete(name, cancel)
+	res := <-resCh
+	err := <-errCh
+	// When resourceGroup not exists, we will get 404
```
weinong: should we simply return the channels to let the caller decide how to deal with them? I can argue that we may want to set a timeout instead of waiting it forever.

rjtsdl: the cancel channel is there, it could be used for timeout scenario.

amanohar: Yes, we should just return the channel and not trying to do too much wrapping and hiding of constructs returned by the clients.

pkg/armhelpers/groupsclient.go
```diff
@@ -18,3 +20,16 @@ func (az *AzureClient) EnsureResourceGroup(name, location string) (resourceGroup
...
+	// Explictly set the error to nil for this scenario
+	if res.Response != nil && res.StatusCode == http.StatusNotFound {
```
amanohar: Why the special casing of not found and response?

rjtsdl: I have removed this part. It is our RP's logic. 

res.StatusCode is actually res.Response.Response.StatusCode. It is the embedding confusion.

seanknox:
Hey @rjtsdl @amanohar @weinong, in the future, could you fill out the PR form with some info about the PR? Short and sweet is fine. We're using this info to plan milestones and releases going forward.

rjtsdl:
This one is somehow for the same purpose of my previous one. Anubhuti already ahve the context. 
But, you are right. I will add more context.

PR #682, opened 2017-05-26
Pass proxy settings to docker build

kaskavalci:
**What this PR does / why we need it**: `docker build` does not work behind proxy. This step is crucial for corporate environments.

**Release note**: None

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/682)

### DIFF
--- a/scripts/devenv.sh
+++ b/scripts/devenv.sh
@@ -3,7 +3,14 @@
 set -eu -o pipefail
 set -x
 
-docker build --pull -t acs-engine .
+docker build --pull -t \
+	--build-arg http_proxy=$http_proxy \
+	--build-arg https_proxy=$https_proxy \
+	--build-arg no_proxy=$no_proxy \
+	--build-arg HTTP_PROXY=$HTTP_PROXY \
+	--build-arg HTTPS_PROXY=$HTTPS_PROXY \
+	--build-arg NO_PROXY=$NO_PROXY \
+	acs-engine .
 
 docker run -it \
 	--privileged \


scripts/devenv.sh
```diff
@@ -3,7 +3,14 @@
 set -eu -o pipefail
 set -x
 
-docker build --pull -t acs-engine .
+docker build --pull -t \
```
anhowe: by adding these build arguments, what happens if the variables like $http_proxy do not exist (for non corp users)

kaskavalci: cURL, Wget, apt-get, npm, pkgin treat empty `http_proxy` as non existent. However, custom implementations may fail if they do not account empty string in proxy variable. Safe approach would to check if http_proxy envvar is present, then call `docker build` with proxy variables. Are you OK with that approach?

shrutir25: @anhowe - how do we want to proceed with this PR ?

jchauncey:
This pr is fairly old now and probably is not the direction we want to proceed with the environment. My opinion would be that if you need this level of customization to work against acs-engine its better to have it as a one off script that gets you into the appropriate environment rather than trying to build a catch all script for everyone.

@anhowe @jackfrancis thoughts?

jackfrancis:
@kaskavalci I agree that this doesn't have general purpose use. If you'd like to re-open a new PR with proxy options as an additive feature, that would be fine. (In other words, we don't want to introduce any change to existing usage patterns of this script.)

PR #684, opened 2017-05-26
Adding scale down group of vms method

JackQuincy:
What this PR does / why we need it:

Will be used in k8s scale down scenarios.
Also switched to using logrus logger so that we can get pass in a logger that has info about the request attached to the log, like request id

We need it to support our release of the RP using acs engine and in the future to support scaling down VMAS acs-engine clusters/

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/684)


pkg/operations/scaledownagentpool_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/Azure/acs-engine/pkg/armhelpers"
+	log "github.com/Sirupsen/logrus"
+	. "github.com/onsi/gomega"
+)
+
+func Test_ScaleDownVMs_ErrorPath(t *testing.T) {
+	RegisterTestingT(t)
+	errs := ScaleDownVMs(&armhelpers.FailingMockClient{}, log.NewEntry(log.New()), "rg", "vm1", "vm2", "vm3", "vm5")
+	Expect(errs.Len()).To(Equal(4))
```
rjtsdl: We should use gingko style testing.

pkg/operations/scaledownagentpool.go
```diff
@@ -0,0 +1,44 @@
...
+type VMScalingErrorDetails struct {
+	Name  string
+	Error error
+}
+
+// ScaleDownVMs removes the vms in the provided list. Returns a list with details on each failure.
+// all items in the list will always be of type *VMScalingErrorDetails
+func ScaleDownVMs(az armhelpers.ACSEngineClient, logger *log.Entry, resourceGroup string, vmNames ...string) *list.List {
+	numVmsToDelete := len(vmNames)
+	errChan := make(chan *VMScalingErrorDetails, numVmsToDelete)
```
rjtsdl: Nice one!

JackQuincy:
@rjtsdl can you take a look I've updated to using ginkgo

seanknox:
@jackfrancis @rjtsdl @amanohar @anhowe hey all, are these changes behind a feature flag? Need to ensure that acs-engine users need to opt in.

JackQuincy:
code isn't reachable from an operation. So just dead code as far as acs-engine is concerned today. I wrote the code for use in our managed service with the anticipation we would use it in a scale down operation in acs-engine in the future.

Issue #464, opened 2017-04-04
What are the plans for upgrading clusters on day 365 for k8s clusters


jkurz:
I love the idea of making k8s installations easier for my team. The issue I am trying to wrap my head around is, what happens on day 365 when kubelet needs to be upgraded. What happens if etcd needs to be updated or flannel needs to be updated? Are you planning on creating a guide to do updates, that we can follow or am I on my own, once I deploy my application to production and I need maintain the clusters.

Is the solution to just reprovision a new cluster and migrate all of the containers over? Sometimes that seems like it could be troublesome and possibly cause issues, especially if etcd nodes are running on the same vms that my kubelets are running on. Which is a whole nother issue. 

Thanks
Josh Kurz

JackQuincy:
Upgrade is on our radar. It is also a priority, but we don't have a ETA yet. I'll update here as more info comes.

jkurz:
@JackQuincy awesome. If there is anything you could make public, so the community could help work on the issue together, that would be great. I am willing to pitch in some elbow grease to possibly come up with a solution to make this a real thing. Would be awesome to come up with an automated solution that upgrades clusters based on some type of scripted method. That way all clusters get upgraded in the same way and it can be tested and proven out for each release.

JackQuincy:
Last thinking I heard was an API to trigger upgrade so that enterprise customers wouldn't be forced to upgrade before they could test compatibility from our RP perspective. That will use some scripts or some other code/magic in ACS Engine to trigger/carryout the upgrade. 
This [PR](https://github.com/Azure/acs-engine/pull/467) is our team changing our setup of the Kubernetes clusters to try to make them easier to upgrade. But the plan is to start moving more of our discussion/progress into github.

amanohar:
@jkurz, regarding:
_Would be awesome to come up with an automated solution that upgrades clusters based on some type of scripted method. That way all clusters get upgraded in the same way and it can be tested and proven out for each release._

Yes, that is the plan. It will be an automated solution based in input from the user and scripts so that all clusters follow the same upgrade model.

Karnich:
Any news on this one? :)

dtzar:
I've started to work on a guide which shows how to manually upgrade nodes and migrate workloads using ACS engine - utilizing new resource pools.  I'll share what I have soon.

JackQuincy:
[...]

PR #707, opened 2017-05-31
Cluster upgrade for Kubernetes 1.5.3 -> 1.6.2

amanohar:
**What this PR does / why we need it**:  Enable cluster upgrade from Kubernetes 1.5.3 -> 1.6.2

This PR sets up upgrade workflow that will be followed by ACS Engine and the service to upgrade Kubernetes clusters created using ACS Engine templates running in Azure. 

This feature is required as a step towards enabling customer initiated cluster lifecycle management and  ACS RP to be a managed service.

This PR supports Kubernetes 1.5.3 -> 1.6.2 upgrade only. More upgrade versions support to come later.

_Design assumptions_: The upgrade will work as designed only for clusters where etcd is on a separate data disk.

_Note_: This is by no means a complete upgrade implementation and other enhancements will follow at later points of time (_not all of the following are guaranteed and are examples_): Drain and cordon node before upgrading it, etcd upgrade, rollback & downgrade, post upgrade validation & health checks, maintaining HA of pools during the upgrade operation, etc.

**Release note**:
```
This PR enables following upgrade scenarios for Kubernetes 1.5.3 -> 1.6.2 clusters:
1. Multi master upgrade
2. Multi agent upgrade
3. Multi agent pool (Linux & Windows pools) upgrade
4. Idempotent upgrade operation (with one pending change to enable full idempotency), i.e. operation can be rerun if it fails and the command will pick up where it left off. 
5. Upgrade operation will only upgrade nodes in the resource group that belong this cluster and skips other VMs that are not part of the cluster.
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/707)


Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM buildpack-deps:xenial
 ENV GO_VERSION 1.8
 ENV KUBECTL_VERSION 1.6.0
 ENV AZURE_CLI_VERSION 2.0.3
+ENV ACSENGINE_EXPERIMENTAL_FEATURES 1
```
colemickens: remove, I think?

pkg/acsengine/transform.go
```diff
@@ -197,62 +199,117 @@ func removeImageReference(logger *logrus.Entry, resourceProperties map[string]in
...
-					logger.Warnf("Template improperly formatted")
-					continue
-				}
+			logger.Infoln(fmt.Sprintf("agentPoolsToPreserve: %v...", agentPoolsToPreserve))
 
-				storageProfile, ok := resourceProperties[storageProfileFieldName].(map[string]interface{})
-				if !ok {
-					logger.Warnf("Template improperly formatted")
-					continue
+			if agentPoolsToPreserve == nil || len(agentPoolsToPreserve) == 0 || agentPoolsToPreserve[poolName] != true {
```
colemickens: are we worried about casing at all here? particularly the map lookup?

pkg/armhelpers/util.go
```diff
@@ -35,18 +37,60 @@ func SplitBlobURI(URI string) (string, string, string, error) {
...
-	vm2Num := vm2NameParts[len(vm2NameParts)-1]
+	if err != nil {
+		return "", "", -1, -1, fmt.Errorf("Error parsing VM Name: %v", err)
+	}
+
+	return poolPrefix, "acs", poolIndex, agentIndex, nil
+}
+
+// GetVMNameIndex return VM index of a node in the Kubernetes cluster
+func GetVMNameIndex(osType compute.OperatingSystemTypes, vmName string) (int, error) {
```
colemickens: nit, feel free to ignore: Why not make this take the VM properties blob? Then it could reach in and look at the OS Disk OSType. No need for the caller to pull it out.

pkg/operations/kubernetes162upgradeagentnode.go
```diff
@@ -1,25 +1,74 @@
...
+	agentCount, _ := poolCountParameter["value"]
+	log.Infoln(fmt.Sprintf("Agent pool: %s, set count to: %d temporarily during upgrade...", poolName, agentCount))
+
+	poolOffsetVarName := poolName + "Offset"
+	templateVariables := kan.TemplateMap["variables"].(map[string]interface{})
+	templateVariables[poolOffsetVarName] = agentNo
+	agentOffset, _ := templateVariables[poolOffsetVarName]
+	log.Infoln(fmt.Sprintf("Agent offset: %v", agentOffset))
+
+	WriteTemplate(kan.UpgradeContainerService, kan.TemplateMap, kan.ParametersMap)
```
colemickens: Is this mostly for debug/troubleshooting? Since it will get written over many times during a full cluster upgrade?

pkg/operations/upgradecluster.go
```diff
@@ -61,38 +76,122 @@ func (uc *UpgradeCluster) UpgradeCluster(subscriptionID uuid.UUID, resourceGroup
...
 	if err != nil {
 		return err
 	}
 
 	orchestratorTypeVersion := fmt.Sprintf("%s:%s", uc.DataModel.Properties.OrchestratorProfile.OrchestratorType,
 		uc.DataModel.Properties.OrchestratorProfile.OrchestratorVersion)
+	targetOrchestratorTypeVersion := fmt.Sprintf("%s:%s", uc.UpgradeModel.OrchestratorProfile.OrchestratorType,
+		uc.UpgradeModel.OrchestratorProfile.OrchestratorVersion)
 
+	// TODO: *vm.Tags["resourceNameSuffix"] ==  Read VM NAME SUFFIX and filter out resources
```
colemickens: seems like this TODO is implemented already?

pkg/operations/upgradecluster.go
```diff
@@ -61,38 +76,122 @@ func (uc *UpgradeCluster) UpgradeCluster(subscriptionID uuid.UUID, resourceGroup
...
+		*uc.AgentPools[poolIdentifier].UpgradedAgentVMs = append(*uc.AgentPools[poolIdentifier].UpgradedAgentVMs, vm)
+	}
+
+	return nil
+}
+
+// WriteTemplate writes upgrade template to a folder
+func WriteTemplate(upgradeContainerService *api.ContainerService,
+	templateMap map[string]interface{}, parametersMap map[string]interface{}) {
+	// ***********Save upgrade template*************
```
colemickens: seems like this is also already implemented.

pkg/acsengine/transform.go
amanohar: I am pulling the pool name out of the VM name and comparing it with the tag that was created during template creation so they should be the same unless someone modifies it. I can write a custom comparator for this to ignore case when searching a map but don't see an immediate issue for now.

pkg/operations/kubernetes162upgradeagentnode.go
amanohar: Removed

pkg/operations/upgradecluster.go
amanohar: You are right, removed.

pkg/operations/upgradecluster.go
amanohar: Removed

amanohar:
@acs-bot test this please

amanohar:
@acs-bot test this please

amanohar:
#464

PR #709, opened 2017-05-31
feat(e2e/kubernetes): Adds a suite of e2e kubernetes tests

jchauncey:
* This adds the start of an e2e test suite built in Ginkgo.
* Tests are modeled after the current {ORCHESTRATOR}/test.sh suite.
* For now this is only a port of the kubernetes suite.
* It also is not currently being run in CI.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/709)


glide.lock
```diff
@@ -38,11 +40,11 @@ imports:
 - name: github.com/Sirupsen/logrus
   version: ba1b36c82c5e05c4f912a88eab0dcd91a171688f
 - name: github.com/spf13/cobra
-  version: 4cdb38c072b86bf795d2c81de50784d9fdd6eb77
+  version: ca57f0f5dba473a8a58765d16d7e811fb8027add
```
colemickens: Why are other packages being updated here?

test/e2e/azure/account.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"encoding/json"
+	"log"
+	"os/exec"
+
+	"github.com/kelseyhightower/envconfig"
+)
+
+// Account holds the values needed to talk to the Azure API
+type Account struct {
```
colemickens: Could we just let `acs-engine` do all of this since we can now do `acs-engine deploy ...` that will do the login, generation and deployment?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,128 @@
+package kubernetes
```
colemickens: Why don't we just run the full k8s e2e suite, or run subsets of it? Otherwise it seems like we're potentially needlessly duplicating effort that has already gone on upstream.

glide.lock
jchauncey: There is no defined version constraint for cobra so when the glide.lock was generated to add envconfig it must have updated that version.

test/e2e/azure/account.go
jchauncey: Right now its about porting what is currently being done in the bash scripts to something a bit more manageable.

test/e2e/kubernetes/kubernetes_test.go
jchauncey: We will be porting the other orchestration engines which may or may not have upstream e2e tests.

glide.lock
colemickens: Did you run `glide up`? I don't understand how it would've just seemingly randomly picked a new version?

test/e2e/azure/account.go
colemickens: Okay, it just sort of duplicates the auth stuff that `acs-engine deploy` now does directly and is a fair amount of code that could just be skipped if we called `acs-engine deploy` directly.

test/e2e/kubernetes/kubernetes_test.go
colemickens: I don't see why that means we wouldn't use an existing suites of tests... especially a suite that already exists, already needs to be run anyway and is run using the ginkgo tech we're using here.

glide.lock
jchauncey: Because you dont have a version specified in the glide.yaml when I did `glide get github.com/kelseyhightower/envconfig` it updated the lock file and pulled in a new version. 

You can reproduce this easily on master by just running the above glide get command and diffing the .lock file.

test/e2e/kubernetes/kubernetes_test.go
jchauncey: Sure I see no reason not to eventually have a job that runs those tests for us since it is a validation that the kubernetes release will run on azure. But as it stands I'm not I can think of a scenario where those tests would prove anything that should stop a release of acs-engine (or maybe even the RP). They would, however, stop a release of kubernetes/kubernetes.

glide.lock
jchauncey: updated PR to pin pflag, go-homedir, and cobra to the versions in the glide.lock before this pr was submitted.

seanknox:
@dmitsh do take a look at this when you can.

jchauncey:
@dmitsh can you take a look at this please?

dmitsh:
sorry for delay. will take a look.

test/e2e/kubernetes/kubernetes_test.go
colemickens: The e2e tests in Kubernetes are:
* already in Ginkgo
* maintained by someone else
* a super set of the validation tests that have been ported from bash
* (are already all being run against ACS-Engine in a few different Jenkins jobs that are not used to gate PRs)

So why not just invoke the relevant small subset of the Kubernetes e2es instead of authoring our own copy to now maintain?

jchauncey: I would leave that up to @seanknox, @sgoings, and @anhowe to figure out. That was not what this PR was about.

jchauncey: Part of the reason not to run the kubernetes e2e tests is time. We aren't trying to validate the entirety of a kube install just that when we provision an acs cluster that it comes up in a valid state. The e2e tests for kubernetes should be validating that kubernetes itself works on azure which is an entirely different set of tests. I think we should move forward with porting our validation tests to ginkgo.

seanknox: @colemickens Good point about the existing Kubernetes e2e tests, I think we should re-orient as soon as possible to re-using existing validation tests and maintain only what is not already covered. That said, I think these changes will improve the e2e tests are in use in acs-engine right now. @jchauncey @dmitsh what do you think?

shrutir25:
@dmitsh - how do we want to proceed with this PR ?

jchauncey:
Im just going to close this. We can revisit it at a later date when we are ready to port the existing bash tests to ginkgo

PR #716, opened 2017-06-01
fix(glide): Pin the pflag, go-homedir, and cobra deps to a specific sha

jchauncey:
This PR locks down the versions for `pflag`, `go-homedir`, and `cobra` to their respective shas that are currently in the glide.lock file on master. This PR is a duplicate of a commit I currently have in #709 but we should do this before we cut the release of acs-engine.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/716)


colemickens:
Why is this needed if the shas are already specified in the lock file?

colemickens:
Per https://github.com/Masterminds/glide

>and glide install will install the versions listed in the glide.lock file

What benefit does it provide to lock them this hard in the glide file? This means we won't get security fixes automatically by just executing `glide up`, for example.

jchauncey:
First if the `glide.yaml` changes the lock file is updated and it will pull down the latest changes from that repo. So because the original yaml file did not have any versions/release set it would always update the version to whatever the head of master is. 

These 3 repos do not cut tagged releases so you can have 1 thing or the other. You either pin it to a sha or you are open to the world and accept that you may get an update that was not intended.

colemickens:
I don't understand at all. Can we back up? What is the goal of this PR?

Simply running `glide install` should never change the revisions installed. AFAICT based on using glide and reading the docs, this only makes it harder for us to be able to run `glide up` and get security updates, etc.

jchauncey:
In my other PR you brought up the fact that the lock file included a new version and that should not have happened. So the only way to stop that from happening is to pin these dependencies to a sha (since they do not do github releases). That way we can pull in new dependencies without unattended consequences.

colemickens:
If you ran `glide install` and it rewrote your lock file, then we should file a bug on `glide`, because it explicitly says that should not happen (and I've never personally had it happen). There's no chance you ran `glide up`?

jchauncey:
I ran a `glide get` that rewrote the lock file

colemickens:
Ugh. I wonder if it's similar to this bug: https://github.com/Masterminds/glide/issues/252. When the `get` runs it maybe triggers the same `up` logic. I still think we should file a bug on `glide` for that weird behavior, but I understand why the PR is needed. I'd be curious to know, if you had put the restrictive SHA into the glide.yaml file and *then* ran `glide get <whatever>` does it respect the SHA specified in `glide.yaml`? I guess it must...

Anyway, sorry for all the questions and thanks for helping me understand.

LGTM to me.

jchauncey:
yes if version restrictions are in place it will honor them to the best of its ability. The more promiscuous version constraints for the other deps may mean that they are updated as well though.

dmitsh:
@acs-bot test this please

seanknox:
Thanks @jchauncey, can you rebase and we'll merge this in ASAP?

seanknox:
@jchauncey Will merge right after CI passes. Ping me if I forget!

seanknox:
(or go ahead and merge yourself, PR is approved)

jchauncey:
I dont have access to merge in this repo

dmitsh:
@acs-bot test this please

PR #744, opened 2017-06-08
modifications to dcos test

shrutir25:
**What this PR does / why we need it**: Fix for errors seen with adding to and validating marathon app.
- adds a check for /web folder before trying to add 
- displays app list in case of validation failure
- blocks teardown of app in case of failure

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/744)


test/cluster-tests/dcos/test.sh
```diff
@@ -38,22 +38,26 @@ ${remote_cp} "${DIR}/marathon.json" azureuser@${INSTANCE_NAME}.${LOCATION}.cloud
...
 done
-if [[ "$retval" != "0" ]]; then
+if [[ $retval -ne 0 ]]; then
   log "gave up waiting for marathon to be added"
   exit -1
 fi
 
 # only need to teardown if app added successfully
-trap teardown EXIT
+#trap teardown EXIT
```
dmitsh: uncomment

test/cluster-tests/dcos/test.sh
```diff
@@ -64,6 +68,8 @@ while [[ ${count} -lt 25 ]]; do
   if [[ "${running}" == "3" ]]; then
     log "Found 3 running tasks"
     break
+  else
+    ${remote_exec} ./dcos marathon app list
```
dmitsh: why do we need this?

shrutir25: This has been put to get a status of app in case of failure. So right now when validation fails, there is no way to check what the current app status looks like

shrutir25: essentially to make it easier to troubleshoot

dmitsh: you will be printing it every cycle?
We already have this here:

if [[ "${running}" != "3" ]]; then
  log "marathon validation failed"
  ${remote_exec} ./dcos marathon app show /web
  exit 1
fi

dmitsh:
@acs-bot test this please

PR #787, opened 2017-06-14
DCOS public agent pools for non-vlabs versions

jackfrancis:

For DCOS cluster templates using API versions `2017-01-31`, `2016-09-30`, and `2016-03-30`, this PR enforces a "public" agent pool listening on TCP 80, 443, and 8080, and converts the single agent pool defined in the cluster template (these API versions require a single agent pool for DCOS) to a "private" agent pool with no publicly accessible DNS.

Unit tests added for all novel functions introduced with this PR.

**Release note**:

`NONE`

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/787)


JackQuincy:
This logic is in the wrong spot. We would want to add the public agent pool as we convert from v20170131 (or the others) to unversioned(api). Then when we convert back we would want to remove it.

pkg/api/converterfromapi.go
```diff
@@ -578,3 +613,30 @@ func convertCertificateProfileToVLabs(api *CertificateProfile, vlabs *vlabs.Cert
...
+	// 1 master => 1 VM
+	// 3, 5 master => 3 VMsize
+	if m.Count == 1 {
+		publicPool.Count = 1
+	} else if m.Count == 3 {
+		publicPool.Count = 3
+	} else if m.Count == 5 {
+		publicPool.Count = 3
+	}
+	// TODO how do we set v20160930PublicPool.Count if p.MasterProfile.Count is neither 1, 3, nor 5?
```
JackQuincy: we have validation that only 1, 3, or 5 is allowed. With no plan of changing at this time.

jackfrancis: 👍

pkg/api/converterfromapi.go
```diff
@@ -578,3 +613,30 @@ func convertCertificateProfileToVLabs(api *CertificateProfile, vlabs *vlabs.Cert
 	vlabs.KubeConfigPrivateKey = api.KubeConfigPrivateKey
 	vlabs.SetCAPrivateKey(api.GetCAPrivateKey())
 }
+
+func addDCOSPublicAgentPool(privateAgentPool *AgentPoolProfile, p *Properties, m *MasterProfile) (*AgentPoolProfile, error) {
+	privateAgentPool.DNSPrefix = "" // - don't use DNS for private pool
```
JackQuincy: before we null this out we should copy it to the public agent pool

pkg/api/converterfromapi.go
```diff
@@ -221,6 +244,18 @@ func convertPropertiesToV20170131(api *Properties, p *v20170131.Properties) {
 		convertAgentPoolProfileToV20170131(apiProfile, v20170131Profile)
 		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20170131Profile)
 	}
+	if api.OrchestratorProfile.OrchestratorType == "DCOS" && len(p.AgentPoolProfiles) == 1 {
+		publicPool, _ := addDCOSPublicAgentPool(api.AgentPoolProfiles[0], api, api.MasterProfile)
+		v20170131PublicPool := &v20170131.AgentPoolProfile{}
+		// TODO verify that we want the public agent pool to be v20170131-specific
```
JackQuincy: I'm confused by these statements. Yes we want to have this only happen on the older versions. The newer versions allow multiple agent pools and the user can just set these themselves explicitly, so there shouldn't be a need for something like this

JackQuincy:
Current changes look right now, we just need to have the conversion from api=>versioned undo this splitting up.

pkg/api/converterfromapi.go
```diff
@@ -182,10 +190,18 @@ func convertPropertiesToV20160330(api *Properties, p *v20160330.Properties) {
...
 	p.AgentPoolProfiles = []*v20160330.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20160330Profile := &v20160330.AgentPoolProfile{}
-		convertAgentPoolProfileToV20160330(apiProfile, v20160330Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160330Profile)
+	// We expect exactly 2 agent pools: index [0] is the private pool, index [1] is the public pool
+	for i, apiProfile := range api.AgentPoolProfiles {
+		// a V20160330 should only have one (private) agent pool
+		if i < 1 {
+			v20160330Profile := &v20160330.AgentPoolProfile{}
```
JackQuincy: I know this will work based on validation but is there a way to detect which is public/private agent pools outside of just going off of the index?  I'm thinking about a future where a cluster is created with version v20170701 then someone calls get with v20170131 afterwards. Will this behave well

jackfrancis: Is there an identifying property inside the agent pool itself that we can use to determine this (i.e., `"type": "private"`)? I'm not sure there is. (unless we use this PR to introduce such an identifying property)

JackQuincy: we would have to add it... I was thinking more of checking their names match what we set when splitting them/only one has a dns prefix/ validating there are only 2 agent pool profiles. It would be less secure then marking them somehow. mostly just thinking out loud at this point

pkg/api/convertertoapi.go
```diff
@@ -640,3 +649,25 @@ func convertVLabsCertificateProfile(vlabs *vlabs.CertificateProfile, api *Certif
 	api.KubeConfigPrivateKey = vlabs.KubeConfigPrivateKey
 	api.SetCAPrivateKey(vlabs.GetCAPrivateKey())
 }
+
+func addDCOSPublicAgentPool(api *Properties) {
+	publicPool := &AgentPoolProfile{}
+	publicPool.Name = api.AgentPoolProfiles[0].Name + "_public"
```
jackfrancis: @JackQuincy Here's something that we are doing when we add the public pool. When doing the conversion to version-specific, we could look out for the `_public` suffix as a blacklist pattern.

JackQuincy: That should work

pkg/api/converterfromapi.go
```diff
@@ -183,9 +193,18 @@ func convertPropertiesToV20160330(api *Properties, p *v20160330.Properties) {
...
+		matched, err := regexp.MatchString(publicAgentPoolSuffix + "$", apiProfile.Name)
+		if !matched && err == nil {
+			v20160330Profile := &v20160330.AgentPoolProfile{}
+			convertAgentPoolProfileToV20160330(apiProfile, v20160330Profile)
+			p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160330Profile)
+		}
+	}
+	// Assign DNS Prefix to private agent pool from public agent pool
+	if len(p.AgentPoolProfiles) > 0 {
+		p.AgentPoolProfiles[0].DNSPrefix = api.AgentPoolProfiles[1].DNSPrefix
```
JackQuincy: mostly looking good. Anyway we can use the above loop to inform what indexes to use here for this replacement?

pkg/api/converterfromapi.go
```diff
@@ -169,10 +170,23 @@ func convertPropertiesToV20160930(api *Properties, p *v20160930.Properties) {
...
+				convertAgentPoolProfileToV20160930(apiProfile, v20160930Profile)
+				p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160930Profile)
+				privIndex = i
+			} else {
+				pubIndex = i
+			}
+		}
+		// Assign DNS Prefix to private agent pool from public agent pool
+		p.AgentPoolProfiles[privIndex].DNSPrefix = api.AgentPoolProfiles[pubIndex].DNSPrefix
 	}
```
JackQuincy: add else that has the old behavior

pkg/api/converterfromapi.go
```diff
@@ -211,10 +225,23 @@ func convertPropertiesToV20160330(api *Properties, p *v20160330.Properties) {
 		convertMasterProfileToV20160330(api.MasterProfile, p.MasterProfile)
 	}
 	p.AgentPoolProfiles = []*v20160330.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20160330Profile := &v20160330.AgentPoolProfile{}
-		convertAgentPoolProfileToV20160330(apiProfile, v20160330Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160330Profile)
+	if api.OrchestratorProfile.IsDCOS() && len(api.AgentPoolProfiles) > 1 {
```
JackQuincy: same

pkg/api/converterfromapi.go
```diff
@@ -245,10 +272,23 @@ func convertPropertiesToV20170131(api *Properties, p *v20170131.Properties) {
 		convertMasterProfileToV20170131(api.MasterProfile, p.MasterProfile)
 	}
 	p.AgentPoolProfiles = []*v20170131.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20170131Profile := &v20170131.AgentPoolProfile{}
-		convertAgentPoolProfileToV20170131(apiProfile, v20170131Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20170131Profile)
+	if api.OrchestratorProfile.IsDCOS() && len(api.AgentPoolProfiles) > 1 {
```
JackQuincy: same

pkg/api/converterfromapi.go
```diff
@@ -169,10 +170,30 @@ func convertPropertiesToV20160930(api *Properties, p *v20160930.Properties) {
 		convertMasterProfileToV20160930(api.MasterProfile, p.MasterProfile)
 	}
 	p.AgentPoolProfiles = []*v20160930.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20160930Profile := &v20160930.AgentPoolProfile{}
-		convertAgentPoolProfileToV20160930(apiProfile, v20160930Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160930Profile)
+	// DCOS conversion logic
+	if api.OrchestratorProfile.IsDCOS() && len(api.AgentPoolProfiles) == 2 {
```
jackfrancis: This condition is meant to be an exact inverse pair with this:

https://github.com/Azure/acs-engine/pull/787/files#diff-fab6595d1729597cff2aeb9912285552R206

pkg/api/converterfromapi.go
```diff
@@ -211,10 +232,30 @@ func convertPropertiesToV20160330(api *Properties, p *v20160330.Properties) {
 		convertMasterProfileToV20160330(api.MasterProfile, p.MasterProfile)
 	}
 	p.AgentPoolProfiles = []*v20160330.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20160330Profile := &v20160330.AgentPoolProfile{}
-		convertAgentPoolProfileToV20160330(apiProfile, v20160330Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20160330Profile)
+	// DCOS conversion logic
+	if api.OrchestratorProfile.IsDCOS() && len(api.AgentPoolProfiles) == 2 {
```
jackfrancis: This set of conditions is meant to be an exact inverse pair with this:

https://github.com/Azure/acs-engine/pull/787/files#diff-fab6595d1729597cff2aeb9912285552R244

pkg/api/converterfromapi.go
```diff
@@ -245,10 +286,30 @@ func convertPropertiesToV20170131(api *Properties, p *v20170131.Properties) {
 		convertMasterProfileToV20170131(api.MasterProfile, p.MasterProfile)
 	}
 	p.AgentPoolProfiles = []*v20170131.AgentPoolProfile{}
-	for _, apiProfile := range api.AgentPoolProfiles {
-		v20170131Profile := &v20170131.AgentPoolProfile{}
-		convertAgentPoolProfileToV20170131(apiProfile, v20170131Profile)
-		p.AgentPoolProfiles = append(p.AgentPoolProfiles, v20170131Profile)
+	// DCOS conversion logic
+	if api.OrchestratorProfile.IsDCOS() && len(api.AgentPoolProfiles) == 2 {
```
jackfrancis: This set of conditions is meant to be an exact inverse pair with this:

https://github.com/Azure/acs-engine/pull/787/files#diff-fab6595d1729597cff2aeb9912285552R296

pkg/acsengine/testdata/v20160330/dcos.json
```diff
@@ -12,13 +12,9 @@
       {
         "name": "agentprivate",
         "count": 3,
-        "vmSize": "Standard_D2_v2"
```
jackfrancis: Modified agent pool data to have only one pool, as per the normal spec for this version.

pkg/acsengine/testdata/v20160930/dcos.json
```diff
@@ -12,13 +12,9 @@
       {
         "name": "agentprivate",
         "count": 3,
-        "vmSize": "Standard_D2_v2"
```
jackfrancis: ditto

pkg/acsengine/testdata/v20170131/dcos.json
```diff
@@ -12,13 +12,9 @@
       {
         "name": "agentprivate",
         "count": 3,
-        "vmSize": "Standard_D2_v2"
```
jackfrancis: ditto

PR #794, opened 2017-06-16
add nil check for properties

rjtsdl:
Add one nil pointer check in the convertfromapi

Ready to review

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/794)


rjtsdl:
My idea is, if it is nil, then we return an empty versoined.Properties is the right behavior.

weinong:
I dont think we should take this fix. 1) acs-engine should never has nil properties. 2) if for some reason it unexpectedly is nil, we should not silently let it through.

PR #802, opened 2017-06-19
Fix deployment command for azure-cli in README

wbuchwalter:
With `azure-cli` `2.0.7` and above, when doing `az group deployment create` the `--parameters` argument takes a path to a json file instead of the content of the file, so the `@` prefix should be removed.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/802)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ $ az group deployment create \
     --name "<DEPLOYMENT NAME>" \
     --resource-group "<RESOURCE_GROUP_NAME>" \
     --template-file "./_output/<INSTANCE>/azuredeploy.json" \
-    --parameters "@./_output/<INSTANCE>/azuredeploy.parameters.json"
+    --parameters "./_output/<INSTANCE>/azuredeploy.parameters.json"
 ```
 
 ### Deploying with Powershell


rjtsdl:
I think instead of 
```
az acs group deployment create
```

You mean
```
az group deployment create
```

README.md
```diff
@@ -102,7 +102,7 @@ $ az group deployment create \
     --name "<DEPLOYMENT NAME>" \
     --resource-group "<RESOURCE_GROUP_NAME>" \
     --template-file "./_output/<INSTANCE>/azuredeploy.json" \
-    --parameters "@./_output/<INSTANCE>/azuredeploy.parameters.json"
```
JackQuincy: was there a change in the CLI? before you had to prefix with an @ to get the cli to look for a file instead of the raw values

wbuchwalter:
@rjtsdl Indeed, thanks. I edited my first comment.

wbuchwalter:
See https://github.com/Azure/azure-cli/commit/3753b503aa8f9d21a28ca8fbeccd5d1623eafdda#diff-f38a6221a66bc6c47742ed49e98f2439R289

If you deploy using `@` with >=`2.0.7`, azure-cli will basically just ignore the parameters file and ask the user to enter each parameter value manually.

PR #808, opened 2017-06-19
WIP accept language header

JackQuincy:
This adds accept language  header on calls to ARM
Allows us to get localized error messages from the services we call
This is how their team suggested we add these headers here https://github.com/Azure/azure-sdk-for-go/issues/658 Looking at it. It is a little cumbersome and makes me think we should push harder to get native support in the actual sdk. I'll point them here
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/808)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -60,6 +60,7 @@ type authArgs struct {
 	ClientSecret    string
 	CertificatePath string
 	PrivateKeyPath  string
+	language        string
 }
 
 func addAuthFlags(authArgs *authArgs, f *flag.FlagSet) {
@@ -70,6 +71,7 @@ func addAuthFlags(authArgs *authArgs, f *flag.FlagSet) {
 	f.StringVar(&authArgs.ClientSecret, "client-secret", "", "client secret (used with --auth-mode=client_secret)")
 	f.StringVar(&authArgs.CertificatePath, "certificate-path", "", "path to client certificate (used with --auth-method=client_certificate)")
 	f.StringVar(&authArgs.PrivateKeyPath, "private-key-path", "", "path to private key (used with --auth-method=client_certificate)")
+	f.StringVar(&authArgs.language, "language", "en-us", "language to return error messages in")
 }
 
 func (authArgs *authArgs) getClient() (*armhelpers.AzureClient, error) {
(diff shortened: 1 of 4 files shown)


pkg/armhelpers/compute.go
```diff
@@ -2,24 +2,116 @@ package armhelpers
...
 import (
 	"github.com/Azure/azure-sdk-for-go/arm/compute"
+	"github.com/Azure/go-autorest/autorest"
 )
 
 // ListVirtualMachines returns (the first page of) the machines in the specified resource group.
-func (az *AzureClient) ListVirtualMachines(resourceGroup string) (compute.VirtualMachineListResult, error) {
-	return az.virtualMachinesClient.List(resourceGroup)
+func (az *AzureClient) ListVirtualMachines(resourceGroup string) (result compute.VirtualMachineListResult, err error) {
+	client := az.virtualMachinesClient
```
anhowe: Put a comment on each of these calls, to say replace with the one liner, once they added the accept language header.

JackQuincy: This PR was more to serve as a case in point that we needed something better. There is discussion in the issue on the SDK and they provided a better way we can use to do this with much less code.

weinong:
Reviewed 3 of 9 files at r1, 1 of 6 files at r2.

---

*[cmd/root.go, line 63 at r2](https://reviewable.io:443/reviews/azure/acs-engine/808#-Kn1gKhlhcjCSpEKyvkw:-Kn1gKhlhcjCSpEKyvkx:b-h6cg2w) ([raw file](https://github.com/azure/acs-engine/blob/2074a96a564779e3625182205e9172bde8dd17b5/cmd/root.go#L63)):*
> ```Go
> 	CertificatePath string
> 	PrivateKeyPath  string
> 	language        string
> ```

why is it private?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/808)*

JackQuincy:
*[cmd/root.go, line 63 at r2](https://reviewable.io:443/reviews/azure/acs-engine/808#-Kn1gKhlhcjCSpEKyvkw:-Kn1iXdBx6h1tE6TLDPk:bhkv38t) ([raw file](https://github.com/azure/acs-engine/blob/2074a96a564779e3625182205e9172bde8dd17b5/cmd/root.go#L63)):*
<details><summary><i>Previously, weinong (Weinong Wang) wrote…</i></summary><blockquote>

why is it private?
</blockquote></details>

mostly because I didn't need it anywhere outside of this file. I could export it if you feel strongly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/808)*

weinong:
Reviewed 1 of 6 files at r2.

---

*[cmd/root.go, line 63 at r2](https://reviewable.io:443/reviews/azure/acs-engine/808#-Kn1gKhlhcjCSpEKyvkw:-Kn1jN-0CWHbdk9W7g68:bflv7fj) ([raw file](https://github.com/azure/acs-engine/blob/2074a96a564779e3625182205e9172bde8dd17b5/cmd/root.go#L63)):*
<details><summary><i>Previously, JackQuincy (Jack) wrote…</i></summary><blockquote>

mostly because I didn't need it anywhere outside of this file. I could export it if you feel strongly
</blockquote></details>

i dont mind either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/808)*

PR #819, opened 2017-06-20
Partition regression to avoid storage account issues

anhowe:
**What this PR does / why we need it**: Partition regression to avoid storage account issues

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/819)


anhowe:
@acs-bot test this please

Issue #823, opened 2017-06-21
acs-engine ci is broken


JackQuincy:
**Is this a request for help?**: No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
Issue

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
all

**What happened**:
az cli rev'ed and the current version is broken.
**What you expected to happen**:
ci to be able to deploy a template

**How to reproduce it** (as minimally and precisely as possible):
run @acs-bot tests

**Anything else we need to know**:
We are currently working on getting a working version of the cli back so that we can restore to normal operations

anhowe:
This is now fixed by #826

PR #826, opened 2017-06-21
Fix Azure CLI in docker by installing with the bundled installer

colemickens:

This fixes the recent breakage of `az` under the acs-engine devenv docker container. This uses the reproducible bundled installer rather than relying on pip and transitive unpinned dependencies.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: n/a  

**Special notes for your reviewer**: none 

**Release note**:

```release-note
* more correctly pin version of azure-cli in devenv docker container
```

Note, this installs the `0.2.9` bundle because the `0.2.10` bundle doesn't install correctly (this is [bug #3800 on azure-cli](https://github.com/Azure/azure-cli/issues/3800))

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/826)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
-acs-engine
-acs-engine.exe
-_output/
-.git/
+./acs-engine
+./acs-engine.exe
+./_output
+./.git
+./test/acs-engine-test/acs-engine-test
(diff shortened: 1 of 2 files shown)


anhowe:
this fixes #823

PR #846, opened 2017-06-23
Add schema and content header to parameters

### LINKED ISSUE #723: Generated azuredeploy.parameters.json missing properties

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes 1.6.2

**What happened**:
Once `azuredeploy.json` and `azuredeploy.parameters.json` are generated, the parameters file starts with parameters block without heading fields (like `azuredeploy.json`):
```
{
  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
```
As result, calling for azure sdk (using azure-sdk-for-python) fails. When the above parameters are inserted into `azuredeploy.parameters.json` the call to sdk works.

The generated files should be used "as is" without additional altering.

We use `ansible` for creating a cluster. It is possible to look what SDK API the [azure deployment](https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/azure/azure_rm_deployment.py) module of the ansible is used to call directly and pass URLs to the generated files.

Azure SDK for python version: azure (2.0.0rc5)

anhowe:
**What this PR does / why we need it**: Add schema and content header to parameters

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #723

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/846)

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -348,6 +348,7 @@ func GetCloudTargetEnv(location string) string {
 func getParameters(cs *api.ContainerService, isClassicMode bool) (map[string]interface{}, error) {
 	properties := cs.Properties
 	location := cs.Location
+	parametersAll := map[string]interface{}{}
 	parametersMap := map[string]interface{}{}
 
 	// Master Parameters
@@ -453,7 +454,12 @@ func getParameters(cs *api.ContainerService, isClassicMode bool) (map[string]int
 		}
 	}
 
-	return parametersMap, nil
+	// add the header
+	parametersAll["$schema"] = "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#"
+	parametersAll["contentVersion"] = "1.0.0.0"
+	parametersAll["parameters"] = parametersMap
+
+	return parametersAll, nil
 }
 
 func addValue(m map[string]interface{}, k string, v interface{}) {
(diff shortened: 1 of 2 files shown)


pkg/acsengine/json.go
```diff
@@ -28,6 +28,22 @@ func PrettyPrintArmTemplate(template string) (string, error) {
...
 }
 
+// PrettyPrintArmParameters will pretty print the arm template ensuring the header is $schema and contentVersion are at top.
+func PrettyPrintArmParameters(template string) (string, error) {
+	translateParams := [][]string{
+		{"\"contentVersion\"", "\"$zcontentVersion\""},
+	}
+
+	template = translateJSON(template, translateParams, false)
+	var err error
```
jackfrancis: Because you're not re-using the error reference `err` anywhere other than the if block below, you don't have to declare it here; `if template, err := PrettyPrintJSON...` is probably more idiomatic.

anhowe: I was able to remove the whole function

anhowe:
@acs-bot test this please

PR #861, opened 2017-06-24
Delay wrapping of parameters

anhowe:
**What this PR does / why we need it**: The parameters were wrapped too early with the schema and content version causing problems for the deploy and upgrade command.  This delays the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/861)

### DIFF
--- a/pkg/acsengine/json.go
+++ b/pkg/acsengine/json.go
@@ -41,6 +41,24 @@ func PrettyPrintJSON(content string) (string, error) {
 	return string(prettyprint), nil
 }
 
+// BuildAzureParametersFile will add the correct schema and contentversion information
+func BuildAzureParametersFile(content string) (string, error) {
+	var parametersMap map[string]interface{}
+	if err := json.Unmarshal([]byte(content), &parametersMap); err != nil {
+		return "", err
+	}
+	parametersAll := map[string]interface{}{}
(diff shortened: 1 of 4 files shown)


pkg/acsengine/json.go
```diff
@@ -41,6 +41,24 @@ func PrettyPrintJSON(content string) (string, error) {
...
+// BuildAzureParametersFile will add the correct schema and contentversion information
+func BuildAzureParametersFile(content string) (string, error) {
+	var parametersMap map[string]interface{}
+	if err := json.Unmarshal([]byte(content), &parametersMap); err != nil {
+		return "", err
+	}
+	parametersAll := map[string]interface{}{}
+	parametersAll["$schema"] = "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#"
+	parametersAll["contentVersion"] = "1.0.0.0"
+	parametersAll["parameters"] = parametersMap
```
rjtsdl: so, this be something like parameters : { parameters :{ ...}}

Are we sure, this is what we want?

JackQuincy: there are multiple formats that things expect the parameters in. this one is more correct but the go sdk for azure doesn't properly handle it

JackQuincy: also in the file it looks like 

> { "$schema": "...",...,"parameters":{...}}

rjtsdl: thx

PR #874, opened 2017-06-26
pre-test init script framework

dmitsh:
- added pre-test init script framework
- added pre-test init scripts for vnet tests (k8s, dcos, swarmmode)

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/874)


examples/vnet/kubernetesvnet.json.pretest.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+az network vnet create -g ${RESOURCE_GROUP} -n KubernetesCustomVNET --address-prefix 10.239.0.0/16 --subnet-name KubernetesSubnet --subnet-prefix 10.239.0.0/16
```
colemickens: Does the script need to be duplicated? Seems like it's the same exact the name of the orchestrator. Can it be removed or parameterized?

This is a nit, if you think this copy/paste is okay, then I'm okay with it.

dmitsh: The scripts for DCOS and SwarmMode are identical, but for k8s is slightly different.

The reason for duplication is to make the pre-test generic. If the "api-model".json.pretest.sh exists, we will execute it.

I'm open to other suggestions.

PR #883, opened 2017-06-27
Allocate static IP addresses from the last /24 of cluster subnet

### LINKED ISSUE #882: Kubernetes deployments fail when azure-vnet plugin is enabled and clusterSubnet is set too small
No.

Issue.

Kubernetes

**What happened**:
When VNET integration is enabled (i.e. pods are assigned private IP addresses from VNET address space), static (for masters) and dynamic (for agents and pods) addresses are being allocated from the same subnet. This results in race conditions where a dynamic address allocation can execute before a static address allocation for the same IP address. In that case, the static allocation will fail and result in deployment failures.

This is more likely to reproduce for clusters with small subnets (prefix length >16) and lots of agents.

Deployments should succeed regardless of subnet size or agent count.

Set "kubernetesConfig.clusterSubnet": 10.240.0.0/17 and networkPolicy: "azure".
The issue has been noticed in #546 and root cause found by @dmccaffery.

Fix is coming in a few minutes.

ofiliz:

When VNET integration is enabled (pods are assigned private IP addresses from VNET address space), static (for masters) and dynamic (for agents and pods) addresses are being allocated from the same subnet. This results in race conditions where a dynamic address allocation can execute before a static address allocation for the same IP address. In that case, the static allocation will fail and result in deployment failures.

This PR fixes 
* When the cluster subnet prefix is /17 or longer, getFirstConsecutiveStaticIPAddress returns IP addresses from the same /24 as the pods. Update that function to return from the very last /24 subnet, clearly separating static addresses from dynamic addresses.
* Update validation code to fail if the cluster subnet is too small (less than 9 bits for host IDs). The last 8 bits are used for dynamic addresses for all nodes and pods. The 9th bit is set for static addresses.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:
fixes #882

**Release note**:
NONE

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/883)

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -261,16 +261,17 @@ func getFirstConsecutiveStaticIPAddress(subnetStr string) string {
 		return DefaultFirstConsecutiveKubernetesStaticIP
 	}
 
-	// Round up the prefix length to the nearest octet boundary.
+	// Find the first and last octet of the host bits.
 	ones, bits := subnet.Mask.Size()
-	if ones%8 != 0 {
-		ones += 8 - ones%8
-	}
+	firstOctet := ones / 8
+	lastOctet := bits/8 - 1
+
+	// Set the remaining host bits in the first octet.
+	subnet.IP[firstOctet] |= (1 << byte((8 - (ones % 8)))) - 1
 
 	// Fill the intermediate octets with 1s and last octet with offset. This is done so to match
 	// the existing behavior of allocating static IP addresses from the last /24 of the subnet.
-	lastOctet := bits/8 - 1
-	for i := ones / 8; i < lastOctet; i++ {
+	for i := firstOctet + 1; i < lastOctet; i++ {
 		subnet.IP[i] = 255
 	}
 	subnet.IP[lastOctet] = DefaultKubernetesFirstConsecutiveStaticIPOffset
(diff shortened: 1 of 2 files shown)


JackQuincy:
@acs-bot test this please

pkg/acsengine/defaults.go
```diff
@@ -261,16 +261,17 @@ func getFirstConsecutiveStaticIPAddress(subnetStr string) string {
...
+	lastOctet := bits/8 - 1
+
+	// Set the remaining host bits in the first octet.
+	subnet.IP[firstOctet] |= (1 << byte((8 - (ones % 8)))) - 1
 
 	// Fill the intermediate octets with 1s and last octet with offset. This is done so to match
 	// the existing behavior of allocating static IP addresses from the last /24 of the subnet.
-	lastOctet := bits/8 - 1
-	for i := ones / 8; i < lastOctet; i++ {
+	for i := firstOctet + 1; i < lastOctet; i++ {
```
JackQuincy: it seems like the +1  here is what is keeping us from getting the conflicts we were having. Is that correct?

ofiliz: It is actually line 270. Previously, we skipped the remaining host bits in the first octet. Line 270 sets them.
For example, given input 10.240.0.0/12, first host bit is 13, so firstOctet points to 240.
Before this change: 10.240.255.5 After this change: 10.255.255.5

dmitsh:
@acs-bot test this please

PR #886, opened 2017-06-28
add validate tag for the purpose of required field

### LINKED ISSUE #884: [Proposal]:2017-07-01 version should have tag to know if the field is optional
No
---

FEATURE REQUEST

Currently, from the type definition, we cannot deduct which field is required, which is optional.
By adding a new field tag `validate`, it could have `required`, `optional`, etc. We could refer to this tag field in all the validate functions. 
The motivation is, it is clear to user, which field is required.

**Update**
After more exploration, I noticed many third party validator package available for the same purpose but much richer. I have picked [this](https://github.com/go-playground/validator).

rjtsdl:
Currently in the versioned api definition, no way to know whether a field is required from user's input. The omitempty is mainly used for marshal. We need another tag for validation. By doing this, we could give user(the developer) a clear view of our api model definition without checking lines of validation code. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #884 

I propose to add the tag. Check [this](https://play.golang.org/p/yqd1jQxZ6p)
I picked this validator package https://github.com/go-playground/validator. 

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/886)


pkg/api/v20170701/types.go
```diff
@@ -18,7 +32,7 @@ type ResourcePurchasePlan struct {
 // resource definition in a JSON template.
 type ContainerService struct {
 	ID       string                `json:"id,omitempty"`
```
rjtsdl: Please also help check if i missed any required field.

rjtsdl:
@acs-bot test this please

JackQuincy:
Any plans to enforce this in validation?

weinong:
is it an experiment on 2017-07-01?

JiangtianLi:
what go validator package do you plan to use? Is there anything more than "required/optional" we could add to tag?

rjtsdl:
@weinong , could be an experiment. 
@JackQuincy , yes, that's one of the next things i want to do. I mentioned it in the description. But i think it could be a different pr, if we think the `validate` tag makes sense.
@JiangtianLi , current i don't have more. It is all defined and handled by ourselves. Let me know if you have any idea.

pkg/api/v20170701/types.go
```diff
@@ -50,31 +64,31 @@ type Properties struct {
...
-	Secret   string `json:"secret,omitempty"`
+	ClientID string `json:"clientId,omitempty" validate:"required"`
+	Secret   string `json:"secret,omitempty" validate:"required"`
 }
 
 // CustomProfile specifies custom properties that are used for
 // cluster instantiation.  Should not be used by most users.
 type CustomProfile struct {
-	Orchestrator string `json:"orchestrator,omitempty"`
+	Orchestrator string `json:"orchestrator,omitempty" validate:"required"`
```
amanohar: Is this a required field?

pkg/api/v20170701/types.go
```diff
@@ -98,15 +112,15 @@ const (
...
+	OrchestratorType    OrchestratorType    `json:"orchestratorType" validate:"required"`
 	OrchestratorVersion OrchestratorVersion `json:"orchestratorVersion"`
 }
 
 // MasterProfile represents the definition of master cluster
 type MasterProfile struct {
-	Count                    int    `json:"count"`
-	DNSPrefix                string `json:"dnsPrefix"`
-	VMSize                   string `json:"vmSize"`
+	Count                    int    `json:"count" validate:"required"`
```
amanohar: Doesn't master count has a default value? Same goes for VMSize,

pkg/api/v20170701/types.go
```diff
@@ -5,6 +5,20 @@ import (
 	"strings"
 )
 
+// Add the validate tag to add a mark, if this is a required or optional field
+// Current it is not used in any logic yet
+// TODO, we could possibly use this tag in all the Validate functions
+const (
+	ValidateTag = "validate"
```
amanohar: Instead of calling it "validate" , why not "required" or "isrequired"?
E.g.:
Location string     `json:"location,omitempty" isrequired:"true" 

or

Location string     `json:"location,omitempty" required:"true"

rjtsdl:
I have looked a few popular validator package. 
I have listed them in the order of popularity. Any preference here?
@amanohar , after we choose the package, i will customize the tag in their way.
@JiangtianLi , which one you have tried? And yes, there is many more features, other than required and optional.

https://github.com/asaskevich/govalidator
https://github.com/go-playground/validator
https://github.com/go-validator/validator

pkg/api/v20170701/types.go
rjtsdl: I will use whatever the validator package specified. My other comment has the options. Pick one that you favor more. They have different syntax.

pkg/api/v20170701/types.go
rjtsdl: I mark it as required, in the sense, CustomProfile is there.
However CustomProfile itself is not required.

pkg/api/v20170701/types.go
rjtsdl: Master Count has a default value in the implementation of `az cli`. 
You can think of `az cli` is the helper for user to put together the api model. This is the validation tag for this api model.

It works, after `az cli` generate the api model. We validate if it provide the required fields. 
Keep in mind, this validate tag only works on validate the generated api model.

JiangtianLi:
@rjtsdl I haven't tried but just looked at its README. I think if the package can meet our container service validation,  such as range, set, or even regex, that should be fine.

rjtsdl:
I will pick this one. https://github.com/go-playground/validator 
I will update this PR after apply this third party package

rjtsdl:
@JackQuincy  @amanohar  @JiangtianLi , i have imported and applied https://github.com/go-playground/validator.v9. I don't expect any behavior change. I only did minimal change in v20170701/validate.go file. It proves the concept that, by introducing the package, things get clear and easier to do validation. We should be able to leverage this package further, when we get more and more familiar with it.

pkg/api/v20170701/types.go
```diff
@@ -98,19 +101,19 @@ const (
...
 	OrchestratorVersion OrchestratorVersion `json:"orchestratorVersion"`
 }
 
 // MasterProfile represents the definition of master cluster
 type MasterProfile struct {
-	Count                    int    `json:"count"`
-	DNSPrefix                string `json:"dnsPrefix"`
-	VMSize                   string `json:"vmSize"`
-	OSDiskSizeGB             int    `json:"osDiskSizeGB,omitempty"`
+	Count                    int    `json:"count" validate:"required,eq=1|eq=3|eq=5"`
```
rjtsdl: If we specify `count` with tags like this. It will be automatically validated. No need to write functions to handle it. TAL this file, and compare with what i have removed in the validate.go file. You will get the idea here.

seanknox:
@rjtsdl From the end-user perspective, what changes?

pkg/api/v20170701/validate.go
```diff
@@ -201,20 +171,6 @@ func (a *Properties) Validate() error {
...
-func validateName(name string, label string) error {
-	if name == "" {
-		return fmt.Errorf("%s must be a non-empty value", label)
-	}
-	return nil
-}
-
 func validatePoolName(poolName string) error {
 	// we will cap at length of 12 and all lowercase letters since this makes up the VMName
 	poolNameRegex := `^([a-z][a-z0-9]{0,11})$`
```
rjtsdl: We can take a look at regex later, when this package get more mature. This is the current state for regex.

```
regex
	a regex validator won't be added because commas and = signs can be part
	of a regex which conflict with the validation definitions. Although
	workarounds can be made, they take away from using pure regex's.
	Furthermore it's quick and dirty but the regex's become harder to
	maintain and are not reusable, so it's as much a programming philosiphy
	as anything.

        In place of this new validator functions should be created; a regex can
	be used within the validator function and even be precompiled for better
	efficiency within regexes.go.

	And the best reason, you can submit a pull request and we can keep on
	adding to the validation library of this package!
```

rjtsdl:
@seanknox , Should be no behavior change for end user. 
For developers, it would be clear API model declaration, and less customized validate functions.

pkg/api/v20170701/types.go
weinong: i'm curious about the error message. any example?

rjtsdl: Yes, when i specify count=2

```
FATA[0000] error parsing the api model: Key: 'MasterProfile.Count' Error:Field validation for 'Count' failed on the 'eq=
1|eq=3|eq=5' tag
```

pkg/api/v20170701/types.go
rjtsdl: Take a closer look, CustomProfile doesn't have any validate function. I don't want to add required here anymore.

pkg/api/v20170701/types.go
JackQuincy: Was that a fatal error log or a error message returned? if it is a fatal error log we can't use this framework

rjtsdl: It is from the cmd below. The package itself doesn't panic or fatal.

```
./cmd/deploy.go:                log.Fatalf("error parsing the api model: %s", err.Error())
./cmd/generate.go:              log.Fatalf("error parsing the api model: %s", err.Error())
./cmd/upgrade.go:               log.Fatalf("error parsing the api model: %s", err.Error())
```


Is something from our engine.go.

pkg/api/v20170701/types.go
```diff
@@ -124,14 +127,14 @@ type MasterProfile struct {
...
+	Name           string `json:"name" validate:"required"`
+	Count          int    `json:"count" validate:"required,min=1,max=100"`
+	VMSize         string `json:"vmSize" validate:"required"`
+	OSDiskSizeGB   int    `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1023"`
 	DNSPrefix      string `json:"dnsPrefix"`
 	FQDN           string `json:"fqdn"`
 	Ports          []int  `json:"ports,omitempty"`
-	StorageProfile string `json:"storageProfile"`
+	StorageProfile string `json:"storageProfile" validate:"StorageAccount|ManagedDisks|len=0`
 	VnetSubnetID   string `json:"vnetSubnetID,omitempty"`
```
amanohar: Add 'eq' condition to the tag

rjtsdl: Fixed, thx

pkg/api/v20170701/types.go
weinong: the error msg at current form is not possible to be localized. So i'm strong against doing this.

JackQuincy: The returned error points out the struct/field/tag that failed so we could take that info than use it to generate a localized error message. But it would require us to make a bunch of if statements or a big switch  case that works like an if else chain.

rjtsdl: yes, that is exactly what i am doing. @JackQuincy , i should look at your comment earlier :(

rjtsdl: @weinong , localization is handled now. It is not the blocker anymore. PTAL

pkg/api/v20170701/validate.go
```diff
@@ -106,36 +89,59 @@ func (a *AgentPoolProfile) Validate(orchestratorType OrchestratorType) error {
...
-		return e
+	// Don't need to call validate.Struct(l)
+	// It is handled by Properties.Validate()
+	if e := validate.Var(l.SSH.PublicKeys[0].KeyData, "required"); e != nil {
+		return fmt.Errorf("KeyData in LinuxProfile.SSH.PublicKeys cannot be empty string")
 	}
 	return nil
 }
 
+func handleValidationErrors(e validator.ValidationErrors) error {
```
rjtsdl: @JackQuincy @JiangtianLi @weinong 
Now, this function will be called to handle the `validator.ValidationErrors`, it is called only in one place.
I noticed, I have a few hard coded strings here. However, it will be no difference with before, when we call
```
validateName(l.AdminUsername, "LinuxProfile.AdminUsername")
```
We also need to pass a hard coded string as a parameter.

This function is better in the sense, it has all hard-code strings in one place.

pkg/api/v20170701/types.go
rjtsdl: The error message for the same MasterProfile.Count=2, would be like
```
FATA[0000] error parsing the api model: MasterProfile count needs to be 1, 3, or 5
```

amanohar:
Outside of the scope of this PR but discussed with @rjtsdl that some of the fields like count, vmSize that were not required before in the ACS service are not required. This impacts the model of asking the users for minimal set of inputs to stand up a cluster (an issue was opened to track discussion around this).

rjtsdl:
@acs-bot test this please

weinong:
Reviewed 1654 of 1656 files at r2, 1 of 2 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[pkg/api/v20170701/validate.go, line 105 at r5](https://reviewable.io:443/reviews/azure/acs-engine/886#-KoJaZgrJK1nR3O4Sx-w-r5-105:-KoTH6l_jdLLbmfQBXd-:b-yai5ye) ([raw file](https://github.com/azure/acs-engine/blob/2d114e73b18b53e073f318f8827051923a4458e3/pkg/api/v20170701/validate.go#L105)):*
<details><summary><i>Previously, rjtsdl (Jingtao Ren) wrote…</i></summary><blockquote>

@JackQuincy @JiangtianLi @weinong 
Now, this function will be called to handle the `validator.ValidationErrors`, it is called only in one place.
I noticed, I have a few hard coded strings here. However, it will be no difference with before, when we call
```
validateName(l.AdminUsername, "LinuxProfile.AdminUsername")
```
We also need to pass a hard coded string as a parameter.

This function is better in the sense, it has all hard-code strings in one place. 
</blockquote></details>

this is pretty cool

---

*[pkg/api/v20170701/validate.go, line 107 at r6](https://reviewable.io:443/reviews/azure/acs-engine/886#-KoTHhFfOdMHuY2HYq7D:-KoTHhFfOdMHuY2HYq7E:b-3utf5q) ([raw file](https://github.com/azure/acs-engine/blob/19d47605e4d4988f19b7b7e61200365937c3c6e6/pkg/api/v20170701/validate.go#L107)):*
> ```Go
> func handleValidationErrors(e validator.ValidationErrors) error {
> 	err := e[0]
> 	switch err.Namespace() {
> ```

err.Namespace() can be called once and re-used throughout this function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/886)*

rjtsdl:
@weinong , address err.Namespace() suggestion. Thx. 

Btw, don't know how to reply a comment from Reviewable :(

weinong:
don't overwrite the commits please.

rjtsdl:
Ah, sure, will add new commits.

rjtsdl:
checked with everybody offline. I will merge this.

PR #887, opened 2017-06-28
update regression test due to SEV2

anhowe:
**What this PR does / why we need it**: westcentralus has lsi with disks, updating regression test.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/887)


rjtsdl:
@acs-bot test this please

PR #892, opened 2017-06-29
Enable cloudprovider rate limit / backoff features

### LINKED ISSUE #922: add cloudprovider backoff configuration

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes v1.6.6

**Feature Description**

Kubernetes 1.6.6 release brought exponential backoff and rate limiting to azure cloudprovider API calls. We want to power up acs-engine to be able to generate templates that enable this cluster behavior.

jackfrancis:

Delivers kubernetes configuration related to managing large cluster performance. Specifically, the following configuration data is now exposed as parameters in the acs-engine template generation output (e.g., `_output/<dnsPrefix>/azuredeploy.parameters.json`:

```
    "cloudProviderBackoff": {
      "value": "false" // if true, enable backoff
    },
    "cloudProviderBackoffDuration": {
      "value": "5" // how many seconds for initial backoff retry attempt
    },
    "cloudProviderBackoffExponent": {
      "value": "1.5" // exponent for successive backoff retries
    },
    "cloudProviderBackoffJitter": {
      "value": "1" // non-1 values add jitter to retry intervals
    },
    "cloudProviderBackoffRetries": {
      "value": "6" // maximum retry attempts before failure
    },
    "cloudProviderRatelimit": {
      "value": "false" // if true, enforce rate limits for azure API calls
    },
    "cloudProviderRatelimitBucket": {
      "value": "10" // number of requests in queue
    },
    "cloudProviderRatelimitQPS": {
      "value": "3" // rate limit QPS
    },
    "kubernetesCtrlMgrNodeMonitorGracePeriod": {
      "value": "5m" // duration after which controller manager marks an AWOL node as NotReady
    },
    "kubernetesCtrlMgrPodEvictionTimeout": {
      "value": "1m" // grace period for deleting pods on failed nodes
    },
    "kubernetesCtrlMgrRouteReconciliationPeriod": {
      "value": "1m" // how often to reconcile cloudprovider-originating node routes 
    },
    "kubernetesNodeStatusUpdateFrequency": {
      "value": "1m" // how often kubelet posts node status to master
    }
```

Note that the `cloudProviderBackoff<Setting>` and `cloudProviderRatelimit<Setting>` configuration parameters only take effect if `cloudProviderBackoff` and `cloudProviderRatelimit`, respectively, are set to `"true"`. For more context, here is the original `kubernetes/kubernetes` PR where these features were introduced:

https://github.com/kubernetes/kubernetes/pull/46660

Authoritative documentation for rate limiter implementation is here:

https://godoc.org/k8s.io/client-go/util/flowcontrol#RateLimiter

And authoritative backoff implementation docs here:

https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Backoff

~

Additionally, the `kubernetesCtrlMgr<Setting>` configuration parameters are documented here:

https://kubernetes.io/docs/admin/kube-controller-manager/

(See `--node-monitor-grace-period`, `--pod-eviction-timeout`, and `--route-reconciliation-period`.)

And `kubernetesNodeStatusUpdateFrequency` maps to the kubelet `--node-status-update-frequency` configuration option, documented here:

https://kubernetes.io/docs/admin/kubelet/

~

Finally, some guidance on changing the `kube-controller-manager` and `kubelet` knobs is provided here:

https://github.com/kubernetes-incubator/kubespray/blob/c07d60bc906f3db97d79fffa6bc49c4b3a24970c/docs/kubernetes-reliability.md

The above document outlines the importance of turning all the various knobs holistically: a configuration change for one behavior may have adverse side-effects if dependent configuration(s) is not appropriately modified in concert.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #922

**Release note**:

```enable azure cloudprovider backoff configuration
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/892)


parts/kuberneteskubelet.service
```diff
@@ -45,6 +45,7 @@ ExecStart=/usr/bin/docker run \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
         --hairpin-mode=promiscuous-bridge \
         --network-plugin=${KUBELET_NETWORK_PLUGIN} \
+        --node-status-update-frequency=1m \
```
jackfrancis: TODO: enable configurable `node-status-update-frequency`, with a default value of `10s` if unspecified

parts/kubernetesmaster-kube-controller-manager.yaml
```diff
@@ -24,6 +24,9 @@ spec:
         - "--service-account-private-key-file=/etc/kubernetes/certs/apiserver.key"
         - "--leader-elect=true"
         - "--v=2"
+        - "--node-monitor-grace-period=5m"
+        - "--pod-eviction-timeout=1m"
+        - "--route-reconciliation-period=10s"
```
jackfrancis: TODO: ditto above with these three properties as well, with appropriate defaults

parts/kubernetesmastercustomscript.sh
```diff
@@ -19,6 +19,16 @@ SERVICE_PRINCIPAL_CLIENT_SECRET="${11}"
...
+# Default values for backoff configuration
+CLOUDPROVIDER_BACKOFF=true
+CLOUDPROVIDER_BACKOFF_RETRIES=6
+CLOUDPROVIDER_BACKOFF_EXPONENT=1.5
+CLOUDPROVIDER_BACKOFF_DURATION=5 # in seconds
+CLOUDPROVIDER_BACKOFF_JITTER=1.0
+# Default values for rate limit configuration
+CLOUDPROVIDER_RATELIMIT=true
+CLOUDPROVIDER_RATELIMIT_QPS=3
+CLOUDPROVIDER_RATELIMIT_BUCKET=10
```
jackfrancis: TODO: convert the above statically assigned `ENV` values into configurable `ENV` values, i.e., suss out how the `${n}` assignments work.

test-activity.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
jackfrancis: TODO: we probably don't want to commit this in the acs-engine repo, but perhaps there's an appropriate place for L&P scripts like this one.

parts/kubernetesparams.t
```diff
@@ -110,6 +110,90 @@
...
+        "description": "Kubernetes controller manager route reconciliation period."
+      },
+      "type": "string"
+    },
+    "cloudProviderBackoff": {
+      {{PopulateClassicModeDefaultValue "cloudProviderBackoff"}}
+      "metadata": {
+        "description": "Enable cloudprovider backoff?"
+      },
+      "type": "boolean"
```
jackfrancis: TODO: what's the proper way to type this variable?

parts/kubernetesparams.t
```diff
@@ -110,6 +110,90 @@
...
+        "description": "If backoff enabled, how long until timeout"
+      },
+      "type": "int"
+    },
+    "cloudProviderBackoffJitter": {
+      {{PopulateClassicModeDefaultValue "cloudProviderBackoffJitter"}}
+      "metadata": {
+        "description": "If backoff enabled, jitter factor between retries"
+      },
+      "type": "float"
```
jackfrancis: ditto type question

parts/kubernetesparams.t
```diff
@@ -110,6 +110,90 @@
...
+    },
+    "cloudProviderRatelimit": {
+      {{PopulateClassicModeDefaultValue "cloudProviderRatelimit"}}
+      "metadata": {
+        "description": "Enable cloudprovider rate limiting?"
+      },
+      "type": "boolean"
```
jackfrancis: ditto type question

parts/kubernetesparams.t
```diff
@@ -110,6 +110,90 @@
...
+    },
+    "cloudProviderRatelimitQPS": {
+      {{PopulateClassicModeDefaultValue "cloudProviderRatelimitQPS"}}
+      "metadata": {
+        "description": "If rate limiting enabled, target maximum QPS"
+      },
+      "type": "float"
```
jackfrancis: ditto type question

parts/kubernetesparams.t
JackQuincy: available types are 
string
secureString
int
bool
object 
secureObject
array,
So pick the apporiate one. for floats you might have to parse a string

seanknox:
This will be a huge boon to folks once it lands. Excited to see it get out there!

yangl900:
@JackQuincy @anhowe anyone can help review this PR?

examples/kubernetes.large.json
```diff
@@ -0,0 +1,84 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "orchestratorVersion": "1.6.2"
```
JackQuincy: should be 1.6.6

parts/kubernetesmastercustomscript.sh
```diff
@@ -29,6 +29,17 @@ KUBECONFIG_CERTIFICATE="${19}"
 KUBECONFIG_KEY="${20}"
 ADMINUSER="${21}"
 
+# Default values for backoff configuration
+CLOUDPROVIDER_BACKOFF="${21}"
```
JackQuincy: we already have 21 above this should start with 22

pkg/acsengine/const.go
```diff
@@ -90,6 +102,12 @@ var KubeImages = map[api.OrchestratorVersion]map[string]string{
 		"dnsmasq":      "k8s-dns-dnsmasq-amd64:1.13.0",
 		"pause":        "pause-amd64:3.0",
 		"windowszip":   "v1.6.0intwinnat.zip",
+		"nodestatusfreq":   "10s",
+		"nodegraceperiod":   "40s",
+		"podeviction":   "5m0s",
+		"routeperiod":   "10s",
```
JackQuincy: These aren't images and none of them vary by k8s version. Could we move these default consts to be siblings with the KubeImages map?  We can still override in the future when we allow the user to specifiy in the UX.

parts/kubernetesparams.t
JackQuincy: should be bool

pkg/acsengine/const.go
jackfrancis: I think the idea is that different versions of k8s will get different config values here. My latest commit, for example, only includes the `backoff*` and `ratelimit*` config properties in `1.6.6`. Does that address your question?

JackQuincy: Sounds good. I'd just want to make sure when we add parameters /use them in the custom data that nothing breaks. I don't imagine it would, just would want to verify.

seanknox: @jackfrancis @JackQuincy Hey Jacks! I'd like this feature to only be available on v1.6 k8s. It should not be toggleable on v1.5.7. Doable?

jackfrancis: The features you're identifying above are the default values of existent features. In other words, those values will change *after* we enable backoff/rate limiting on large clusters. Because we're passing in this config as a command line option now, we have to provide values. (I'm pretty sure passing these CLI options in is not toggleable per-version.)

parts/kubernetesagentcustomdata.yml
```diff
@@ -109,6 +109,10 @@ write_files:
     KUBELET_REGISTER_SCHEDULABLE=true
     KUBELET_NODE_LABELS={{ GetKubernetesLabels . }}
     KUBELET_POD_INFRA_CONTAINER_IMAGE={{WrapAsVariable "kubernetesPodInfraContainerSpec"}}
+    KUBELET_NODE_STATUS_UPDATE_FREQUENCY={{WrapAsVariable "kubernetesNodeStatusUpdateFrequency"}}
```
JackQuincy: do we need any similar changes in master customdata?

JackQuincy:
Seems like you have missed your most recent make build and that tempaltes.go is going to be out of date.

JackQuincy:
unit tests are failing. Please use make ci to repro and fix the failing tests

examples/kubernetes.large.json
```diff
@@ -0,0 +1,84 @@
+{
```
anhowe: merge this file with examples/largeclusters/kubernetes.json

parts/kubernetesagentcustomdata.yml
jackfrancis: Yes, done.

PR #898, opened 2017-06-30
exclude australiaeast from the regression test

dmitsh:
exclude australiaeast from the regression test

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/898)


dmitsh:
will re-submit

PR #901, opened 2017-07-03
Added retries for apt-get install and systemctl enable commands for kubernetes custom script.

### LINKED ISSUE #853: Docker Install should be retried; CSE should not swallow errors
This is a bad cluster that was provisioned, but the ARM template deployment was successful.

1. We should retry the docker installer.

2. Something swallowed the `systemctl enable docker` error. Probably, either the script isn't run with `set -e` or the enable command is being ` || true`d incorrectly.

`/var/log/cloud-init-output.log`:

```
The following additional packages will be installed:
  aufs-tools cgroupfs-mount libltdl7
Suggested packages:
  mountall
The following NEW packages will be installed:
  aufs-tools cgroupfs-mount docker-engine libltdl7
0 upgraded, 4 newly installed, 0 to remove and 55 not upgraded.
Need to get 19.4 MB/19.5 MB of archives.
After this operation, 102 MB of additional disk space will be used.
Get:1 http://azure.archive.ubuntu.com/ubuntu xenial/main amd64 libltdl7 amd64 2.4.6-0.1 [38.3 kB]
Get:2 https://aptdocker.azureedge.net/repo ubuntu-xenial/main amd64 docker-engine amd64 1.12.6-0~ubuntu-xenial [19.4 MB]
Err:2 https://aptdocker.azureedge.net/repo ubuntu-xenial/main amd64 docker-engine amd64 1.12.6-0~ubuntu-xenial
  GnuTLS recv error (-54): Error in the pull function.
Fetched 38.3 kB in 0s (45.5 kB/s)
E: Failed to fetch https://aptdocker.azureedge.net/repo/pool/main/d/docker-engine/docker-engine_1.12.6-0~ubuntu-xenial_amd64.deb  GnuTLS recv error (-54): Error in the pull function.

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Failed to restart docker.service: Unit docker.service not found.
usermod: group 'docker' does not exist
```

`/var/log/azure/cluster-provision.log`:
```
waiting for runcmd to finish
runcmd finished
APISERVER_PRIVATE_KEY is non-empty, assuming master node
+ ensureDocker
+ systemctl enable docker
Failed to execute operation: No such file or directory
+ true
+ configNetworkPolicy
+ [[ none = \a\z\u\r\e ]]
+ [[ none = \c\a\l\i\c\o ]]
+ setNetworkPlugin kubenet
+ sed -i 's/^KUBELET_NETWORK_PLUGIN=.*/KUBELET_NETWORK_PLUGIN=kubenet/' /etc/default/kubelet
+ setDockerOpts ''
+ sed -i 's#^DOCKER_OPTS=.*#DOCKER_OPTS=#' /etc/default/kubelet
```

bobbypage:

Added retries for apt-get install commands as well as systemctl commands for enabling services such as docker and the kubelet. Fixes #853 

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/901)


parts/kubernetesmastercustomscript.sh
```diff
@@ -200,8 +200,30 @@ function configNetworkPolicy() {
     fi
 }
 
+function systemctlEnableAndCheck() {
+    systemctl enable $1
+    systemctl is-enabled $1
```
rjtsdl: qq: did you see lots of times when enable is called. But it is not enabled?

bobbypage: I didn't personally, but Cole mentioned he saw this on the original issue (#853), so that's why I added this check...

colemickens: That's only because the install failed and we weren't failing this script.

I don't really think this is necessary... Instead we should be setting `set -e` in this script so that any uncaught/unhandled errors just fail the entire script.

JackQuincy:
We've had an issue that this would of helped with and david isn't reponding and fixing the merge conflict so I'm going to start a new PR

PR #909, opened 2017-07-05
k8s: update to 1.7.0

### LINKED ISSUE #975: Support Kubernetes v1.7.0
## Acceptance Criteria
- Users can install Kubernetes v1.7.0 on Linux and Windows clusters.

martell:
update to kubernetes 1.7.0

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #975

I tested this on Azure on eu-west.
Installed a bunch of stuff with helm and all was working as intended.
**Release note**:

```release-note
```
None

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/909)

### DIFF
--- a/parts/kuberneteswindowssetup.ps1
+++ b/parts/kuberneteswindowssetup.ps1
@@ -178,9 +178,11 @@ Write-KubernetesStartFiles($podCIDR)
 c:\k\kubelet.exe --hostname-override=`$global:AzureHostname --pod-infra-container-image=kubletwin/pause --resolv-conf="" --allow-privileged=true --enable-debugging-handlers --api-servers=https://`${global:MasterIP}:443 --cluster-dns=`$global:KubeDnsServiceIp --cluster-domain=cluster.local  --kubeconfig=c:\k\config --hairpin-mode=promiscuous-bridge --v=2 --azure-container-registry-config=c:\k\azure.json --runtime-request-timeout=10m
 "@
 
-    if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7")
+    if ($global:KubeBinariesVersion -ge "1.6.0")
     {
+        # stop using container runtime interface from 1.6.0+ (officially deprecated from 1.7.0)
         $KubeletArgList += "--enable-cri=false"
+        # more time is needed to pull windows server images (flag supported from 1.6.0)
         $KubeletCommandLine += " --enable-cri=false --image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`""
     }
     $KubeletArgListStr = "`"" + ($KubeletArgList -join "`",`"") + "`""
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -80,6 +80,28 @@ const (
 
 // KubeImages represents Docker images used for Kubernetes components based on Kubernetes version
 var KubeImages = map[api.OrchestratorVersion]map[string]string{
+	api.Kubernetes170: {
+		"hyperkube":    "hyperkube-amd64:v1.7.0",
+		"dashboard":    "kubernetes-dashboard-amd64:v1.6.1",
+		"exechealthz":  "exechealthz-amd64:1.2",
+		"addonresizer": "addon-resizer:2.0",
+		"heapster":     "heapster:v1.4.0",
+		"dns":          "k8s-dns-kube-dns-amd64:1.14.4",
+		"addonmanager": "kube-addon-manager-amd64:v6.1.2",
+		"dnsmasq":      "k8s-dns-dnsmasq-amd64:1.14.4",
+		"pause":        "pause-amd64:3.0",
+		"windowszip":   "v1.7.0intwinnat.zip",
+		"nodestatusfreq":  DefaultKubernetesNodeStatusUpdateFrequency,
+		"nodegraceperiod": DefaultKubernetesCtrlMgrNodeMonitorGracePeriod,
+		"podeviction":     DefaultKubernetesCtrlMgrPodEvictionTimeout,
+		"routeperiod":     DefaultKubernetesCtrlMgrRouteReconciliationPeriod,
+		"backoffretries":  strconv.Itoa(DefaultKubernetesCloudProviderBackoffRetries),
+		"backoffjitter":   strconv.FormatFloat(DefaultKubernetesCloudProviderBackoffJitter, 'f', -1, 64),
+		"backoffduration": strconv.Itoa(DefaultKubernetesCloudProviderBackoffDuration),
+		"backoffexponent": strconv.FormatFloat(DefaultKubernetesCloudProviderBackoffExponent, 'f', -1, 64),
+		"ratelimitqps":    strconv.FormatFloat(DefaultKubernetesCloudProviderRateLimitQPS, 'f', -1, 64),
+		"ratelimitbucket": strconv.Itoa(DefaultKubernetesCloudProviderRateLimitBucket),
+	},
 	api.Kubernetes166: {
 		"hyperkube":       "hyperkube-amd64:v1.6.6",
 		"dashboard":       "kubernetes-dashboard-amd64:v1.6.1",
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -850,7 +850,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) map[str
 				case "kubeClusterCidr":
 					val = "10.244.0.0/16"
 				case "kubeBinariesVersion":
-					val = string(api.KubernetesLatest)
+					val = string(api.KubernetesDefaultVersion)
 				case "caPrivateKey":
 					// The base64 encoded "NotAvailable"
 					val = "Tm90QXZhaWxhYmxlCg=="
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -61,8 +61,10 @@ const (
 	Kubernetes162 OrchestratorVersion = "1.6.2"
 	// Kubernetes166 is the string constant for Kubernetes 1.6.6
 	Kubernetes166 OrchestratorVersion = "1.6.6"
-	// KubernetesLatest is the string constant for latest Kubernetes version
-	KubernetesLatest OrchestratorVersion = Kubernetes166
+	// Kubernetes166 is the string constant for Kubernetes 1.7.0
+	Kubernetes170 OrchestratorVersion = "1.7.0"
+	// KubernetesDefaultVersion is the string constant for current Kubernetes version
+	KubernetesDefaultVersion OrchestratorVersion = Kubernetes166
 )
 
 const (
--- a/pkg/api/vlabs/const.go
+++ b/pkg/api/vlabs/const.go
@@ -84,7 +84,7 @@ const (
 const (
 	// Kubernetes153 is the string constant for Kubernetes 1.5.3
 	Kubernetes153 OrchestratorVersion = "1.5.3"
-	// Kubernetes157 is the string constant for Kubernetes 1.5.3
+	// Kubernetes157 is the string constant for Kubernetes 1.5.7
 	Kubernetes157 OrchestratorVersion = "1.5.7"
 	// Kubernetes160 is the string constant for Kubernetes 1.6.0
 	Kubernetes160 OrchestratorVersion = "1.6.0"
@@ -92,6 +92,8 @@ const (
 	Kubernetes162 OrchestratorVersion = "1.6.2"
 	// Kubernetes166 is the string constant for Kubernetes 1.6.6
 	Kubernetes166 OrchestratorVersion = "1.6.6"
-	// KubernetesLatest is the string constant for latest Kubernetes version
-	KubernetesLatest OrchestratorVersion = Kubernetes166
+	// Kubernetes170 is the string constant for Kubernetes 1.7.0
+	Kubernetes170 OrchestratorVersion = "1.7.0"
+	// KubernetesDefaultVersion is the string constant for current Kubernetes version
+	KubernetesDefaultVersion OrchestratorVersion = Kubernetes166
 )
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -122,7 +122,7 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 			}
 		}
 		// Enforce sane cloudprovider rate limit defaults, if CloudProviderRateLimit is true in KubernetesConfig
-		if a.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimit == true && k8sVersion == api.Kubernetes166 {
+		if a.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimit == true && (k8sVersion == api.Kubernetes170 || k8sVersion == api.Kubernetes166) {
 			if a.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS == 0 {
 				a.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
 			}
--- a/pkg/acsengine/engine_test.go
+++ b/pkg/acsengine/engine_test.go
@@ -198,12 +198,15 @@ func addTestCertificateProfile(api *api.CertificateProfile) {
 
 func TestVersionOrdinal(t *testing.T) {
 	RegisterTestingT(t)
+
+	v170 := api.OrchestratorVersion("1.7.0")
 	v166 := api.OrchestratorVersion("1.6.6")
 	v162 := api.OrchestratorVersion("1.6.2")
 	v160 := api.OrchestratorVersion("1.6.0")
 	v153 := api.OrchestratorVersion("1.5.3")
 	v16 := api.OrchestratorVersion("1.6")
 
+	Expect(v166 < v170).To(BeTrue())
 	Expect(v166 > v162).To(BeTrue())
 	Expect(v162 < v166).To(BeTrue())
 	Expect(v162 > v160).To(BeTrue())
--- a/pkg/api/convertertoapi.go
+++ b/pkg/api/convertertoapi.go
@@ -485,7 +485,7 @@ func convertV20160330OrchestratorProfile(v20160330 *v20160330.OrchestratorProfil
 func convertV20170131OrchestratorProfile(v20170131 *v20170131.OrchestratorProfile, api *OrchestratorProfile) {
 	api.OrchestratorType = OrchestratorType(v20170131.OrchestratorType)
 	if api.OrchestratorType == Kubernetes {
-		api.OrchestratorVersion = KubernetesLatest
+		api.OrchestratorVersion = KubernetesDefaultVersion
 	} else if api.OrchestratorType == DCOS {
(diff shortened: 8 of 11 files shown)


pkg/acsengine/const.go
```diff
@@ -54,6 +54,18 @@ const (
...
+		"hyperkube":    "hyperkube-amd64:v1.7.0",
+		"dashboard":    "kubernetes-dashboard-amd64:v1.6.1",
+		"exechealthz":  "exechealthz-amd64:1.2",
+		"addonresizer": "addon-resizer:2.0",
+		"heapster":     "heapster:v1.4.0",
+		"dns":          "k8s-dns-kube-dns-amd64:1.14.4",
+		"addonmanager": "kube-addon-manager-amd64:v6.1.2",
+		"dnsmasq":      "k8s-dns-dnsmasq-amd64:1.14.4",
+		"pause":        "pause-amd64:3.0",
+		"windowszip":   "v1.7.0intwinnat.zip",
```
JackQuincy: @JiangtianLi I'm guessing this binary doesn't exist yet. Could you help with adding it?

pkg/api/const.go
```diff
@@ -61,8 +61,10 @@ const (
 	Kubernetes162 OrchestratorVersion = "1.6.2"
 	// Kubernetes166 is the string constant for Kubernetes 1.6.6
 	Kubernetes166 OrchestratorVersion = "1.6.6"
+	// Kubernetes166 is the string constant for Kubernetes 1.6.6
+	Kubernetes170 OrchestratorVersion = "1.7.0"
 	// KubernetesLatest is the string constant for latest Kubernetes version
-	KubernetesLatest OrchestratorVersion = Kubernetes166
+	KubernetesLatest OrchestratorVersion = Kubernetes170
```
JackQuincy: while this is true it is the latest version we might want to mark 1.6.6 for a while as default until we have a little bake time on 1.7.0

martell: I was thinking that when I was doing the patch.
I believe this only occurs when you do not specify orchestratorVersion in your json.
Will update to reflect your comments.

pkg/acsengine/const.go
JiangtianLi: sure. we don't have 1.7.0 binary for Windows yet. we need to build and upload.

martell: what can I say, I move fast waiting for no one :)

JiangtianLi:
@martell v1.7.0intwinnat.zip has been uploaded. In kuberneteswindowssetup.ps1, please modify and remove `--enable-cri=false` for v1.7.0 and above

JiangtianLi:
@martell please deploy a k8s windows 1.7.0 cluster and verify workload at https://github.com/Azure/acs-engine/blob/master/docs/kubernetes.windows.md works properly. Please let me know if you hit any issue. Thanks!

JiangtianLi:
@martell Please merge properly with https://github.com/Azure/acs-engine/pull/892
@jackfrancis What do you suggest rate limit parameters for v1.7.0?

jackfrancis:
@JiangtianLi `ratelimit*` and `backoff*` const values should be the same as `1.6.6`. I'll verify that k8s 1.7.0 hasn't changed any of the defaults for these const params:

- `nodestatusfreq`
- `nodegraceperiod`
- `podeviction`
- `routeperiod`

Stay tuned for final word on the above 4.

jackfrancis:
@JiangtianLi Confirmed that none of the default config values for the params we're using have changed from k8s 1.6.6 to 1.7.0 (we're explicitly using those defaults as _our_ defaults in `const.go`). For clarity, here are the values and authoritative references in the k8s 1.7.0 release sha:

- nodestatusfreq: `"10s"`
  - https://github.com/kubernetes/kubernetes/blob/d3ada0119e776222f11ec7945e6d860061339aad/pkg/apis/componentconfig/v1alpha1/defaults.go#L313

- nodegraceperiod: `"40s"`
  - https://github.com/kubernetes/kubernetes/blob/d3ada0119e776222f11ec7945e6d860061339aad/cmd/kube-controller-manager/app/options/options.go#L87

- podeviction: `"5m0s"`
  - https://github.com/kubernetes/kubernetes/blob/d3ada0119e776222f11ec7945e6d860061339aad/cmd/kube-controller-manager/app/options/options.go#L86

- routeperiod: `"10s"`
  - https://github.com/kubernetes/kubernetes/blob/d3ada0119e776222f11ec7945e6d860061339aad/cmd/kube-controller-manager/app/options/options.go#L76

In summary, copy all values over from 1.6.6!

pkg/api/vlabs/const.go
```diff
@@ -84,14 +84,16 @@ const (
...
 	// Kubernetes160 is the string constant for Kubernetes 1.6.0
 	Kubernetes160 OrchestratorVersion = "1.6.0"
 	// Kubernetes162 is the string constant for Kubernetes 1.6.2
 	Kubernetes162 OrchestratorVersion = "1.6.2"
 	// Kubernetes166 is the string constant for Kubernetes 1.6.6
 	Kubernetes166 OrchestratorVersion = "1.6.6"
+	// Kubernetes170 is the string constant for Kubernetes 1.7.0
+	Kubernetes170 OrchestratorVersion = "1.7.0"
 	// KubernetesLatest is the string constant for latest Kubernetes version
 	KubernetesLatest OrchestratorVersion = Kubernetes166
```
SaaldjorMike: Wouldn't we also need to update this?

pkg/api/const.go
```diff
@@ -61,6 +61,8 @@ const (
 	Kubernetes162 OrchestratorVersion = "1.6.2"
 	// Kubernetes166 is the string constant for Kubernetes 1.6.6
 	Kubernetes166 OrchestratorVersion = "1.6.6"
+	// Kubernetes166 is the string constant for Kubernetes 1.6.6
+	Kubernetes170 OrchestratorVersion = "1.7.0"
 	// KubernetesLatest is the string constant for latest Kubernetes version
 	KubernetesLatest OrchestratorVersion = Kubernetes166
```
SaaldjorMike: Wouldn't we also need to update this?

JackQuincy: We like to let a new version back a little while before making it default, possibly we should rename this.

SaaldjorMike: Makes sense. Was just wondering, due to the name of the variable + the variable description in the line above, that points to the fact that it is referring to the latest version and not the "default version" or whatever is a more suited name.

pkg/api/vlabs/const.go
martell: I previously did this but was asked not to so I did in a fixup.
This should be called `KubernetesDefault` instead.
With the go ahead from @JackQuincy I will rename that.

parts/kuberneteswindowssetup.ps1
```diff
@@ -175,7 +175,7 @@ Write-KubernetesStartFiles($podCIDR)
 c:\k\kubelet.exe --hostname-override=`$global:AzureHostname --pod-infra-container-image=kubletwin/pause --resolv-conf="" --allow-privileged=true --enable-debugging-handlers --api-servers=https://`${global:MasterIP}:443 --cluster-dns=`$global:KubeDnsServiceIp --cluster-domain=cluster.local  --kubeconfig=c:\k\config --hairpin-mode=promiscuous-bridge --v=2 --azure-container-registry-config=c:\k\azure.json
 "@
 
-    if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7")
+    if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7" -and $global:KubeBinariesVersion -lt "1.7.0")
```
JiangtianLi: @martell Is `--image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`"` not used by 1.7.0? I think at least `--image-pull-progress-deadline=20m` is needed for downloading windows container image.

JiangtianLi:
@martell Please merge master and test since there is a commit to fix failure in Windows node due to latest Windows version.

JiangtianLi:
@martell The change LGTM. There is an issue with 1.7.0 zip for container internet connection due to that cri is now in 1.7.0 and can't be disabled. So I will need update zip and let you know.

lachie83:
I also have a PR for 1.7.1 that builds on the awesome work that @martell has contributed. #1020 Which one shall we proceed with?

martell:
@lachie83 seems like overkill to duplicate work :)
I rebased from master today to keep this updated.
I can bump the version to 1.7.1 if you really want to skip 1.7.0?

There are some extra thing in here
- rename KubernetesLatest to KubernetesCurrent
- fix typo in Kubernetes157 comment

@JiangtianLi any luck on your updated zips?
FYI I am using this in production quite happily.

JiangtianLi:
@martell what kubelet/kubeproxy binaries are you using? The issue I have is that Windows container lack one NIC and doesn't have outbound connection. Do you see the same case?

martell:
@JiangtianLi I am using this with linux containers currently.
I was waiting for updated zips before I tested windows containers.

JiangtianLi:
@martell I see. Good to know Linux container is already working. Windows containers are built with additional changes. I'll try to solve the issue and update you.

examples/kubernetesversions/kubernetes1.7.0.json
```diff
@@ -0,0 +1,36 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "orchestratorVersion": "1.7.0"
```
seanknox: I want to start removing cluster configs that provide little value—I'd opt to delete this one too.

examples/kubernetesversions/kubernetes1.7.1.json
```diff
@@ -0,0 +1,36 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "orchestratorVersion": "1.7.1"
```
seanknox: Remove this example too please.

parts/kuberneteswindowssetup.ps1
```diff
@@ -178,10 +178,14 @@ Write-KubernetesStartFiles($podCIDR)
 c:\k\kubelet.exe --hostname-override=`$global:AzureHostname --pod-infra-container-image=kubletwin/pause --resolv-conf="" --allow-privileged=true --enable-debugging-handlers --api-servers=https://`${global:MasterIP}:443 --cluster-dns=`$global:KubeDnsServiceIp --cluster-domain=cluster.local  --kubeconfig=c:\k\config --hairpin-mode=promiscuous-bridge --v=2 --azure-container-registry-config=c:\k\azure.json --runtime-request-timeout=10m
 "@
 
-    if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7")
+    if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7")
     {
-        $KubeletArgList += "--enable-cri=false"
-        $KubeletCommandLine += " --enable-cri=false --image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`""
+        if ($global:KubeBinariesVersion -lt "1.7.0")
```
seanknox: This is unclear to me; what versions are we trying to target for this conditional?  Can this be rewritten to be something like

```powershell
 if ($global:KubeBinariesVersion -ge "1.6.0" -and $global:KubeBinariesVersion -le "1.7.0")
```

pkg/acsengine/engine.go
```diff
@@ -850,7 +850,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) map[str
 				case "kubeClusterCidr":
 					val = "10.244.0.0/16"
 				case "kubeBinariesVersion":
-					val = string(api.KubernetesLatest)
+					val = string(api.KubernetesCurrent)
```
seanknox: If you're going to change this, I'd opt for something clearer like `KubernetesDefaultVersion`, or just not change it in this PR.

pkg/acsengine/const.go
```diff
@@ -80,6 +80,50 @@ const (
 
 // KubeImages represents Docker images used for Kubernetes components based on Kubernetes version
 var KubeImages = map[api.OrchestratorVersion]map[string]string{
+	api.Kubernetes171: {
+		"hyperkube":    "hyperkube-amd64:v1.7.1",
+		"dashboard":    "kubernetes-dashboard-amd64:v1.6.1",
```
seanknox: This PR started out as v1.7.0, and I'd like it to continue to be focused on that. Each Kubernetes version needs to be vetted, run through our CI, Windows artifacts created, etc.

examples/kubernetesversions/kubernetes1.7.0.json
martell: Sure can do

examples/kubernetesversions/kubernetes1.7.1.json
martell: sure

pkg/acsengine/engine.go
martell: done

pkg/acsengine/const.go
martell: Was not aware of that process.
Will do

parts/kuberneteswindowssetup.ps1
martell: sure

martell:
@seanknox I wasn't aware of the internal testing procedures involved with vetting each version.
More then happy to have a single PR for each. My concern with having 2 PR's was that 1.7.0 would be skipped when I was running into intermittent issues like `installation: Error: client: etcd cluster is unavailable or misconfigured` on 1.7.1.
I assume the windows issues @JiangtianLi has with the NAT is external to acs-engine source code.
With the default still being 1.6.6 I assume it should be possible to merge this?
The quicker this goes in the quicker @lachie83 can rebase the new PR off master and to get 1.7.1  internal testing kickstarted. :)

parts/kuberneteswindowssetup.ps1
```diff
@@ -178,10 +178,14 @@ Write-KubernetesStartFiles($podCIDR)
...
+    if ($global:KubeBinariesVersion -ge "1.6.0" -and $global:KubeBinariesVersion -le "1.7.0")
+    {
+            $KubeletArgList += "--enable-cri=false"
+            $KubeletCommandLine += " --enable-cri=false"
+    }
     if ($global:KubeBinariesVersion -ne "1.5.3" -and $global:KubeBinariesVersion -ne "1.5.7")
     {
-        $KubeletArgList += "--enable-cri=false"
-        $KubeletCommandLine += " --enable-cri=false --image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`""
+        $KubeletCommandLine += " --image-pull-progress-deadline=20m --cgroups-per-qos=false --enforce-node-allocatable=`"`""
```
seanknox: @martell or @JiangtianLi can you you add a comment above these conditionals so others understand what they do (and why we have them)?

seanknox:
@acs-bot test this please

seanknox:
@martell No worries at all and thanks for your contribution! This looks LGTM, have one request to be document the need for different kubelet flags in `parts/kuberneteswindowssetup.ps1`.

parts/kuberneteswindowssetup.ps1
martell: For clarification I only removed `--enable-cri=false` from these 2 lines to guard it from 1.6x+
We can probably simplify the conditional to just `if ($global:KubeBinariesVersion -ge "1.6.0")`?
and the comment would simply be `we don't use Container Runtime Interface from 1.6.0+`
Would that suffice @seanknox ?

JiangtianLi:
@seanknox @martell --enable-cri is deprecated from 1.7.0. Other options has good explanation in cmd/kubelet/app/options/options.go as we need them for the Window path, i.e., without them, Windows code path are not implemented. --image-pull-progress-deadline=20m is due to full Windows Server image size.

parts/kuberneteswindowssetup.ps1
seanknox: FYI, I want to provide as much context to users here without forcing them to dig through Kubernetes source code.

Based on @JiangtianLi comment below, `--enable-cri` is deprecated starting in v1.7.0. I think just a comment to that effect is fine.

`--image-pull-progress-deadline=20m` is needed due to the size of the container images.  This option is enabled only for k8s versions ~v1.5.3 to ~v1.5.7— @JiangtianLi is that intentional or should this be on for all versions?

JiangtianLi: @seanknox yes, 1.5.* doesn't have --image-pull-progress-deadline option. The timeout is hardcoded. We have that option from 1.6.0 and above.

martell: done

parts/kuberneteswindowssetup.ps1
```diff
@@ -177,11 +177,16 @@ Write-KubernetesStartFiles($podCIDR)
     $KubeletCommandLine = @"
 c:\k\kubelet.exe --hostname-override=`$global:AzureHostname --pod-infra-container-image=kubletwin/pause --resolv-conf="" --allow-privileged=true --enable-debugging-handlers --api-servers=https://`${global:MasterIP}:443 --cluster-dns=`$global:KubeDnsServiceIp --cluster-domain=cluster.local  --kubeconfig=c:\k\config --hairpin-mode=promiscuous-bridge --v=2 --azure-container-registry-config=c:\k\azure.json --runtime-request-timeout=10m
 "@
-
+    # stop using container runtime interface from 1.6.0+ (officially deprecated from 1.7.0)
+    if ($global:KubeBinariesVersion -ge "1.6.0")
```
seanknox: Think you want this to be:
```powershell
if ($global:KubeBinariesVersion -ge "1.6.0" -and $global:KubeBinariesVersion -lt "1.7.0")
```

That ensures users are using between v1.6.0 up until v1.7.0

parts/kuberneteswindowssetup.ps1
seanknox: Ok, so `image-pull-progress-deadline` requires v1.6.0 at a minimum. @martell you should update that logic to be

```powershell
if ($global:KubeBinariesVersion -ge "1.6.0")
```

No need to bookend k8s versions here as it continues to be supported in v1.7.0.

parts/kuberneteswindowssetup.ps1
martell: The flag would still be valid for 1.7.0. The deprecation just makes sure this happens by default. I suppose it might become an invalid flag in 1.8+

seanknox: Fair enough. In that case, I think you could group both checks under one check for v1.6.0 or greater.

martell:
@seanknox done :)

seanknox:
@acs-bot test this please

andrey-moor:
Would it be possible to upgrade from 1.6 without deploying a new cluster?

amanohar:
@andrey-moor ACS Engine today does have an experimental upgrade feature in the work that will enable upgrade with redeploy (but will recreate agent and master nodes). At some point ACS Engine will support upgrade to 1.7.0.

seanknox:
Thanks again @martell @JiangtianLi!

seanknox:
giphy-downsized 4

PR #924, opened 2017-07-06
add Helm Tiller to Kubernetes addon modules

### LINKED ISSUE #854: Add Helm Tiller as Addon
 
No
---

Feature Request

Kubernetes

**Feature**

Add [http://helm.sh](Helm)'s Tiller (in-cluster) component as an add-on module for Kubernetes (_a la_ Heapster). Tiller runs as a Deployment (with an optional Service) inside of the Kubernetes cluster. The installation should be as close as possible to the `helm init` behavior. Ideally, a user should only have to install the `helm` cli client (`helm init --client-only`), and immediately be able to talk to Tiller (`helm version` should show Tiller's version with no other configuration steps).

Other than image version, there are not many commonly used configuration options on Tiller's deployment, so this should be fairly straightforward.

seanknox:
**What this PR does / why we need it**: This implements feature request #854 by adding the Tiller manifest and supporting configuration to the Kubernetes addon modules.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #854 

**Special notes for your reviewer**: @technosophos rules

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/924)


martell:
@technosophos @seanknox Would it be possible to add this to the ubuntu host script?
`curl https://storage.googleapis.com/kubernetes-helm/helm-vX.Y.Z-linux-amd64.tar.gz`
`tar -zxvf helm-vX.Y.Z-linux-amd64.tgz`
`mv linux-amd64/helm /usr/local/bin/helm`
Otherwise we still have to install helm on the client when we ssh into the host.

Also thank you for creating helm, lets keep this ship sailing :)

seanknox:
Thanks @martell, great suggestion. Added a followup item here: https://github.com/Azure/acs-engine/issues/974

technosophos:
None of the errors on that Jenkins job appear to have anything to do with changed code. Can someone please explain them? Are they flakes or does this need a rebase or what?

seanknox:
I've been trying to track this down too. Here's the actual error from text template fun:

```
Deployment failed. {
  "error": {
    "code": "InvalidTemplate",
    "message": "Unable to process template language expressions for resource '/subscriptions/e766d9ee-d3d9-4b63-a912-8963dcfdf655/resourceGroups/dabigsean-5966938d/providers/Microsoft.Compute/virtualMachines/k8s-master-37197068-0' at line '1' and column '51661'. 'The template variable 'kubernetesTillerSpec' is not found. Please see https://aka.ms/arm-template/#variables for usage details.'"
  }
}
```

parts/kubernetesmastercustomdata.yml
```diff
@@ -247,6 +247,7 @@ write_files:
     sed -i "s|<kubernetesKubeDNSSpec>|{{WrapAsVariable "kubernetesKubeDNSSpec"}}|g; s|<kubernetesDNSMasqSpec>|{{WrapAsVariable "kubernetesDNSMasqSpec"}}|g; s|<kubernetesExecHealthzSpec>|{{WrapAsVariable "kubernetesExecHealthzSpec"}}|g" "/etc/kubernetes/addons/kube-dns-deployment.yaml"
     sed -i "s|<kubernetesHeapsterSpec>|{{WrapAsVariable "kubernetesHeapsterSpec"}}|g; s|<kubernetesAddonResizerSpec>|{{WrapAsVariable "kubernetesAddonResizerSpec"}}|g" "/etc/kubernetes/addons/kube-heapster-deployment.yaml"
     sed -i "s|<kubernetesDashboardSpec>|{{WrapAsVariable "kubernetesDashboardSpec"}}|g" "/etc/kubernetes/addons/kubernetes-dashboard-deployment.yaml"
+    sed -i "s|<kubernetesTillerSpec>|{{WrapAsVariable "kubernetesTillerSpec"}}|g" "/etc/kubernetes/addons/kube-tiller-deployment.yaml"
```
lachie83: I believe you also need to lay these files down in disk. eg L130
```
- path: /etc/kubernetes/addons/kubernetes-dashboard-deployment.yaml
  permissions: "0644"
  encoding: gzip
  owner: "root"
  content: !!binary |
    MASTER_ADDON_KUBERNETES_DASHBOARD_DEPLOYMENT_B64_GZIP_STR
```

parts/kubernetesmasteraddons-tiller-service.yaml
```diff
@@ -0,0 +1,29 @@
+# Produced by 'helm init --dry-run --debug'
```
lachie83: Can we please clean this manifest up given that most of it is null. It also needs the following label for addonmanager to pick it up

```
  labels:
    kubernetes.io/cluster-service: "true"
```

parts/kubernetesmasteraddons-tiller-deployment.yaml
```diff
@@ -0,0 +1,117 @@
+# Produced by 'helm init --dry-run --debug'
+apiVersion: extensions/v1beta1
```
lachie83: Can we please clean this manifest up given that most of it is null. It also needs the following label for addonmanager to pick it up

```
  labels:
    kubernetes.io/cluster-service: "true"
```

pkg/acsengine/const.go
```diff
@@ -65,6 +65,7 @@ var KubeImages = map[api.OrchestratorVersion]map[string]string{
 		"dnsmasq":      "k8s-dns-dnsmasq-amd64:1.13.0",
 		"pause":        "pause-amd64:3.0",
 		"windowszip":   "v1.6.6intwinnat.zip",
+		"tiller":       "tiller:v2.5.0",
```
lachie83: Let's bump this to 2.5.1

lachie83:
@seanknox this is ready for review

pkg/acsengine/defaults.go
```diff
@@ -17,6 +17,7 @@ var (
 		//KubernetesSpecConfig is the default kubernetes container image url.
 		KubernetesSpecConfig: KubernetesSpecConfig{
 			KubernetesImageBase:    "gcrio.azureedge.net/google_containers/",
+			TillerImageBase:		"gcrio.azureedge.net/kubernetes-helm/",
```
martell: indentation

pkg/acsengine/types.go
```diff
@@ -42,6 +42,7 @@ type DCOSSpecConfig struct {
 //KubernetesSpecConfig is the kubernetes container images used.
 type KubernetesSpecConfig struct {
 	KubernetesImageBase    string
+	TillerImageBase 	   string
```
martell: indentation

pkg/acsengine/defaults.go
lachie83: Thanks

martell:
namespace seems fine, resources seem fine, probes look good.
`beta.kubernetes.io/os: linux` didn't know multiple OS were supported in the one cluster.
Or at least they must be planned seeing this flag.
LGTM other then the indentation nits, excited to see this land.
Hope you guys give this a speedy review so we can all be happy helming.

lachie83:
@martell thanks for the review. We do support hybrid clusters. I'm now wondering if this is really required because it should function the same on a windows host right?

martell:
hey @lachie83 :) hmm, docs seem out of date also.
```Kubelet populates this with runtime.GOOS as defined by Go. This can be handy if you are mixing operating systems in your cluster (although currently Linux is the only OS supported by Kubernetes).
```
Not sure if this means will only run on linux or try to request linux first.

lachie83:
Yeah those are out of date and nodeSelector means that it will only run on node that match that label.

seanknox:
You need to keep the node selector, it was part of helm work to support running tiller in mixed OS agent pools. The selector ensures tiller lands on a Linux node.

seanknox:
@acs-bot test this please

seanknox:
@lachie83 I rebased after merging in k8s v1.7.2 and other features yesterday. With luck this should be good to go today!

lachie83:
Thanks @seanknox Let me double check this morning. I think I may have to split out the tiller-deployment for 1.5 as rbac was only in alpha there and I'm provisioning with a CRB. I'll report back

pkg/acsengine/const.go
```diff
@@ -90,6 +90,7 @@ var KubeImages = map[string]map[string]string{
 		"addonmanager":    "kube-addon-manager-amd64:v6.4-beta.2",
 		"dnsmasq":         "k8s-dns-dnsmasq-amd64:1.14.4",
 		"pause":           "pause-amd64:3.0",
+		"tiller":          "tiller:v2.5.1",
```
jackfrancis: Let's replace the `"tiller:v2.5.1"` string literal with a `const`, like `DefaultTillerImage` (or something) here and in all the k8s versions below.

parts/kubernetesmasteraddons-tiller-deployment1.5.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: v1
```
jackfrancis: How does this 1.5 version of the deployment spec fit in?

pkg/acsengine/const.go
lachie83: Would we override if there is a specific k8s version depedency?

parts/kubernetesmasteraddons-tiller-deployment1.5.yaml
lachie83: Was missing a commit but the main difference is L9 using the alpha rbac API as beta isn't available in 1.5

lachie83: It's beta in 1.6,7

pkg/acsengine/const.go
seanknox: > Let's replace the "tiller:v2.5.1" string literal with a const, like DefaultTillerImage (or something) here and in all the k8s versions below. 

Oh, I like that. Would opt to get that refactoring in in a subsequent PR, as we're following the existing pattern. New PR would have the refactoring, and maybe some new unit tests (whoa!).

parts/kubernetesmasteraddons-tiller-deployment1.5.yaml
jackfrancis: 👍

lachie83: I've been deploying all addons with specific service accounts and associated CRB even if RBAC isn't enabled. That way it all works when RBAC is enabled.

seanknox: The API differences between k8s v1.5.x and later versions are substantial enough to warrant a new deployment spec. Hopefully soon we'll be able to remove k8s v1.5.x entirely from acs-engine.

pkg/acsengine/const.go
jackfrancis: Yeah, the idea behind the version-specific dictionaries are to carry version-specific config values for the various params. Based on 2.5.1 being present in all k8s versions, I would assume that that value would evolve universally; so, for example, if we decide in the future to bake a `v2.6.0` into acs-engine-curated clusters, we'd want to change the tiller version string value across all versions. (And thus the advantage to defining it one place as a constant.)

And yes, if we evolve tiller versions going forward only for newer versions of k8s, then we'd revisit how exactly to decompose this collection of string variables.

Hope that makes sense!

jackfrancis: @seanknox hey man what about my preëxisting `DefaultKubernetesNodeStatusUpdateFrequency` pattern?

But seriously, yes, whether or not we deliver the tiller version in this PR as a `const`, we would benefit from a small housekeeping PR that refactors all of these copy pasta string literals as single sources of `const` truth.

lachie83: Perfect sense. I've updated to use constants.

pkg/acsengine/const.go
```diff
@@ -67,6 +67,8 @@ const (
 	DefaultKubernetesCloudProviderRateLimitQPS = 3.0
 	// DefaultKubernetesCloudProviderRateLimitBucket is 10, takes effect if DefaultKubernetesCloudProviderRateLimit is true
 	DefaultKubernetesCloudProviderRateLimitBucket = 10
+	// DefaultTillerImage defines the Helm Tiller deployment version on Kubernetes Clusters
+	DefaultTillerImage = "tiller:v2.5.1"
```
seanknox: Ohh you fancy! Seriously this will be nice to refactor the other string settings into consts in future work.

Issue #877, opened 2017-06-27
Pin to specific versions of azure-cli


seanknox:
Recently we discovered that a [newer version of azure-cli broke scripts](https://github.com/Azure/acs-engine/issues/811) we use for internal development and testing. We should pin to the latest "blessed" version of azure-cli to prevent upstream changes from breaking acs-engine.

seanknox:
We already do this in the Dockerfile, so maybe it's enough to call out a specific version in the README.

jackfrancis:
This is done.

PR #931, opened 2017-07-07
Fix syntax of passing files to --parameters

yuvipanda:
Without this it'll give you a very non-descriptive `invalid syntax: ./_output/datahubfa17/azuredeploy.parameters.json`
error!

I spent quite a bit of time trying to understand this mysterious error message :( I'd have liked to not.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/931)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ $ az group deployment create \
     --name "<DEPLOYMENT NAME>" \
     --resource-group "<RESOURCE_GROUP_NAME>" \
     --template-file "./_output/<INSTANCE>/azuredeploy.json" \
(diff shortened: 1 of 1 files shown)


seanknox:
@yuvipanda The confusing "@ or not to @" problem was is from a newer [azure-cli](https://github.com/Azure/azure-cli). We're planning to pin versions to prevent this kind of problem from happening again: https://github.com/Azure/acs-engine/issues/877

@yuvipanda can you try upgrading to the latest version of azure-cli and report back if the problem goes away?

shrutir25:
@yuvipanda - we will be closing this PR since there has been no response for over 2 weeks. Please feel free to reopen it.

PR #939, opened 2017-07-07
populate default value in unmarshal

### LINKED ISSUE #936: Close the gap of acs-engine api and ACS public document
No
---

ISSUE

I am noticing the api model on acs-engine is not aligned with what published [here](https://docs.microsoft.com/en-us/rest/api/compute/containerservices). For example, acs-engine requires user pass in a valid MasterProfile.Count field, but the document clearly states that is not required, and be default to 1. 

All

**What happened**:
Once we rollout acs-engine with acs rp in the next one or two months, broader people will notice the breaking change even with the same api-version, such as 2016-09-30. The behavior will be different.

User shouldn't notice we swapped the backend service. As we still conform the same api specs.

Check 2016-09-30's validate function on `MasterProfile.Count`, apparently 0 is not allowed. The document allows 0, or empty.

rjtsdl:
close api definition gap to public document

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #936 

One sample for masterCount, if looks good, I will apply to all the other fields.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/939)

### DIFF
--- a/pkg/api/v20170131/types.go
+++ b/pkg/api/v20170131/types.go
@@ -121,7 +121,7 @@ func (m *MasterProfile) UnmarshalJSON(b []byte) error {
 	}
 	*m = MasterProfile(mm)
 	if m.Count == 0 {
-		// When MasterProfile.Count is missing or 0, set to default 1
+		// if MasterProfile.Count is missing or 0, set to default 1
 		m.Count = 1
 	}
 	return nil
@@ -145,6 +145,31 @@ type AgentPoolProfile struct {
 
 	// subnet is internal
 	subnet string
+}
+
+// UnmarshalJSON unmarshal json using the default behavior
+// And do fields manipulation, such as populating default value
+func (a *AgentPoolProfile) UnmarshalJSON(b []byte) error {
+	// Need to have a alias type to avoid circular unmarshal
+	type aliasAgentPoolProfile AgentPoolProfile
+	aa := aliasAgentPoolProfile{}
+	if e := json.Unmarshal(b, &aa); e != nil {
+		return e
+	}
+	*a = AgentPoolProfile(aa)
+	if a.Count == 0 {
+		// if AgentPoolProfile.Count is missing or 0, set it to default 1
+		a.Count = 1
+	}
+
+	if string(a.OSType) == "" {
+		// OSType is the operating system type for agents
+		// Set as nullable to support backward compat because
+		// this property was added later.
+		// If the value is null or not set, it defaulted to Linux.
+		a.OSType = Linux
+	}
+	return nil
 }
 
 // JumpboxProfile dscribes properties of the jumpbox setup
--- a/pkg/api/v20170131/types.go
+++ b/pkg/api/v20170131/types.go
@@ -1,6 +1,7 @@
 package v20170131
 
 import (
+	"encoding/json"
 	"fmt"
 	neturl "net/url"
 	"strings"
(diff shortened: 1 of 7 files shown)


rjtsdl:
@JiangtianLi , i saw OrchestratorType is using `UnmarshalText`, which needs a custom type `OrchestratorType`. If you think it is ok, i want to change all of them using `UnmarshalJSON`, which does addition decoding compared with `UnmarshalText`

Then i can use OrchestratorProfile's UnmarshalJSON to do what we have now. We can eliminate a type definition.

pkg/api/v20170131/types.go
```diff
@@ -109,6 +110,23 @@ type MasterProfile struct {
...
+// And do fields manipulation, such as populating default value
+func (m *MasterProfile) UnmarshalJSON(b []byte) error {
+	// Need to have a alias type to avoid circular unmarshal
+	type aliasMasterProfile MasterProfile
+	mm := aliasMasterProfile{}
+	if e := json.Unmarshal(b, &mm); e != nil {
+		return e
+	}
+	*m = MasterProfile(mm)
+	if m.Count == 0 {
```
rjtsdl: If other fields in MasterProfile need to populate default value, it can be done inside this function.

JiangtianLi: Does json.Unmarshal fill the empty field with 0?

pkg/api/v20170131/types.go
```diff
@@ -109,6 +110,23 @@ type MasterProfile struct {
 	subnet string
 }
 
+// UnmarshalJSON unmarshal json using the default behavior
+// And do fields manipulation, such as populating default value
+func (m *MasterProfile) UnmarshalJSON(b []byte) error {
+	// Need to have a alias type to avoid circular unmarshal
+	type aliasMasterProfile MasterProfile
```
JiangtianLi: why use alias?

pkg/api/v20170131/types.go
rjtsdl: Yes

pkg/api/v20170131/types.go
rjtsdl: otherwise, when you call MasterProfile.UnmarshalJSON, it causes a circular call to itself.

weinong:
LGTM. let me know when you are ready to checkin.

rjtsdl:
There maybe concerns on why do we need to populate default values when we have `SetPropertiesDefaults` when GenerateTemplate. 

My principle would be 
`SetPropertiesDefaults` operates on api.ContainerService, which is agnostic to api version. When the api doesn't expose all the vlab features, we use SetPropertiesDefaults to fill in vlab.ContainerService's defaults.

The populating in the unmarshal function is to fulfill the promise we made on the versioned api documents. Think about the scenario, in version 2018-09-30, we probably will switch default storageProfile to managedDisk. However, we won't change the existing versions default, which is to storageAccount. Make sense now?

rjtsdl:
@weinong, thx for reviewing. I will do this trick for the other versions. And also, I am inclined to do this for 2017-07-01. I will have the PR ready by Monday.

rjtsdl:
@acs-bot test this please

rjtsdl:
@weinong , i get code ready. PTAL

weinong:
Reviewed 2 of 2 files at r2, 5 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/azure/acs-engine/939)*

JiangtianLi:
@rjtsdl LGTM. Yes, we can change orchestratorType in a separate PR.

rjtsdl:
I will just merge it. Thx

PR #951, opened 2017-07-10
need to populate more default values for 2017-07-01

### LINKED ISSUE #936: Close the gap of acs-engine api and ACS public document

rjtsdl:
Need to populate more default values for 2017-07-01

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #936 

Ready to review

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/951)

### DIFF
--- a/pkg/api/v20170701/types.go
+++ b/pkg/api/v20170701/types.go
@@ -149,8 +149,8 @@ func (m *MasterProfile) UnmarshalJSON(b []byte) error {
 		m.StorageProfile = StorageAccount
 	}
 
-	// OSDiskSizeGB
-	// The current behavior is, when OSDiskSizeGB is 0, we just don't pass it in the arm template
+	// OSDiskSizeGB is an override value. vm sizes have default OS disk sizes.
+	// If it is not set. The user should get the default for the vm size
 	return nil
 }
 
@@ -205,8 +205,8 @@ func (a *AgentPoolProfile) UnmarshalJSON(b []byte) error {
 		a.OSType = Linux
 	}
 
-	// OSDiskSizeGB
-	// The current behavior is, when OSDiskSizeGB is 0, we just don't pass it in the arm template
+	// OSDiskSizeGB is an override value. vm sizes have default OS disk sizes.
+	// If it is not set. The user should get the default for the vm size
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/v20170701/types.go
```diff
@@ -139,9 +139,18 @@ func (m *MasterProfile) UnmarshalJSON(b []byte) error {
...
+		m.FirstConsecutiveStaticIP = "10.240.255.5"
+	}
+
+	if m.StorageProfile == "" {
+		// if StorageProfile is missing, set to default StorageAccount
+		m.StorageProfile = StorageAccount
+	}
+
+	// OSDiskSizeGB
+	// The current behavior is, when OSDiskSizeGB is 0, we just don't pass it in the arm template
```
rjtsdl: @anhowe @amanohar ,
Wondering what the default value would be used for OSDiskSizeGB? OSDiskSizeGB is different with other fields, such as "StorageProfile". I can deduce StorageProfile from either template, or SetDefaultProperties function. However, OSDiskSizeGB will be strip out in the arm template if it is 0.

Options:
1. We don't have a default value defined in the api spec. (I don't like this. User would be confused)
2. We pick a value as default. (What the value would be 200GB?)

JackQuincy: OSDIskSizeGB is an override value. vm sizes have default OS disk sizes. We should leave it at 0 if it isn't set and the user should get the default for the vm size.

rjtsdl: I see. I think i can change the comment to be what you have here.

rjtsdl:
@acs-bot test this please

rjtsdl:
@acs-bot test this please

Issue #608, opened 2017-05-12
HA Master lose connection on ILB restart due to kubernetes issue 41184


anhowe:
The agents may report NodeNot Ready due to single master restart, or HA Masters on ILB restart. 

This is to track the following kubernetes issue: https://github.com/kubernetes/kubernetes/issues/41185.

Per @colemickens, this is a related issue: https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-299177721.

To avoid this, an alternative solution is multi-host client: https://github.com/kubernetes/kubernetes/pull/40674.

seanknox:
To add more color on this, one person's research indicates the delay is due to the way the Linux kernel handles dead TCP connections: https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731

PR #953, opened 2017-07-10
added kubelet restart and restart-cancel scripts

dmitsh:
added kubelet restart and restart-cancel scripts

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/953)


seanknox:
FYI This work tracks against https://github.com/Azure/acs-engine/issues/608, where kubelets take 17m to reconnect to the master when the master is replaced/rebooted.

parts/kuberneteskubelet.service
```diff
@@ -13,6 +13,8 @@ ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le
 ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
 ExecStartPre=-/sbin/ebtables -t nat --list
 ExecStartPre=-/sbin/iptables -t nat --list
+ExecStartPre=/opt/azure/containers/kubelet-restart-cancel.sh
+ExecStartPost=/usr/bin/systemd-run --on-active=40 --on-unit-active=20 --unit=kubelet-restart /opt/azure/containers/kubelet-restart.sh
```
seanknox: Does this mean the script will run every 20s?

parts/kubelet-restart.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if [ ! -e /opt/azure/containers/runcmd.complete ]; then
+  exit 0
+fi
```
seanknox: Does `runcmd.complete`  refer to https://github.com/Azure/acs-engine/blob/master/parts/kubernetesagentcustomdata.yml#L158, which signifies the server has been provisioned? Want to make sure I understand the significance here.

parts/kubelet-restart.sh
```diff
@@ -0,0 +1,10 @@
...
+
+ready=$(kubectl --kubeconfig /var/lib/kubelet/kubeconfig get nodes/$(hostname) -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}')
```
seanknox: Not sure I understand the check here...the kubelet is seeing if it shows up as "Ready" via `kubectl`?

parts/kubelet-restart.sh
```diff
@@ -0,0 +1,10 @@
...
+if [ "$ready" = "Unknown" ]; then
+  systemctl restart kubelet
+fi
```
seanknox: What happens if the master doesn't response?

seanknox:
@dmitsh Could use some overview of the architecture behind these checks.

parts/kuberneteskubelet.service
dmitsh: Yes, the script will first be executed 40 seconds after the kubelet service is up and running.
After that it will run every 20 seconds

parts/kubelet-restart.sh
dmitsh: Correct, I didn't want to manipulate kubelet before installation is completed

parts/kubelet-restart.sh
dmitsh: Correct, this command shows kubelet status running on the local node

parts/kubelet-restart.sh
dmitsh: Do you mean what happens if kubectl won't run?

seanknox:
Closed in favor of using Dmitry's TCP retry fix in #1032

PR #957, opened 2017-07-10
Agent Pools Only

### LINKED ISSUE #900: Agent Pools Only
### Overview

We would like a way to generate the ARM templating for agent pools only. We will need to pass in a Kubernetes master API endpoint at runtime. We will also need to pass in TLS certificate information at run time as well.

### Success Criteria

 - As an ACS engine user I can generate arm templates that will deploy an arbitrary agent pool independently of a master infrastructure.
 - As an ACS engine user I can conveniently pass in TLS certificate data, as well as a publicly resolvable API endpoint for an existing master. The new agent pools will automatically be configured to work with the existing master infrastructure.
 - As an ACS engine user I can easily find documentation on how to achieve this
 - As an ACS engine developer I can easily find and run tests that ensure this functionality works

kris-nova:
This feature is aimed at allowing users of `acs-engine` a unique ability into deploying Kubernetes nodes (configured to their liking) against an arbitrary Kubernetes control plane.

# Problem Space

- We needed to create ARM templates for agent pools only, using a unique data model
- The model is unique, in that not other models accurately capture what we need to represent
- As the feature set grows we want the model to scale independently, and thus the underlying mapping into an ARM template

# Technical Decisions

- By keeping the template mapping in it's own directory we can match the mapping logic to this model, and keep the ARM template definitions explicit
- We remove complexity in adding more templating functions to the existing codepath, and introduce confidence as we have removed the potential points of failure in mapping configuration through to an ARM template

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/957)

Closes #900


pkg/api/vlabs/types.go
```diff
@@ -37,6 +37,11 @@ type Properties struct {
 	WindowsProfile          *WindowsProfile          `json:"windowsProfile,omitempty"`
 	ServicePrincipalProfile *ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
 	CertificateProfile      *CertificateProfile      `json:"certificateProfile,omitempty"`
+	CustomProfile           *CustomProfile           `json:"customProfile,omitempty"`
+}
+
+type CustomProfile struct {
```
JackQuincy: Based on our discussion I thought kubernetesagentpool was completely separate from vlabs? this little bit is confusing me

pkg/api/kubernetesagentpool/types.go
```diff
@@ -0,0 +1,111 @@
+package kubernetesagentpool
+
+const (
+	APIVersion = "kubernetesagentpool"
```
amanohar: I though this is the internal/in memory API model just like github.com/Azure/acs-engine/pkg/api/types.go. Public/version API model is converted to internal model that ACS Engine then uses to generate the template.

Versioning and validation is done on the public API model (and the validation code is also called by the service).

kris-nova: This is still a WIP (Work in progress) 😄

.gitignore
```diff
@@ -10,7 +10,8 @@ test/acs-engine-test/acs-engine-test
 .editorconfig
 _dist/
 bin/
-
+dev.json
```
brendandburns: Remove these.

pkg/api/apiloader.go
```diff
@@ -3,13 +3,12 @@ package api
...
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-
 	"github.com/Azure/acs-engine/pkg/api/v20160330"
 	"github.com/Azure/acs-engine/pkg/api/v20160930"
 	"github.com/Azure/acs-engine/pkg/api/v20170131"
 	"github.com/Azure/acs-engine/pkg/api/v20170701"
 	"github.com/Azure/acs-engine/pkg/api/vlabs"
+	"io/ioutil"
```
brendandburns: I don't think this should move down here, keep standard libraries separate from app libraries.

parts/kubernetes/agentpool/kubelet.service
```diff
@@ -0,0 +1,53 @@
+[Unit]
```
brendandburns: An equivalent file for a kubelet.service already exists here:

https://github.com/Azure/acs-engine/blob/master/parts/kuberneteskubelet.service

Let's keep this centralized so that we aren't continually flipping flags in multiple places.

pkg/acsengine/engine.go
```diff
@@ -487,12 +487,16 @@ func getParameters(cs *api.ContainerService, isClassicMode bool) (map[string]int
...
 }
 
+func AddValue(m map[string]interface{}, k string, v interface{}) { addValue(m, k, v) }
 func addValue(m map[string]interface{}, k string, v interface{}) {
 	m[k] = map[string]interface{}{
 		"value": v,
 	}
 }
 
+func AddSecret(m map[string]interface{}, k string, v interface{}, encode bool) {
```
brendandburns: As above wrt to externalization.

pkg/acsengine/engine.go
```diff
@@ -487,12 +487,16 @@ func getParameters(cs *api.ContainerService, isClassicMode bool) (map[string]int
 	return parametersMap, nil
 }
 
+func AddValue(m map[string]interface{}, k string, v interface{}) { addValue(m, k, v) }
```
brendandburns: Why are we externalizing this function?

(and if we need to, why not just externalize the existing function)

brendandburns:
In general, I'm a little confused by this PR. I think that a better approach would be add a `createMaster` or some-such parameter to the existing API model. If `createMaster` is false, then we will drop all of the master creation code, and substitute in pre-calculated values for certificates, ip addresses etc.

Unless I'm mistaken, that's going to result in a lot more code-reuse and will mean that we don't have to make similar changes in multiple different places.

I could definitely be wrong, or misunderstanding complexities. If so, please let me know. In general a brief high-level design in the description of this PR would be useful to understanding how it fits together with the existing ApiModels.

Thanks!

pkg/api/apiloader.go
kris-nova: Just my IDE playing tricks again.. will fix

parts/kubernetes/agentpool/kubelet.service
kris-nova: So we intentionally decided we wanted to keep the agent pool definitions separate from the rest of `/parts` to remove complexity and so we could have cleaner definitions.

The logic here is that this unit file is different than the rest of the kubernetes work, so we wanted to keep it separated.

Also this API is coupled with the parts directory `/parts/kubernetes/agentpool` so it will include those definitions and *only* those definitions while parsing. That way we can just define the pieces here explicitly and there is no need for logic in them.

Also, no switch flipping necessary - just use the other API and this is what you get

pkg/acsengine/engine.go
kris-nova: Again, another intentional design decision. We needed this exported so we could use it in the new interpolator packages. At the end of the day it's a convenience function, and I am happy to copy this into the other packages as necessary.

pkg/acsengine/engine.go
kris-nova: I didn't want to make changes to existing code, so this allowed me to call the exported function without having to change existing code. See note above, happy to change just explaining my reasoning.

Makefile
```diff
@@ -58,6 +58,11 @@ checksum:
 clean:
 	@rm -rf $(BINDIR) ./_dist
 
+dev:
+	go generate $(GOFILES)
```
kris-nova: This is just a target I made for compiling for quick iteration, I used it as a convenience and thought others would find it useful. It just compiles, and generates, without testing

examples/agentpools.json
```diff
@@ -0,0 +1,38 @@
...
+          {
+            "keyData": ""
+          }
+        ]
+      }
+    },
+    "servicePrincipalProfile": {
+      "servicePrincipalClientID": "",
+      "servicePrincipalClientSecret": ""
+    }
```
kris-nova: no

pkg/acsengine/engine.go
```diff
@@ -329,7 +329,7 @@ func GetCloudSpecConfig(location string) AzureEnvironmentSpecConfig {
 	switch GetCloudTargetEnv(location) {
 	case azureChinaCloud:
 		return AzureChinaCloudSpec
-	//TODO - add cloud specs for germany and usgov
+		//TODO - add cloud specs for germany and usgov
```
kris-nova: No idea, not my todo - just snagged it with `go fmt` can undo the change if we want

pkg/interpolator/agentpool/params.go
```diff
@@ -0,0 +1,87 @@
...
+	acsengine.AddSecret(parametersMap, "caCertificate", properties.CertificateProfile.CaCertificate, true)
+	acsengine.AddSecret(parametersMap, "caPrivateKey", properties.CertificateProfile.CaPrivateKey, true)
+	acsengine.AddSecret(parametersMap, "clientCertificate", properties.CertificateProfile.ClientCertificate, true)
+	acsengine.AddSecret(parametersMap, "clientPrivateKey", properties.CertificateProfile.ClientPrivateKey, true)
+	acsengine.AddSecret(parametersMap, "kubeConfigCertificate", properties.CertificateProfile.KubeConfigCertificate, true)
+	acsengine.AddSecret(parametersMap, "kubeConfigPrivateKey", properties.CertificateProfile.KubeConfigPrivateKey, true)
+
+	acsengine.AddValue(parametersMap, "dockerEngineDownloadRepo", cloudSpecConfig.DockerSpecConfig.DockerEngineRepo)
+	acsengine.AddValue(parametersMap, "kubernetesHyperkubeSpec", KubernetesImagebase+acsengine.KubeImages[KubernetesVersion]["hyperkube"])
+	//acsengine.AddValue(parametersMap, "kubernetesAddonManagerSpec", cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase+acsengine.KubeImages[KubernetesVersion]["addonmanager"])
```
kris-nova: They were handy to have while working on pulling agent pools params out into their own package... if I hit a bug where some data was missing I could usually just find it here, and uncomment and get it working again..

Kind of a minimalistic approach.. include the bare minimum instead of all of them (but leave the others around as we're developing)

I can take them out

examples/agentpools.json
kris-nova: ahh.. but the line below it does 😄

cmd/root.go
```diff
@@ -38,8 +38,10 @@ func NewRootCmd() *cobra.Command {
 	p := rootCmd.PersistentFlags()
 	p.BoolVar(&debug, "debug", false, "enable verbose debug logs")
 
+	genCmd := newGenerateCmd()
```
kris-nova: Done see latest commit in PR: https://github.com/Azure/acs-engine/pull/957/commits/c4a7bb6f5511ac101745d868e9da9571bfd15183

pkg/api/kubernetesagentpool/types.go
kris-nova: Okay the versioned API models are in place now

JackQuincy:
High level: I still really don't like the name agent pool. As this should never be used with a standard cluster to just deploy its agent pools.

seanknox:
> High level: I still really don't like the name agent pool. As this should never be used with a standard cluster to just deploy its agent pools.

The to deploy agent pools. Not sure the disconnect here.

JackQuincy:
A standard acs cluster can never use these templates or api models. Or any of the code path as designed right now. The naming implies that it can.

kris-nova:
@brendanburns 

> In general, I'm a little confused by this PR. I think that a better approach would be add a createMaster or some-such parameter to the existing API model. If createMaster is false, then we will drop all of the master creation code, and substitute in pre-calculated values for certificates, ip addresses etc.


We looked at that approach first and found it was a little trickier than that after trying for a few days. For instance there were a number of changes we would need to have made in the `/parts` directory for that workL
 - `kubelet.service` needs custom directives defined
 - Renaming master -> jumpbox
    - There would have been sizeable changes in each of the `.t` files in parts just changing names/flipping values over as we started coded we caught ourselves wanted to create new parts all together (hence why they are now in a new directory)
   - We also would need to turn off a lot of the parameters conditionally as well (none of the kubernetes and master parameters are needed)
   - We also would need to turn on all the jump box parameters as well (size, count, etc)
   - We also needed to build in logic for passing in values like kubernetes version, and kubernetes master endpoint

All of this would have been introducing a ton of logic into `parts`

Adding more logic and complexity seemed like more work than just taking the minimalist approach here. We talked about it for quite some time up in Redmond, and this is what the team agreed on. I hope that is cool? I am really happy with this design, and think it solves our original problem set of:

1. Needing to create agent pools only, and have control over the inputs/outputs and how they map to an ARM template
2. Gives us the ability to scale/version our own API as the feature set increases over time without trying to add more to `vlabs`

> Unless I'm mistaken, that's going to result in a lot more code-reuse and will mean that we don't have to make similar changes in multiple different places.

I think part of the intent here was to explicitly keep things separated as they are two independent concern sets. The agent pool only PR is intended to be ran in a very unique way. We don't always want changes going from one to the other... right?

Just sharing my thoughts 😄 happy to jump on a call to discuss more if you think that would help!

Cheers

pkg/interpolatorwriter/writer.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func (i *InterpolatorWriter) AddFile(name string, buffer []byte) {
+	i.otherFiles[name] = buffer
+}
+
+var InterpolatorWriterMutex sync.Mutex
+
+func (i *InterpolatorWriter) Write() error {
```
kris-nova: I am wondering if we should make this implement `io.writer` I kind of feel like we should

pkg/api/kubernetesagentpool/v20170727/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	// Kubernetes Version
+	if p.Properties.KubernetesVersion == "" {
+		return fmt.Errorf("Empty Kubernetes version")
+	}
+	if !strings.Contains(p.Properties.KubernetesVersion, ".") {
+		return fmt.Errorf("Invalid Kubernetes version")
+	}
+
+	// -------------------------------
+	// Kubernetes Version
```
sgoings: Kubernetes endpoint?

pkg/api/kubernetesagentpool/v20170727/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	if p.Properties.KubernetesEndpoint == "" {
+		return fmt.Errorf("Empty Kubernetes endpoint")
+	}
+	if !strings.Contains(p.Properties.KubernetesVersion, ".") {
```
sgoings: This check mixes up KubernetesVersion and KubernetesEndpoint as well.

pkg/api/kubernetesagentpool/v20170727/marshal.go
```diff
@@ -0,0 +1,39 @@
...
+	version := m.APIVersion
+	agentPool := &AgentPool{}
+	if err := json.Unmarshal(contents, &agentPool); err != nil {
+		return nil, "", err
+	}
+	//service, err := LoadAgentPool(contents, version)
+	return agentPool, version, nil
+}
+
+//func LoadAgentPool(contents []byte, version string) (*AgentPool, error) {
```
sgoings: Why is this commented out?

docs/agentpools.md
```diff
@@ -0,0 +1,77 @@
+# Deploying Agent Pools Only
```
sgoings: Awesome. Thanks for this doc.

Makefile
sgoings: Yeah - I get that. My suggestion:

```
dev: generate
```

or 

```
dev: build
```

pkg/acsengine/engine.go
jackfrancis: My preference would be to matriculate the existing `addValue` to `AddValue` (and then update all existing usages of `addValue` to `AddValue` accordingly). I understand the "change no existing code" directive, but this kind of change is a minor compromise to that strategy with the most long-term benefit, in my view.

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	// -------------------------------
+	// Agent pools
+	if len(p.Properties.AgentPoolProfiles) < 1 {
+		return fmt.Errorf("Missing agent pool definition")
+	}
+	for _, agentPool := range p.Properties.AgentPoolProfiles {
+		if agentPool.Count < 1 {
+			return fmt.Errorf("Must using a value greater than 0 for agent pool count")
+		}
+		if agentPool.OSDiskSizeGb != 0 && agentPool.OSDiskSizeGb < 2 {
```
sgoings: Why the `!=0` check? Additionally... if I set disk size == 2, this check would pass without error. Error message should therefore probably read:

"Must use a disk size greater than or equal to 2gb for agent pool"

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("Empty service CIDR in network profile")
+	}
+
+	// -------------------------------
+	// Certificate Profile
+	if p.Properties.CertificateProfile.APIServerCertificate == "" {
+		return fmt.Errorf("Empty API certificate")
+	}
+	if p.Properties.CertificateProfile.APIServerPrivateKey == "" {
+		return fmt.Errorf("Empty API key")
```
sgoings: "Empty API server private key"?

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
...
+		return fmt.Errorf("Empty pod CIDR in network profile")
+	}
+	if p.Properties.NetworkProfile.ServiceCIDR == "" {
+		return fmt.Errorf("Empty service CIDR in network profile")
+	}
+
+	// -------------------------------
+	// Certificate Profile
+	if p.Properties.CertificateProfile.APIServerCertificate == "" {
+		return fmt.Errorf("Empty API certificate")
```
sgoings: "Empty API server certificate"?

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if p.Properties.CertificateProfile.APIServerPrivateKey == "" {
+		return fmt.Errorf("Empty API key")
+	}
+	if p.Properties.CertificateProfile.CaCertificate == "" {
+		return fmt.Errorf("Empty CA certificate")
+	}
+	if p.Properties.CertificateProfile.CaPrivateKey == "" {
+		return fmt.Errorf("Empty CA key")
```
sgoings: "Empty CA private key"?

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if p.Properties.CertificateProfile.ClientCertificate == "" {
+		return fmt.Errorf("Empty client certificate")
+	}
+	if p.Properties.CertificateProfile.ClientPrivateKey == "" {
+		return fmt.Errorf("Empty client key")
```
sgoings: Same comment as above with adding "Private" into the message mix.

pkg/api/kubernetesagentpool/validate.go
```diff
@@ -0,0 +1,99 @@
+package kubernetesagentpool
+
+import (
+	"fmt"
+	"strings"
+)
+
+func (p *AgentPool) Validate() error {
```
sgoings: I wonder if these validations would be more helpful to the user if the exact string that would need to be set in json would be contained in the error message? (Or does it already and I just didn't know some magic?)

pkg/api/kubernetesagentpool/v20170727/convert.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/Azure/acs-engine/pkg/agentPoolOnlyApi/v20170831"
+)
+
+
+// Convertv20170831ToAgentPool will convert a Convertv20170831 object into AgentPool=
+func Convertv20170831ToAgentPool(hostedMaster *v20170831.HostedMaster) (*AgentPool, error) {
```
jackfrancis: Do we want to have two packages (`pkg/api/kubernetesagentpool` and `pkg/api/kubernetesagentpool/v20170727`) with this same exportable function?

pkg/interpolator/agentpool/params.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/Azure/acs-engine/pkg/acsengine"
+	"github.com/Azure/acs-engine/pkg/api/kubernetesagentpool"
+)
+
+const (
+	KubernetesImagebase = "gcrio.azureedge.net/google_containers/"
+)
+
+// getParameters is an unexported function that will create the parameters for the azuredeploy.parameters.json file
+// This was intentionally pulled out of acsenging so we can have a unique and decoupled grouping of parameters for agent pools only.
```
sgoings: "acsenging" --> "acsengine"

pkg/api/kubernetesagentpool/v20170727/types.go
```diff
@@ -0,0 +1,114 @@
...
+	ClientPrivateKey string `json:"clientPrivateKey,omitempty"`
+	// KubeConfigCertificate is the client certificate used for kubectl cli and signed by the CA
+	KubeConfigCertificate string `json:"kubeConfigCertificate,omitempty"`
+	// KubeConfigPrivateKey is the client private key used for kubectl cli and signed by the CA
+	KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
+}
+
+type ServicePrincipalProfile struct {
+	ClientID string `json:"servicePrincipalClientID,omitempty"`
+	Secret   string `json:"servicePrincipalClientSecret,omitempty"`
```
weinong: the service principal profile in vlab has been updated to include KeyVaultSecretRef

pkg/api/kubernetesagentpool/types.go
```diff
@@ -0,0 +1,114 @@
...
+	ClientCertificate string `json:"clientCertificate,omitempty"`
+	// ClientPrivateKey is the private key used by the client kubelet services and signed by the CA
+	ClientPrivateKey string `json:"clientPrivateKey,omitempty"`
+	// KubeConfigCertificate is the client certificate used for kubectl cli and signed by the CA
+	KubeConfigCertificate string `json:"kubeConfigCertificate,omitempty"`
+	// KubeConfigPrivateKey is the client private key used for kubectl cli and signed by the CA
+	KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
+}
+
+type ServicePrincipalProfile struct {
```
weinong: the service principal profile in vlab has been updated to include KeyVaultSecretRef

pkg/api/kubernetesagentpool/v20170727/validate.go
kris-nova: I don't understand what you are asking here :/

kris-nova: derp see it now!

pkg/acsengine/engine.go
kris-nova: I completely agree with this.. the reason I didn't make that change is because I think there is a lot more work involved once we start looking at the `acsengine` package as a library.. was hoping we could have another PR that cleaned some of that up and offered docs on how `interpolator`'s moving forward could take advantage of the freshly exported `acsengine` functions..

I know it's hacky, and will change if we really want it :) just sharing my thoughts

pkg/api/kubernetesagentpool/types.go
kris-nova: If we need it for agent pools we can add it later, the use case here is pretty slim at the moment

pkg/api/kubernetesagentpool/v20170727/convert.go
kris-nova: I removed it - it was a copy/pasta problem

pkg/api/kubernetesagentpool/v20170727/marshal.go
kris-nova: see note above on copy/pasta it's out

pkg/api/kubernetesagentpool/v20170727/types.go
kris-nova: see note above on the other comment like this

pkg/api/kubernetesagentpool/v20170727/validate.go
kris-nova: fixed thanks!

pkg/api/kubernetesagentpool/validate.go
kris-nova: It could be for sure, is that out of scope though? I feel like we could have an error-message beautifying PR at another time?

pkg/api/kubernetesagentpool/validate.go
kris-nova: You can pass 0, and it defaults downstream - I am not sure of the default off my head.. I just didn't want people trying to run k8s on a 1gb drive.

error message updated, thanks

pkg/acsengine/engine.go
jackfrancis: I think we should just deliver it in this PR without the hacky fun! Vive la `AddValue`!

pkg/api/kubernetesagentpool/validate.go
kris-nova: I mean the model only has keys, and certs for the API server.. it's a nit but happy to add it

pkg/api/kubernetesagentpool/validate.go
kris-nova: see above

pkg/api/kubernetesagentpool/validate.go
kris-nova: there are only certs and keys, private is implicit

Makefile
kris-nova: Ah I see,

So the `generate` and `build` targets to thing I didn't wan't in my dev target:

Primarily `-v`

pkg/api/kubernetesagentpool/validate.go
kris-nova: see above

pkg/acsengine/engine.go
kris-nova: Updated in https://github.com/Azure/acs-engine/pull/957/commits/5777eff342c6bddae5aa5dd0e07fc69a092ab432

parts/kubernetes/agentpool/kubelet.service
brendandburns: There are all kinds of kubelet flags below that we're going to have to flip in multiple places. This is a recipe for skew and errors. I still don't see a concrete example where we can't have a shared `kubelet.service` file.  Happy to talk this through in person if that's easier.

pkg/api/kubernetesagentpool/types.go
```diff
@@ -0,0 +1,114 @@
...
+	ClientID string `json:"servicePrincipalClientID,omitempty"`
+	Secret   string `json:"servicePrincipalClientSecret,omitempty"`
+}
+
+type JumpBoxProfile struct {
```
brendandburns: Most of this is duplicated from other types: https://github.com/Azure/acs-engine/blob/master/pkg/api/v20160330/types.go

Let's not duplicate these things unnecessarily.

pkg/api/kubernetesagentpool/v20170727/types.go
weinong: ok

pkg/api/kubernetesagentpool/types.go
weinong: make sense

pkg/api/kubernetesagentpool/v20170727/types.go
```diff
@@ -0,0 +1,114 @@
+package v20170727
```
amanohar: What is the difference between: pkg/api/kubernetesagentpool/types.go  and pkg/api/kubernetesagentpool/v20170727/types.go? Why does agent pool only have another API version? The only publicly exposed API version should be 20170831 that I added to expose HostedMaster type.

pkg/api/kubernetesagentpool/types.go
kris-nova: So while some of this is similar, it's use case is called from a very unique point during an HTTPs transaction with the ACS RP.. which means (if you look closely) there are a number of differences between the structs defined here, and the structs defined in the "public" API model..

For instance:
1. Kubernetes version
2. Kubernetes endpoints
3. TLS Certificate material

Are all unique to this request.. The underlying code that parses the new API was deliberately chosen to uncouple itself from the concept of a `ContainerService{}` as the use case here is different (we aren't creating a container service)

See my notes https://github.com/Azure/acs-engine/pull/957#issuecomment-318453486 that explain why we decided to do things this way.

TLDR; It's similar code, but it's not duplicating anything

pkg/api/kubernetesagentpool/v20170727/types.go
kris-nova: See my note here https://github.com/Azure/acs-engine/pull/957#discussion_r130453111 explaining the subtle differences in the structs, and the fact that this code is being called at a unique point during an HTTPs transaction with the ACS RP

seanknox:
@acs-bot test this please

seanknox:
Please rebase off of master—we just merged in CircleCI support for acs-engine. Rebasing will kick off a job in CI.

kris-nova:
@seanknox rebased

slack:
Rebased on master and re-pushed.

jackfrancis:
Closing this PR. Agent pools implementation are in PR: #1290 + additional fixes and was merged to master on 8/18.

unicorns

PR #967, opened 2017-07-12
Fix container start timeout in latest Windows image

### LINKED ISSUE #959: Container on Windows agent doesn't have outbound internet connection
Yes
---

Issue
---

v0.3.0
---
The container on Windows agent doesn't have outbound internet connection. Bascially, it cannot resolve the DNS.

```
PS C:\Agent> Invoke-WebRequest -Uri http://bing.com -UseBasicParsing
Invoke-WebRequest : The remote name could not be resolved: 'bing.com'
At line:1 char:1
+ Invoke-WebRequest -Uri http://bing.com -UseBasicParsing
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```

However, if I RDP to the machine, and use the docker to create the same image, it can resolve the DNS and get the result.

Kubernetes 1.6.6

**What happened**:
No outbound internet connection from container.

Outbound internet connections should be allow in the container.

Deploy a cluster with Windows agent.

JiangtianLi:
We switch to use "latest" Windows Server image and in recent Windows Update, the container start time takes more than the default 2 minutes. It causes subsequent code path to be executed to setup Windows container. This PR extends the timeout for Windows, specifically for start container here.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #959 

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/967)


seanknox:
@JiangtianLi would this fix the outbound internet issue (https://github.com/Azure/acs-engine/issues/959)?

JiangtianLi:
@seanknox  Yes, it is exactly to fix that issue.

seanknox:
I'm going to manually test myself first and then merge in.

JiangtianLi:
@seanknox Thanks. I also am doing manual test. I'll see if it is quick to add to regression test.

robbiezhang:
@JiangtianLi does it mean we need to expect up to 10 minutes to start the Windows container?

JiangtianLi:
@robbiezhang That parameter is the timeout limit according to [here](https://github.com/JiangtianLi/kubernetes/blob/k8swin166/cmd/kubelet/app/options/options.go#L197). So I guess the bigger, the safer. It may affect other scenario so that it wait longer time to timeout. To mitigate the issue, we need to choose some number in-between. If 5 minutes has worked for most scenario, we can use that.

seanknox:
My understanding of this toggle is it's to short-circuit long-running Docker operations, e.g. an image pull or starting a (Windows) container. It looks like we have a performance regression with the Windows VM we're using on agents that requires us to increase the default 2m timeout to something larger. Waiting (up to) 10m on Docker operations seems reasonable in this case (sluggish Windows performance) though I'm not sure what the other operational impacts there might be. Long-term, we should of course improve the VM performance.

PR #915, opened 2017-07-06
ignore templates.go

jsilberm:
**What this PR does / why we need it**:
Removes the need for including "templates.go" in future PRs where template files have changed.
This should remove much friction associated with community contributions to acs-engine.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

[878](https://github.com/Azure/acs-engine/issues/878)

**Special notes for your reviewer**:

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/915)

PR #973, opened 2017-07-12
ref(*): remove gitignored templates.go

seanknox:
This is a follow-up to https://github.com/Azure/acs-engine/pull/915, which added `./pkg/acsengine/templates.go` to gitignore. This removes it from the repo.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/973)


seanknox:
@dmitsh looks like we need to add a `make build` somewhere in the CI tasks. where's the right spot?

rjtsdl:
I think it failed at `make validate-generated`. I think we can deprecate 'validate-generated'. How do you think @dmitsh

colemickens:
Assuming this passes and we have a strategy for regurgitating this in the new RP, this LGTM.

JackQuincy:
@colemickens we are now generating it as part of our make build-binary command.

colemickens:
If I were better at GIFs I would insert a very celebratory jubilant GIF here. This was so painful when dealing with multi-commit PRs. This will improve life greatly. Thanks @seanknox and @JackQuincy !

seanknox:
@colemickens 
giphy-downsized

PR #989, opened 2017-07-13
[WIP] Add NVIDIA drivers for k8s linux agents

### LINKED ISSUE #828: Automatically Install NVidia Drivers on provisioning
Currently I use Kubernetes and William Buchwalters branch of ACS engine located here: https://github.com/Azure/acs-engine such that I can have the scripts automatically run and install the nvidia drivers on GPU nodes in my cluster.

We do batch machine learning training on GPU nodes in Azure, we serve ML models from GPU nodes running in Kubernetes and also host and provide a variety of GPU services as a SaaS type offering using the GPU nodes.  GPU nodes are critical for us.  William's fork is starting to fall quite a bit behind this one.  It would be really great if we could just use Azure/acs-engine mainline for provisioning clusters.

Right now this is painful because your updates and other updates fix bugs in clusters that we currently run.

Thanks!

Kubernetes

**What happened**:
I have to currently ssh into each node (we use an auto-scaler as well making this challenging).

drivers get installed as part of provisioning

wbuchwalter:
Automatically install NVIDIA drivers on linux agents for kubernetes.
Fixes #828

**To Do:**
- [x] Add scripts to install NVIDIA drivers 
- [x] Ensure we are allowed to do this from a legal perspective 
- [x] Add unit tests
- [X] Update GPU documentation

**Implementation Details**:

* There are two ways to install the NVIDIA drivers on Ubuntu:
  * By using the `*.run` file available on NVIDIA's website. This is the faster method, it takes ~2-3 minutes to install, but all the NVIDIA libraries are installed under `/usr/lib/x86_64-linux-gnu/`, which means mounting the drivers into the container can be painful and create some conflicts. A workaround is to map `/usr/lib/x86_64-linux-gnu/` to a different directory in the container, but it can **probably** still create some issues if you need to add both `/usr/lib/x86_64-linux-gnu/` and the mounted path to your `LD_LIBRARY_PATH`. [Here is an example of a template](https://gist.github.com/wbuchwalter/4f28a4b2f03f11c00265aeaa51d6fa30).

  * By using `apt` with the `ppa` repository. Much slower to install, about ~12 minutes, so not ideal when scaling a lot, but the libraries are installed under a separate directory `/usr/lib/nvidia-*`. This allows us to mount only the drivers into the container and nothing else, which should prevent any dependency/versioning issue.  [Here is an example of a template](https://gist.github.com/wbuchwalter/c066ca55fd1e9099bdc16dd9ef4865de). **This is the way I used in this PR**. I assume most people using GPUs will run jobs that are quite long (in ML we are typically talking about hours) so 12 minutes shouldn't be a big deal, and this should cause less issues.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/989)


wbuchwalter:
I updated this PR to allow for different drivers installation scripts depending on the `VMSize`.
Currently only K80 and M60 VMs will have the drivers automatically installed.
P40 and P100 will need manual installation for now.
Also at some point there might be some customization needed for the `NC24r` and `NV24r` for infiniband.

I didn't add any integration test, as this would add significant time overhead to any e2e test that would run it (~+ 12 minutes), let me know if I should add one.

PR #999, opened 2017-07-14
acs-engine CI integration with MDM

dmitsh:
Collecting metrics on acs-engine test failures and sending it to Geneva/MDM pipeline.
Metrics includes
 - test scenario name
 - region
 - error class (deployment, validation, etc.)
 - error

Once logged in MDM, it is possible to aggregate errors and create monitoring alerts

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/999)


test/acs-engine-test/report/report.go
```diff
@@ -28,52 +33,69 @@ type ReportMgr struct {
...
-		"DCOS marathon validation failed": "dcos/test.sh] marathon validation failed",
-		"DCOS marathon not added":         "dcos/test.sh] gave up waiting for marathon to be added",
-		"DCOS marathon-lb not installed":  "Failed to install marathon-lb",
-
-		"DockerCE failed to create network": "DockerCE: gave up waiting for network to be created",
-		"DockerCE failed to create service": "DockerCE: gave up waiting for service to be created",
-		"DockerCE service unreachable":      "DockerCE: gave up waiting for service to be externally reachable",
+	errors = []errorInstance{
+		{errName: "azcli run", errClass: errClassAzcli, errRegex: "_init__.py"},
+		{errName: "azcli load", errClass: errClassAzcli, errRegex: "Error loading command module"},
```
seanknox: A common error type is defined with `errorInstance`, but it didn't look like these are surfaced as an actual golang error, which seems like an antipattern. Maybe I'm missing something.

seanknox: A common error type is defined with `errorInstance`, but it didn't look like these are surfaced as an actual golang error, which seems like an antipattern. Maybe I'm missing something.

seanknox: (specifically, my thought is errors are best returned using the `error` type).

glide.yaml
```diff
@@ -39,6 +39,8 @@ import:
 - package: golang.org/x/crypto
   subpackages:
   - ssh
+- package: github.com/alexcesaro/statsd
+  version: ^2.0.0
```
seanknox: We need to register every dependency with Microsoft OSS. Is statsd 2.0.0 registered?

dmitsh: statsd is used in RP ...

seanknox: Legal rules are different for public facing projects versus internal Microsoft projects. This version was not registered with Microsoft CELA yet. I went ahead and did so: https://ossmsft.visualstudio.com/_oss?requestId=41301&_a=progress

In the future you'll need to register any new OSS component or any OSS version change here: https://ossmsft.visualstudio.com/_oss (requires MSFT login).

test/acs-engine-test/report/report.go
jackfrancis: If the `errorInstance` type (and the usage of it in array form as `errors`) is fulfilling a MDM specification, please mention that in the type definition! (I'm assuming that's what's going on there)

seanknox:
To add some color to this PR:
1. We currently run full suite of regression tests twice: once as a full suite, and again individually but called out as individual feature regressions. This means we're duplicating tests at the cost of significant time and against CI subscription limits.
1. Dmitry and I talked about ways to run the same full regression suite only once, while still providing granularity and notification to the team on issues. That's what this PR enables: we can deployment data generated via CI to Geneva/MDM, which allows us to easily visualize problems as well as setup automatically notification, including to on-call peeps via ICM.
1. Additionally, we'll be able to delete a bunch of individual feature tests from the codebase and maintain them in a single spot (the full regression suite).

Very excited about this one. Thanks @dmitsh!

test/acs-engine-test/report/report.go
dmitsh: addressed these comments

test/acs-engine-test/report/report.go
```diff
@@ -28,52 +33,69 @@ type ReportMgr struct {
...
+		{name: "DockerCE failed to create service", class: errClassValidation, regex: "DockerCE: gave up waiting for service to be created"},
+		{name: "DockerCE service unreachable", class: errClassValidation, regex: "DockerCE: gave up waiting for service to be externally reachable"},
 	}
 }
 
-func New(jobName string, buildNum int, nDeploys int) *ReportMgr {
+func New(metricsNS, metricName, jobName string, buildNum int, nDeploys int) *ReportMgr {
 	h := &ReportMgr{}
+	h.metricsNS = metricsNS
+	h.metricName = metricName
```
amanohar: @dmitsh what would be an example of metricName that will be used in ACS Engine? E,g, in the service we have ApiExpectedFailures, ApiDuration, etc.?

test/acs-engine-test/metrics/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+		statsd.Network("udp"),
+		statsd.ErrorHandler(
+			func(err error) {
+				fmt.Println(err.Error())
+			}))
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+	client.Count(string(data), count)
```
amanohar: A bit more description of how metrics logging is going to work will be helpful (asking because I have done MDM metrics logging in a couple of Azure services). What does count refer to here?

test/acs-engine-test/report/report.go
```diff
@@ -97,14 +119,16 @@ func (h *ReportMgr) Copy() *ReportMgr {
 	return n
 }
 
-func (h *ReportMgr) Process(txt, location string) {
-	for key, regex := range errorRegexpMap {
-		if match, _ := regexp.MatchString(regex, txt); match {
-			h.addFailure(key, map[string]int{location: 1})
+func (h *ReportMgr) Process(txt, testName, location string) {
+	for _, logErr := range logErrors {
+		if match, _ := regexp.MatchString(logErr.regex, txt); match {
```
amanohar: It's best to log metrics error codes instead of making the determination using string match as strings can change.

test/acs-engine-test/metrics/metrics.go
seanknox: Good call out—believe Dmitry and Jack F are going to pair on writing some unit tests which would be a good opportunity to make naming a bit clearer.

test/acs-engine-test/report/report.go
dmitsh: I anticipate two metrics: TestError and TestDuration.

test/acs-engine-test/report/report.go
dmitsh: Deployment errors are presented as text. Validation tests are controlled by us, so we will know when the text is changing.

test/acs-engine-test/metrics/metrics.go
dmitsh: For TestError the count represents number of errors. For TestDuration it is duration in seconds.

test/acs-engine-test/report/report.go
dmitsh: But I agree, there are different ways to retrieve error codes, for example sending REST API request to ARM to query deployment status.

test/acs-engine-test/metrics/metrics.go
amanohar: We should discuss more on the design if you are planning to use this as a generic class for ACS RP. Plus, I am not sure if MDM metrics code for RP needs to live in ACS Engine. IMO that's out of scope of what ACS Engine should do.

Also please check the (old) ACS RP code to see examples of how metrics are logged and used currently (CRP metrics logging also has some good examples).

test/acs-engine-test/report/report.go
amanohar: Irrespective (even know when text is changing) no MDM metric should be dependent on random error strings to determine failure/success. This method is error prone. and non deterministic and I wouldn't want it to be used in the RP.

test/acs-engine-test/metrics/metrics.go
dmitsh: Sure, we can have a discussion if ACSRP is interested in adopting this class.

test/acs-engine-test/report/report.go
dmitsh: Log files serve multiple purposes, and the error strings aren't random, but rather unique.

amanohar: And error strings can be error prone hence the need for error codes.

PR #1005, opened 2017-07-15
feat(custom hyperkube): support customHyperkubeImage in kubernetesConfig

### LINKED ISSUE #1003: vlabs should support customHyperkubeImage specified in kubernetesConfig

We should support allowing the user to specify a customHyperkubeImage in the apimodel.

This is not something I expect to ever be pulled into a versioned API, but it will make it easier to support easy `./acs-engine deploy ...` scenarios where the user wants to override the custom hyperkube image.

This would also be one of the things necessary to deprecate/remove `./test/deploy.sh` in favor of `./acs-engine deploy`, which I'm eager to do.

colemickens:
**What this PR does / why we need it**: Fixes #1003 by adding a `customHyperkubeImage` field to `kubenetesConfig` the apimodel (vlabs only). This is not something I expect to end up in a released/supported api-version but it's very useful for co-development of features with acs-engine, and allowing advanced users to test custom Kubernetes builds.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1003

**Special notes for your reviewer**: n/a

**Release note**:

```release-note
feat(custom hyperkube): support customHyperkubeImage in kubernetesConfig
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1005)


examples/managed-identity/kubernetes-msi.json
```diff
@@ -42,8 +43,8 @@
       }
     },
     "servicePrincipalProfile": {
-      "servicePrincipalClientID": "",
-      "servicePrincipalClientSecret": ""
+      "servicePrincipalClientID": "msi",
```
amanohar: Is this an example or can you now actually specify msi as a value? If yes, seems confusing to me that id can be a real value or a string. If useManagedIdentity is true then this value does not need to be set.

seanknox: I like the flow Anubhuti purposes and in general want to slim down cluster configs wherever we can...

colemickens: Actually, this is easier to do than I thought, and I agree with it in principle. I'll do that right now as well.

colemickens:
@seanknox we already have tests that validate back/forth serialization/deserialization of fields, as I understand it. This again ties into the validation Issue I filed. I'll add a test for this field in vlabs serialization at least. (But it would be better if I were adding a unit test for validation of the internal model. Imagine this field was backported to other APIs, I'd be groaning about having to unit test 5-6-7 different validation code paths...)

seanknox:
> (But it would be better if I were adding a unit test for validation of the internal model. Imagine this field was backported to other APIs, I'd be groaning about having to unit test 5-6-7 different validation code paths...)

Ouch, I hear that. Having one place to define acceptable properties sounds like some major rearchitecting. In any case, thanks for whatever added coverage you can provide.

examples/managed-identity/kubernetes-msi.json
colemickens: I'm fixing this in a different PR, since this is unrelated to this PR.

colemickens:
Few things; ready for another look, please.

1. I added a test for the field, as requested, @seanknox.

2. @amanohar I punted on fixing the fields in this PR since it's not really related to this PR and is more of a bugfix of the original addition. I will fix it in the other PR or in a third follow-up.

3. I've changed the second commit a fair amount. Please take a look at it. The commit message describes why I made the choices I made about the test configurations. (cc: @dmitsh as FYI for the test config changes.)

colemickens:
Two more notes:

1. I split the changes to the MSI test config out of this PR. I'll send another PR that stacks on this that also includes the change to `servicePrincipalProfile` that @amanohar brought up.

2. @amanohar I did undo the places where I had put `msi` into the SPP, at least. Like I said, will followup with fuller fix.

colemickens:
To be super clear, this is ready for review as well. cc: @seanknox @amanohar @jackfrancis @dmitsh

pkg/acsengine/engine.go
```diff
@@ -378,6 +378,12 @@ func getParameters(cs *api.ContainerService, isClassicMode bool) (map[string]int
 	// Kubernetes Parameters
 	if properties.OrchestratorProfile.OrchestratorType == api.Kubernetes {
 		KubernetesVersion := properties.OrchestratorProfile.OrchestratorVersion
+
+		kubernetesHyperkubeSpec := properties.OrchestratorProfile.KubernetesConfig.KubernetesImageBase + KubeImages[KubernetesVersion]["hyperkube"]
+		if cs.Properties.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage != "" {
```
jackfrancis: is there a reason why we're dereferencing `cs.Properties` here (and the line below) instead of `properties`?

colemickens: Hm. Nope. I'll change, I'd prefer to be consistent. Thanks.

PR #1006, opened 2017-07-15
fix(autodeploy+msi): don't autogen creds when using msi

### LINKED ISSUE #1004: autodeploy should skip credential creation when UseManagedIdentity is true
This is an **Issue**.

Right now, if the user specifies `kubernetesConfig->useManagedIdentity: true` then we still create credentials unnecessarily. We should be smarter.

colemickens:
**What this PR does / why we need it**: Fixes #1004 by skipping automatic credential creation when `useManagedIdentity` is `true` in the apimodel.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1004

**Special notes for your reviewer**: n/a

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1006)


cmd/deploy.go
```diff
@@ -177,31 +177,43 @@ func (dc *deployCmd) validate(cmd *cobra.Command, args []string) {
...
 		}
 
-		log.Warnln("apimodel: ServicePrincipalProfile was empty, creating application...")
+		if dc.containerService.Properties.OrchestratorProfile != nil &&
+			dc.containerService.Properties.OrchestratorProfile.KubernetesConfig != nil &&
+			dc.containerService.Properties.OrchestratorProfile.KubernetesConfig.UseManagedIdentity {
+			log.Warnln("apimodel: ServicePrincipalProfile was empty, but user is using msi. Skipping ServicePrincipal creation.")
+
+			dc.containerService.Properties.ServicePrincipalProfile = &api.ServicePrincipalProfile{
+				ClientID: "msi",
```
amanohar: Added a comment in the other PR. I don't think ServicePrincipalProfile fields should accept "msi" as a valid value. It confusing from end user and documentation standpoint.

colemickens:
I can add at least this unit test:
- confirm that for a given "incomplete" api-model (for fields we expect to auto-fill) that after completing the setup, the resulting model is valid (but that's going to impact the comments in the other PR...)

cmd/deploy.go
colemickens: Let's discuss it over there.

colemickens:
So, actually, I'm going to argue a basic sanity test for auto-field-population is separate from what I'm doing here. Further, the change to remove the SPP when in MSI mode corresponds with the changes I'm making to test the auto-deploy code, where I want to validate that creds are skipped when MSI is enabled.

So I'm going to send the auto-deploy test (with specific test for this case here) in a follow-up PR.

colemickens:
Sorry, changing my mind a lot. I'm going to make the SPP optional and add the auto-deploy test (with specific test for MSI creds/no-creds). Will update PR shortly.

cmd/deploy.go
colemickens: Removed this. Made it optional in a later commit.

colemickens:
@acs-bot test this please

seanknox:
No worries. Flag @dmitsh @jackfrancis and others when you're ready for review.

colemickens:
It's ready for review, should've been more clear. cc: @dmitsh @jackfrancis

colemickens:
@jackfrancis The field itself in the apimodel is bool. I suspect you're looking at the templatized string... it's outputting a string value, but it's still a bool if you look closer.

colemickens:
I'm going to leave this sit since it includes commits for changes that @amanohar requested.

colemickens:
Discussed in person with @amanohar but to be clear, I've done these things to "remove" SPP in MSI mode:
1. Made it not required when useManagedIdentity is true.
2. Made it so that the parameters do not include any SP credentials in the user-facing portion. They're simply not listed, no default, nothing. This is because we presumably want to transition to completely removing those things.
3. Internal to the ARM template, in the master variables, we hoist the SP creds into a variable which is then used for actual resources. There, if MSI is enabled, we still just hardcode to `msi/msi` rather than trying to remove them downstream. Frankly, there isn't a good way to do it - even if we made the input to the CSE script conditional, we can't really just omit the SP creds since they're positional arguments. So internally msi/msi will still be placed, but this is more or less a black box to the user, and so is probably fine.

PR #1014, opened 2017-07-17
add anti-affinity to kube-dns deployment

lachie83:
Currently kube-dns pods could be placed on the same node. This change adds anti-affinity to prefer that pods be scheduled to different nodes that match the label k8s-app=kube-dns

Note that this change only is supported in Kubernetes 1.6 and beyond

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1014)


seanknox:
@acs-bot test this please

PR #1039, opened 2017-07-19
fix using user supplied serviceprinciple in apimodel

lachie83:
Fixes the use of user supplied SP credentials in the API model. If supplied in apimodel this is current experience
```
./acs-engine deploy --subscription-id xxx --location eastus --resource-group xxx --api-model examples/classic/cluster.json
FATA[0005] apimodel invalid: ServicePrincipalProfile is missing either the clientid or secret.
```

```
    "servicePrincipalProfile": {
      "servicePrincipalClientID": "xxxx",
      "servicePrincipalClientSecret": "xxxx"
    }
```

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1039)

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -181,36 +181,41 @@ func autofillApimodel(dc *deployCmd) {
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
 		if spp != nil && !(spp.ClientID == "" && spp.Secret == "") {
-			log.Fatal("apimodel invalid: ServicePrincipalProfile is missing either the clientid or secret.")
-		}
+			log.Info("using user supplied ServicePrincipalProfile in apimodel.")
 
-		log.Warnln("apimodel: ServicePrincipalProfile was empty, creating application...")
-
-		// TODO: consider caching the creds here so they persist between subsequent runs of 'deploy'
-		appName := dc.containerService.Properties.MasterProfile.DNSPrefix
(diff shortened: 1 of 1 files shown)


cmd/deploy.go
```diff
@@ -181,36 +181,41 @@ func autofillApimodel(dc *deployCmd) {
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
 		if spp != nil && !(spp.ClientID == "" && spp.Secret == "") {
-			log.Fatal("apimodel invalid: ServicePrincipalProfile is missing either the clientid or secret.")
-		}
+			log.Info("using user supplied ServicePrincipalProfile in apimodel.")
 
-		log.Warnln("apimodel: ServicePrincipalProfile was empty, creating application...")
+			dc.containerService.Properties.ServicePrincipalProfile = &api.ServicePrincipalProfile{
```
colemickens: I don't think this is needed. Should be sufficient to just fix the condition checks.

colemickens: The point of the conditional on line 183 is to check if only ONE of the two values is set. It looks like it one of my internal refactors when writing the regressing PR removed the original check. It was originally:

- if either is not set {
   - if one is set { throw a more specific error }
- }

seanknox:
@acs-bot test this please

lachie83:
Spoken to @colemickens. He's going to implement a fix. Closing

Issue #1032, opened 2017-07-18
High TCP timeout and retry values


seanknox:
**Is this a request for help?**:

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

Because the Kubelet does not perform application-layer heartbeating, the only way it knows the connection has been lost is when the kernel decides so. As mentioned above, Kubernetes does configure keepalives but these aren't at play here since they are only sent after 2 hours of inactivity by default (`net.ipv4.tcp_keepalive_time`).

Instead, TCP retransmissions account for the 15 minute delay. The default value for `net.ipv4.tcp_retries2` means that with exponential backoff, it takes ≈15m25s before the kernel kills the connection. Kubernetes places no timeout on its HTTP requests so the "update node status" request hangs for this time. When the 15 minutes expire, all the requests fail and Kubernetes reconnects. This situation is not entirely specific to the use of an ELB – if a directly-connected apiserver becomes partitioned or dies ungracefully, it will take 15 minutes before the Kubelet attempts to reconnect.

## AC
- Lower `net.ipv4.tcp_keepalive_time` and `net.ipv4.tcp_retries2` on agents to more aggressive values.
- Important: research what these should be

https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731

narasi:
Hello,

Just a concerned soul who is facing the `NodeNotReady` issue intermittently on our Kubernetes cluster ( v1.5.3)

Since the upstream issue is fixed, do we have a fix for this yet? (apart from the partial mitigation)

PR #1044, opened 2017-07-19
set net.ipv4.tcp_retries2=8 to mitigate broken tcp connections

dmitsh:
Configure TCP on the nodes to reduce number of synack_retries.
This will mitigate "broken" tcp connections, resulting with agent nodes being in "NotReady" state for 15-17 minutes. 
This PR addresses issue #1032

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1044)


seanknox:
@dmitsh What were your findings with this enabled?

dmitsh:
@acs-bot test this please

dmitsh:
With this change the node goes into "NotReady" state for 75 seconds, then gets back to "Ready".

lachie83:
Can we document this somewhere? I don't want this getting lost in the ether. Also reference the upstream issue.

seanknox:
Discussed in standup that because this is such a significant change that we'll need to monitor for a bit, we're going to merge this after we cut the v0.4.0 release.

seanknox:
75s is much better than 17m. :) Agree that we need to document this somewhere in `docs/` and then this should be good to go.

dmitsh:
I confirmed that the attached disk hasn't been offloaded within 75 seconds.

seanknox:
What does "offloaded + disk" mean in this case? Lachie and I were wondering.

dmitsh:
I was using Anthony's terminology. By "offloaded + disk" I meant that the pod hasn't been rescheduled to run on another node, and therefore the disk hasn't been detached and reattached to another node.

docs/kubernetes.md
```diff
@@ -197,16 +197,25 @@ read and **write** permissions to the target Subscription.
...
+While [Managed disks](../examples/disks-managed/README.md) are supported for the node OS disks, they are currently not supported for persistent volumes. See https://github.com/kubernetes/kubernetes/pull/46360 for details.
+
+## Known issues and mitigations
+
+### Node "NotReady" due to lost TCP connection
+
+Nodes might appear in the "NotReady" state for approx. 15 minutes if master stops receiving updates from agents.
+This is a known upstream kubernetes [issue #41916](https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731). This fixing PR is currently under review.
+
+To mitigate this problem in ACS-Engine, we implemented the fix in the kubelet service by setting **net.ipv4.tcp_retries2=8** prior to starting the kubelet.
```
seanknox: Change to: 

> ACS-Engine partially mitigates this issue on Linux by detecting dead TCP connections more quickly via **net.ipv4.tcp_retries2=8**.

parts/kuberneteskubelet.service
```diff
@@ -11,6 +11,7 @@ ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
 ExecStartPre=/bin/mkdir -p /var/lib/kubelet
 ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
 ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
+ExecStartPre=/sbin/sysctl -w net.ipv4.tcp_retries2=8
```
seanknox: Can you add this comment above this line:
```
# This is a partial workaround to this upstream Kubernetes issue: https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731
ExecStartPre=/sbin/sysctl -w net.ipv4.tcp_retries2=8
```

lachie83:
Can you fix the couple of items @seanknox suggested then I think we are good to go

dmitsh:
done

lachie83:
LGTM

Issue #771, opened 2017-06-14
Support AAD authentication to the Kubernetes API


seanknox:
AAD authentication support just landed in Kubernetes upstream. We should support using AAD to authenticate to the Kubernetes API. This is a meta issue to track this feature work.

- [x] AAD authentication support in Kubernetes upstream and scheduled Kubernetes v1.7 release
- [x] Kubernetes v1.7.0 release
- [ ] acs-engine provisions Azure authentication apps, one for client (kubectl) and one for server (apiserver itself)
- [ ] acs-engine configures apiserver with options required for AAD
- [ ] acs-engine configures kubeconfig to use Azure authentication provider

Step by step details located here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/plugin/pkg/client/auth/azure/README.md

colemickens:
If it's helpful, here's some code that can do the Application creation: https://github.com/colemickens/azkube/blob/master/util/ad.go.

ACS-Engine will need to generate two Apps and then configure them appropriately for the client/server behaviors (I don't have code for that latter part but @lachie83 has done it manually and the manifest update should be straightforward.)

One of these also have to be round-tripped into the VM similar to the SP creds that are plumbed in today. This may require extending the ServicePrincipalProfile section even further, or it may be the push to start having ACS-Engine stage everything in KeyVault and then have a bootstrap service on the VM that can use MSI to pull down all of the assets it thinks it needs to pull down.

seanknox:
Awesome, and thanks for calling out the AAD app creation step. Added that work to our internal tracker.

PR #918, opened 2017-07-06
Activate k8s rbac

Globegitter:
**What this PR does / why we need it**:
Activates RBAC - this seems to be the way forward with k8s.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

Partially fixes #680 as it enables RBAC per default but does not help for upgrading existing clusters.

**Special notes for your reviewer**:
This haven't spent too much time to check that this is the most ideal way to set RBAC up, but it gets it working for now in our test cluster and thought I would share that work.

This depends on #909 which upgrades heapster to v1.4.0

I tested this on Azure on northeurope with a few of our services and seems to be all working.

##
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/918)

PR #1049, opened 2017-07-20
AAD integration support for Kuberenetes

karataliu:
Add AAD integration support for Kubernetes

**Which issue this PR fixes**
Issue #771
Addresses the following:
- acs-engine configures apiserver with options required for AAD
- acs-engine configures kubeconfig to use Azure authentication provider

Please refer to new added `docs/kubernetes.aad.md` for usage details.

Related upstream PR: https://github.com/kubernetes/kubernetes/pull/43987
@cosmincojocar

**Release note**:

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1049)


seanknox:
@acs-bot test this please

colemickens:
Wow, this is great; thank you!

Regarding the apimodel changes...There's no "enabled" field. It's implicit by nature of the other fields being populated... but here's what I'm thinking about:

* ServicePrincipalProfile was required since inception because we made it a hard pre-req for the cluster
* Thus, we chose to make auto-deploy populate those fields automatically. (Unless `useManagedIdentity`)

We can't use that heuristic with `aadProfile` unless we want to start requiring it absolutely. Instead, I'm not sure of a good way of determining in ACS-Engine if we want to try auto-populate those fields..

At the same time `aadProfile->enabled` seems overly verbose, but I don't think a command-line flag is appropriate either, given the other consumers that may want to use this functionality.

Any thoughts/ideas?

colemickens:
Didn't want to block on the above, code change looks good. Docs look great. One other comment: 

AAD AuthN + AllowAll is a LARGE foot-gun. One that we might want to disallow. We may want to gate this on Lachie's PR. Maybe we can use a bool to control both RBAC+AAD integration together and solve the other thing I mentioned ^, *and* avoid the foot-gun of AllowAll.

karataliu:
Thanks for reviewing.
- For auto populating:
Maybe one option is to have something like:
  ```
  "kubernetesConfig": { 
    "enableAad": true 
  }
  ```
  It would control the auto registration of two new applications, and also be the switch for whether we should look at 'aadProfile' section at all. And we can consider a single property to control both AAD and RBAC.

- For default AllowAll
Yes, I think there's more we can do once having RBAC. 
In order to be secure, the AAD integration should always be enabled together with RBAC. We might add a new parameter 'clusterAdmins' there, and allow users to pass in a list of initial AAD user identities to be configured as cluster admins.

lachie83:
@karataliu This is fantastic. I would like to review this as well given that I've been working on enabling RBAC. I was thinking that we might nest the enable AAD under the enableRBAC or something like that. Take a look at #918 at what we are introducing in the apimodel in this PR.

seanknox:
giphy 3

Yeah, this is great! Will take a look at this today, though I'm happy if Cole and Lachie are happy.

kyliel:
@lachie83 How's going on for the review? Anything else should Dong take care of? Thanks.

lachie83:
@kyliel -- Going to get this merged before I review this one

seanknox:
Hi @karataliu, please rebase off of master—we just merged in CircleCI support. Rebasing will kick off a job in CI (which you'll be able to see).

karataliu:
@seanknox Rebased version pushed. Thanks.

docs/README.md
```diff
@@ -12,6 +12,7 @@ This cluster definition examples demonstrate how to create a customized Docker E
 * [Kubernetes Walkthrough](kubernetes.md) - shows how to create a Kubernetes enabled Docker cluster on Azure
```
seanknox: Linter checks failed, looks like a typo and need to run `gofmt`.

karataliu: fixed

parts/kubernetesmaster-kube-apiserver.yaml
```diff
@@ -29,6 +29,9 @@ spec:
         - "--tls-private-key-file=/etc/kubernetes/certs/apiserver.key"
         - "--client-ca-file=/etc/kubernetes/certs/ca.crt"
         - "--service-account-key-file=/etc/kubernetes/certs/apiserver.key"
+        - "--oidc-client-id=<oidcClientId>"
+        - "--oidc-issuer-url=<oidcIssuerUrl>"
+        - "--oidc-username-claim=oid"
```
jackfrancis: It looks like the way that these three `--oidc*` options are defined, and populated below in `kubernetesmastercustomdata.yml`, doesn't accommodate an api model without an AAD profile. Am I incorrect? We want to ensure that this additional configuration has no effect on existing users/api models.

lachie83:
Now that RBAC is merged I will take a look at this one.

pkg/api/convertertoapi.go
```diff
@@ -382,6 +382,11 @@ func convertVLabsProperties(vlabs *vlabs.Properties, api *Properties) {
 		api.CertificateProfile = &CertificateProfile{}
 		convertVLabsCertificateProfile(vlabs.CertificateProfile, api.CertificateProfile)
 	}
+
```
jackfrancis: I think we need all this conversion logic in `converterfromapi.go` in the reverse direction as well.

parts/kubernetesmaster-kube-apiserver.yaml
karataliu: oidc Auth is enabled when both 'oidc-client-id' and 'oidc-issuer-url' are not null:
https://github.com/kubernetes/kubernetes/blob/b78fc209a49049312ae038ef0235141f6c6eb929/pkg/kubeapiserver/authenticator/config.go#L152

For api model without AAD profile, those value would be empty, so that oidc auth is disabled:
https://github.com/Azure/acs-engine/pull/1049/files#diff-b7e839779245b296d836bd5663a27941R272

pkg/api/convertertoapi.go
karataliu: Good catch. Fixed.

pkg/api/vlabs/types.go
```diff
@@ -38,6 +38,7 @@ type Properties struct {
 	WindowsProfile          *WindowsProfile          `json:"windowsProfile,omitempty"`
 	ServicePrincipalProfile *ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
 	CertificateProfile      *CertificateProfile      `json:"certificateProfile,omitempty"`
+	AadProfile              *AadProfile              `json:"aadProfile,omitempty"`
```
amanohar: To keep this consistent with the field naming scheme in the API Model e.g: APIServerCertificate, SSH, DNSPrefix - please rename to: "AADProfile"

karataliu: fixed.

parts/kubernetesmaster-kube-apiserver.yaml
jackfrancis: I think we'd prefer that these options be entirely omitted at runtime if not enabled. This is a more explicit approach, and also ensures that if a future bug is introduced that breaks variable injection we are more able to easily identify `--oidc-client-id=` (empty value) as a symptom of something wrong.

Here is an example in a recent PR of an option line being removed entirely during template generation if a given condition isn't met:

https://github.com/Azure/acs-engine/commit/f5c988a5031de8ce39009351d6801a0bd8b95eeb#diff-b7e839779245b296d836bd5663a27941R237

I hope that makes sense!

docs/kubernetes.aad.md
```diff
@@ -0,0 +1,100 @@
+# Microsoft Azure Container Service Engine - Kubernetes AAD integration Walkthrough
```
lachie83: Docs directory structure has changed. I suggest this is moved to docs/kubernetes/aad.md or features.md.

docs/kubernetes.aad.md
```diff
@@ -0,0 +1,100 @@
...
+- `tenantID`      : (optional) the `AAD tenant`'s ID. If not specified, will use the tenant of the deployment subscription.
+
+After template generation, the local generated kubeconfig file (`_output/<instance>/kubeconfig/kubeconfig.<location>.json`) will have the default user using AAD.
+Initially it isn't assoicated with any AAD user yet. To get started, try any kubectl command (like `kubectl get pods`), and you'll be prompted to the device login process. After login, you will be able to operate the cluster using your AAD identity.
+
+### Note
+Please note that as of Kubernetes 1.7, the default is authorization mode is `AlwaysAllow`, which means any authenticated user have full access of the cluster.
+OpenID Connect is an authentication protocol responsible for identify users only, so initally all active accounts under the tenant will be able to login and have full admin privilege of the cluster.
+
+In this case you may want to also turn on RBAC for your cluster. Following instructions are for turnning on RBAC together with AAD integration:
```
lachie83: Link to RBAC documentation in acs-engine?

docs/kubernetes.aad.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Since we use AAD object ID as OpenID Connect identity.
+    You'll first need to figure out your account's object ID. Here is how to do it using Azure Portal:
+    Navigate to `Azure Active Directory` -> `Users and groups` -> `All users`. And choose your account in right pannel. Switch to `Manage` -> `Profile`, and you can see the `Object ID` property.
+2. Figure out your user name. The user name would be in form of `IssuerUrl#ObjectID` format.
+    You can navigate to `https://login.microsoftonline.com/{tenantid}/.well-known/openid-configuration`, and find the `IssuerUrl` under `issuer` property.
+3. Add your account as admin role
+```
+kubectl create clusterrolebinding aad-default-cluster-admin-binding --clusterrole=cluster-admin --user={UserName}
```
lachie83: Please provide an example with a real username alongside the example

docs/kubernetes.aad.md
```diff
@@ -0,0 +1,100 @@
...
+```
+4. Turn on RBAC on master nodes.
```
lachie83: Link to the acs-engine rbac docs. This is now irrelevant

parts/kubernetesmaster-kube-apiserver.yaml
lachie83: Is OID the best field to pin the username to?

pkg/acsengine/engine.go
```diff
@@ -297,8 +297,25 @@ func GenerateKubeConfig(properties *api.Properties, location string) (string, er
 	kubeconfig = strings.Replace(kubeconfig, "{{WrapAsVerbatim \"variables('caCertificate')\"}}", base64.StdEncoding.EncodeToString([]byte(properties.CertificateProfile.CaCertificate)), -1)
 	kubeconfig = strings.Replace(kubeconfig, "{{WrapAsVerbatim \"reference(concat('Microsoft.Network/publicIPAddresses/', variables('masterPublicIPAddressName'))).dnsSettings.fqdn\"}}", FormatAzureProdFQDN(properties.MasterProfile.DNSPrefix, location), -1)
 	kubeconfig = strings.Replace(kubeconfig, "{{WrapAsVariable \"resourceGroup\"}}", properties.MasterProfile.DNSPrefix, -1)
-	kubeconfig = strings.Replace(kubeconfig, "{{WrapAsVerbatim \"variables('kubeConfigCertificate')\"}}", base64.StdEncoding.EncodeToString([]byte(properties.CertificateProfile.KubeConfigCertificate)), -1)
-	kubeconfig = strings.Replace(kubeconfig, "{{WrapAsVerbatim \"variables('kubeConfigPrivateKey')\"}}", base64.StdEncoding.EncodeToString([]byte(properties.CertificateProfile.KubeConfigPrivateKey)), -1)
+
```
lachie83: We need to restrict this functionality to only run in > 1.7 clusters.

parts/kubernetesmaster-kube-apiserver.yaml
karataliu: @jackfrancis Makes sense, fixed. Those lines with 'oidc' will get removed if 'aadProfile' is not present.

karataliu: @lachie83 
`sub` is recommended, it is bound to both user identity and the auth request, but does not have a directly mapped entry in AAD

`email` and `upn`, those are usually user email addresses, but not guaranteed to be available for every user. Based on a recent test:
For organizational users, they have 'upn', but not 'email'.
For Microsoft Account external users, they have 'email', but not 'upn'.
Service principal users have neither of them.

`oid` is the one present for all kinds of users, and have a directly mapped entry in AAD.

Refer to:
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-protocols-oauth-code

docs/kubernetes.aad.md
karataliu: Moved to docs/kubernetes/aad.md

docs/kubernetes.aad.md
karataliu: Link added.

docs/kubernetes.aad.md
karataliu: Added.

docs/kubernetes.aad.md
karataliu: I've put a link to the rbac doc. And add a note for that this section is for how to enable rbac manually.

pkg/acsengine/engine.go
karataliu: Open ID auth has been available since v1.1.0:
https://github.com/kubernetes/kubernetes/commit/6376e41850c29af4954cf5f44bda292e3ab9521f

Aad auth plugin is available for kubectl > v1.7.0, but that's the client side.

I've tested against server v1.6.0 and it works well.

Is there any other consideration that we need the restriction?

cosmincojocar:
@lachie83 @colemickens  Please have a look at this [issue](https://github.com/kubernetes/kubernetes/issues/23720) for RBAC. The OIDC plugin of apiserver supports only a fixed claim **group** which should carry a list of groups assigned to a subject. This is not really the case with AAD which typically embeds a URI to Graph API from where the groups can be fetched. 

The solution discussed in the issue is to use a separate webhook service that can accept a token review request and validate the token on behalf of apiserver. It can also fetch the groups from Graph API if the token is valid.

I'm thinking to implement such a webhook that might be deployed as a HA service or locally next to each apiserver.

karataliu:
@cosmincojocar Looks like what you mentioned is about is group membership, and the working approach is still under discussion. This PR addresses current upstream enhancement which targets at user identity only. Not groups would be assigned to the user yet.

Any more comments for this PR?

travisnielsen:
@cosmincojocar Are you sure? AAD will indeed provide group membership out-of-the-box if you set the 'groupMembershipClaims' field to 'All' in the manifest of the AAD application that represents the API server. I've used this claim for role bindings in k8s and it works. My suggestion would be to leave this in.

cosmincojocar:
@travisnielsen That's right for users which are assigned in less than 200 groups. In that case AAD will inline the groups into the token. Beyong this number, it will embed a link to Graph API. I would say that the current implementation works for most of the users. I would love to see this PR merged.

jchauncey:
@karataliu could you rebase this so I can kick off CI?

lachie83:
I'll take another pass at review of this today

karataliu:
@jchauncey rebase pushed

PR #1064, opened 2017-07-20
bumped addonmanager tag to be consistent with other releases

lachie83:
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1064)

### DIFF
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -109,7 +109,7 @@ var KubeImages = map[api.OrchestratorVersion]map[string]string{
 		"addonresizer": "addon-resizer:2.0",
 		"heapster":     "heapster:v1.4.0",
 		"dns":          "k8s-dns-kube-dns-amd64:1.14.4",
-		"addonmanager": "kube-addon-manager-amd64:v6.1.2",
+		"addonmanager": "kube-addon-manager-amd64:v6.4-beta.2",
 		"dnsmasq":      "k8s-dns-dnsmasq-amd64:1.14.4",
 		"pause":        "pause-amd64:3.0",
 		"windowszip":   "v1.7.0intwinnat.zip",


seanknox:
@acs-bot test this please

pkg/acsengine/const.go
```diff
@@ -109,7 +109,7 @@ var KubeImages = map[api.OrchestratorVersion]map[string]string{
 		"addonresizer": "addon-resizer:2.0",
 		"heapster":     "heapster:v1.4.0",
 		"dns":          "k8s-dns-kube-dns-amd64:1.14.4",
-		"addonmanager": "kube-addon-manager-amd64:v6.1.2",
+		"addonmanager": "kube-addon-manager-amd64:v6.4-beta.2",
```
JackQuincy: is there a reason we are using a beta version?

lachie83: It's the version that currently supports using kubectl 1.6 which is required for creation of RBAC resource types using addon-manager . See this for the changelog --
 https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/CHANGELOG.md#version-64-beta2--mon-june-12-2017-jeff-grafton-jgraftongooglecom. We used the 6.4 tag but which I thought would cover this beta since 6.4 hasn't actually been releases and the timestamp on 6.4 is much older than the 6.4-beta.2 -- https://gcr.io/v2/google_containers/kube-addon-manager/tags/list

JackQuincy: So we need 6.4 and it is still in beta. Is that correct? I guess this is okay, but just makes me kind of nervous to use beta versions.

lachie83:
Yes that's correct and I don't think we have any other option at this point if we would like to run this tool.

seanknox:
Kubernetes components at "beta" are already heavily tested—Deployments are one of many examples. Kubernetes beta is many more times rigorous than just about any other software project in existence.

PR #1088, opened 2017-07-24
Adding CoreOS Support for Kubernetes Agent Nodes 

uday31in:
Adding CoreOS Support for Kubernetes Agent Nodes

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1088)


examples/coreos/coreos-agent.json
```diff
@@ -0,0 +1,49 @@
...
+     "agentPoolProfiles":  [
+								 {
+									 "name":  "coreos",
+									 "count":  1,
+									 "vmSize":  "Standard_DS2_v2",
+									 "availabilityProfile":  "AvailabilitySet",
+									 "osImageOffer": "CoreOS",
+									 "osImagePublisher": "CoreOS",
+									 "osImageSKU": "Stable",
+									 "osImageVersion": "latest"
```
rjtsdl: Not sure if it is just me. The format here looks weird.

pkg/api/vlabs/types.go
```diff
@@ -305,6 +307,12 @@ func (a *AgentPoolProfile) IsLinux() bool {
 	return a.OSType == Linux
 }
 
+// IsUbuntu returns true if OSImagePublisher is Canonical
```
rjtsdl: The comment doesn't match the function implementation. I prefer `IsUbuntu`

rjtsdl: Or changename to IsCoreOS.

parts/kubernetesagentcustomdataforcoreos.yml
```diff
@@ -0,0 +1,120 @@
+#cloud-config
```
rjtsdl: Why this cannot be integrated into "kubernetesagentcustomdata.yml" ?

uday31in: CoreOS  cloudconfig  is different than normal CloudConfig. For Example, Packages are not valid config when targeting CoreOS. I also propose to keep it separate such that this will eventually Ignition based config as CoreOS is going that way.

uday31in:
@rjtsdl - thank you for reviewing the PR. I have made changes as requested. Thanks.

JackQuincy:
@anhowe Thoughts on this?

pkg/api/vlabs/types.go
```diff
@@ -195,10 +195,12 @@ type AgentPoolProfile struct {
 	DiskSizesGB         []int  `json:"diskSizesGB,omitempty"`
 	VnetSubnetID        string `json:"vnetSubnetID,omitempty"`
 	IPAddressCount      int    `json:"ipAddressCount,omitempty"`
-
+	OsImageOffer      	string `json:"osImageOffer,omitempty"`
```
amanohar: - The flattened approach to providing offer, sku, version (versus being in a struct) in agentpool is concerning. Also, not having this in LinuxProfile doesn't seem correct (and has long term impact to the API model).
- I see that masters are left to be Ubuntu which might be ok (and convenient) in the short terms but introduces complexity around interactions between agents and masters & what happens when master will be CoreOS as well.
- Also, what about testing? How do we make sure this feature doesn't fall behind in the test matrix?

pkg/api/vlabs/validate.go
```diff
@@ -136,6 +137,13 @@ func (a *AgentPoolProfile) Validate(orchestratorType OrchestratorType) error {
 			return e
 		}
 	}
+	
+	//CoreOS is only applicable for Kubernetes
```
amanohar: This is empty. Is there plan to add validation?

JackQuincy:
More basic level do we have an issue for this? As this is a change we would need to discuss the right way to update the api model etc. for. Can you create one and we can have a discussion there.

pkg/api/vlabs/types.go
uday31in: - ARM uses the same model so i guess it is OK to follow the same approach. Intention is to first introduce change for agent nodes and let it bake before changing the masters. When we have Ignition support we can move master and these settings will go to linux profile otherwise it will be exactly same as using defaults.  Currently it is hard coded to use Image in variables but ultimately we should be allowing users to customise it from the acs-engine itself.

- Are there any issues for mixed mode cluster (CoreOS  Agent + Ubuntu Masters). My understanding is this is fine and supported. 

- Please point me to existing tests specific to Ubuntu. I will do the same for CoreOS.

pkg/api/vlabs/validate.go
uday31in: Added the support. https://github.com/uday31in/acs-engine/commit/d64b5085a056fdf168ee2173bd75423fc9a4f2fb

pkg/api/vlabs/types.go
uday31in: @amanohar  Also looking at how it is organised currently, we should ideally leave the OS choice at agent/master pool level instead of putting it in Linux Profile (one per cluster definition :-(). At agent level we have a choice to mix the OS For Example, CoreOS and Ubuntu or (in future) different version of Windows RS1 and RS3 in same cluster.

uday31in:
Closing this PR in favour of #1128

PR #1093, opened 2017-07-24
auto create sp when kv ref is not used

### LINKED ISSUE #1092: SP auto-generation clashing with move of key-vault URL into new key

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):  Issue

---

**What version of acs-engine?**: master

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** All 

**What happened**:

These two changes:

https://github.com/Azure/acs-engine/commit/0b5962459a832f17b6fec109aefdd1244370669e
https://github.com/Azure/acs-engine/commit/6a2da10dc4444015e56f1c000698ac8fec74583f

Have caused the following change in behavior I’m seeing off master is that it will now try to auto-generate a new SP because it’s only checking the existence of the old key 
“servicePrincipalClientSecret” which used to support both key-vault resource URL and a literal key string.
 

use the secret from keyvault if present.

api-model with servicePrincipalClientSecret missing and servicePrincipalClientKeyvaultSecretRef present. How do we want to handle the existance of both in the API model?

weinong:
Skip the auto create SP logic when KV secret is used.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1092 

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1093)

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -180,7 +180,7 @@ func autofillApimodel(dc *deployCmd) {
 
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
-		if spp != nil && (spp.ClientID == "" || spp.Secret == "") {
+		if spp != nil && spp.KeyvaultSecretRef == "" && (spp.ClientID == "" || spp.Secret == "") {
 			// they didn't specify one or the other
 			if spp != nil && !(spp.ClientID == "" && spp.Secret == "") {
 				// they didn't specify just one
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -180,13 +180,7 @@ func autofillApimodel(dc *deployCmd) {
 
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
-		if spp != nil && spp.KeyvaultSecretRef == "" && (spp.ClientID == "" || spp.Secret == "") {
-			// they didn't specify one or the other
-			if spp != nil && !(spp.ClientID == "" && spp.Secret == "") {
-				// they didn't specify just one
-				log.Fatal("apimodel invalid: ServicePrincipalProfile is missing either the clientid or secret.")
-			}
-
+		if spp != nil && spp.ClientID == "" && spp.Secret == "" && spp.KeyvaultSecretRef == "" {
 			log.Warnln("apimodel: ServicePrincipalProfile was missing or empty, creating application...")
 
 			// TODO: consider caching the creds here so they persist between subsequent runs of 'deploy'


weinong:
@acs-bot test this please

cmd/deploy.go
```diff
@@ -180,7 +180,7 @@ func autofillApimodel(dc *deployCmd) {
 
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
-		if spp != nil && (spp.ClientID == "" || spp.Secret == "") {
+		if spp != nil && spp.KeyvaultSecretRef == "" && (spp.ClientID == "" || spp.Secret == "") {
```
JackQuincy: Feels like this should be `(spp.ClientID == "" || (spp.KeyvaultSecretRef == "" && spp.Secret == ""))`

cmd/deploy.go
```diff
@@ -180,7 +180,7 @@ func autofillApimodel(dc *deployCmd) {
...
 			// they didn't specify one or the other
 			if spp != nil && !(spp.ClientID == "" && spp.Secret == "") {
```
JackQuincy: we should update this condition as well.

weinong: this logic is to auto populate the sp profile. there is another check in validation already.

cmd/deploy.go
lachie83: You're still going to match this condition if Secret is "" which is valid if KeyvaultSecretRef is not empty.

weinong: when KeyvaultSecretRef is not empty and Secret is empty, we don't automatically create sp.

lachie83: Ah yes. You're correct!

JackQuincy: Conceptually spp.Secert can now be found in two places. So places that used to check for its existence now need to check in two places. If only the secret ref was specified the code here will think every thing is peachy and not fail the request until something fails later since the client id is empty. If we just change this line then we will come in and generate a new SP/secret not realizing the user had tried to set the secret via KV. and both will be set which should cause issues elsewhere in the code.

weinong: the code in deploy.go is to conveniently create spn if it's not specified. this will only be applicable when keyvault secret ref is not used.

lachie83:
LGTM. I've provisioned clusters testing all combinations.

weinong:
@JackQuincy Does it make sense to you?

PR #1094, opened 2017-07-24
remove unnecessary customized types

### LINKED ISSUE #921: Intent to Implement: Change how versioning is handled
Versioning will be changed such that:
* User now specifies `orchestratorVersionHint` (examples: `1.5`, `1.6`). At deploy time, it is resolved to a concrete version (`1.6.6`).
* Upgrade API allows for same minor version to be specified. It can then either do a full upgrade, or just a patch version level upgrade, based on what it knows is the highest resolvable version at the time of upgrade being called

This will require changes to the api models for vlabs, and the 2017-07-01 API. But this is fine because we haven't really fully gone live with these fields in that api version.

rjtsdl:
To fix #921, i need to lay out things a little bit. I noticed, OrchestratorType, OrchestratorVersion are defined as custom type. If i want to add OrchestratorVersionHint, won't it be another custom type? I don't think so. They could all be strings. Avoid unnecessary types. This is not fully solved issue #921. But it can be checked in independently, without any behavior change.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: relates #921 

Ready to review
**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1094)

### DIFF
--- a/pkg/api/v20170701/types.go
+++ b/pkg/api/v20170701/types.go
@@ -107,8 +107,8 @@ const (
 
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
-	OrchestratorType    OrchestratorType    `json:"orchestratorType" validate:"required"`
-	OrchestratorVersion OrchestratorVersion `json:"orchestratorVersion"`
+	OrchestratorType    string `json:"orchestratorType" validate:"required"`
+	OrchestratorVersion string `json:"orchestratorVersion"`
 }
 
 // MasterProfile represents the definition of master cluster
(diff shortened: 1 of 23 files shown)


pkg/api/v20170701/types.go
```diff
@@ -215,29 +215,31 @@ func (a *AgentPoolProfile) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-// OrchestratorType defines orchestrators supported by ACS
```
rjtsdl: @JiangtianLi 

I was thinking to remove OrchestratorType custom type in my previous `validate tag` PR. Now i take it into action. And also enforce the same tradition across the old versions.

rjtsdl:
@acs-bot test this please

pkg/api/vlabs/types.go
```diff
@@ -226,32 +254,6 @@ type KeyVaultCertificate struct {
 	CertificateStore string `json:"certificateStore,omitempty"`
 }
 
-// OrchestratorType defines orchestrators supported by ACS
-type OrchestratorType string
```
JiangtianLi: Any reason we remove type define for OrchestratorType and OrchestratorVersion?

rjtsdl: 1. Unnecessary
2. I am going to add a `OrchestratoryTypeHint`. If i follow the tradition, it will be another custom type. Don't feel right.

JiangtianLi: I see. custom type is good for type-checking but if it is more cumbersome than beneficial here, then we don't need the custom type. There are other custom types such as OsType too. We can leave it for now.

rjtsdl: 3. OrchestratorVersion is not going to be a field taking input from api model. It will be a read-only field. I am going to do this change, not in this PR yet.

JackQuincy: one small input it should be OcherstratorVersionHint, not Type. We should still have a short list of allowed Orchestrators.

rjtsdl: OrchestratorType has a list of allowed orchestrators for sure. We just don't need a custom type for it. It will be checked in validation.
True, It should be OrchestratorVersionHint, thx.

PR #1096, opened 2017-07-25
use validate tag for vlab version as well

rjtsdl:
Update vlab api to adopt validate tag. For #921, i will add OrchestratorVersionHint, don't want to have different validate functions for the new field. So I just convert it to use the tag approach as well.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: relates #921

Ready to review
**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/v20170701/validate.go
+++ b/pkg/api/v20170701/validate.go
@@ -114,10 +114,11 @@ func handleValidationErrors(e validator.ValidationErrors) error {
 	err := e[0]
 	ns := err.Namespace()
 	switch ns {
-	case "Properties.OrchestratorProfile", "Properties.MasterProfile",
+	case "Properties.OrchestratorProfile", "Properties.OrchestratorProfile.OrchestratorType",
+		"Properties.MasterProfile",
 		"Properties.MasterProfile.DNSPrefix", "Properties.MasterProfile.VMSize",
 		"Properties.LinuxProfile", "Properties.ServicePrincipalProfile.ClientID",
-		"Properties.ServicePrincipalProfile.Secret", "Properties.WindowsProfile.AdminUsername",
+		"Properties.WindowsProfile.AdminUsername",
 		"Properties.WindowsProfile.AdminPassword":
 		return fmt.Errorf("missing %s", ns)
 	case "Properties.MasterProfile.Count":
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -6,12 +6,19 @@ import (
 	"net"
 	"net/url"
 	"regexp"
+	"strings"
 	"time"
+
+	validator "gopkg.in/go-playground/validator.v9"
 )
 
-var keyvaultSecretPathRegex *regexp.Regexp
+var (
+	validate                *validator.Validate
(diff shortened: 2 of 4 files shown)


rjtsdl:
@acs-bot test this please

pkg/api/v20170701/validate.go
```diff
@@ -114,10 +114,11 @@ func handleValidationErrors(e validator.ValidationErrors) error {
 	err := e[0]
 	ns := err.Namespace()
 	switch ns {
-	case "Properties.OrchestratorProfile", "Properties.MasterProfile",
+	case "Properties.OrchestratorProfile", "Properties.OrchestratorProfile.OrchestratorType",
+		"Properties.MasterProfile",
 		"Properties.MasterProfile.DNSPrefix", "Properties.MasterProfile.VMSize",
 		"Properties.LinuxProfile", "Properties.ServicePrincipalProfile.ClientID",
-		"Properties.ServicePrincipalProfile.Secret", "Properties.WindowsProfile.AdminUsername",
+		"Properties.WindowsProfile.AdminUsername",
```
JiangtianLi: I may not be up to date on this: why is ServicePrincipalProfile.Secret not needed?

pkg/api/vlabs/validate.go
```diff
@@ -193,31 +157,63 @@ func validateKeyVaultSecrets(secrets []KeyVaultSecrets, requireCertificateStore
...
+	if e := validate.Var(l.SSH.PublicKeys[0].KeyData, "required"); e != nil {
+		return fmt.Errorf("KeyData in LinuxProfile.SSH.PublicKeys cannot be empty string")
 	}
 	if e := validateKeyVaultSecrets(l.Secrets, false); e != nil {
 		return e
 	}
 	return nil
 }
 
+func handleValidationErrors(e validator.ValidationErrors) error {
```
JiangtianLi: Seems there are many overlap with handleValidationErrors in other api version. Is there way to consolidate them, i.e. common check into one util function?

pkg/api/v20170701/validate.go
rjtsdl: Oh, we have Secret or KeyvaultSecretRef. We only need either of them to be there.

And also for vlab specific, looks we allow a ServicePrincipalProfile with all the fields empty for non-kubernetes scenario. I don't want to break the existing behavior. Although i don't think it is right. But this is more of a correction PR for that purpose. This PR is scoped to keep the existing behavior, only change how validate handled.

pkg/api/vlabs/validate.go
rjtsdl: There is potential for this. We only need a superset of `handleValidationErrors ` to handle all the scenarios. But the assumption would be vlab and v20170701 will have the same constraint on certain field. However, even they are different we can use version to switch it. So i agree we can consolidate them. Let me do it here.

rjtsdl: I moved the function to pkg/api/common package. And give each version the freedom to override common validation error messages.

rjtsdl:
@acs-bot test this please

PR #1108, opened 2017-07-26
[API] Change how orchestratorVersion is handled

### LINKED ISSUE #921: Intent to Implement: Change how versioning is handled

### LINKED ISSUE #1114: Refactor Kubernetes upgrade code
No
---

FEATURE REQUEST
---

master/HEAD
---

1. After the change to user OrchestratorVersionHint as user input, upgrade code still treat OrchestratorVersion as user's input. Need to refactor that
2. Also upgrade is focus on 1.5.3 -> 1.6.2, we should work on the scenario to upgrade  1.5 -> 1.6 regardless of the patch part. 

Kubernetes

More aligned with OrchestratorVersionHint change. #921

rjtsdl:
Solve #921 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #921, fixes #1114 

Ready to review.
**TODO**
- [x] Refactor Upgrade code

**Release note**:

```release-note
```

**Context**:
The underlying idea of this change is, to make acs-engine support no more than one patched version of k8s. Say, there is 1.6.2, and 1.6.6, acs-engine will support only one of them, and that is not an option for the user. Most likely, we will support 1.6.6, as it is latest. 

Benefit is, the list of k8s versions we need to support will drop significantly. And we there is no point to still support 1.6.2 while 1.6.6 is also there, considering they should be compatible.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1108)


pkg/api/const.go
```diff
@@ -51,38 +51,33 @@ const (
...
-	// Kubernetes166 is the string constant for Kubernetes 1.7.0
-	Kubernetes170 string = "1.7.0"
-	// Kubernetes166 is the string constant for Kubernetes 1.7.1
-	Kubernetes171 string = "1.7.1"
-	// KubernetesDefaultVersion is the string constant for current Kubernetes version
-	KubernetesDefaultVersion string = Kubernetes166
+	KubernetesVersionHint17 string = "1.7"
+	KubernetesVersionHint16 string = "1.6"
+	KubernetesVersionHint15 string = "1.5"
+	// Why not 1.7?
```
rjtsdl: I notice the default version is lagging behind. It is using 1.6.6, shall we use latest?

rjtsdl:
@acs-bot test this please

pkg/acsengine/const.go
```diff
@@ -78,9 +78,10 @@ const (
 	DCOSPublicAgent DCOSNodeType = "DCOSPublicAgent"
 )
 
-// KubeImages represents Docker images used for Kubernetes components based on Kubernetes version
+// KubeImages represents Docker images used for Kubernetes components based on Kubernetes version hint
+// version hint here is the major and minor part of the version. For instance, version 1.6.6 has version hint 1.6
 var KubeImages = map[string]map[string]string{
-	api.Kubernetes171: {
+	api.KubernetesVersionHint17: {
```
JiangtianLi: Is this the place if patched version is bumped up, we update here? kuberneteswindowssetup.ps1 currently depends on version and not versionHint for version comparison. Can we update kuberneteswindowssetup.ps1 too?

pkg/acsengine/engine.go
```diff
@@ -520,65 +517,18 @@ func addSecret(m map[string]interface{}, k string, v interface{}, encode bool) {
...
 		},
 		"IsKubernetesVersionGe": func(version string) bool {
-			targetVersion := version
-			targetVersionOrdinal := VersionOrdinal(targetVersion)
-			orchestratorVersionOrdinal := VersionOrdinal(cs.Properties.OrchestratorProfile.OrchestratorVersion)
+			targetVersion, _ := goversion.NewVersion(version)
+			orchestratorVersion, _ := goversion.NewVersion(cs.Properties.OrchestratorProfile.OrchestratorVersion)
 			return cs.Properties.OrchestratorProfile.OrchestratorType == api.Kubernetes &&
-				orchestratorVersionOrdinal >= targetVersionOrdinal
+				orchestratorVersion.Compare(targetVersion) != -1
```
JiangtianLi: How does goversion Compare work? Is it more than string comparison, which is enough for kuberneteswindowssetup.ps1.

pkg/acsengine/const.go
```diff
@@ -102,29 +103,7 @@ var KubeImages = map[string]map[string]string{
...
-		"hyperkube":       "hyperkube-amd64:v1.7.0",
-		"dashboard":       "kubernetes-dashboard-amd64:v1.6.1",
-		"exechealthz":     "exechealthz-amd64:1.2",
-		"addonresizer":    "addon-resizer:2.0",
-		"heapster":        "heapster:v1.4.0",
-		"dns":             "k8s-dns-kube-dns-amd64:1.14.4",
-		"addonmanager":    "kube-addon-manager-amd64:v6.4-beta.2",
-		"dnsmasq":         "k8s-dns-dnsmasq-amd64:1.14.4",
-		"pause":           "pause-amd64:3.0",
-		"windowszip":      "v1.7.0intwinnat.zip",
```
JiangtianLi: where do we get the version of windowszip? I saw it is removed but not added somewhere in your change?

pkg/acsengine/engine.go
rjtsdl: It is more than string comparison. Before, we use a function called VersionOrdinal to format the version strings, and then do string comparison. That's not very intuitive. I am inclined to use this package for version comparison.

pkg/acsengine/const.go
rjtsdl: This is the place, when patched version is bumped up, the value of this map need to be updated. Let me look at ps1 file. I may missed it.

pkg/acsengine/const.go
rjtsdl: When we have v1.7.1, we don't need to have v1.7.0 anymore. This is the whole point of only maintaining one patched version of a minor. 

Yes, you would expect to see i removed quite a few code without adding back.

pkg/acsengine/const.go
rjtsdl: Actually, parts/ files still using version instead of versionHint. It is just the version is derived from versionHint when it is converted to unversioned api model.

rjtsdl: Btw, i have split KubeImages into KubeImages in api package, and KubeConfigs in place.

pkg/api/vlabs/types.go
```diff
@@ -128,9 +128,10 @@ const (
 
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
-	OrchestratorType    string            `json:"orchestratorType" validate:"required"`
-	OrchestratorVersion string            `json:"orchestratorVersion"`
-	KubernetesConfig    *KubernetesConfig `json:"kubernetesConfig,omitempty"`
+	OrchestratorType        string            `json:"orchestratorType" validate:"required"`
+	OrchestratorVersionHint string            `json:"orchestratorVersionHint"`
```
amanohar: This will be a breaking change in the public (preview) API version. Are we ok with this?

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -70,7 +70,7 @@ func (uc *UpgradeCluster) UpgradeCluster(subscriptionID uuid.UUID, resourceGroup
 	}
 
 	switch ucs.OrchestratorProfile.OrchestratorVersion {
```
amanohar: Can you make sure upgrade works with your changes? E.g. OrchestratorVersion  was an input provided by the customer. Now OrchestratorVersionHint is the input, this is not the right field to be validated but it should flow into the upgrade operation to create the right template.

pkg/api/vlabs/types.go
rjtsdl: AFAIK, we are aware of that. @anhowe to confirm though.

pkg/operations/kubernetesupgrade/upgradecluster.go
rjtsdl: @JiangtianLi and I discussed this briefly yesterday. 

My understanding is, I will not touch the logic of upgrade in general. Only in a few places, where I replace version consts with plain strings. The current upgrade feature is not working anyway. So I will leave it later few. 

Another dimension is, upgrade so far is only handling scenario of kubernete 1.5.3 -> 1.6.2. However, 1.5.3 will not be supported, 1.5.7 is the latest of 1.5 minor. And it will never upgrade to 1.6.2, because 1.6.6 is the latest of 1.6 minor.

amanohar: @rjtsdl,

> The current upgrade feature is not working anyway

Where did you get this info? Can you provide more details. Upgrade works today and please make sure that this PR doesn't break it.

amanohar: Also @rjtsdl 

> Another dimension is, upgrade so far is only handling scenario of kubernete 1.5.3 -> 1.6.2. 

Most important part of the upgrade operation is that it verifies the logic that upgrade from one version to another works. The fact that 1.5.3 will be discontinued doesn't mean that upgrade op code is not usable. It's a matter of changing the target version and then it will work.

rjtsdl: I can open an issue to refactor upgrade code. Will update here when i have the issue created.

amanohar: Sure, main point being let's not break one feature because of another feature. Note the upgrade is the most important scenario for this change.

rjtsdl: #1114 , just created the issue. Jiangtian or I will work on it right after this change.
I will make sure it passed pr test, so it should not break any existing scenario.

rjtsdl: Actually, it should be irrelevant in term of upgrade and the change i made. upgrade has its own api model. So it won't break for sure.

rjtsdl:
@acs-bot test this please

pkg/operations/kubernetesupgrade/upgradecluster.go
amanohar: It's not irrelevant to upgrade. Upgrade uses the version specified in its API model to set the container service API model to generate the template.
Following lines need to be changed and test:
1) https://github.com/Azure/acs-engine/blob/master/pkg/operations/kubernetesupgrade/v162upgrader.go#L42

2) https://github.com/Azure/acs-engine/blob/master/pkg/operations/kubernetesupgrade/v162upgrader.go#L27-L30

3) https://github.com/Azure/acs-engine/blob/master/pkg/operations/kubernetesupgrade/upgradecluster.go#L73-79

rjtsdl: Synced offline

The upgrade code path is working out of the box. So it will not be the blocker here.
I will align upgrade code with this change later, as documented #1114. 

Thx for illustrating me how to use upgrade function. And seeing the upgrade function works is exciting.

amanohar: Thanks for running the upgrade op (now you know it works) :)

As also mentioned in person is that upgrade operation should be changed as part of this PR because it important to verify that assumptions in upgrade works with this API change (given that this API change is being made for upgrade). 

So while it doesn't break upgrade operation it also doesn't verify that the API change works/is usable/makes sense for upgrade.

rjtsdl: OK, i will make the upgrade change along with the existing change.

seanknox:
@acs-bot test this please

pkg/operations/kubernetesupgrade/upgradecluster.go
rjtsdl: @amanohar, i made the minimal change for upgrade to align with OrchestratorVersionHint change. I tried to upgrade, now it will upgrade 1.5.x directly to 1.6.6, and it is SUCCESS. I keep the cluster, if you want to take a look. It is exciting for me to see it work !

rjtsdl:
@acs-bot test this please

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -48,7 +48,8 @@ var _ = Describe("Upgrade Kubernetes cluster tests", func() {
 		ucs := api.UpgradeContainerService{}
 		ucs.OrchestratorProfile = &api.OrchestratorProfile{}
 		ucs.OrchestratorProfile.OrchestratorType = api.Kubernetes
-		ucs.OrchestratorProfile.OrchestratorVersion = "1.6.2"
+		ucs.OrchestratorProfile.OrchestratorVersionHint = api.KubernetesVersionHint16
+		ucs.OrchestratorProfile.OrchestratorVersion = api.KubeImages[api.KubernetesVersionHint16]["version"]
```
amanohar: This will get the latest version?

rjtsdl: Yes, this will get the latest version we support under 1.6, so it is 1.6.6.

amanohar: And that's because we will always only have one patch version of a minor version (so there will be no need to choose between one patch versus another)?

rjtsdl: Yes, choose under minor is not an option anymore.

rjtsdl:
@acs-bot test this please

rjtsdl:
@acs-bot test this please

rjtsdl:
@acs-bot test this please

rjtsdl:
@acs-bot test this please

rjtsdl:
@seanknox , there is one remaining question. How do we communicate this to acs-engine users?

For ACS RP, the impact will be 2017-07-01 api model only. And this one is in preview, so it is ok to change.

For ACS Engine users, most of them use vlabs version. This change will certainly break their existing  setup.

seanknox:
Can you give a 1-2 sentence elevator pitch summarizing this change, pretending we're speaking to a non-ACS engineer? I looked at the issues and couldn't quite make sense of it.

rjtsdl:
@seanknox , updated **Context** section in this pr's description.

seanknox:
Also, please rebase off of master—we just merged in CircleCI support. Rebasing will kick off a job.

jackfrancis:
@seanknox 👍

jackfrancis:
@seanknox @jchauncey I'm stumped why the most recent `ci/circleci: test` failed due to missing `go-version`. It's committed in this PR under `vendor/`:

https://github.com/Azure/acs-engine/pull/1108/files#diff-eb066af97cb10032e5af1303d0fe48de

My local `make build` runs are returning `0` exit code.

seanknox:
> I'm stumped why the most recent ci/circleci: test failed due to missing go-version.

It's because CI also runs `make bootstrap` before running tests. If you try that on your machine, you'll see that `go-version` and its dependencies are dropped from `vendor/` because `go-version` is missing from `glide.yaml`. 

To fix:
1. add `go-version` to `glide.yaml`:
```yaml
- package: github.com/hashicorp/go-version
  version: 03c5bf6be031b6dd45afec16b1cf94fc8938bc77
```
2. run `glide update`
3. commit changes to `glide.yaml`, `glide.lock`, and `vendor/`. You'll see changes to ginkgo and gomega which should be fine.
4. run `make build test` and unit tests should pass.

pkg/operations/kubernetesupgrade/v162upgrader.go
```diff
@@ -26,7 +26,7 @@ type Kubernetes162upgrader struct {
 // ClusterPreflightCheck does preflight check
 func (ku *Kubernetes162upgrader) ClusterPreflightCheck() error {
 	// Check that current cluster is 1.5.3
-	if ku.DataModel.Properties.OrchestratorProfile.OrchestratorVersion != api.Kubernetes153 {
+	if ku.DataModel.Properties.OrchestratorProfile.OrchestratorVersion != "1.5.3" {
```
amanohar: Looks like the change the @rjtsdl made to pass orchestrator release instead of version got reverted. Adding comment here to add that change back in (because ACS Engine should determine what the patch version should be).

jackfrancis: @amanohar I'm not sure what should be changed here, should we do it in this PR?

jackfrancis:
@seanknox post-`0.5.0` release wave

pkg/operations/kubernetesupgrade/v162upgrader.go
rjtsdl: I just addressed with the latest change. In upgrade, we use OrchestratorRelease through out the code instead of OrchestratorVersion.

examples/kubernetesversions/kubernetes1.5.3.json
```diff
@@ -3,7 +3,7 @@
   "properties": {
     "orchestratorProfile": {
       "orchestratorType": "Kubernetes",
-      "orchestratorVersion": "1.5.3"
```
amanohar: Looks like 1.5.3 does not exist in ACS Engine anymore. If that's the case this file should be renamed to: kubernetes1.5.json

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -196,7 +196,7 @@ func createContainerService(containerServiceName string, masterCount int, agentC
 
 	cs.Properties.OrchestratorProfile = &api.OrchestratorProfile{}
 	cs.Properties.OrchestratorProfile.OrchestratorType = api.Kubernetes
-	cs.Properties.OrchestratorProfile.OrchestratorVersion = api.Kubernetes153
+	cs.Properties.OrchestratorProfile.OrchestratorVersion = "1.5.3"
```
amanohar: This needs to be updated to OrchestratorRelease as well

pkg/api/common/const.go
```diff
@@ -37,3 +37,39 @@ const (
...
+	KubernetesRelease1Dot6 string = "1.6"
+	// KubernetesRelease1Dot5 is the major.minor string prefix for 1.5 versions of kubernetes
+	KubernetesRelease1Dot5 string = "1.5"
+	// KubernetesDefaultRelease is the default major.minor version for kubernetes
+	KubernetesDefaultRelease string = KubernetesRelease1Dot6
+)
+
+// KubeReleaseToVersion maps a major.minor release to an full major.minor.patch version
+var KubeReleaseToVersion = map[string]string{
+	KubernetesRelease1Dot7: "1.7.1",
```
rjtsdl: @jackfrancis , i think, it should be 1.7.2

PR #1111, opened 2017-07-26
added metrics for deployment and validation durations

dmitsh:
added deployment and validation duration metrics for better test experience and diagnostics

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1111)

### DIFF
--- a/test/acs-engine-test/main.go
+++ b/test/acs-engine-test/main.go
@@ -24,12 +24,24 @@ import (
 const (
 	script = "test/step.sh"
 
+	stepInitAzure        = "set_azure_account"
+	stepCreateRG         = "create_resource_group"
+	stepPredeploy        = "predeploy"
+	stepGenerateTemplate = "generate_template"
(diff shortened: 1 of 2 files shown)


test/acs-engine-test/main.go
```diff
@@ -380,6 +425,9 @@ func mainInternal() error {
 		fmt.Println("Warning: BUILD_NUMBER is not set or invalid. Assuming 0")
 		buildNum = 0
 	}
+	if os.Getenv("SKIP_METRICS") == "y" {
```
JiangtianLi: where is SKIP_METRICS set? seems not in this PR?

dmitsh: it is an environment variable to control test flow. It might be set in a Jenkins job.

JiangtianLi: I see. Can you add a comment here about how to use that env var?

PR #1142, opened 2017-07-31
Add pod subnet to non masquerade

### LINKED ISSUE #425: How to set up Container to Internal (Peered) Network
Hi, I have an Azure ACS Kubernetes cluster running. Its a custom vnet which I have working correctly.

I have older legacy applications running in the same Azure subscription on VMs (different vnet) with connectivity through Peering. I can ping the cluster from the legacy machines and communicate with pods if I expose them on via NodePort. I can also ping the legacy machines from the cluster nodes and the master.

The pods / containers running inside the cluster can access each other properly and access the internet if necessary. They can not access the peered network / legacy apps. I suspect its either a route i need to add into the docker / kubernetes network or iptables / firewall stopping the communication

How can I get my containers to talk to the peered subnet?

For reference:
Kubernetes cluster 10.10.x.x/16 (service network 10.10.0.0/20)
Agents 10.10.16.x/20
Container internal network (auto created - the ips the pods get from the host) 10.240.1.x/24
Peered network 10.12.x.x/16

A pod (10.240.1.5) can communicate with the internet, the master ip, agent (host) ips and other containers

But it can't connect to 10.12.x.x/16 the peered network.

sylr:

If this option is not provided the default used by kubelet is 10.0.0.0/8 (according to https://github.com/Azure/acs-engine/issues/425#issuecomment-291854852). 

In a Custom VNet context (VNet space which is not 10.0.0.8), we want to set this value to the subnet of the pods (kubeClusterCidr) otherwise Kubernetes will think that all traffic directed towards 10.0.0.0/8 from the pods is intended to stay within Kubernetes.

By setting the correct subnet all traffic which is outside kubeClusterCidr will be "masqueraded" and will be able to exit kubernetes and reach the intended resources in the Custom VNet. 

**Which issue this PR fixes**: fixes #425

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1142)

### DIFF
--- a/parts/kubernetesagentcustomdata.yml
+++ b/parts/kubernetesagentcustomdata.yml
@@ -113,6 +113,7 @@ write_files:
     KUBE_CTRL_MGR_NODE_MONITOR_GRACE_PERIOD={{WrapAsVariable "kubernetesCtrlMgrNodeMonitorGracePeriod"}}
     KUBE_CTRL_MGR_POD_EVICTION_TIMEOUT={{WrapAsVariable "kubernetesCtrlMgrPodEvictionTimeout"}}
     KUBE_CTRL_MGR_ROUTE_RECONCILIATION_PERIOD={{WrapAsVariable "kubernetesCtrlMgrRouteReconciliationPeriod"}}
+    KUBELET_POD_SUBNET={{WrapAsVariable "kubeClusterCidr"}}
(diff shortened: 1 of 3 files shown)


mrosterm:
What is the status on this one?

bigmyx:
I am also having the same issue, would appreciate an update!

PR #1165, opened 2017-08-02
revert changes in test/common.sh

dmitsh:
reverting the change that broke regression tests

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1165)


seanknox:
We'll need to setup Key Vault for acs-engine CI or refactor common.sh to not require them. Can you work with @jackfrancis and @jchauncey on this?

dmitsh:
You still can use plain text. I will work with the guys on that.

PR #1166, opened 2017-08-02
Add Ginkgo unit test & JUnit reporting framework

### LINKED ISSUE #1167: Add Ginkgo unit test & JUnit reporting framework
We want to try out using Ginkgo as the test runner for both unit and integration/functional/regression tests, which will make tests easier to write, refactor, and reuse.

## acceptance criteria:

- Lay down Ginkgo boilerplate/setup.
- Use JUnit reporting. This should be in one config that's re-used by tests.
- Add single unit test with ginkgo/gomega syntax

dtzar:
Code is written with @OguzPastirmaci @TsuyoshiUshio @peterhuene @dcaro @dtzar 

Helps to have consistency for unit testing using Ginkgo/Gomega, and centralizing JUnit format output into CircleCI dashboard

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1167 

**Release note**:
```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1166)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,9 +64,7 @@ ifneq ($(GIT_BASEDIR),)
 endif
 
 test: test-style
-	go get github.com/onsi/ginkgo/ginkgo
-	go get github.com/onsi/gomega
-	ginkgo -v -ldflags='$(LDFLAGS)' $(GOFILES)
+	ginkgo -r -ldflags='$(LDFLAGS)' .
 
 .PHONY: test-style
 test-style:
@@ -81,6 +79,7 @@ HAS_GOX := $(shell command -v gox;)
 HAS_GIT := $(shell command -v git;)
 HAS_GOBINDATA := $(shell command -v go-bindata;)
 HAS_GOMETALINTER := $(shell command -v gometalinter;)
+HAS_GINKGO := $(shell command -v ginkgo;)
 
 .PHONY: bootstrap
 bootstrap:
@@ -101,6 +100,10 @@ ifndef HAS_GOMETALINTER
 	gometalinter --install
 endif
 	glide install
+ifndef HAS_GINKGO
+	go get -u github.com/onsi/ginkgo/ginkgo
+endif
+
 
 ci: bootstrap test-style build test lint
 	./scripts/coverage.sh --coveralls
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ clean:
 
 GIT_BASEDIR    = $(shell git rev-parse --show-toplevel 2>/dev/null)
 ifneq ($(GIT_BASEDIR),)
-	LDFLAGS += -X github.com/Azure/acs-engine/cmd_test.JUnitOutDir=${GIT_BASEDIR}/test/junit
+	LDFLAGS += -X github.com/Azure/acs-engine/pkg/test.JUnitOutDir=${GIT_BASEDIR}/test/junit
 endif
 
 test: test-style
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,9 @@ ifneq ($(GIT_BASEDIR),)
 endif
 
 test: test-style
-	go test -v -ldflags '$(LDFLAGS)' $(GOFILES)
+	go get github.com/onsi/ginkgo/ginkgo
+	go get github.com/onsi/gomega
+	ginkgo -v -ldflags='$(LDFLAGS)' $(GOFILES)
 
 .PHONY: test-style
 test-style:
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,13 @@ checksum:
 clean:
 	@rm -rf $(BINDIR) ./_dist
 
+GIT_BASEDIR    = $(shell git rev-parse --show-toplevel 2>/dev/null)
+ifneq ($(GIT_BASEDIR),)
+	LDFLAGS += -X github.com/Azure/acs-engine/cmd_test.JUnitOutDir=${GIT_BASEDIR}/test/junit
+endif
+
 test: test-style
-	go test -v $(GOFILES)
+	go test -v -ldflags '$(LDFLAGS)' $(GOFILES)
 
 .PHONY: test-style
 test-style:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,8 @@ jobs:
       - run:
           name: Run style and unit tests
           command: make test-style build test
+      - store_test_results:
+          path: test/junit
       - run:
           name: Generate test coverage results
           command:  make build coverage
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ _dist/
 bin/
 .env
 
+test/junit/
 test/acs-engine-test/acs-engine-test.exe
 pkg/operations/junit.xml
 pkg/operations/kubernetesupgrade/junit.xml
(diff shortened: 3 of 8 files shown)


seanknox:
@jchauncey please take a look.

Makefile
```diff
@@ -58,8 +58,13 @@ checksum:
...
 	@rm -rf $(BINDIR) ./_dist
 
+GIT_BASEDIR    = $(shell git rev-parse --show-toplevel 2>/dev/null)
+ifneq ($(GIT_BASEDIR),)
+	LDFLAGS += -X github.com/Azure/acs-engine/pkg/test.JUnitOutDir=${GIT_BASEDIR}/test/junit
+endif
+
 test: test-style
-	go test -v $(GOFILES)
+	go test -v -ldflags '$(LDFLAGS)' $(GOFILES)
```
jchauncey: we should probably use the ginkgo cli to run the tests

dtzar: Ok, would you just want to run two separate lines for `make test` (one with `go test` and one with `ginkgo`?)

dtzar: Or since it seems ginkgo runs all go tests too, how about switching the `make test` to run `ginkgo -v -ldflags='$(LDFLAGS)' $(GOFILES)` only?  Should we add any switches defined here: https://onsi.github.io/ginkgo/#ginkgo-and-continuous-integration

seanknox: I believe you can use `gingko` to run unit or e2e BDD-style tests. For running only unit tests, pass `$GOFILES` to ginkgo.

seanknox: yep! what you said. :)

Makefile
```diff
@@ -58,8 +58,15 @@ checksum:
...
+GIT_BASEDIR    = $(shell git rev-parse --show-toplevel 2>/dev/null)
+ifneq ($(GIT_BASEDIR),)
+	LDFLAGS += -X github.com/Azure/acs-engine/pkg/test.JUnitOutDir=${GIT_BASEDIR}/test/junit
+endif
+
 test: test-style
-	go test -v $(GOFILES)
+	go get github.com/onsi/ginkgo/ginkgo
+	go get github.com/onsi/gomega
+	ginkgo -v -ldflags='$(LDFLAGS)' $(GOFILES)
```
jchauncey: you dont have to do $(GOFILES) here you can just do `ginkgo -r .` at the root

jchauncey: i see @seanknox comment about doign this just to run the unit test files. we can use `-skipPackage` to skip the e2e tests

Makefile
```diff
@@ -58,8 +58,15 @@ checksum:
...
 
+GIT_BASEDIR    = $(shell git rev-parse --show-toplevel 2>/dev/null)
+ifneq ($(GIT_BASEDIR),)
+	LDFLAGS += -X github.com/Azure/acs-engine/pkg/test.JUnitOutDir=${GIT_BASEDIR}/test/junit
+endif
+
 test: test-style
-	go test -v $(GOFILES)
+	go get github.com/onsi/ginkgo/ginkgo
+	go get github.com/onsi/gomega
```
seanknox: Move installation of ginkgo and gomega to `make bootstrap`.

jchauncey: you know if we relied on the deis/go-dev image we wouldnt have to worry about this. it already has all of these installed

jchauncey: plus if github is down this will fail

Makefile
seanknox: Nice. Just confirmed that `ginkgo -r` does the right thing. No need for $GOFILES!

Makefile
seanknox: The Docker dev experience in acs-engine needs to be thought through, and I'd like to use something like `deis/go-dev`. For now, let's improve the experience for those running on their own machine.

jchauncey: then i would separate installation of needed tools from installing the code dependencies of the project.

dtzar: IMO we should have the acs-engine image already have ginkgo/gomega installed.  This would be a simple update to the Dockerfile.  For now, how about I just move these lines to the `make boostrap` and change the test command line to `ginkgo -r .`?

seanknox: Agreed on both fronts @dtzar.

jchauncey: the image used for distribution really shouldnt have our dev tools on it. Thats just added bloat and baggage we shouldnt have to worry about maintaining. The pattern most people follow is to have a development image that you can use for local and ci runs that contains all the necessary tooling while the image you build for distribution just has the binaries. 

Doing this has multiple benefits:

1 - it saves us from having to build an image during the CI run just to build binaries/run tests
2 - we reduce the image size we ship to our customers by not having unnecessary tooling installed
3 - when we need to update testing/development dependencies we dont have to cut an unnecessary release of a distribution image
4 - allows us to independently version the development image outside of the project so it can keep pace with tools as they come out (like updating ginkgo/gomega) but a project can stay pinned to a particular image version taht works for them until they are ready to upgrade.
5 - it means that for ANY given commit you will have the appropriate versioned development image that can build the corresponding binaries and run the tests. This allows someone who needs to build an older version of the project an easy way to do so without having to worry about installing old tools

seanknox: Agreed and good points @jchauncey. At the moment we our own dev process doesn't have clear instructions  on developing with the Docker image, so de facto almost all development is happening outside of containers. In a subsequent issue/PR (not this one) let's flesh out how developing acs-engine in a container can be better.

jchauncey: my e2e pr will bring some of that to the table

dtzar: My thinking is that the Docker Hub microsoft/acs-engine image IS the development environment image anyone can use or we just use ourselves.  We could automate the push of the images to Docker hub easily using CircleCI and tag the image with the short git commit when it hits master - then you can match any image in Docker hub to any commit in Github off master (your point 5).   The idea is that the public docker hub image makes it easier to get the development environment for anyone versus building the docker image every time locally.  If we don't want them to have to update every commit to master, we could be more selective of when we push to docker hub (i.e. per release or when a significant change happens with Dockerfile).

Per point 1 - technically we don't have to build an image during CI (which I know we're doing now) to do the e2e tests.  It is really a safeguard measure for if/when someone upgrades the depdendent components (kubectl, az cli, etc).  If we would extract these type of components from the build image, then I think this would be great as they aren't really needed to build acs-engine binary itself or generate templates, right?  This would also help with your point #2 and 3.

In general, I'm all about lightweight docker image which just has binaries to simply use acs-engine (versus download the binary).  This perhaps would be best in a separate image.

jchauncey: right what im getting at is that msft/acs-engine is a distribution image not a developmental image. its a separation of concerns that cleans up how we manage ci and the overall development environment as a whole. the development image in question is definitely public and can be used by anyone to build and generate everythign they need within the project regardless of their environment.

seanknox:
@jchauncey I think this is good to go. can we get a thumbs up?

dtzar:
I amended the last commit with removing the `-v' flag from the ginkgo runner (per the request of @seanknox) and did a force push to keep commit history cleaner.

seanknox:
Thank you @OguzPastirmaci @TsuyoshiUshio @peterhuene @dcaro @dtzar! We are very happy to have this and sets the project up to greatly improve test coverage and test clarity.

PR #1076, opened 2017-07-21
remove version specific pieces of the heapster deployment. Rollback resizer to 1.7

lachie83:
Fixes #1075 

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1076)

PR #1174, opened 2017-08-03
Fix heapster deployment

malekpour:
Remove deprecated parameters from heapster deployments that prevent pods from starting successfully.

More at: https://github.com/kubernetes/kubernetes/commit/63ccedcfa7a34ba4bed8b697a07eab9e1e14896d

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1174)


pkg/acsengine/const.go
```diff
@@ -87,7 +87,7 @@ var KubeImages = map[string]map[string]string{
 		"dashboard":       "kubernetes-dashboard-amd64:v1.6.1",
 		"exechealthz":     "exechealthz-amd64:1.2",
 		"addonresizer":    "addon-resizer:1.7",
-		"heapster":        "heapster:v1.4.0",
+		"heapster":        "heapster-arm64:v1.4.0",
```
SaaldjorMike: arm64?

malekpour: Just to be aligned with other images. Also kubernetes started using arm64. https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/google/heapster-controller.yaml#L48

SaaldjorMike: you should you dont mean amd64 instead of arm64?

malekpour: oops, typo. just fixed it.

lachie83:
@malekpour This was updated as part of #1076

parts/kubernetesmasteraddons-heapster-deployment.yaml
```diff
@@ -40,11 +40,9 @@ spec:
         - "--extra-cpu=0.5m"
         - "--memory=140Mi"
         - "--extra-memory=4Mi"
-        - "--threshold=5"
```
lachie83: We rolled the addon-resizer back across the board currently so this change is irrelevant -- #1076

pkg/acsengine/const.go
```diff
@@ -87,7 +87,7 @@ var KubeImages = map[string]map[string]string{
 		"dashboard":       "kubernetes-dashboard-amd64:v1.6.1",
 		"exechealthz":     "exechealthz-amd64:1.2",
 		"addonresizer":    "addon-resizer:1.7",
-		"heapster":        "heapster:v1.4.0",
+		"heapster":        "heapster-amd64:v1.4.0",
```
lachie83: These appear to be the same image but I'm happy to keep the naming consistent. Can you please change on all released versions?

malekpour: Sure, I did for `heapster-amd64:v1.4.0` and `heapster-amd64:v1.3.0` but left `heapster:v1.2.0` unchanged since there is no amd64 available. This is also the kubernetes patters.

parts/kubernetesmasteraddons-heapster-deployment.yaml
malekpour: I just saw https://github.com/Azure/acs-engine/pull/1076. I still think fixing `acs-engine` to work with `addon-resizer:v2.0` is better solution compared to keep using `addon-resizer:v1.7`

If we are using `kubernetes:1.7` and `heapster:v1.4.0`, it makes sense to use `addon-resizer:v2.0`.

https://github.com/kubernetes/kubernetes/commit/63ccedcfa7a34ba4bed8b697a07eab9e1e14896d

lachie83: Okay. Let's split the resizer change out into a different PR and make this PR simply updating the container image

lachie83: Also can you let me know the changelog of features between 1.7 and 2.0 for the resizer?

malekpour: It seems it has rewritten entirely and references changed from 1.7 to 2.0 everywhere https://github.com/kubernetes/autoscaler/commit/93942efb7530a7d266dd7d7a0c83b0417b049471

lachie83:
LGTM

pkg/acsengine/azureconst.go
```diff
@@ -2,7 +2,7 @@ package acsengine
 
 import "fmt"
 
-// AUTOGENERATED FILE - last generated 2017-05-19 20:39:35
+// AUTOGENERATED FILE - last generated 2017-08-04 15:59:31
```
lachie83: This should be in a separate PR

malekpour: I agree and didn't mean to add this. I thought the other one is merged and just pushed this to my forked repo.

PR #1182, opened 2017-08-04
Alternate approach to agent pools.

brendandburns:
Test PR to visualize diffs.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1182)

### DIFF
--- a/parts/kubernetesmastervars.t
+++ b/parts/kubernetesmastervars.t
@@ -42,14 +42,18 @@
     "username": "[parameters('linuxAdminUsername')]",
     "masterFqdnPrefix": "[tolower(parameters('masterEndpointDNSNamePrefix'))]",
     "masterPrivateIp": "[parameters('firstConsecutiveStaticIP')]",
+{{ if .HasMaster }}
     "masterVMSize": "[parameters('masterVMSize')]",
-    "sshPublicKeyData": "[parameters('sshRSAPublicKey')]",
-{{if  GetClassicMode}}
-    "masterCount": "[parameters('masterCount')]",
-{{else}}
-    "masterCount": {{.MasterProfile.Count}},
 {{end}}
+    "sshPublicKeyData": "[parameters('sshRSAPublicKey')]",
+{{if .HasMaster}}
+  {{if GetClassicMode}}
+    "masterCount": "[parameters('masterCount')]",
+  {{else}}
+    "masterCount": {{.MasterProfile.Count}},
+  {{end}}
     "masterOffset": "[parameters('masterOffset')]",
+{{end}}
     "apiVersionDefault": "2016-03-30",
     "locations": [
          "[resourceGroup().location]",
@@ -64,7 +68,11 @@
     "osImageSKU": "16.04-LTS",
     "osImageVersion": "16.04.201706191",
     "resourceGroup": "[resourceGroup().name]",
+{{if .HasMaster}}
     "routeTableName": "[concat(variables('masterVMNamePrefix'),'routetable')]",
+{{else}}
+    "routeTableName": "[concat(variables('agentNamePrefix'), 'routetable')]",
+{{end}}
     "routeTableID": "[resourceId('Microsoft.Network/routeTables', variables('routeTableName'))]",
     "sshNatPorts": [22,2201,2202,2203,2204],
     "sshKeyPath": "[concat('/home/',variables('username'),'/.ssh/authorized_keys')]",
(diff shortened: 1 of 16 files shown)


pkg/api/apiloader.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 
+	"github.com/Azure/acs-engine/pkg/agentPoolOnlyApi/v20170831"
```
anhowe: for another PR, but "github.com/Azure/acs-engine/pkg/agentPoolOnlyApi/v20170831", should be under "github.com/Azure/acs-engine/pkg/api/v20170831"

amanohar: It still makes sense for all "agentPoolOnlyApi" to be under separate folder (could be moved under api) but shouldn't be a peer of container service type. 

The reason is both api models support similarly named fields/properties but shouldn't necessarily be shared because each of those properties could have their own fields and could cause conflicts it cs and cp types are under the same version (e.g. ServicePrincipalProfile). Placing them directly under "github.com/Azure/acs-engine/pkg/api/v20170831" means cs and cp resource types could be forced to be in the same package name and that's not desired.

anhowe: thanks, makes sense.  I like the idea of having it under "api", but not having it as peer of cs type.

amanohar: Yes, this is just temporarily placement (and probably will remain so for 2-3 weeks or so) to not disrupt/confuse with the prod changes already in place.

parts/kubernetesmastervars.t
```diff
@@ -64,7 +68,11 @@
     "osImageSKU": "16.04-LTS",
     "osImageVersion": "16.04.201706191",
     "resourceGroup": "[resourceGroup().name]",
+{{if .HasMaster}}
     "routeTableName": "[concat(variables('masterVMNamePrefix'),'routetable')]",
+{{else}}
+    "routeTableName": "FIXME",
```
amanohar: Could this route table live in the same resource group as the agents like it does today for a master+Agents cluster and so same naming scheme can be used?

pkg/api/apiloader.go
```diff
@@ -41,6 +42,12 @@ func (a *Apiloader) DeserializeContainerService(contents []byte, validate bool)
 // LoadContainerService loads an ACS Cluster API Model, validates it, and returns the unversioned representation
 func (a *Apiloader) LoadContainerService(contents []byte, version string, validate bool) (*ContainerService, error) {
 	switch version {
+	case v20170831.APIVersion:
```
amanohar: At some point it might make sense to move this to it's separate function like: 
`func (a *Apiloader) LoadAgentPoolOnlyContainerService(contents []byte, version string, validate bool) (*ContainerService, error) { 
...
}
`
The reason is that the deserialized types of the input (`contents`) are different and both containerservice and agentpoolonly type could have same API version depending on release schedule.

amanohar: Also, in the case of AgentPoolsOnly type, certificates will need to be passed as a separate params and not auto generated as is the case for ContainerService

brendandburns:
@amanohar @jackfrancis @sgoings @slack This is ready for review.

Thanks!

pkg/api/apiloader.go
```diff
@@ -104,6 +111,10 @@ func (a *Apiloader) LoadContainerService(contents []byte, version string, valida
 // SerializeContainerService takes an unversioned container service and returns the bytes
 func (a *Apiloader) SerializeContainerService(containerService *ContainerService, version string) ([]byte, error) {
 	switch version {
+	case v20170831.APIVersion:
+		// TODO: implement this...
```
jackfrancis: @brendanburns  @amanohar I assume this PR has value with this unimplemented? (and the resultant conversions that will ostensibly go in `converterfromapi.go`)

pkg/api/types.go
```diff
@@ -310,6 +310,11 @@ type UpgradeContainerService struct {
 	OrchestratorProfile *OrchestratorProfile `json:"orchestratorProfile,omitempty"`
 }
 
+// HasWindows returns true if the cluster contains a master
```
mboersma: Typo: `HasMaster`

amanohar:
@brendandburns more changes/updates on top of this PR at: https://github.com/Azure/acs-engine/pull/1290

amanohar:
Closing this PR. All changes from this PR were part of this PR: https://github.com/Azure/acs-engine/pull/1290 + additional fixes and was merged to master on 8/18.

PR #1207, opened 2017-08-08
feat(outputs): add additional values to output block

slack:

Much of the underlying IaaS resources are computed as part of the ARM template expansion. External consumers of this information don't currently have a structured way to access this data. This PR adds additional values to the `"outputs"` section of the template.

**Release note**:
```release-note
Record additional data for resources created during the evaluation of the ARM template.

Adds the following values:
- resourceGroup
- subnetName
- securityGroupName
- virtualNetworkName
- routeTableName
- primaryAvailablitySetName
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1207)


amanohar:
@slack, LGTM. It might be useful to add to this the agent pool only PR as well.

slack:
One outstanding question is agentPools in different subnets. This layout won't accommodate that configuration, but I'm not sure that is something we support.

amanohar:
@slack yes we do support agent pools in different subnets. ACS Engine also supports outputting params per agent pool. One e.g. is how the storage account names info containing agent pool VHDs are outputted (so that RP can reconstruct storage account names). 
The formula is: <AgentPoolName>StorageAccountCount. e.g.: agentpool1StorageAccountCount
Similarly subnet can be: <AgentPoolName>Subnetname

parts/iaasoutputs.t
```diff
@@ -0,0 +1,24 @@
+    "resourceGroup": {
+        "type": "string",
+        "value": "[variables('resourceGroup')]"
+    },
```
amanohar: From the RP standpoint outputting resource group is not needed since the RP creates the resource group.

amanohar:
@slack, yes we do support agent pools in different subnets.

ACS Engine also supports outputting params per agent pool. One e.g. is how the storage account names info containing agent pool VHDs are outputted (so that RP can reconstruct storage account names).

The formula is: [AgentPoolName] StorageAccountCount. e.g.: agentpool1StorageAccountCount
Similarly subnet can be: [AgentPoolName] Subnetname e.g.: agentpool1Subnetname

This is the file: https://github.com/Azure/acs-engine/blob/master/parts/agentoutputs.t#L8-L15

Following needs to be added to it:
```
"{{.Name}}SubnetName": {
"type": "string",
"value": "[variables('{{.Name}}SubnetName')]"
}
```

seanknox:
@jchauncey @jackfrancis FYI.

slack:
Last question... Since these outputs will be consumed by another system, I'd like to get a test in place... Any recommendations for a good spot?

amanohar:
Best place to do this IMO would be in static tests like: https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/transform_test.go that use: https://github.com/Azure/acs-engine/tree/master/pkg/acsengine/transformtestfiles for expected file contents.

slack:
I ended up placing the tests in engine_test, seemed like that was the spot that was generating fresh templates.

pkg/acsengine/engine_test.go
```diff
@@ -273,3 +274,73 @@ func TestGetStorageAccountType(t *testing.T) {
...
+	}
+	armTemplate, _, _, err := templateGenerator.GenerateTemplate(containerService)
+	if err != nil {
+		t.Fatalf("Failed to generate arm template: %v", err)
+	}
+
+	var template TestARMTemplate
+	json.Unmarshal([]byte(armTemplate), &template)
+
+	// fmt.Printf("Result: %+v\n", template.Outputs)
```
seanknox: commented code you can whack

PR #1246, opened 2017-08-14
Add TillerImageBase for AzureChinaCloudSpec

clement-buchart:
Currently when generating for China Cloud, tillerSpec doesn't have a repository basename : 
```
"kubernetesTillerSpec": {
      "value": "tiller:v2.5.1"
    },
```
This fix it by setting a missing default for AuchanChinaCloud

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1246)


rjtsdl:
@clement-buchart , just curious, what is `AuchanChinaCloud` stand for?

clement-buchart:
@rjtsdl I meant `AzureChinaCloudSpec` which is the `AzureEnvironmentSpecConfig` holding the default configuration for the Chinese Azure cloud (from what i gathered there are actually 4 Azure clouds : the public one, a german one, a chinese one, and one specific to the us gov).

The main reason for this differences in default configuration in the chinese context would be the Great FireWall, blocking access to a lot of ressources, in particular google-hosted gcr.io docker registry.

PR #1222, opened 2017-08-09
[k8s] update --register-schedule with --register-with-taints flags

rjtsdl:
**What this PR does / why we need it**:
Deprecate register-schedule flag from k8s 1.6, keep k8s 1.5 use what it is now

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1215 

**Special notes for your reviewer**:
Ready for review

**Release note**:

```release-note
Using taint is the recommended way going forward.
We should expect to see 

Taints:			node-role.kubernetes.io/master=true:NoSchedule

After 'kubectl describe node masternode'
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1222)

PR #1253, opened 2017-08-15
fix regression introduced by register-with-taints pr

rjtsdl:
Fix E2E regression introduced by #1222 

1. fix "registerWithTaints" variable not available in some cases
2. add a 1.5 kube-proxy template file like other addons

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1253)

### DIFF
--- a/parts/kubernetesmasteraddons-kube-proxy-daemonset1.5.yaml
+++ b/parts/kubernetesmasteraddons-kube-proxy-daemonset1.5.yaml
@@ -1,0 +1,53 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  labels:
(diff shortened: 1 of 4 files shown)


rjtsdl:
I marked it as P1, since it is a fix of regression.

seanknox:
Thanks @rjtsdl, LGTM. After we get https://github.com/Azure/acs-engine/pull/1249 to add the missing e2e coverage we can rebase this branch to run with those tests.

rjtsdl:
@seanknox, one question in my mind is, if #1249 will have better coverage, considering the fix is not in before that, the e2e test for #1249 can never pass, right? 

So i thought, we should get this in first, and let #1249 rebase on that.

PR #1254, opened 2017-08-15
fix(pki_test, upgrader.go): Fix the styling errors in both of these files

jchauncey:
On a clean checkout of master if you run `make test-styles` these two files fail that check. This should fix those errors.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1254)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   test:
     working_directory: /go/src/github.com/Azure/acs-engine
     docker:
-      - image: microsoft/acs-engine:latest
+      - image: quay.io/deis/go-dev:v1.2.0
         environment:
           GOPATH: /go
     steps:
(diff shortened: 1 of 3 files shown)


seanknox:
Good catch—how these sneak in?

jchauncey:
I have no idea how these got in.

seanknox:
Thanks for fixing the gap in the style checks, too! LGTM.

jchauncey:
this is still inflight while i figure out how to better fix the styling check in circle.

scripts/validate-go.sh
```diff
@@ -19,7 +19,6 @@ exit_code=0
 
 if ! hash gometalinter 2>/dev/null ; then
   go get -u github.com/alecthomas/gometalinter
-  gometalinter --install
```
seanknox: I think we need to keep this as first we 1) install gometalinter itself and 2) install the linters themselves.

PR #1256, opened 2017-08-15
Delivers a `make test-kubernetes` ginkgo e2e test

jackfrancis:
with circle config

**What this PR does / why we need it**: an attempt to deliver the minimum necessary diff for @jchauncey's new ginkgo-driven kubernetes e2e

**Special notes for your reviewer**: delivers a new `vendor/` 3rd party dependency: `github.com/kelseyhightower/envconfig`

**Release note**:

```
ginkgo kubernetes e2e tests
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1256)


.circleci/config.yml
```diff
@@ -177,6 +177,27 @@ jobs:
           no_output_timeout: "30m"
       - store_artifacts:
           path: /go/src/github.com/Azure/acs-engine/_logs
+  ginkgo-k8s-e2e:
```
jchauncey: This is missing the part of the configuration that tells circle to run this.

jchauncey:
``` 
    - ginkgo-k8s-e2e:
          requires:
            - e2e-hold
          filters:
            branches:
              ignore: /master/
```

glide.lock
```diff
@@ -40,6 +40,8 @@ imports:
   version: 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
 - name: github.com/JiangtianLi/gettext
   version: a8983c062be4b565d723c478922d7736e04fdba4
+- name: github.com/kelseyhightower/envconfig
```
jackfrancis: I'm open to any way to add `envconfig` to `glide.yaml`, `glide.lock`, and `vendor/` so long as we don't also deliver unrelated dependency changes (such as deleting `glide.lock` and re-`glide install`'ing would do).

jchauncey: the only way to do that is to pin every dep to a sha. so that they never move. As it stands we have some dependencies with patch level constraints which means they could get updated if the glide.lock file needs to be rebuilt (such as adding a new dep).

PR #1257, opened 2017-08-15
adding 'orchestrators' command to print releases/versions for supported orchestrators

dmitsh:
Adding 'info' command to acs-engine, which prints release/version info for supported orchestrators.

The new command will be used for informational purpose, and in cluster validations

Usage examples:

$ ./bin/acs-engine orchestrators
Kubernetes{Release: 1.5, Version: 1.5.7, Default: false}
Kubernetes{Release: 1.7, Version: 1.7.2, Default: false}
Kubernetes{Release: 1.6, Version: 1.6.6, Default: true}
DCOS{Release: 1.9, Version: 1.9.0, Default: true}
DCOS{Release: 1.8, Version: 1.8.8, Default: false}
DCOS{Release: 1.7, Version: 1.7.3, Default: false}
Swarm{Version: 1.1.0, Docker Compose Version: 1.6.2}
DockerCE{Version: 17.03, Docker Compose Version: 1.14.0}

$ ./bin/acs-engine orchestrators --orchestrator DCOS
DCOS{Release: 1.8, Version: 1.8.8, Default: false}
DCOS{Release: 1.7, Version: 1.7.3, Default: false}
DCOS{Release: 1.9, Version: 1.9.0, Default: true}

$ ./bin/acs-engine orchestrators --orchestrator kubernetes --release 1.6
Kubernetes{Release: 1.6, Version: 1.6.6, Default: true}

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1257)


seanknox:
I'm a little confused what need driving this command, which I'm pretty sure will confuse non-ACS users of acs-engine. Are there feature requests for a command by users or is this just to be used in cluster validation tests (and if for validation tests, how will this be used?)?

dmitsh:
Format is not an issue, and can be changed.
I also considered incorporating 'info' with 'version', but wasn't sure.
My concern was overloading the term of "version": acs-engine version vs. versions of supported orchestrators.

To make one command out of these two, we can, for example, use "--full" flag to display both acs-engine version and orchestrator releases/versions.

seanknox:
Based on the PR description, this command doesn't tell a user what version is actually installed, only what is potentially installed. Maybe I'm missing something?

dmitsh:
That is exactly what it does - informs users which orchestration versions are supported, and which version is installed by default per orchestrator. Extra flags are used to customize the query.

sauryadas:
I think this is a useful command to show all the orchestrator release/ versions supported in acs-engine. There is no other way of finding this out today.

cmd/info.go
```diff
@@ -0,0 +1,143 @@
...
+		ver, ok := common.DCOSReleaseToVersion[release]
+		if !ok {
+			return fmt.Errorf("DCOS release %s is not supported", release)
+		}
+		printInfo(release, ver, common.DCOSDefaultRelease)
+	}
+	return nil
+}
+
+func swarmInfo(release string) error {
```
rjtsdl: My concern for swarm and dockerce is, we hard code the version in the fmt.Println. It is not resilience of changes. Since orchestratorRelease option is not available to swarm and dockerce. We can just say "Not an option"?

dmitsh: I think it is important to provide version info for swarm and dockerce.
I will add comments in the scripts where we have that hard-coded.

seanknox:
> Based on the PR description, this command doesn't tell a user what version is actually installed, only what is potentially installed. Maybe I'm missing something?

> That is exactly what it does - informs users which orchestration versions are supported, and which version is installed by default per orchestrator.

actually installed is not the same as what is supported.

dmitsh:
@seanknox Is that a question or a statement? I understand that "actually installed is not the same as what is supported". This command is not about "actually installed" cluster version, but rather "what is supported".

jackfrancis:
This looks great, and a useful feature. My request is that we either provide a heading above each version, e.g.:

```
$ ./bin/acs-engine info
Supported Kubernetes versions:
Kubernetes{Release: 1.5, Version: 1.5.7, Default: false}
Kubernetes{Release: 1.7, Version: 1.7.2, Default: false}
Kubernetes{Release: 1.6, Version: 1.6.6, Default: true}

Supported DCOS versions:
DCOS{Release: 1.9, Version: 1.9.0, Default: true}
DCOS{Release: 1.8, Version: 1.8.8, Default: false}
DCOS{Release: 1.7, Version: 1.7.3, Default: false}

Supported Swarm versions:
Swarm{Version: 1.1.0, Docker Compose Version: 1.6.2}

Supported DockerCE versions:
DockerCE{Version: 17.03, Docker Compose Version: 1.14.0}
```

Or we change the option string from `info` to something a little more narrowly defined, e.g., `supported`, or `orchestrators`. My thinking is that `info` is very general, and we may want to add additional descriptive build data in the future, so having a heading visually guides the user through the various sub-types of informational data.

dmitsh:
@jackfrancis agree, "orchestrators" is way descriptive. Changing now.

parts/configure-swarm-cluster.sh
```diff
@@ -6,6 +6,7 @@ echo "starting swarm cluster configuration"
 date
 ps ax
 
+# **NOTE** Keep version numbers in sync with constants in cmd/info.go
 SWARM_VERSION="swarm:1.1.0"
```
amanohar: If we really want the "info/orchestrator" cmd to be in sync and up to date then this should be passed as inputs to the shell script files (and have a const in go code). Otherwise these values can easily go out of sync. Otherwise LGTM @dmitsh

jackfrancis: I agree, is there any way to eliminate manually managing these values in two places?

dmitsh: "The sky's the limit" :)
I tried to keep it simple, but your guys perfectionism took over :)

PR #1258, opened 2017-08-15
Enable the StorageAccount Encryption and Enforce HTTPS

robbiezhang:
The PR enables the [encryption at rest](https://docs.microsoft.com/en-us/azure/storage/storage-security-guide#encryption-at-rest) and enforces [encryption in transit](https://docs.microsoft.com/en-us/azure/storage/storage-security-guide#encryption-in-transit) for the StorageAccount provisioned by acs-engine for Kubernetes.

It's a requirement from the internal security review of CloudShell.

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1258)

### DIFF
--- a/parts/kubernetesagentresourcesvmas.t
+++ b/parts/kubernetesagentresourcesvmas.t
@@ -52,7 +52,7 @@
         {
             "platformFaultDomainCount": "2",
             "platformUpdateDomainCount": "3",
-		"managed" : "true"
+            "managed" : "true"
         },
   
       "type": "Microsoft.Compute/availabilitySets"
@@ -77,14 +77,14 @@
           "keySource": "Microsoft.Storage",
           "services": {
             "blob": {
-              "enabled": "true"
+              "enabled": true
             },
             "file": {
-              "enabled": "true"
+              "enabled": true
             }
           }
         },
-        "supportsHttpsTrafficOnly": "true"
+        "supportsHttpsTrafficOnly": true
       },
       "sku": {
         "name": "[variables('vmSizesMap')[variables('{{.Name}}VMSize')].storageAccountType]"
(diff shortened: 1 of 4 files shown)


parts/kubernetesagentresourcesvmas.t
```diff
@@ -67,10 +67,25 @@
...
+            "file": {
+              "enabled": "true"
+            }
+          }
+        },
+        "supportsHttpsTrafficOnly": "true"
+      },
+      "sku": {
+        "name": "[variables('vmSizesMap')[variables('{{.Name}}VMSize')].storageAccountType]"
       },
```
rjtsdl: qq: where is the spec of ARM template documented?

rjtsdl: Well, i think i found it.
https://docs.microsoft.com/en-us/rest/api/storagerp/storageaccounts#StorageAccounts_Update

parts/kubernetesmasterresources.t
```diff
@@ -24,10 +24,25 @@
...
+      "kind": "Storage",
       "location": "[variables('location')]",
       "name": "[variables('masterStorageAccountName')]",
       "properties": {
-        "accountType": "[variables('vmSizesMap')[variables('masterVMSize')].storageAccountType]"
+        "encryption": {
+          "keySource": "Microsoft.Storage",
+          "services": {
+            "blob": {
+              "enabled": "true"
```
rjtsdl: As documented in the spec, "enabled" is a boolean type, do we need to use "true", or it can be just true.
It maybe ok, because SRP is mostly written in c# code. When unmarshal, string or boolean can both be handled nicely. 

However, it is not true for other unmarshal packages. Like in golang, it need to be exactly matched. So I suggest to use true, not "true"

robbiezhang: I thought it was a convention to always use string in the ARM templateI, because I saw several int and bool properties using string.

Anyway, I'll change it to boolean type.

rjtsdl: Thx, i approved.

rjtsdl:
@anhowe , to also TAL

rjtsdl:
@robbiezhang , looks the e2e tests failed. AndI took a look at the error, it is something we fixed yesterday. So you should just need to rebase on the master, the test should pass afterwards.

rjtsdl:
I guess for this scenario, an E2E test makes more sense?

clementnero:
This is the following of this closed PR #406.
Good idea to add "supportsHttpsTrafficOnly" property.
I agree with @anhowe, enabling this feature via a flag would be great.
Thx a lot.

PR #1259, opened 2017-08-15
Update CNI to the 0.9 release

lachie83:
This change updates the following:

- Azure CNI plugin to 0.9
- Introduces new vnetCidr key to exclude from the network masquerade block

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1259)

### DIFF
--- a/parts/kubernetesmastercustomscript.sh
+++ b/parts/kubernetesmastercustomscript.sh
@@ -191,7 +191,8 @@ function configAzureNetworkPolicy() {
     mkdir -p $CNI_BIN_DIR
 
     # Mirror from https://github.com/Azure/azure-container-networking/releases/tag/$AZURE_PLUGIN_VER/azure-vnet-cni-linux-amd64-$AZURE_PLUGIN_VER.tgz
-    downloadUrl https://github.com/ofiliz/hello/releases/download/pre-release/azure-vnet-cni-linux-amd64-v0.8-7-g8acef7d.tgz | tar -xz -C $CNI_BIN_DIR
+    # TODO: Migrate back to acs-mirror.azureedge.net/cni/azure-vnet-cni-linux-amd64-latest.tgz once available
+    downloadUrl https://github.com/Azure/azure-container-networking/releases/download/v0.9/azure-vnet-cni-linux-amd64-v0.9.tar.gz | tar -xz -C $CNI_BIN_DIR
     # Mirror from https://github.com/containernetworking/cni/releases/download/$CNI_RELEASE_VER/cni-amd64-$CNI_RELEASE_VERSION.tgz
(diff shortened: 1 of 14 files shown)


parts/kubernetesmastercustomscript.sh
```diff
@@ -186,7 +186,7 @@ function configAzureNetworkPolicy() {
     mkdir -p $CNI_BIN_DIR
 
     # Mirror from https://github.com/Azure/azure-container-networking/releases/tag/$AZURE_PLUGIN_VER/azure-vnet-cni-linux-amd64-$AZURE_PLUGIN_VER.tgz
-    downloadUrl https://acs-mirror.azureedge.net/cni/azure-vnet-cni-linux-amd64-latest.tgz | tar -xz -C $CNI_BIN_DIR
+    downloadUrl https://github.com/ofiliz/hello/releases/download/pre-release/azure-vnet-cni-linux-amd64-v0.8-7-g8acef7d.tgz | tar -xz -C $CNI_BIN_DIR
```
jackfrancis: I don't think we don't want to merge this to master until we test/vet the external `azure-vnet-cni-linux-amd64` download and then host it in `https://acs-mirror.azureedge.net/cni`.

parts/kubernetesmastervars.t
```diff
@@ -102,6 +102,7 @@
     "subnetName": "[split(parameters('masterVnetSubnetID'), '/')[variables('subnetNameResourceSegmentIndex')]]",
     "vnetNameResourceSegmentIndex": 8,
     "virtualNetworkName": "[split(parameters('masterVnetSubnetID'), '/')[variables('vnetNameResourceSegmentIndex')]]",
+    "vnetCidr": "10.0.0.0/8",
```
lachie83: I will make this configurable in the api model

parts/kubernetesmastercustomscript.sh
lachie83: Correct!

pkg/acsengine/defaults.go
```diff
@@ -158,6 +158,7 @@ func setMasterNetworkDefaults(a *api.Properties) {
 				// When VNET integration is enabled, all masters, agents and pods share the same large subnet.
 				a.MasterProfile.Subnet = a.OrchestratorProfile.KubernetesConfig.ClusterSubnet
 				a.MasterProfile.FirstConsecutiveStaticIP = getFirstConsecutiveStaticIPAddress(a.MasterProfile.Subnet)
+				a.MasterProfile.VnetCidr = (a.MasterProfile.VnetCidr)
```
lachie83: @jackfrancis is this required given that I set a default for the param in kubernetesparams.t

jackfrancis: Yeah, I'm not sure what that statement will accomplish. Do you get the expected behavior in both cases (default and user-provide) without that statement?

lachie83: Yes. I've yanked this

PR #1260, opened 2017-08-16
chore(circleci/config.yml): Move to using deis/go-dev in CI instead of the acs-engine image

jchauncey:


seanknox:
LGTM!

jchauncey:
deis/go-dev is a versioned development image that has versioned components needed for ci and local development. This makes our CI environment more predictable. Which means we shouldnt have failures related to a misconfigured CI env

jackfrancis:
Filed this to migrate the deis dev image to a longer-term ACS-sponsored image: https://github.com/Azure/acs-engine/issues/1266

PR #1270, opened 2017-08-16
add merge workflow between load raw api model and validation

### LINKED ISSUE #1269: Validation failed on scale command

---

ISSUE
---

all the versions

---

Currently, there validation of user's request to scale always failed. Because, when user scale, the servicePrincipalProfile.Secret is left empty. The validation need it not empty.

Kubernetes

**What happened**:

For scale, which is a update scenario, we should merge with existing containerService properties. In this case, it is servicePrincipalProfile.Secret. The we do validation.

rjtsdl:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1269

**Release note**:

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1270)

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -50,9 +50,9 @@ import:
   version: a8983c062be4b565d723c478922d7736e04fdba4
 - package: github.com/Masterminds/semver
   version: v1.3.1
+- package: github.com/imdario/mergo
+  version: ^0.2.2
 testImport:
 - package: github.com/onsi/gomega
 - package: github.com/onsi/ginkgo
-  version: v1.3.1
 - package: github.com/kelseyhightower/envconfig
-  version: ~1.3.0
--- a/glide.yaml
+++ b/glide.yaml
@@ -55,4 +55,6 @@ import:
 testImport:
 - package: github.com/onsi/gomega
 - package: github.com/onsi/ginkgo
+  version: v1.3.1
 - package: github.com/kelseyhightower/envconfig
+  version: ~1.3.0
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -108,7 +108,7 @@ func (dc *deployCmd) validate(cmd *cobra.Command, args []string) {
 		},
 	}
 	// skip validating the model fields for now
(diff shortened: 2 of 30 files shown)


rjtsdl:
@seanknox , the package below is under BSD 3-clause. It is also used in acsrp. So if it is not registered, please help.
```
github.com/imdario/mergo
```

jchauncey:
ginkgo tests failed with a known issue that should be fixed with #1292

glide.yaml
```diff
@@ -50,9 +50,9 @@ import:
   version: a8983c062be4b565d723c478922d7736e04fdba4
 - package: github.com/Masterminds/semver
   version: v1.3.1
+- package: github.com/imdario/mergo
+  version: ^0.2.2
 testImport:
 - package: github.com/onsi/gomega
 - package: github.com/onsi/ginkgo
-  version: v1.3.1
```
jackfrancis: we don't want to lose these version declarations

glide.yaml
```diff
@@ -50,9 +50,9 @@ import:
...
 - package: github.com/kelseyhightower/envconfig
-  version: ~1.3.0
```
jackfrancis: ditto

jackfrancis: @JiangtianLi just add back those version strings so the diff is clean for these two packages, but don't worry about re-running glide or anything

glide.yaml
rjtsdl: Well, i just cannot avoid conflict with testImport with version specified.

jchauncey: then you are running into the exact problem i was seeing earlier when i was trying to get glide to reconcile

rjtsdl: How is your workaround? Mine was just remove the versions of testImport. @jchauncey

jchauncey: i eventually stopped fiddling with it since it wasnt pressing.

glide.yaml
rjtsdl: So you are saying, i can keep glide.lock not synced with glide.yaml?

rjtsdl: I guess, it is yes :)

PR #1272, opened 2017-08-16
Notes on 3rd party dependencies

jackfrancis:
**What this PR does / why we need it**: add docs on how we manage 3rd party dependencies in the project

**Release note**:

```
dependencies documentation
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1272)

### DIFF
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -54,6 +54,14 @@ workflow for doing this is as follows:
 5. When you are ready for us to review, push your branch to GitHub, and
    then open a new pull request with us.
 
+### Third Party Dependencies
+
+Third party dependencies reside locally inside the repository under the `vendor/` directory. We use [glide](https://github.com/Masterminds/glide) to enforce our dependency graph, declared in [glide.yaml](https://github.com/Azure/acs-engine/blob/master/CONTRIBUTING.md) in the project root.
+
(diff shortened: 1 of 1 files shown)


jackfrancis:
@anhowe @seanknox @jchauncey FYI, a 1st draft on the language that describes how we intend to express a structured dependency management strategy

seanknox:
This is great. Worth mentioning that all dependencies need to be registered with Microsoft OSS to be cleared for use?

docs/developers.md
```diff
@@ -54,6 +54,14 @@ workflow for doing this is as follows:
 5. When you are ready for us to review, push your branch to GitHub, and
    then open a new pull request with us.
 
+### Third Party Dependencies
```
jchauncey: Ive added an issue about updating labels (#1277) so when people submit prs that need dependency updates/additions we can track them.

jchauncey:
My only comment to this is I wonder if this should be some where more visible?

jackfrancis:
@seanknox I'm inclined to defer on how much info we publish about that. Who would be the best resource to determine how we want to say that?

seanknox:
> @seanknox I'm inclined to defer on how much info we publish about that. Who would be the best resource to determine how we want to say that?

Fair enough. As long as there's a process we (ACS) follow to ensure dependencies are registered, that's what matters.

PR #1285, opened 2017-08-17
default to tiller 2.6.0 for k8s 1.7

lachie83:
Helm 2.6 is compiled against k8s 1.7 https://github.com/kubernetes/helm/releases/tag/v2.6.0

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1285)


rjtsdl:
If helm 2.6 also works with k8s 1.6, we can make tiller 2.6 as default, and change k8s 1.5 to use tiller 2.5.1 as an exception.

bacongobbler:
It works with Kubernetes 1.6. I've been using canary releases of Helm on Kubernetes v1.6 for a while now. No issues.

PR #1287, opened 2017-08-17
change master count to 1 for k8s 1.6/1.7 example files

skriss:
Signed-off-by: Steve Kriss <steve@heptio.com>

The kubernetes 1.6 & 1.7 example files have a master count of 3. All other kubernetes examples have a master count of 1 (which is what I would expect as a user). I believe this was likely just a fat-finger in the past that got carried over. Reverting back to 1.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1287)

### DIFF
--- a/examples/kubernetes-releases/kubernetes1.6.json
+++ b/examples/kubernetes-releases/kubernetes1.6.json
@@ -6,7 +6,7 @@
       "orchestratorRelease": "1.6"
     },
     "masterProfile": {
-      "count": 3,
+      "count": 1,
       "dnsPrefix": "",
       "vmSize": "Standard_D2_v2"
     },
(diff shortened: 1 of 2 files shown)


examples/kubernetes-releases/kubernetes1.6.json
```diff
@@ -6,7 +6,7 @@
       "orchestratorRelease": "1.6"
     },
     "masterProfile": {
-      "count": 3,
+      "count": 1,
       "dnsPrefix": "",
```
rjtsdl: AFAIK, there is difference in template generation according to masterCount 1 or not. 
For masterCount == 1, we don't put a internal load balancer in front. While otherwise, we do.
So i guess having count == 3, is to test more generic scenario.

skriss: OK; I was under the impression these files were examples for end-users to consume rather than test scenarios based on the fact that they were in the examples/ dir.  Might be worth moving them into an appropriately-named directory if they're test scenarios.

rjtsdl: Yeah, we were using them for testing :(

PR #1289, opened 2017-08-17
update 1.7 release to kubernetes v1.7.4

### LINKED ISSUE #1288: Support Kubernetes 1.7.4
Looks like 1.7.4 has further changes that help with setups where the VNET is outside the Kubernetes Resource Group

lachie83:
closes #1288 

This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1289)

### DIFF
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -84,7 +84,6 @@ const (
 // For instance, Kubernetes release "1.7" would contain the version "1.7.2"
 var KubeConfigs = map[string]map[string]string{
 	api.KubernetesRelease1Dot7: {
-		"version":         "1.7.4",
 		"hyperkube":       "hyperkube-amd64:v1.7.4",
 		"dashboard":       "kubernetes-dashboard-amd64:v1.6.1",
 		"exechealthz":     "exechealthz-amd64:1.2",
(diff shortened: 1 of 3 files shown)


pkg/acsengine/const.go
```diff
@@ -84,8 +84,8 @@ const (
 // For instance, Kubernetes release "1.7" would contain the version "1.7.2"
 var KubeConfigs = map[string]map[string]string{
 	api.KubernetesRelease1Dot7: {
-		"version":         "1.7.2",
-		"hyperkube":       "hyperkube-amd64:v1.7.2",
+		"version":         "1.7.4",
```
rjtsdl: This is not used, we can remove "version" from KubeConfigs.

jackfrancis: @lachie83 I think what @rjtsdl is saying is that the `version` property inside this `map[string]string` is no longer used to determine the specific release version for the `1.7` channel. (Note that it's not there for `1.6` or `1.5` below.) This is an oversight, left over from when we deprecated its usage.

I think it's appropriate to use this PR to fix that! So, would you kindly delete the `"version"` string property altogether from `api.KubernetesRelease1Dot7`?

lachie83:
Thanks @rjtsdl. I've yanked the version key under config.

PR #687, opened 2017-05-27
Extensions support

RobBagby:
**What this PR does / why we need it**:
This PR adds extension support to acs-engine.  Extensions provide an easy way for acs-engine users to add pre-packaged functionality into their cluster.  For example, an extension could configure a monitoring solution on an ACS cluster.  Other examples include: OMS, Helm, Helm charts, Deis Workflow, Sysdig, Vamp and more.  The user would not need to know the details of how to install the monitoring solution.  Rather, the user would simply add the extension into the extensionsProfile section of the template and the masterProfile as below:

```jacascript
{
  "masterProfile": {
      "count": 3,
      "dnsPrefix": "dnsprefix",
      "vmSize": "Standard_D2_v2",
      "osType": "Linux",
      "firstConsecutiveStaticIP": "10.240.255.5",
      "extensions": [
        { 
          "name": "hello-world-k8s"
        }
     ]
  },
  "extensionsProfile": [
    {
        "name": "hello-world-k8s",
        "version": "v1",
        "extensionParameters": "parameters"
    }
  ]
}
```

**How are extensions implemented**
Extensions take advantage of the following ARM constructs: Linked Templates, Custom Script Extensions.  A typical extension would consist of the following files:

|File Name|Description|
|-----------------------------|---|
|template.json               |The ARM template used to deploy the extension|
|template-link.json          |The ARM template snippet which will be injected into azuredeploy.json to call template.json|
|EXTENSION-NAME.sh           |The script file that will execute on the VM itself via Custom Script Extension to perform installation of the extension|
|supported-orchestrators.json |Defines what orchestrators are supported by the extension (Swarm, Dcos, or Kubernetes).  This will be extended in the future to contain more than just supported orchestrators.  Dependencies between extensions, for example will be added here.|

**MVP**
The minimum viable product for extensions will run a script via the above ARM constructs on a master node.  This will work for the scenarios where the extension is deploying a container/pod/daemonset/etc.  Future versions will allow for running scripts on one or more agents in agent pools.

**Extensions**
Although this PR only contains 2 hello world extensions, several extensions have been written and will be added by subsequent PRs.  The extensions written include:
- OMS for Kubernetes
- Sysdig for Kubernetes
- Helm for Kubernetes (installs Helm)
- VAMP for DC/OS

**Testing Extensions**
Because extensions will exist in the GitHub repo, a "rootURL" property exists for an extension in the extensionsProfile:
```javascript
"extensionsProfile": [
      { 
        "name": "hello-world-k8s", 
[...]

Issue #599, opened 2017-05-11
Extensions Design


RobBagby:
# Microsoft Azure Container Service Engine - Extensions

Extensions in acs-engine will provide an easy way for acs-engine users to add pre-packaged functionality into their cluster.  For example, an extension could configure a monitoring solution on an ACS cluster.  The user would not need to know the details of how to install the monitoring solution.  Rather, the user would simply add the extension into the extensionsProfile section of the template.

At a high level, extensions simply take advantage of a couple of ARM constructs: linked templates and usually custom script extensions.  The typical extension will simply add a Microsoft.Resources/deployments resource to the ARM template that points to a linked template.  The linked template will typically have a custom script extension that points at a script file.

This issue outlines the design for extensions. 

# extensionsProfile

The extensionsProfile contains the extensions that the cluster will install. The following illustrates a template with a hello-world-k8s extension.

``` json
{ 
  ...
  "extensionsProfile": [
    {
        "name": "hello-world-k8s",
        "version": "v1",
        "extensionParameters": "parameters",
        "rootURL": "http://mytestlocation.com/hello-world-k8s/"
    }
  ]
}
```

|Name|Required|Description|
|---|---|---|
|name|yes|the name of the extension.  This has to exactly match the name of a folder under the extensions folder|
|version|yes|the version of the extension.  This has to exactly match the name of the folder under the extension name folder|
|extensionParameters|optional|extension parameters may be required by extensions.  The format of the parameters is also extension dependant.|
|rootURL|optional|url to the root location of extensions.  The rootURL must have an extensions child folder that follows the extensions convention.  The rootURL is mainly used for testing purposes.|

# rootURL
You normally would not provide a rootURL.  The extensions are normally loaded from the extensions folder in GitHub.  However, you may specify the rootURL when testing a new extension.  The rootURL must adhere to the extensions conventions.  For example, in order to use an Azure Storage account to test an extension named extension-one, you would do the following:
- Create a storage account.  For the purposes of this example, we will call it 'mystorageaccount'
- Create a blob container called 'extensions'
- Under 'extensions', create a folder called 'extension-one'
- Under 'extension-one', create a folder called 'v1'
- Under 'v1', upload your files (see Required Extension Files) 
- Set the rootURL to: 'https://mystorageaccount.blob.core.windows.net/'

# masterProfile, agentPoolProfiles
[...]

PR #1304, opened 2017-08-19
Rob bagby extension support

JackQuincy:
This enables custom extensions that can be shared with other devs. 
#599
This is just a restart on this pr
#687

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1304)


glide.lock
```diff
@@ -98,7 +98,7 @@ imports:
   version: 5bf94b69c6b68ee1b541973bb8e1144db23a194b
 - name: github.com/Sirupsen/logrus
   version: ba1b36c82c5e05c4f912a88eab0dcd91a171688f
-  repo: https://github.com/sirupsen/logrus
+  repo: https://github.com/Sirupsen/logrus
```
jackfrancis: Please remove this change from this PR, we'll manage this in a separate `glide` PR, thanks!

glide.yaml
```diff
@@ -20,7 +20,7 @@ import:
   - autorest/date
 - package: github.com/Sirupsen/logrus
   version: v0.11.5
-  repo: https://github.com/sirupsen/logrus
+  repo: https://github.com/Sirupsen/logrus
```
jackfrancis: ditto

pkg/acsengine/const.go
```diff
@@ -158,6 +158,11 @@ var KubeConfigs = map[string]map[string]string{
 }
 
 const (
+	// Root URL for extensions
+	DefaultExtensionsRootURL = "https://raw.githubusercontent.com/Azure/acs-engine/master/"
```
jackfrancis: Could you move this `const` down to the pre-existing `const` block?

glide.lock
JackQuincy: I couldn't build this PR without it, so I'll leave it here and remove it after the other Pr goes through

pkg/acsengine/const.go
JackQuincy: this file has a bunch of them but these are all urls so sure I can move this one.

JackQuincy:
So there has been a change in kubernetes provision script that isn't compatiable with this design. I'm going to be closing this until I can adjust the design for the new changes.

Issue #1310, opened 2017-08-21
Fix timing issue with ginkgo k8s tests


jchauncey:
So we are seeing a weird intermittent issue where the ginkgo tests are unable to ssh into the master node. 

```
ssh -i /Users/chauncey/go/src/github.com/Azure/acs-engine/_output/id_rsa -o ConnectTimeout=10 -o StrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null azureuser@test-7455
Warning: Permanently added 'test-7455,ip' (ECDSA) to the list of known hosts.
Received disconnect from ip port 22:2: Too many authentication failures
Authentication failed.
```

PR #1311, opened 2017-08-21
fix(test/e2e/kubernetes_test): Fix the access dashboard test

jchauncey:
* Update logic around waiting for pod readiness
* Make sure we add the ssh key to the agent before running the tests
* If we get an error in this test output the stdout/stderr

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1311)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ pkg/i18n/translations.go
 _logs/
 test/acs-engine-test/report/TestReport.json
 *.swp
+
+# I have no idea why these get generated when I run the e2e test
+test/e2e/kubernetes/translations/ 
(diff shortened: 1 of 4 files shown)


jchauncey:
closes #1311

.gitignore
```diff
@@ -22,3 +22,6 @@ pkg/i18n/translations.go
 _logs/
 test/acs-engine-test/report/TestReport.json
 *.swp
+
+# I have no idea why these get generated when I run the e2e test
+test/e2e/kubernetes/translations/
```
jackfrancis: Are these generated by these ginkgo tests?

jchauncey: as far as i can tell yes they are generated by running something that is part of the e2e tests (maybe the az cli does it?). it was annoying me that i kept having to remove them so figured i would just add it to the gitignore

jackfrancis:
@jchauncey can you fix the "closes #<this PR>" comment above so we know which issue it fixes?

jchauncey:
closes #1310

PR #1314, opened 2017-08-21
update logrus casing in a couple spots

JackQuincy:
make build fails with out this.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1314)


glide.lock
```diff
@@ -98,7 +98,7 @@ imports:
   version: 5bf94b69c6b68ee1b541973bb8e1144db23a194b
 - name: github.com/Sirupsen/logrus
   version: ba1b36c82c5e05c4f912a88eab0dcd91a171688f
-  repo: https://github.com/sirupsen/logrus
+  repo: https://github.com/Sirupsen/logrus
```
seanknox: Isn't the project sirupsen/logrus (lowercase) now?

JackQuincy: possibly. I know that with name and repo different i had failing make build

JackQuincy: if I try to switch to lower case it fails with a import collision issue at compile time. might be one of our dependencies hasn't switched yet in the version we have vendored.

jchauncey: As far as I can tell if you pull in a transitive dependency update where they have renamed their package import to use the lowercase version of `Sirupsen` then we must also use the lower case version because I dont think glide can handle that translation properly. That being said, changing repo to be upper case will probably break things in a more significant way.

We should probably decide immediately if we are going to port over to the lowercase version of sirupsen or keep trying to triage these issues

JackQuincy: sounds like I messed this up. I had to run glide install --force

JackQuincy:
was a user error not needed

PR #1317, opened 2017-08-22
E2E changes for regional regression testing

jackfrancis:
**What this PR does / why we need it**: Enable regional randomization for E2E deployment configurations. This makes static "location" configuration optional, and randomizes regions by default.

**Release note**:

```NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1317)


test/acs-engine-test/main.go
```diff
@@ -150,6 +153,11 @@ func (m *TestManager) testRun(d config.Deployment, index, attempt int, timeout t
 		rgPrefix = "y"
 		fmt.Printf("RESOURCE_GROUP_PREFIX is not set. Using default '%s'\n", rgPrefix)
 	}
+	// Randomize region if no location was configured
+	if d.Location == "" {
+		randomIndex := rand.Intn(len(m.regions))
```
jchauncey: Ok so I noticed locally that unless I did -

```
r := rand.New(rand.NewSource(time.Now().UnixNano()))
```

I could not rely on the RNG to give me back a truly distributed random number. Not that you need to do that here but just something I noticed when generating random suffixs for resource groups.

test/acs-engine-test/main.go
```diff
@@ -450,6 +458,25 @@ func mainInternal() error {
...
+		case "westcentralus": // TODO re-enable when this region's reliability has been upgraded
+		case "centraluseuap": // TODO determine why this region is flaky
+		case "brazilsouth": // canary region
+		default:
+			regions = append(regions, region)
+		}
+	}
+	testManager.regions = regions
+	// seed random number generator
+	rand.Seed(time.Now().Unix())
```
jackfrancis: @jchauncey we seed only once here, which in my experience was sufficient to yield random distribution in all the subsequent (concurrent) `rand.Intn` invocations.

Issue #1318, opened 2017-08-22
Tiller cannot be upgraded by users


seanknox:
**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: v0.5.0

---

Tiller resources are defined with `kubernetes.io/cluster-service: "true"` label which the addon-manager interprets as an uneditable resource. Users who try to bump the Tiller version will have their bump reverted by addon-manager.

`kubernetes.io/cluster-service: "true"` is deprecated and should be removed from resources for the more explicit `addonmanager.kubernetes.io/mode: "EnsureExists"` or `addonmanager.kubernetes.io/mode: "Reconcile"` notation. 

This is an issue just against our Tiller install, but `kubernetes.io/cluster-service: "true"` should eventually be replaced across all other resources using it.

PR #1319, opened 2017-08-22
fix(tiller): allow tiller resources to be edited by users

seanknox:
**What this PR does / why we need it**: Allows Tiller resources to be upgraded by users.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/Azure/acs-engine/issues/1318 and https://github.com/Azure/ACS/issues/55

**Release note**:

```release-note
Use `addonmanager.kubernetes.io/mode=EnsureExists` label with Tiller
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1319)

### DIFF
--- a/parts/kubernetesmasteraddons-tiller-deployment.yaml
+++ b/parts/kubernetesmasteraddons-tiller-deployment.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: kube-system
   labels:
     kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: "EnsureExists"
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
@@ -12,6 +13,7 @@ metadata:
   name: tiller
   labels:
     kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: "EnsureExists"
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
@@ -28,6 +30,7 @@ metadata:
     app: helm
     name: tiller
     kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: "EnsureExists"
   name: tiller-deploy
   namespace: kube-system
 spec:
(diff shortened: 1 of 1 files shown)


parts/kubernetesmasteraddons-tiller-deployment.yaml
```diff
@@ -4,14 +4,14 @@ metadata:
   name: tiller
   namespace: kube-system
   labels:
-    kubernetes.io/cluster-service: "true"
```
lachie83: I would leave this annotation here and strip it out in another PR

lachie83:
One nit otherwise LGTM

parts/kubernetesmasteraddons-tiller-deployment.yaml
seanknox: Good call, amended.

parts/kubernetesmasteraddons-tiller-deployment.yaml
```diff
@@ -5,13 +5,15 @@ metadata:
   namespace: kube-system
   labels:
     kubernetes.io/cluster-service: "true"
```
jackfrancis: Don't we want to remove this deprecated label?

jackfrancis: @lachie83 do you approve of retaining the `kubernetes.io/cluster-service: "true"` label with the introduction of `addonmanager.kubernetes.io/mode=EnsureExists`?

seanknox: @jackfrancis @lachie83  If they're both present, `addonmanager.kubernetes.io/mode=EnsureExists` will take precedence. If only `kubernetes.io/cluster-service: "true"` is present then addon-manager will treat it as `addonmanager.kubernetes.io/mode=Reconcile`. 

Lachie suggested removing the deprecation in another PR which is OK by me. That said, I think it's a lot clearer to remove the deprecated label and use the actual, used value for new cluster.

PR #1323, opened 2017-08-22
fix(test/e2e/kubernetes): Bring k8s e2e tests to parity with bash tests

### LINKED ISSUE #1278: Bring ginkgo tests to feature parity with current bash tests
iteration 0 of the ginkgo runner was just to get the ball rolling on an e2e suite written in go. Now we need to fill in the holes between the two test suites so we can cut over the testing to the new ginkgo runner.

I will update this issue with the coverage that needs to be added as I investigate. 

* Add checks for heapster, kube-addonmanager, kube-apiserver, kube-controller-manager, kube-scheduler, tiller
* Make node count check utilize the cluster definition model 
* test for windows deployments

jchauncey:
* closes #1278
* add tests for heapster, kube-addon-manager, kube-apiserver, kube-controller-manager, kube-scheduler, tiller
* adds logic for dynmamically running the linux and windows agent tests
* assertions made about node count use the cluster definition to determine the expected count

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1323)

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"net/http"
 	"os"
 	"os/exec"
@@ -22,8 +23,9 @@ import (
 
 var (
 	cfg  config.Config
-	err  error
 	acct azure.Account
+	eng  engine.Engine
+	err  error
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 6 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -83,95 +90,163 @@ var _ = AfterSuite(func() {
...
+				r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
seanknox: That is a mighty line. 😆

jchauncey: Its from the `simpleweb-windows` deployment manifest. I would rather not rely on the external file if I can help it so doing this seemed like a better compromise

PR #1332, opened 2017-08-23
unified ServicePrincipalProfile interface between vlab and versioned API

dmitsh:
unified ServicePrincipalProfile interface between vlab and versioned API

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1332)


nalthoff:
@dmitsh  I still see references to "ServiceClientID" in some of the examples like in the "scale up" folder.  Also, I see it in a few template files (example: \acs-engine\parts\kubernetesagentresourcesvmas.t ) 
Do those need to be updated as well?

PR #1352, opened 2017-08-25
e2e(k8s): keep only hybrid windows scenario in e2e

seanknox:
Hybrid clusters (Windows + Linux) e2e tests capture if Windows nodes are functional which frees us to remove Windows-only e2e tests. This removes `windows/kubernetes.json` from 

- `test/acse-conf/acse-feature-validation.json`
- `test/acse-conf/acse-regression.json`

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1352)

### DIFF
--- a/test/acse-conf/acse-feature-validation.json
+++ b/test/acse-conf/acse-feature-validation.json
@@ -5,10 +5,6 @@
       "location": "eastus"
     },
     {
-      "cluster_definition": "windows/kubernetes.json",
-      "location": "eastus"
-    },
-    {
       "cluster_definition": "dcos-releases/dcos1.8.json",
       "location": "eastus"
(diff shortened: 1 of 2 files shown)


jackfrancis:
@jchauncey would you like to note the ginkgo failure before we merge?

jchauncey:
thanks. that node count failure has started cropping up some. Ill deflake shortly

jackfrancis:
Clarified with @anhowe that we can't remove the windows-only test because it is indirectly proving out windows masters as schedulable (which hybrid clusters don't).

Long-term, we'll incorporate all this coverage into the minimum number of clusters using the next-gen ginkgo runner.

PR #1356, opened 2017-08-25
added e2e test cleanup script

jackfrancis:
**What this PR does / why we need it**: Adds a resource group cleanup script so we can periodically clean up e2e resources when our e2e runs fail in unexpected ways and don't clean up after themselves

**Release note**:

```NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1356)

### DIFF
--- a/test/e2e/cleanup.sh
+++ b/test/e2e/cleanup.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+####################################################
+
+if [ -z "$SERVICE_PRINCIPAL_CLIENT_ID" ]; then
+    exit 1;
+fi
+
+if [ -z "$SERVICE_PRINCIPAL_CLIENT_SECRET" ]; then
+    exit 1;
+fi
+
+if [ -z "$TENANT_ID" ]; then
+    exit 1;
+fi
+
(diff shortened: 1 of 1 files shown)


test/e2e/cleanup.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+####################################################
+
+if [ -z "$SERVICE_PRINCIPAL_CLIENT_ID" ]; then
+    exit 1;
```
jchauncey: Probably good idea to say what you are missing in each of these checks

test/e2e/cleanup.sh
```diff
@@ -0,0 +1,44 @@
...
+if [ -z "$TENANT_ID" ]; then
+    exit 1;
+fi
+
+if [ -z "$SUBSCRIPTION_ID_TO_CLEANUP" ]; then
+    exit 1;
+fi
+
+if [ -z "$EXPIRATION_IN_HOURS" ]; then
+    EXPIRATION_IN_HOURS=2
```
jchauncey: good idea to echo this value so people looking at the console output know how far back we start

PR #1365, opened 2017-08-29
feat(runner.go): port bash ginkgo runner to go

jchauncey:
* Ports bash runner to golang
* Updates the version check for the k8s test so that it uses the version specified in the cluster definition if provided
* Adds some retry logic to the pod networking test since that was failing locally
* Adds retry logic when provisioning fails (up to 3 times overridden using PROVISION_RETRIES)

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1365)


jchauncey:
going to close and reopen this pr to see if i can get circle to pick up the right commit.

PR #1378, opened 2017-08-31
Set osdisk name unconditionally

malekpour:
This PR will set osdisk name for all storage types regardless of their types. Currently if the osdisk is not using storage account, a random name will be generated which is not consistent with other resources.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1378)


jackfrancis:
@anhowe do you have any thoughts on this change? Your PR here initially introduced the semantic storage `name` value based on managed disk option being true:

https://github.com/Azure/acs-engine/pull/625

Is there a good reason to keep the existing condition in place for making this `name` value assignment?:

`"[concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')),'-osdisk')]"`

anhowe:
@jackfrancis this should be fine to have all os disks named the same way.

anhowe:
@jackfrancis Thinking further I recommend verifying the following:
1. deploy a cluster with storage accounts
2. re-generate the template using apimodel.json from output directory, but now with the name, and redeploy.

If the above two work, then we should be good.  Otherwise, this should only be applied to vlabs.

PR #1381, opened 2017-08-31
K8s v1.7.5

seanknox:
Bumps v1.7 release to v1.7.5:

- Supports retrieving Azure access tokens via Managed Identity Extension
- Bumped Heapster version to 1.4.2
- Made v1.7.5 default version in Dockerfile 

I've built the Windows zip which is available at https://acsmirror.blob.core.windows.net/wink8s/v1.7.5-1intwinnat.zip (non-CDN URL). 

@lachie83 @jackfrancis @JiangtianLi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN mkdir /tmp/azurecli \
 
 RUN curl -fsSL https://get.docker.com/ | sh
 
-ENV KUBECTL_VERSION 1.6.6
+ENV KUBECTL_VERSION 1.7.5
 RUN curl "https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl" > /usr/local/bin/kubectl \
     && chmod +x /usr/local/bin/kubectl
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN mkdir /tmp/azurecli \
 
 RUN curl -fsSL https://get.docker.com/ | sh
 
-ENV KUBECTL_VERSION 1.6.6
+ENV KUBECTL_VERSION 1.7.5
```
jackfrancis: @seanknox @lachie83 are we confident that this is the best kubectl version for testing all supported versions of k8s? (i.e., are we expressing with this change that it's preferable to have newer kubectl's talk to older k8s clusters vs older kubectl's talk to newer k8s clusters?

seanknox: Dockerfile used only in dev workflow AFAIK, and yes I think it's better for dev workflow to orient toward newer versions. Not attached to this, can remove in this PR.

jackfrancis: 👍 I actually like it here.

PR #1384, opened 2017-09-01
Remove deprecated kubelet/apiserver/etc options

seanknox:
* remove deprecated `--register-schedulable` kubelet option for k8s >= v1.6.0
* Removed explicit setting of `--enable-debugging-handlers` and `--hairpin-mode=promiscuous-bridge` which are both default options since v1.6.0 at least.

@jackfrancis @jchauncey

### DIFF
--- a/parts/kuberneteskubelet.service
+++ b/parts/kuberneteskubelet.service
@@ -38,7 +38,6 @@ ExecStart=/usr/bin/docker run \
         --address=0.0.0.0 \
         --allow-privileged=true \
         --enable-server \
-        --enable-debugging-handlers \
         --pod-manifest-path=/etc/kubernetes/manifests \
         --cluster-dns=${KUBELET_CLUSTER_DNS} \
         --cluster-domain=cluster.local \
@@ -46,7 +45,6 @@ ExecStart=/usr/bin/docker run \
         --cloud-provider=azure \
         --cloud-config=/etc/kubernetes/azure.json \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
-        --hairpin-mode=promiscuous-bridge \
         --network-plugin=${KUBELET_NETWORK_PLUGIN} \
         --max-pods=${KUBELET_MAX_PODS} \
         --node-status-update-frequency=${KUBELET_NODE_STATUS_UPDATE_FREQUENCY} \
--- a/parts/kuberneteswindowssetup.ps1
+++ b/parts/kuberneteswindowssetup.ps1
@@ -301,7 +301,7 @@ catch
 
     $kubeProxyStartStr = @"
 `$env:INTERFACE_TO_ADD_SERVICE_IP="vEthernet (forwarder)"
-c:\k\kube-proxy.exe --v=3 --proxy-mode=userspace --hostname-override=$AzureHostname --kubeconfig=c:\k\config
+c:\k\kube-proxy.exe --proxy-mode=userspace --hostname-override=$AzureHostname --kubeconfig=c:\k\config
 "@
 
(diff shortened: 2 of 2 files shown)


parts/kuberneteskubelet.service
```diff
@@ -38,15 +38,13 @@ ExecStart=/usr/bin/docker run \
         --address=0.0.0.0 \
         --allow-privileged=true \
         --enable-server \
-        --enable-debugging-handlers \
```
jackfrancis: verified that docs say this option in enabled in `1.5`, `1.6`, and `1.7`

parts/kuberneteskubelet.service
```diff
@@ -38,15 +38,13 @@ ExecStart=/usr/bin/docker run \
...
         --pod-manifest-path=/etc/kubernetes/manifests \
         --cluster-dns=${KUBELET_CLUSTER_DNS} \
         --cluster-domain=cluster.local \
         --node-labels="${KUBELET_NODE_LABELS}" \
         --cloud-provider=azure \
         --cloud-config=/etc/kubernetes/azure.json \
         --azure-container-registry-config=/etc/kubernetes/azure.json \
-        --hairpin-mode=promiscuous-bridge \
```
jackfrancis: verified that (according to official docs) this is already the default setting for `1.5`, `1.6`, and `1.7`

parts/kubernetesmastercustomdata.yml
```diff
@@ -201,8 +201,6 @@ write_files:
     KUBELET_REGISTER_NODE=--register-node=true
     KUBELET_REGISTER_WITH_TAINTS=--register-with-taints={{WrapAsVariable "registerWithTaints"}}
   {{end}}
-{{else}}
```
jackfrancis: As far as I can tell we still need this to support 1.5 clusters (which the codebase as-is currently still supports).

parts/kubernetesmastervars.t
```diff
@@ -148,12 +148,6 @@
     "dockerBridgeCidr": "[parameters('dockerBridgeCidr')]",
 {{if IsKubernetesVersionGe "1.6.0"}}
     "registerWithTaints": "node-role.kubernetes.io/master=true:NoSchedule",
-{{else}}
```
jackfrancis: ditto my above comment about 1.5 clusters

parts/kuberneteskubelet.service
seanknox: Yes, though it's on by default already. That's why I removed it.

jackfrancis: Yes! That was my way of 👍 this line removal!

parts/kuberneteskubelet.service
jackfrancis: 👍 this line removal too!

parts/kubernetesmastercustomdata.yml
seanknox: reverted changes that affected v1.5.

parts/kubernetesmastervars.t
seanknox: ditto here.

seanknox:
Calling out some non-intuitive logic with regards to Kubernetes version support in acs-engine.

We define manifests for Kubernetes v1.5 resources—e.g. Kubernetes services/deployments etc such as`parts/kuberneteskubelet1.5.service`—but we use the same custom data files for the Kubernetes daemons like the API server etc. Thus changing options to API server affects all versions in acs-engine:
- Kubernetes v1.5.x
- Kubernetes v1.6.x
- Kubernetes v1.7.x

PR #1389, opened 2017-09-01
Fix 'Deployment Template Validation Failed" for DC/OS extensions

### LINKED ISSUE #1385: Error deploying extensions on DC/OS

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: Current master at time of opening issue (https://github.com/Azure/acs-engine/commit/99802def22fd76c7b74d681076ec8b3a6755d260)

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** DC/OS 

**What happened**: When deploying the a DC/OS cluster with either the hello-world-dcos or vamp-dcos extensions the following error 

```
Deployment failed. {
  "error": {
    "code": "InvalidTemplate",
    "message": "Deployment template validation failed: 'The template parameters 'masterVMNamePrefix' in the parameters file are not valid; they are not present in the original template and can therefore not be provided at deployment time. The only supported parameters for this template are 'artifactsLocation, apiVersionDefault, targetVMName, extensionParameters'. Please see https://aka.ms/arm-deploy/#parameter-file for usage details.'."
  }
}
```

**What you expected to happen**: The cluster and extensions should be successfully deployed

* Fill out the required fields in `examples/extensions/dcos.json`
* Remove the `rootUrl` property for the extension: https://github.com/Azure/acs-engine/blob/master/examples/extensions/dcos.json#L50
* generate the template and deploy

The error seems to be because the template expansion ends up pulling in the `masterOffset` variable which is defined for the k8s templates but not for DC/OS.

By default, the content for the scripts etc are pulled from the raw links on github. The `rootUrl` allows this to be overridden which simplifies development.

Since the examples have the `rootUrl` specified to point to the storage account that @robbagby created for the original PR, the examples will use those versions of the templates and scripts rather than the ones in the repo.

It might be worth removing the `rootUrl` from the examples to avoid this confusion.

stuartleeks:
**What this PR does / why we need it**: Addresses issues with DC/OS extensions as per #1385

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: Fixes #1385 

**Release note**:

```release-note
```

### DIFF
--- a/extensions/hello-world-dcos/v1/template-link.json
+++ b/extensions/hello-world-dcos/v1/template-link.json
@@ -1,14 +1,10 @@
 {
-    "name": "[concat(EXTENSION_TARGET_VM_NAME_PREFIX, copyIndex(EXTENSION_LOOP_OFFSET), 'HelloWorldDcos')]",
+    "name": "[concat(EXTENSION_TARGET_VM_NAME_PREFIX, sub(variables('masterCount'), 1), 'HelloWorldDcos')]",
     "type": "Microsoft.Resources/deployments",
     "apiVersion": "[variables('apiVersionLinkDefault')]",
     "dependsOn": [
         "[resourceId('Microsoft.Compute/virtualMachines/extensions', concat(variables('masterVMNamePrefix'), sub(variables('masterCount'), 1)), 'waitforleader')]"
     ],
-    "copy": {
-        "count": "EXTENSION_LOOP_COUNT",
-        "name": "helloWorldExtensionLoop"
-    },
     "properties": {
         "mode": "Incremental",
         "templateLink": {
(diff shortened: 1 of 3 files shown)


stuartleeks:
@JackQuincy - do you want me to remove the `rootUrl` properties from the examples here, or send in a separate PR?

stuartleeks:
@JackQuincy let me know how you want to proceed with this PR

jackfrancis:
Happy to re-open this if we can confirm that PR #1392 did not already address this. Thanks!

PR #1391, opened 2017-09-01
Add CNI plugin download URLs for Windows and Linux

ofiliz:

Adds separate configurable download URLs for official CNI plugins and azure-vnet CNI plugins for Linux and Windows. This enables easy configuration of plugin URLs and also is necessary to bring Windows support to parity with Linux.

**Release note**:

```release-note
```


parts/kubernetesmastervars.t
```diff
@@ -34,8 +34,9 @@
     "kubernetesKubeDNSSpec": "[parameters('kubernetesKubeDNSSpec')]",
     "kubernetesDNSMasqSpec": "[parameters('kubernetesDNSMasqSpec')]",
     "networkPolicy": "[parameters('networkPolicy')]",
-    "azureVnetCniURL":"[parameters('azureVnetCniURL')]",
-    "azureCniURL":"[parameters('azureCniURL')]",
+    "cniPluginsURL":"[parameters('cniPluginsURL')]",
+    "vnetCniLinuxPluginsURL":"[parameters('vnetCniLinuxPluginsURL')]",
+    "vnetCniWindowsPluginsURL":"[parameters('vnetCniWindowsPluginsURL')]",
```
jackfrancis: @ofiliz Where are we actually using the `vnetCniWindowsPluginsURL` variable? Mainly: how are we implementing the conditional CNI URL whether or not the platform in Linux or Windows?

ofiliz: @jackfrancis I am using the Windows plugin URL in a separate PR coming next week. There won't be a conditional check anywhere, the kubernetesmastercustomscript.sh will use the Linux version and the kuberneteswindowssetup.ps1 scripts will use the Windows version.

Because we are already renaming these, I thought I'll add the correct "Linux" vs. "Windows" qualifiers to the names now, to prevent further churn next week.

jackfrancis: Understood, so the Windows artifacts will be unused but referable with a follow-up PR. 👍

PR #1396, opened 2017-09-04
bug(agentpool): use current acs-engine defaults for agent subnet range

slack:
Agent pool only cluster definitions use 10.0.0.0/16 range for the VM subnet rather than 10.240.0.0/16.

This PR instead uses the kubernetes master defaults for VMs leaving 10.0.0.0/16 open for K8S service addresses and kube-dns to take 10.0.0.10.

/cc @amanohar Would love feedback on this approach, not entirely sure I implemented the defaults in the right way!

### DIFF
--- a/parts/masterparams.t
+++ b/parts/masterparams.t
@@ -29,6 +29,13 @@
   {{end}}
 {{end}}
 {{if IsHostedMaster}}
+    "masterSubnet": {
+      "defaultValue": "{{.HostedMasterProfile.Subnet}}",
+      "metadata": {
+        "description": "Sets the subnet for the VMs in the cluster."
+      },
+      "type": "string"
+    },
     "kubernetesEndpoint": {
       "defaultValue": "{{.HostedMasterProfile.FQDN}}",
       "metadata": {
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -123,6 +123,8 @@ func SetPropertiesDefaults(cs *api.ContainerService) (bool, error) {
 
 	setMasterNetworkDefaults(properties)
 
+	setHostedMasterNetworkDefaults(properties)
+
 	setAgentNetworkDefaults(properties)
 
 	setStorageDefaults(properties)
@@ -230,6 +232,14 @@ func setExtensionDefaults(a *api.Properties) {
 			extension.RootURL = DefaultExtensionsRootURL
 		}
 	}
+}
+
+// SetHostedMasterNetworkDefaults for hosted masters
+func setHostedMasterNetworkDefaults(a *api.Properties) {
+	if a.HostedMasterProfile == nil {
+		return
+	}
+	a.HostedMasterProfile.Subnet = DefaultKubernetesMasterSubnet
 }
 
 // SetMasterNetworkDefaults for masters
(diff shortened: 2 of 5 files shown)


pkg/acsengine/defaults.go
```diff
@@ -296,6 +306,10 @@ func setAgentNetworkDefaults(a *api.Properties) {
 			profile.OSType = api.Linux
 		}
 
+		if a.OrchestratorProfile.OrchestratorType == api.Kubernetes && a.HostedMasterProfile != nil {
+			profile.Subnet = a.HostedMasterProfile.Subnet
```
amanohar: I am assuming the RP sets the master subnet value somehow during conversion from 20170831 to pkg/api/types.go and then agents get these values? I wouldn't recommend doing it this way.

Another way to do this is have the default value in: github.com\Azure\acs-engine\pkg\acsengine\const.go and then set that value for both agent (if VnetSubnetID is empty) and master subnet.

pkg/acsengine/defaults.go
```diff
@@ -296,6 +306,10 @@ func setAgentNetworkDefaults(a *api.Properties) {
 			profile.OSType = api.Linux
 		}
 
+		if a.OrchestratorProfile.OrchestratorType == api.Kubernetes && a.HostedMasterProfile != nil {
```
amanohar: I am assuming the RP sets the master subnet value somehow during conversion from 20170831 to pkg/api/types.go and then agents get these values? I wouldn't recommend doing it this way.

Another way to do this is have the default value in: github.com\Azure\acs-engine\pkg\acsengine\const.go and then set that value for both agent (if VnetSubnetID is empty) and master subnet.

parts/masterparams.t
```diff
@@ -29,6 +29,13 @@
   {{end}}
 {{end}}
 {{if IsHostedMaster}}
+    "masterSubnet": {
```
jackfrancis: Could we change this param to `hostedMasterSubnet` from `masterSubnet`, because it appears to be intended only in "hosted master" scenarios?

amanohar: I would vote for not making those kinds of specializations. masterSubnet is already being used thoughout (and in line 27) above as well. I don't see a use case regarding why this param should be called by a different name.

jackfrancis: I didn't catch that, thanks! It would help if the `if not IsHostedMaster` was coupled with an `else` clause (or `if IsHostedMaster`--`else`) to more clearly communicate that only one of those branches are entered into during a single execution.

jackfrancis: That remark was intended for another PR, not this one!

pkg/api/types.go
```diff
@@ -296,6 +296,7 @@ type HostedMasterProfile struct {
 	// Not used during PUT, returned as part of GETFQDN
 	FQDN      string `json:"fqdn,omitempty"`
 	DNSPrefix string `json:"dnsPrefix"`
+	Subnet    string `json:"subnet"`
```
amanohar: IMO a comment is needed here that this is not really the subnet of the master/control plane but actually used during template generation to assign the subnet to agents.

slack:
Added comment for `HostedMasterProfile.Subnet`, PTAL.

pkg/api/types.go
```diff
@@ -296,6 +296,10 @@ type HostedMasterProfile struct {
 	// Not used during PUT, returned as part of GETFQDN
 	FQDN      string `json:"fqdn,omitempty"`
 	DNSPrefix string `json:"dnsPrefix"`
+	// Subnet holds the CIDR which defines the Azure Sbunet in which
```
amanohar: spelling

PR #1408, opened 2017-09-06
RHEL support for Docker swarm mode

tomconte:

Add RHEL support for Docker swarm mode.

This is more of a "strawman" PR to get feedback about the approach. A partner of mine wants to run Docker swarm mode on RHEL using ACS-Engine and I implemented this as a prototype. It probably needs polishing and testing (but it seems to work :).

Basically, I added an element "distro" to LinuxProfile, where you can request a different distro than the default e.g. "rhel". ACS-Engine then provides the supported SKU/version for that distro.

The main issue with RHEL is that it does not support Custom Data / cloud-init. I used regular CSE, pointing the fileUri to my fork at the moment. We could assemble a URL pointing to a specific release on the acs-engine repo instead.

**Release note**:

NONE

### DIFF
--- a/examples/swarmmode-rhel.json
+++ b/examples/swarmmode-rhel.json
@@ -1,0 +1,37 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "SwarmMode"
+    },
+    "masterProfile": {
+      "count": 3,
+      "dnsPrefix": "",
+      "vmSize": "Standard_D2_v2"
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "agentpublic",
+        "count": 3,
+        "vmSize": "Standard_D2_v2",
+        "dnsPrefix": "",
+        "ports": [
+          80,
+          443,
+          8080
+        ]
+      }
+    ],
+    "linuxProfile": {
+      "adminUsername": "azureuser",
+      "ssh": {
+        "publicKeys": [
+          {
+            "keyData": ""
+          }
+        ]
+      },
+      "distro": "rhel"
+    }
+  }
+}
--- a/examples/swarmmode-rhel.json
+++ b/examples/swarmmode-rhel.json
@@ -7,7 +7,8 @@
     "masterProfile": {
       "count": 3,
       "dnsPrefix": "",
-      "vmSize": "Standard_D2_v2"
+      "vmSize": "Standard_D2_v2",
+      "distro": "rhel"
     },
     "agentPoolProfiles": [
       {
@@ -19,7 +20,8 @@
           80,
           443,
           8080
-        ]
+        ],
+        "distro": "rhel"
       }
     ],
     "linuxProfile": {
@@ -30,8 +32,7 @@
             "keyData": ""
           }
         ]
-      },
-      "distro": "rhel"
+      }
     }
   }
 }
--- a/parts/swarmagentresourcesvmss.t
+++ b/parts/swarmagentresourcesvmss.t
@@ -185,7 +185,7 @@
                   "settings": {
                     "commandToExecute": "[variables('agentCustomScript')]",
                     "fileUris": [
-                      "[concat('https://raw.githubusercontent.com/tomconte/acs-engine/tco-rhel-swarmmode/parts/', variables('configureClusterScriptFile'))]"
+                      "[concat('{{ GetConfigurationScriptRootURL }}', variables('configureClusterScriptFile'))]"
                     ]
                   },
                   "type": "CustomScript",
--- a/parts/swarmagentresourcesvmss.t
+++ b/parts/swarmagentresourcesvmss.t
@@ -127,7 +127,9 @@
             "adminUsername": "[variables('adminUsername')]",
             "computerNamePrefix": "[variables('{{.Name}}VMNamePrefix')]",
 {{if IsSwarmMode}}
+  {{if not IsRHEL}}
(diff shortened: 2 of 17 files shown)


stuartleeks:
@avodovnik - didn't we have a conversation around something like this a while back?

jackfrancis:
@tomconte could you `gofmt` the two files that unit tests are complaining about?

```
==> Running static validations <==
pkg/api/const.go:1::warning: file is not gofmted with -s (gofmt)
pkg/api/vlabs/const.go:1::warning: file is not gofmted with -s (gofmt)
```

I'm curious about running some E2E against this branch to see what regressions this introduces. (need unit tests to pass though!)

pkg/api/convertertoapi.go
```diff
@@ -464,6 +464,7 @@ func convertVLabsLinuxProfile(vlabs *vlabs.LinuxProfile, api *LinuxProfile) {
 		convertVLabsKeyVaultSecrets(&s, secret)
 		api.Secrets = append(api.Secrets, *secret)
 	}
+	api.Distro = Distro(vlabs.Distro)
```
jackfrancis: We need the inverse in `pkg/api/converterfromapi.go` as well.

jackfrancis:
@tomconte my out-of-band E2E investigations suggest no regressions from this PR 🍾 

Could you add an example API model that demonstrates a RHEL Swarm cluster and put under the `examples/` directory? Perhaps `examples/swarmmode-RHEL.json`. Use the existing `examples/swarmmode.json` as a template; we can probably just copy that file exactly and just add a `"RHEL"`-valued `distro` property to the `linuxProfile`.

parts/swarmagentresourcesvmss.t
```diff
@@ -173,6 +175,26 @@
...
+          ,"extensionProfile": {
+            "extensions": [
+              {
+                "name": "configure{{.Name}}",
+                "properties": {
+                  "publisher": "Microsoft.Azure.Extensions",
+                  "settings": {
+                    "commandToExecute": "[variables('agentCustomScript')]",
+                    "fileUris": [
+                      "[concat('https://raw.githubusercontent.com/tomconte/acs-engine/tco-rhel-swarmmode/parts/', variables('configureClusterScriptFile'))]"
```
jackfrancis: @tomconte this is looking good overall, how can we clean up these static links to your github fork?

tomconte: I think the simplest is to just point to the master branch of the acs-engine repo for now... The only risk there is that if the configuration script is broken for some reason on master, deployments will fail. And at some point we might be able to point to a stable release branch instead?

stuartleeks: For the extensions work with @robbagby we had an approach where it would look to github *unless* a parameter/property was overridden. This allowed you to override in dev (and even to resolve from a storage account while you're working on it rather than needing to push to github to test it)

Could something similar work here?

tomconte: Makes sense! I will check it out.

tomconte:
As suggested by @stuartleeks I have added a script root parameter so the default can be overridden.

examples/swarmmode-rhel.json
```diff
@@ -0,0 +1,37 @@
...
+    "linuxProfile": {
+      "adminUsername": "azureuser",
+      "ssh": {
+        "publicKeys": [
+          {
+            "keyData": ""
+          }
+        ]
+      },
+      "distro": "rhel"
```
jackfrancis: I reviewed this with @anhowe, and we agreed that the `distro` property should be a agent pool profile property (in the `agentPoolProfiles` array). The reason for this is that we want to support different distributions for different agent pools (there may be use cases where a heterogeneous cluster makes sense: _n_ nodes using distro ABC, and _n_ nodes using distro XYZ).

We also want the master profile to have this option as well.

Thank you for all this work, we really *do* want this feature! (We just want to be able to re-use it, and I understand your use-case is swarm mode-specific; so apologies if we're asking you to go the extra mile!)

tomconte: Thanks Jack, that's the kind of feedback I was looking for :) Indeed my partner's requirement is to run all nodes on RHEL because most of their customers have standardized on that distribution exclusively. They don't have a requirement for heterogeneous clusters. I will evaluate the work needed to move the distro parameter at the master/agent level and get back to you.

tomconte: Since we are going to support multiple distros per deployment, and the generated template can contain bits specific to one distro (e.g. cloud-init vs CSE), I will need to remove the OS specs from the deployment parameters and move them to template variables. There will be one set for the master, and then one set per agent pool. Is that all right with you guys?

tomconte:
@jackfrancis as discussed I have moved the distro parameter into the master/agent config blocks.

tomconte:
@jackfrancis of course the other orchestrators still need the OS image parameters... I have restored them and successfully ran `make test-e2e` locally.

jackfrancis:
@tomconte I had to revert this master merge because additional tests showed regressions for the following api models:

https://github.com/Azure/acs-engine/blob/master/examples/disks-managed/swarmmode-vmas.json
https://github.com/Azure/acs-engine/blob/master/examples/disks-storageaccount/swarmmode.json

Simply, it appears that existing api models with managed disk config butwithout a `distro` property are not working. (We need to ensure those old api model expressions stay valid as we move the code forward.)

Assuming you have your forked branch, I'd just reopen a new PR and we can iterate our way through this next week.

Thanks!

PR #1428, opened 2017-09-12
Pass extension parameters as parameter allow using keyvault to do so

JackQuincy:
This causes extension parameters  to be put in into the template as a secure string parameter as well allowing them to be passed in using a keyvault ref. This allows makes it a better way to handle the secrets that will often be passed this way when an extension like the k8s oms extension is used.

**Release note**:

```release-note
Adding ability to pass extension parameters in through key vault
Removing extension loop index capability in extension parameters as it wouldn't work in key vault and can easily be parsed from the host name.
```


JackQuincy:
@jackfrancis @anhowe PTAL

PR #1433, opened 2017-09-13
use overlay2 docker graph driver in k8s

### LINKED ISSUE #555: Switch Docker Graph driver to overlay2 or aufs

tanmaykm:
**What this PR does / why we need it**: Uses overlay2 as the docker graph driver instead of overlay.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #555

**Release note**:

```release-note
use overlay2 docker graph driver in k8s clusters (instead of overlay)
```

### DIFF
--- a/parts/kubernetesagentcustomdata.yml
+++ b/parts/kubernetesagentcustomdata.yml
@@ -14,7 +14,7 @@ write_files:
   content: |
     [Service]
     ExecStart=
-    ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay --bip={{WrapAsVariable "dockerBridgeCidr"}}
+    ExecStart=/usr/bin/docker daemon -H fd:// --storage-driver=overlay2 --bip={{WrapAsVariable "dockerBridgeCidr"}}
 
 - path: "/etc/docker/daemon.json"
   permissions: "0644"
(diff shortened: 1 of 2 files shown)


jackfrancis:
@anhowe @JackQuincy any concerns with this change?

parts/kubernetesagentcustomdata.yml
```diff
@@ -147,6 +147,18 @@ write_files:
   content: !!binary |
     {{WrapAsVariable "provisionScript"}}
 
+- path: "/opt/azure/containers/useoverlay2.sh"
```
anhowe: why do you need this script?  You already set this file above.

anhowe:
We will need some bake time, but if overlay2 is an improvement, then this should not be a problem.  Any known issues with overlay2?

parts/kubernetesagentcustomdata.yml
tanmaykm: Oh yes, it's not needed. Sorry for the mixup. Will fix it in a bit.

jackfrancis:
@yuvipanda Could you kindly add any thoughts on your manual usage of overlay2? Which versions of kubernetes have you tested this out successfully on?

yuvipanda:
@jackfrancis I've been running a 20-node, 1000 pod with moderate container and image churn for about a month on kubernetes 1.7.2 with manual overlay2 and it's been working great!

jackfrancis:
@yuvipanda Thanks for the anecode! Do you have any experience with this on `v1.6` clusters?

PR #1438, opened 2017-09-14
Option to specify agent Windows OS image version

### LINKED ISSUE #1383: Specify the Windows version in API Model
Yes
---

Feature request
---

Any version
---

Kubernetes

**What happened**:
The ARM template generated by the acs-engine always uses the "latest" as the agentWindowsVersion

It's better that we can specify the agentWindowsVersion in the API model

itowlson:
**What this PR does / why we need it**: The ARM template generated by acs-engine always specifies "latest" as the agentWindowsVersion.  We have a user request to be able to specify a specific version of the WindowsServer-2016-with-Containers OS image when provisioning Windows agents.

**Which issue this PR fixes**: fixes #1383 

1. Is this feature desirable?  Most customers should use 'latest' so as to pick up security patches, though it's possible some customers need to lock to a specific version for compliance or validation reasons.

2. Is this the right scope?  I have put it at the global `windowsProfile` level but it could be per-agent-pool (we allow different OSes on different agent pools so why not different OS versions).

3. Is it the right property name in the apimodel?  `imageVersion` might not be specific enough...  (Also, have I put it in the right version?  I added it to vlabs; please let me know if that's not right.)

**Release note**:

```release-note
Added new `windowsProfile` option `imageVersion` to control the version of the WindowsServer 2016-Datacenter-with-Containers OS disk image for agent VMs.
```

### DIFF
--- a/pkg/api/convertertoapi.go
+++ b/pkg/api/convertertoapi.go
@@ -498,6 +498,7 @@ func convertV20170701WindowsProfile(v20170701 *v20170701.WindowsProfile, api *Wi
 func convertVLabsWindowsProfile(vlabs *vlabs.WindowsProfile, api *WindowsProfile) {
 	api.AdminUsername = vlabs.AdminUsername
 	api.AdminPassword = vlabs.AdminPassword
+	api.ImageVersion = vlabs.ImageVersion
 	api.Secrets = []KeyVaultSecrets{}
 	for _, s := range vlabs.Secrets {
 		secret := &KeyVaultSecrets{}
--- a/examples/windows/dcos-win-version.json
+++ b/examples/windows/dcos-win-version.json
@@ -1,0 +1,44 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "DCOS"
+    },
+    "masterProfile": {
+      "count": 1,
+      "dnsPrefix": "dcos-mstr",
+      "vmSize": "Standard_D2_v2"
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "win2",
+        "count": 2,
+        "vmSize": "Standard_D2_v2",
+        "availabilityProfile": "AvailabilitySet",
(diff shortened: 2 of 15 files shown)


pkg/api/convertertoapi.go
```diff
@@ -498,6 +498,7 @@ func convertV20170701WindowsProfile(v20170701 *v20170701.WindowsProfile, api *Wi
 func convertVLabsWindowsProfile(vlabs *vlabs.WindowsProfile, api *WindowsProfile) {
 	api.AdminUsername = vlabs.AdminUsername
 	api.AdminPassword = vlabs.AdminPassword
+	api.ImageVersion = vlabs.ImageVersion
```
jackfrancis: We'll need the same conversion in the reverse direction in `pkg/api/converterfromapi.go`.

itowlson:
@jackfrancis As discussed on Slack, here are diff files to show how the generated files change in the default case (i.e. where the user does _not_ specify the new `imageVersion` attribute).  These were generated from ./examples/windows/dcos-win.json and ./examples/windows/kubernetes-D2.json, using acs-engine built from current master and from this PR.  Note that there are additional changes in the k8s apimodel and parameters files -- I believe these are down to random certificate generation rather than a regression.

**DCOS**

[dcos-azuredeploy.json.diff.txt](https://github.com/Azure/acs-engine/files/1304880/dcos-azuredeploy.json.diff.txt)

The DCOS apimodel.json and azuredeploy.parameters.json diffs were empty and GitHub wouldn't let me upload them.

**Kubernetes**

[k8s-apimodel.json.diff.txt](https://github.com/Azure/acs-engine/files/1304879/k8s-apimodel.json.diff.txt)
[k8s-azuredeploy.json.diff.txt](https://github.com/Azure/acs-engine/files/1304882/k8s-azuredeploy.json.diff.txt)
[k8s-azuredeploy.parameters.json.diff.txt](https://github.com/Azure/acs-engine/files/1304881/k8s-azuredeploy.parameters.json.diff.txt)

PR #1444, opened 2017-09-14
(docs) Update monitoring doc

bobbypage:
Changed the monitoring doc instructions for Grafana setup. Instead of deleting heapster, include instructions on how to reconfigure the heapster deployment to use InfluxDB as it's storage backend. This is a bit safer than deleting the heapster deployment and recreating it.


jackfrancis:
Thanks @bobbypage !

PR #1446, opened 2017-09-15
Deprecate DCOS version: 1.7

jackfrancis:
**What this PR does / why we need it**: Remove DCOS version 1.7 functionality in acs-engine. Supported versions in `vlabs` going forward are `1.8` – `1.10`

**Release note**:

```
Deprecate DCOS 1.7 version functionality
```


jackfrancis:
@anhowe @weinong @JackQuincy would love a sanity check of this, thanks!

PR #1453, opened 2017-09-19
Don't present agent-pool-only validation results for full-cluster apimodels

### LINKED ISSUE #1427: acs-engine 0.6.0 generate fails on DNSPrefix  and  ServicePrincipalProfile validation
No
---

ISSUE
---

0.6.0
---

Kubernetes

**What happened**:
Error trying to generate template from latest [kubernetes-hybrid.json](https://github.com/Azure/acs-engine/blob/master/examples/windows/kubernetes-hybrid.json) template with populated values for all fields.
```
> acs-engine.exe generate ./kubernetes-hybrid.json
INFO[0000] Error returned by LoadContainerService: the service principal client ID must be specified with Orchestrator Kubernetes. Attempting to load container service using LoadContainerServiceForAgentPoolOnlyCluster
FATA[0000] error parsing the api model: Namespace Properties.DNSPrefix is not caught, Key: 'Properties.DNSPrefix' Error:Field validation for 'DNSPrefix' failed on the 'required' tag
Key: 'Properties.ServicePrincipalProfile.ClientID' Error:Field validation for 'ClientID' failed on the 'required' tag
Key: 'Properties.CertificateProfile' Error:Field validation for 'CertificateProfile' failed on the 'required' tag
```
Template generated

acs-engine version 0.6.0 (downloaded as binary from https://github.com/Azure/acs-engine/releases/download/v0.6.0/acs-engine-v0.6.0-windows-amd64.zip)
example template with redacted values
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes"
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "mydnsname",
      "vmSize": "Standard_D2_v2"
    },
    "agentPoolProfiles": [
      {
        "name": "linuxpool1",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "availabilityProfile": "AvailabilitySet"
      },
      {
        "name": "windowspool2",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "osType": "Windows"
      }
    ],
    "windowsProfile": {
      "adminUsername": "azureuser",
      "adminPassword": "supersecretpasswordhere"
    },
    "linuxProfile": {
      "adminUsername": "azureuser",
      "ssh": {
        "publicKeys": [
          {
            "keyData": "<SSH public key data here>"
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId ": "<service principal app id>",
      "secret": "<service principal password>"
    }
  }
}
```

itowlson:
**What this PR does / why we need it**: If a user creates an apimodel for an ACS cluster, but omits a required field such as the service principal client ID, then `acs-engine generate` produces a spurious error message that the required DNS prefix is missing.  This occurs because if an apimodel fails validation then it is retried as an agent-pool-only apimodel, and if it fails validation against _that_ schema, then it is always the agent-pool-only validation error that is returned.  Since the APO schema has `dnsPrefix` as its first required field, and a full cluster has `dnsPrefix` under `masterProfile` rather than at the top level, this validation fails on `dnsPrefix` and so that is the error that gets displayed.

(Alternatively, if the user has specified a stable apiVersion instead of vlabs, they get a spurious error that they are using an unrecognised apiVersion, because the full cluster API versions are not valid for APO.)

An ideal solution to this would be to use different API namespaces, or to add a `specType` field or something like that, so that the user could clearly indicate the intent of the apimodel JSON.  This would be a breaking change for vlabs though.  Therefore, this PR takes a heuristic approach: if the apimodel validates as a full cluster or APO, we use that, but if it validates as _neither_, then we try to infer intent based first on the API version and then on which properties are present in the apimodel JSON.  Whichever intent we infer, we report only the error from the appropriate validation.

**Which issue this PR fixes**: fixes #1427 

**Special notes for your reviewer**: None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/api/apiloader.go
+++ b/pkg/api/apiloader.go
@@ -42,8 +42,15 @@ func (a *Apiloader) DeserializeContainerService(contents []byte, validate bool, 
 	version := m.APIVersion
 	service, err := a.LoadContainerService(contents, version, validate, existingContainerService)
 	if service == nil || err != nil {
-		log.Infof("Error returned by LoadContainerService: %+v. Attempting to load container service using LoadContainerServiceForAgentPoolOnlyCluster", err)
-		service, err = a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate)
+		if isAgentPoolOnlyClusterJSON(contents) {
+			log.Info("No masterProfile: interpreting API model as agent pool only")
+			service, err := a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate)
+			if service == nil || err != nil {
+				log.Infof("Error returned by LoadContainerServiceForAgentPoolOnlyCluster: %+v", err)
+			}
+			return service, version, err
+		}
+		log.Infof("Error returned by LoadContainerService: %+v", err)
 	}
 
 	return service, version, err
--- a/pkg/api/convertertoagentpoolonlyapi.go
+++ b/pkg/api/convertertoagentpoolonlyapi.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"encoding/json"
+
 	"github.com/Azure/acs-engine/pkg/api/agentPoolOnlyApi/v20170831"
 	"github.com/Azure/acs-engine/pkg/api/agentPoolOnlyApi/vlabs"
 )
@@ -263,4 +265,24 @@ func convertVLabsAgentPoolOnlyCertificateProfile(vlabs *vlabs.CertificateProfile
 	api.ClientPrivateKey = vlabs.ClientPrivateKey
 	api.KubeConfigCertificate = vlabs.KubeConfigCertificate
 	api.KubeConfigPrivateKey = vlabs.KubeConfigPrivateKey
+}
+
+func isAgentPoolOnlyClusterJSON(contents []byte) bool {
+	properties, propertiesPresent := propertiesAsMap(contents)
+	if !propertiesPresent {
+		return false
+	}
+	_, masterProfilePresent := properties["masterProfile"]
+	return !masterProfilePresent
+}
+
+func propertiesAsMap(contents []byte) (map[string]interface{}, bool) {
+	var raw interface{}
+	json.Unmarshal(contents, &raw)
+	jsonMap := raw.(map[string]interface{})
+	properties, propertiesPresent := jsonMap["properties"]
+	if !propertiesPresent {
+		return nil, false
+	}
+	return properties.(map[string]interface{}), true
 }
--- a/pkg/api/convertertoagentpoolonlyapi_test.go
+++ b/pkg/api/convertertoagentpoolonlyapi_test.go
@@ -1,0 +1,41 @@
+package api
+
+import (
+	"testing"
+)
+
+func TestIfMasterProfileIsMissingThenApiModelIsAgentPoolOnly(t *testing.T) {
+	json := `
+	{
+		"apiVersion": "vlabs",
+		"properties": {
+			"dnsPrefix": "dp",
+			"fqdn": "fqdn",
+			"agentPoolProfiles": [],
+			"servicePrincipalProfile": {}
+		}
+	}
+	`
+	isAgentPool := isAgentPoolOnlyClusterJSON([]byte(json))
+	if !isAgentPool {
+		t.Error("Expected JSON without masterProfile to be interpreted as agent pool, but it was not")
+	}
+}
+
+func TestIfMasterProfileIsPresentThenApiModelIsFullCluster(t *testing.T) {
+	json := `
+	{
+		"apiVersion": "vlabs",
+		"properties": {
+			"orchestratorProfile": {},
+			"masterProfile": {},
+			"agentPoolProfiles": [],
+			"servicePrincipalProfile": {}
+		}
+	}
+	`
+	isAgentPool := isAgentPoolOnlyClusterJSON([]byte(json))
+	if isAgentPool {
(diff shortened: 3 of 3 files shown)


jackfrancis:
@itowlson Thanks for this! btw `gofmt -s pkg/api/convertertoagentpoolonlyapi_test.go` should fix the linting test failure.

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -264,3 +266,63 @@ func convertVLabsAgentPoolOnlyCertificateProfile(vlabs *vlabs.CertificateProfile
 	api.KubeConfigCertificate = vlabs.KubeConfigCertificate
 	api.KubeConfigPrivateKey = vlabs.KubeConfigPrivateKey
 }
+
+func isAgentPoolOnlyCluster(version string, contents []byte) bool {
```
amanohar: Please remove this function. This is not the right way to check if it's an agent pool only cluster or not.

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -264,3 +266,63 @@ func convertVLabsAgentPoolOnlyCertificateProfile(vlabs *vlabs.CertificateProfile
...
+	agentPoolOnlyClusterCount := 0
+	for k := range properties {
+		switch k {
+		case "orchestratorProfile", "masterProfile", "extensionProfiles", "aadProfile":
+			fullClusterCount++
+		case "dnsPrefix", "fqdn", "kubernetesVersion", "kubernetesRelease", "accessProfiles":
+			agentPoolOnlyClusterCount++
+		}
+	}
+	return distinctivePropertyCounts{
```
amanohar: This can be removed as well.

Safer/guaranteed way is:
If MasterProfile != nil or it contains property MasterProfile  then it's not a agent pool only cluster. Relying on count could break the logic in future because e.g. extensionProfiles could be added to agentpoolonly cluster and accessProfiles could be added to full cluster.

pkg/api/apiloader.go
```diff
@@ -147,30 +154,32 @@ func (a *Apiloader) LoadContainerService(
...
 
 // LoadContainerServiceForAgentPoolOnlyCluster loads an ACS Cluster API Model, validates it, and returns the unversioned representation
-func (a *Apiloader) LoadContainerServiceForAgentPoolOnlyCluster(contents []byte, version string, validate bool) (*ContainerService, error) {
+func (a *Apiloader) LoadContainerServiceForAgentPoolOnlyCluster(contents []byte, version string, validate bool) (*ContainerService, bool, error) {
 	switch version {
 	case v20170831.APIVersion:
 		managedCluster := &v20170831.ManagedCluster{}
 		if e := json.Unmarshal(contents, &managedCluster); e != nil {
-			return nil, e
+			return nil, true, e
```
amanohar: It's not impossible that ManagedCluster and ContainerService types could have same API versions so relying on API version to return true is error prone.

pkg/api/apiloader.go
```diff
@@ -42,8 +42,15 @@ func (a *Apiloader) DeserializeContainerService(contents []byte, validate bool,
 	version := m.APIVersion
 	service, err := a.LoadContainerService(contents, version, validate, existingContainerService)
 	if service == nil || err != nil {
-		log.Infof("Error returned by LoadContainerService: %+v. Attempting to load container service using LoadContainerServiceForAgentPoolOnlyCluster", err)
-		service, err = a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate)
+		service, matchedAgentPoolOnly, agentPoolErr := a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate)
```
amanohar: My recommendation would be to update isAgentPoolOnlyClusterFromProperties to check if MasterProfile is present. If it is not present call: LoadContainerServiceForAgentPoolOnlyCluster else call: LoadContainerService.

Regarding: 
 > This occurs because if an apimodel fails validation then it is retried as an agent-pool-only apimodel, and if it fails validation against that schema, then it is always the agent-pool-only validation error that is returned. 

This is not entirely correct from ACS Engine perspective since generate is a command, error is printed and not swallowed. But I understand that it could be confusing. ACS Engine is commonly cmds are commonly used only for ContainerService type (and not agent pool only type) so it makes sense to optimize for that.

pkg/api/convertertoagentpoolonlyapi.go
itowlson: Thanks @amanohar for the advice and feedback. If we can use masterProfile alone then that's much nicer - it greatly simplifies the logic.  I've made that change; please let me know if you would like to see further changes.

jackfrancis:
@itowlson Linter is complaining (!):

`pkg/api/convertertoagentpoolonlyapi.go:270:6:warning: func isAgentPoolOnlyClusterJson should be isAgentPoolOnlyClusterJSON (golint)`

Go idioms generally prefer upper-casing all acronyms like NASA in a camelCase setting (e.g., `iLoveNASA`) in favor of imposing camel-case upon an acronym (e.g., `iLoveNasa`).

itowlson:
@jackfrancis Aargh, sorry about that, too long in .NET-land!  I ran `gofmt` but not `make test-style` - my bad.  Fix pushed, will check CI results.

itowlson:
@amanohar @jackfrancis Does this need re-review since I pushed the linting fix?

jackfrancis:
@itowlson You're good but github review process doesn't love commit squashing ;)

PR #1455, opened 2017-09-19
Add RHEL support for Swarm Mode

tomconte:
Allow deploying Swarm Mode clusters on Red Hat by adding a "distro": "rhel" element to the master and agent configurations.

This is Take Two. I have rebased and squashed the changes from the previous PR, and added the changes required to get the two failing tests to work (examples/disks-managed/swarmmode-vmas.json and examples/disks-storageaccount/swarmmode.json). Tested using `e2e.sh` on the various swarmm configurations.

### DIFF
--- a/examples/swarmmode-rhel.json
+++ b/examples/swarmmode-rhel.json
@@ -1,0 +1,38 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "SwarmMode"
+    },
+    "masterProfile": {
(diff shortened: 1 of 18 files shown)


jackfrancis:
This is almost there! Our vanilla swarm api model example appears to have regressed: see `examples/swarm.json`. If you want, you could build a cluster using that api model from this branch, and then run `test/cluster-tests/swarm/test.sh` using your cluster context, or use that shell script as a reference for how to manually validate.

tomconte:
Hopefully that should be it 🤞

PR #1462, opened 2017-09-20
Hybrid k8s 1.6

jackfrancis:
**What this PR does / why we need it**: Sets hybrid k8s cluster example config to use version 1.6

**Release note**:

```
NONE
```


jackfrancis:
@JiangtianLi FYI, we want to move forward defaulting to Kubernetes 1.7 for new clusters. The api model at `examples/windows/kubernetes-hybrid.json` was not passing E2E for a 1.7 build, so I've explicitly set that api model to use 1.6 instead (which passes).

PR #1468, opened 2017-09-21
k8s: fix cherry picking

martell:
git cherry-pick A..B does not include A in the list
git cherry-pick A^..B includes A in the list

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/build-windows-k8s.sh
+++ b/scripts/build-windows-k8s.sh
@@ -69,7 +69,7 @@ k8s_16_cherry_pick() {
 	# 5fa0725025 Use adapter vEthernet (HNSTransparent) on Windows host network to find node IP
 	# 79cf9963f7 Merge pull request #51126 from chen-anders/anders/port-47991-to-release-1.6
 
-	git cherry-pick 79cf9963f7..232fa6e5bc
+	git cherry-pick 79cf9963f7^..232fa6e5bc
 }
 
 k8s_17_cherry_pick() {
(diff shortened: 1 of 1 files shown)


jackfrancis:
@JiangtianLi I believe these changes are meant to enable a working windows k8s v1.6.10 build flow, could you kindly validate?

martell:
@jackfrancis @JiangtianLi It was originally for v1.6.9 because this commit was not actually cherry picked. https://github.com/kubernetes/kubernetes/commit/79cf9963f7
This is not needed for v1.6.10 but we should apply it anyway to fix older versions and in case this happens with another commit in future.

scripts/build-windows-k8s.sh
```diff
@@ -83,8 +83,8 @@ k8s_17_cherry_pick() {
 	# adeb88d774 Use adapter vEthernet (HNSTransparent) on Windows host network to find node IP
 	# 02549d6647 Merge pull request #50914 from shyamjvs/add-logging-to-logdump
 
-	git cherry-pick 02549d6647..45ba7bb0fb
-        git cherry-pick 51fab673e1..72b9c8f519
+	git cherry-pick 02549d6647^..45ba7bb0fb
```
JiangtianLi: @martell Thanks for the suggestion. However, on A..B, we intentionally use a commit A before the commit we want to include. So for example, 02549d6647..45ba7bb0fb, we actually want to include the commit after 02549d6647, not 02549d6647. So if we go with your approach, we would need to change 02549d6647 to the commit after it.
@seanknox Please double check.

JiangtianLi:
@martell I think we should go with your approach but please modify not to include the left side commit too.

scripts/build-windows-k8s.sh
martell: hmm, I see. read my comment above regarding 79cf996.
I think A^..B and changing 02549d6647 to the commit after if that is what you want to do here because most developers won't be aware of this behaviour, increasing the barrier to contribution entry.

JiangtianLi: I agree to use A^..B. Please change 79cf9963f7, etc. accordingly. We used to have 5fa0725025 (the one after 79cf9963f7) without ^ and  we fixed it. But your way is a better fix.

JiangtianLi:
I have https://github.com/Azure/acs-engine/pull/1520 to address this issue. We can close this.
@martell Thanks for your suggestions and initializing this!

martell:
@JiangtianLi Thanks for following up.
Sorry I didn't get around to this sooner, my free OSS time is limited ATM :(
Closing in favour of #1520

PR #1478, opened 2017-09-22
using distinct cluster wait times for windows or linux

jackfrancis:
**What this PR does / why we need it**: increase wait time for Windows E2E clusters

**Release note**:

```
NONE
```

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -21,9 +21,12 @@ import (
 )
 
 var (
-	cfg config.Config
-	eng engine.Engine
-	err error
+	cfg                    config.Config
+	eng                    engine.Engine
+	linuxClusterWaitTime   time.Duration
+	windowsClusterWaitTime time.Duration
+	clusterWait            time.Duration
+	err                    error
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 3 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -21,16 +21,21 @@ import (
...
+	err                    error
 )
 
 var _ = BeforeSuite(func() {
 	cwd, _ := os.Getwd()
 	rootPath := filepath.Join(cwd, "../../..") // The current working dir of these tests is down a few levels from the root of the project. We should traverse up that path so we can find the _output dir
 	c, err := config.ParseConfig()
 	c.CurrentWorkingDir = rootPath
+	linuxClusterWaitTime = 10 * time.Minute
+	windowsClusterWaitTime = 15 * time.Minute
```
jchauncey: There's really no reason to make these different. If anything we should make it take a value from the outside with a default of 10 mins. We can use the golang `time.ParseDuration` func that allows you to do something like `10m` or `30s`. This way for windows tests we increase if we want or it just defaults to `10s` when building the configuration object. This way we have 1 wait time for all objects because honestly we shouldn't wait around indefinitely which is what these types of waits encourage.

jackfrancis:
Closing in favor of @jchauncey's approach

PR #1480, opened 2017-09-23
Add VnetIntegration switch to Kubernetes

### LINKED ISSUE #1499: Add switch in KubernetesConfig to enable VNET integration
No.
---

Feature Request.
---

Latest, v0.7.0
---

Kubernetes, all versions.

**What happened**:
Azure VNET CNI plugin now has two operational modes:
* Bridge mode: This mode assigns real IP addresses from VNET space to containers. This makes pods first class citizens like VMs, enabling scale and allowing all SDN features to apply to pods.
* Tunnel mode: Also known as the "network policy" mode, this mode does everything the bridge mode does, plus it also hairpins all pod traffic through the Hyper-V host so that VNET network policy can be applied.

Currently, the only way to enable the CNI plugin is by setting the KubernetesConfig.networkPolicy to "azure". This is not optimal for two reasons:

1. We want other network policy providers like Calico to also be able to use real IP addresses for pods and take advantage of SDN features. Because customers can't set networkPolicy to both "azure" and "calico" at the same time, this is currently not possible.
2. CNI plugin's "bridge" operational mode does not actually enforce networkPolicy, so turning on the feature by using the KubernetesConfig.networkPolicy property is misleading.

To solve these problems, we need to have separate knobs for the network policy functionality and the "bridge" functionality. See #1480 for more details.

Feature implemented in #1480.

ofiliz:
To solve the problems described in issue #1499, this PR adds a "vnetIntegration" property to KubernetesConfig. It defaults to disabled.

The existing "networkPolicy" property controls what the network policy provider is. The new "vnetIntegration" property controls whether VNET CNI plugins are enabled. This allows Calico to implement their network policy on top of our IPAM CNI plugin by setting vnetIntegration:"enabled" and networkPolicy:"calico". It also enables customers to immediately start using our VNET CNI plugin (with pods having access to Internet) by just setting vnetIntegration:"enabled".

The azure networkPolicy requires vnetIntegration to be enabled. This is forced in acs-engine. New test cases are added to validate behavior of all possible combinations.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1499

I reordered some fields in KubernetesConfig to group similar fields together and make sure they are always read/parsed/passed in the same order for consistency. This isn't a must, but results in better code IMHO.

**Release note**:

```release-note
```
@nisheeth-ms, @sharmasushant, @tamilmani1989, @lachie83, @anhowe 
fixes #1499


lachie83:
Thanks @ofiliz I'll take a look

ofiliz:
Wow, I need to sign a CLA now! 😄 It took Microsoft just two days to forget who I am. 😆

ofiliz:
@lachie83 Can we please get this done soon? The Windows CNI PR depends on this one.

PR #1490, opened 2017-09-25
added PoolUpgradeProfile and UpgradeProfile types to support upgrade operation

dmitsh:
added PoolUpgradeProfile and UpgradeProfile types to support upgrade operation


pkg/api/v20170930/orchestratorversiontypes.go
```diff
@@ -1,13 +1,42 @@
...
+	OSType   OSType                 `json:"osType,omitempty"`
+	Upgrades []*OrchestratorProfile `json:"upgrades,omitempty"`
+}
+
+// UpgradeProfile contains cluster properties:
+//  - orchestrator type and version for the cluster
+//  - list of pool profiles, constituting the cluster
+type UpgradeProfile struct {
+	ControlPlaneProfile *PoolUpgradeProfile   `json:"controlPlaneProfile"`
+	AgentPoolsProfiles  []*PoolUpgradeProfile `json:"agentPoolsProfiles"`
```
amanohar: Should be `AgentPoolProfiles`

pkg/api/v20170930/const.go
```diff
@@ -16,3 +16,9 @@ const (
 	// DockerCE is the string constant for the Docker CE orchestrator type
 	DockerCE string = "DockerCE"
 )
+
+// the OSTypes supported by vlabs
```
jackfrancis: Could you kindly remove the vlabs reference here?

pkg/api/orchestratorversiontypes.go
```diff
@@ -1,6 +1,29 @@
 package api
 
-// OrchestratorVersionProfile contains orchestrator version info
+// PoolUpgradeProfile contains pool properties:
+//  - orchestrator type and version
+//  - pool name (for agent pool)
+//  - OS type of the VMs in the pool
+//  - list of applicable upgrades
+type PoolUpgradeProfile struct {
```
amanohar: Why do we need internal API model for this? Couldn't think just we construct and return the public API model. Internal API model mainly exists because ACS Engine needs a standardized model to operation on to generate templates and also to save normalized data in the DB.

pkg/api/orchestratorversiontypes.go
```diff
@@ -1,6 +1,9 @@
 package api
 
-// OrchestratorVersionProfile contains orchestrator version info
+// OrchestratorVersionProfile contains information of a supported orchestrator version:
+//  - orchestrator type and version
+//  - whether this orchestrator version is deployed by default if orchestrator release is not specified
+//  - list of available upgrades for this orchestrator version
 type OrchestratorVersionProfile struct {
```
amanohar: Is this still needed in the unversioned model?

PR #1496, opened 2017-09-26
update ubuntu image to 16.04.201709190

jackfrancis:
**What this PR does / why we need it**: Updates ubuntu image to September 16.04-LTS release

**Release note**:

```
Updates ubuntu image to September 16.04-LTS release
```

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -41,7 +41,7 @@ var (
 		ImageOffer:     "UbuntuServer",
 		ImageSku:       "16.04-LTS",
 		ImagePublisher: "Canonical",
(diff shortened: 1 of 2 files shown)


pkg/acsengine/defaults.go
```diff
@@ -83,7 +83,7 @@ var (
 				ImageOffer:     "UbuntuServer",
 				ImageSku:       "16.04-LTS",
 				ImagePublisher: "Canonical",
-				ImageVersion:   "16.04.201701130",
+				ImageVersion:   "16.04.201709190",
```
anhowe: This is specific to German cloud, and you probably want to verify this version exists in Germany, or leave the German cloud change for a separate PR.

PR #1498, opened 2017-09-27
Mitigate service IP not update in time issue in windows kubeproxy sin…

JiangtianLi:
…ce 1.7.0 kubeproxy uses event-based service configuration

Windows kubeproxy adds service IP to forwarder NIC OnServiceUpdate. Since 1.7.0, kubeproxy uses event-based service configuration and therefore kube-system services are not passed during POD service creation. It has to wait for the default 15m to get update and before that, cluster DNS is not reachable. This is a mitigation of the issue.


jackfrancis:
@JiangtianLi Could we set this even lower, to 5mins? Are there cluster performance ramifications to refreshing that frequently?

JiangtianLi:
@jackfrancis I've changed the duration.

jackfrancis:
👍 Looks like that the mitigation is entirely in the post-deployment setup script? In other words, we won't be building a new win k8s image to remediate?

JiangtianLi:
Yes, this will be sufficient and we don't need to fix in kube-proxy, which is non-trivial.

PR #1502, opened 2017-09-27
feat(kubernetes_test): Add outbound internet checks for both agent types

jchauncey:
This adds outbound checks for both windows and linux agent pools.


jchauncey:
@jackfrancis this is ready to merge

PR #1519, opened 2017-10-03
Vendor kubernetes client

JackQuincy:
This vendors k8s.io/client-go so that I can add a drain and cordon change in another review.
I just added these lines to the glide.yaml then referenced them so glide would pull dependencies. called glide up -v. then removed the reference.

- package: k8s.io/apiserver
  version: 366b072768b4e6d93c7de236464c0abe85d0b7c6
- package: k8s.io/client-go
  version: ~4.0.0


jackfrancis:
There were some artifacts here so I created #1530 instead. (sorry this process is so painful)

PR #1521, opened 2017-10-03
changed ubuntu image to 16.04.201708151

CecileRobertMichon:
**What this PR does / why we need it**: Updates the Ubuntu image to August 16.04-LTS release

**Release note**:


jackfrancis:
This is good news!, as the 1.8 tests are known to fail until we stabilize our 1.8 cluster deployments. Should we continue incrementing the ubuntu image until we get one that fails? Would be nice to get as recent as possible.

PR #1529, opened 2017-10-03
feat(v20170831): agent pools for v20170831 must use ManagedDisks

seanknox:
This ensures that Kubernetes agents created via the v20170831 API use ManagedDisks.

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20170831/const.go
+++ b/pkg/api/agentPoolOnlyApi/v20170831/const.go
@@ -26,8 +26,6 @@ const (
 
 // storage profiles
 const (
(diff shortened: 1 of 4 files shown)


seanknox:
@jackfrancis This breaks on K8s v1.6, which is expected. For v20170831 users we're ok with dropping k8s 1.6 support as we're focused on v1.7 and v1.8.

jackfrancis:
@seanknox Your comment doesn't suggest that other kubernetes 1.6 API versions (incl. vlabs) will break, so I've restarted the 1.6 E2E job (assuming the failure is transient). Which is to say, we can't break 1.6 for api models that resemble this:

https://github.com/Azure/acs-engine/blob/master/examples/kubernetes-releases/kubernetes1.6.json

PR #1537, opened 2017-10-04
increase wait time for node readiness to 15 mins

jackfrancis:
```
NONE
```

### DIFF
--- a/test/e2e/runner.go
+++ b/test/e2e/runner.go
@@ -84,7 +84,7 @@ func main() {
 		os.Setenv("KUBECONFIG", cfg.GetKubeConfig())
 		log.Printf("Kubeconfig:%s\n", cfg.GetKubeConfig())
 		log.Println("Waiting on nodes to go into ready state...")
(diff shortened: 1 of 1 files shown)


test/e2e/runner.go
```diff
@@ -84,7 +84,7 @@ func main() {
 		os.Setenv("KUBECONFIG", cfg.GetKubeConfig())
 		log.Printf("Kubeconfig:%s\n", cfg.GetKubeConfig())
 		log.Println("Waiting on nodes to go into ready state...")
-		ready := node.WaitOnReady(eng.NodeCount(), 10*time.Second, 10*time.Minute)
+		ready := node.WaitOnReady(eng.NodeCount(), 10*time.Second, 15*time.Minute)
```
jchauncey: do we want to make this `cfg.Timeout`? that way we can configure it as an env var?

jackfrancis:
@jchauncey yes, thanks!, see 2nd commit

PR #1541, opened 2017-10-05
Cordon and Drain agent nodes during upgrade

JackQuincy:
Adds Draining pods from nodes before we delete them during upgrade. Making it less disruptive to the cluster.
Looking at this diff there are a couple little changes I need to revert. I'll make those tomorrow morning, but I wanted to get this in CI.
**Release note**:

```release-note
```


JackQuincy:
In adhoc testing of this I found two issues with running this in upgrade.  I'll pull it out of upgrade for now as it will break upgrade until at least #1547 is fixed. Other issue is #1548 might not be related to this but they are associated with each other. On a side note draining worked great.

parts/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
   name: kube-dns
   namespace: kube-system
 spec:
-  clusterIP: <kubeDNSServiceIP>
```
amanohar: Are you sure this is the right change? I remember there was work done to parametrize: `kubeDNSServiceIP`. Why remove it?

JackQuincy: Mistake. Must of been part igg a merge mistake. I'll revert it

jackfrancis:
@amanohar If this looks good to you I'll merge.

JackQuincy:
I just rebased on master as I had gotten pretty far off. Hopefully tests all pass and we can move forward from there.

JackQuincy:
@amanohar when you have time take a look. I managed to pass tests with retry. I updated to make the timeout passed in like we talked about offline.

JackQuincy:
@amanohar @jackfrancis ping post weekend

pkg/armhelpers/kubeclient.go
```diff
@@ -0,0 +1,143 @@
...
+
+//WaitForDelete waits until all pods are deleted. Returns all pods not deleted and an error on failure
+func (c *KubernetesClientSetClient) WaitForDelete(logger *log.Entry, pods []v1.Pod, usingEviction bool) ([]v1.Pod, error) {
+	var verbStr string
+	if usingEviction {
+		verbStr = "evicted"
+	} else {
+		verbStr = "deleted"
+	}
+	err := wait.PollImmediate(c.interval, c.timeout, func() (bool, error) {
```
amanohar: @JackQuincy will this fail the operation if wait for delete times out (the comment above mentions so)? I don't expect scale to fail with pod deletion fails. This should be best effort IMO.

pkg/armhelpers/kubeclient.go
```diff
@@ -0,0 +1,143 @@
...
+	return c.clientset.Nodes().Get(name, metav1.GetOptions{})
+}
+
+//UpdateNode updates the node in the api server with the passed in info
+func (c *KubernetesClientSetClient) UpdateNode(node *v1.Node) (*v1.Node, error) {
+	return c.clientset.Nodes().Update(node)
+}
+
+//SupportEviction queries the api server to discover if it supports eviction, and returns supported type if it is supported
+func (c *KubernetesClientSetClient) SupportEviction() (string, error) {
```
amanohar: Should some of this be written upstream? Advantages I can think of is that bugs fixes could be done by the community. We don't need to keep up with new features in the client. If we went the new client then `SupportEviction` comes with it. Otherwise with each new version of client this code could fall behind.

PR #1544, opened 2017-10-05
[ACS RP] remove orchestratorRelease for versioned api model

### LINKED ISSUE #1545: [ACS RP] OrchestratorRelease doesn't conform with ARM spec

---

ISSUE
---

ALL
---

OrchestratorRelease will by mapped to a specific orchestratorVersion, based on the map at the time. The map is changed frequently. ARM need to make everything clear in the PUT operation, and the same template should result the same output. They don't expect any underlying change.

Kubernetes, DCOS

**What happened**:

rjtsdl:
This change is scoped under versioned api model. It has nothing to do with vlabs.
versioned api model need to be conformed with ARM spec, we need to make it clear.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1545
Feel free to correct the issue, I only put in my understanding.

**Release note**:

```release-note
```


pkg/api/v20170701/types.go
```diff
@@ -115,8 +115,7 @@ const (
 // OrchestratorProfile contains Orchestrator properties
 type OrchestratorProfile struct {
```
amanohar: This needs to be changed for ManagedCluster type as well (2017-08-31 version).

pkg/api/upgrade/v20170930/orchestratorversiontypes.go
```diff
@@ -9,7 +9,6 @@ type OSType string
 //  - version: major, minor, and patch version numbers
 type OrchestratorProfile struct {
```
amanohar: `UpgradeProfile` and `PoolUpgradeProfile` needs to be moved out of API version 2017-09-30 and added to 2017-07-01 and 2017-08-31 versions because these procy child resource types under `containerservice` and `managedcluster` resource types.

pkg/api/v20170701/types.go
```diff
@@ -115,8 +115,7 @@ const (
...
 	OrchestratorType    string `json:"orchestratorType" validate:"required"`
-	OrchestratorRelease string `json:"orchestratorRelease"`
```
amanohar: KubeConfig for loading values based on version (in addition to release) needs to be enabled: https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/const.go#L94.

pkg/api/convertertoapi.go
```diff
@@ -562,17 +548,19 @@ func convertV20170701OrchestratorProfile(v20170701cs *v20170701.OrchestratorProf
 
 	switch api.OrchestratorType {
 	case Kubernetes:
-		switch v20170701cs.OrchestratorRelease {
+		release, _ := common.GetReleaseFromVersion(v20170701cs.OrchestratorVersion)
+		switch release {
 		case KubernetesRelease1Dot8, KubernetesRelease1Dot7, KubernetesRelease1Dot6, KubernetesRelease1Dot5:
-			api.OrchestratorRelease = v20170701cs.OrchestratorRelease
+			api.OrchestratorRelease = release
```
dmitsh: We don't need OrchestratorRelease in unversioned API. Release (if someone needs it) can be derived from OrchestratorVersion. Also, have both OrchestratorRelease  and OrchestratorVersion is prone to bugs

pkg/api/v20170701/types.go
rjtsdl: Ah, i missed it. Just found it is called `KubernetesRelease`. Will make the change.

rjtsdl:
Close this, as I don't know if i want to push forward.

pkg/api/upgrade/v20170930/orchestratorversiontypes.go
rjtsdl: Can we do this later? I don't know the context of this. @dmitsh can you fix this after this PR merged?

pkg/api/convertertoapi.go
rjtsdl: As latest discussion, we don't change internal model.

rjtsdl:
close again.

PR #1564, opened 2017-10-09
Etcd3 Support

khenidak:
**What**: This PR enables etcd to use different versions of etcd (via opt in property on KubernetesConfig). If the user does not use set etcdVersion property it will default to current value (version 2.5.2)
 
**Why((: Encrypting secrets via KeyVault (1.8 EncryptionConfig) requires etcd3 API (api-server settings). Because it uses protobuf and this is not compatible with etcd2 API

@jdumars @seanknox @itowlson 

make test fails because i believe it does not load the defaults..


seanknox:
@lachie83  too

parts/kubernetesmastercustomdata.yml
```diff
@@ -335,7 +338,57 @@ write_files:
...
+    [Service]
+    Environment=DAEMON_ARGS=
+    Environment=ETCD_NAME=%H
+    Environment=ETCD_DATA_DIR=/var/lib/etcd/default
+    EnvironmentFile=-/etc/default/%p
+    Type=notify
+    User=etcd
+    PermissionsStartOnly=true
+    ExecStart=/usr/bin/etcd $DAEMON_ARGS
+    Restart=on-abnormal
```
anhowe: Restart=always

parts/kubernetesmastercustomdata.yml
```diff
@@ -335,7 +338,57 @@ write_files:
...
+- path: "/opt/azure/containers/setup-etcd.sh"
+  permissions: "0744"
+  owner: "root"
+  content: |
+    #!/bin/bash
+    # if provisioning using any etcd version other than 2.5.2
+    # We download it and set it up here 
+    set -x
+    ETCD_VER=v{{ .OrchestratorProfile.KubernetesConfig.EtcdVersion }}
+    DOWNLOAD_URL=https://github.com/coreos/etcd/releases/download # we default to github
```
anhowe: this should go through CDN (Dmitry can show you) so that this is not down if github goes down.

parts/kubernetesmastercustomdata.yml
```diff
@@ -335,7 +338,57 @@ write_files:
...
+    systemctl enable etcd.service
+{{end}}
+
 runcmd: 
+{{ if .OrchestratorProfile.IsCustomEtcdVersion }}
+- /opt/azure/containers/setup-etcd.sh
+{{end}}
 - apt-mark hold walinuxagent {{GetKubernetesMasterPreprovisionYaml}}
 - /bin/echo DAEMON_ARGS=--name "{{WrapAsVerbatim "variables('masterVMNames')[copyIndex(variables('masterOffset'))]"}}" --initial-advertise-peer-urls "{{WrapAsVerbatim "variables('masterEtcdPeerURLs')[copyIndex(variables('masterOffset'))]"}}" --listen-peer-urls "{{WrapAsVerbatim "variables('masterEtcdPeerURLs')[copyIndex(variables('masterOffset'))]"}}" --advertise-client-urls "{{WrapAsVerbatim "variables('masterEtcdClientURLs')[copyIndex(variables('masterOffset'))]"}}" --listen-client-urls "{{WrapAsVerbatim "concat(variables('masterEtcdClientURLs')[copyIndex(variables('masterOffset'))], ',http://127.0.0.1:', variables('masterEtcdClientPort'))"}}" --initial-cluster-token "k8s-etcd-cluster" --initial-cluster "{{WrapAsVerbatim "variables('masterEtcdClusterStates')[div(variables('masterCount'), 2)]"}} --data-dir "/var/lib/etcddisk"" --initial-cluster-state "new" | tee -a /etc/default/etcd
 - sudo /bin/chown -R etcd:etcd /var/lib/etcd/default
```
anhowe: Does this still work? from line 394-401.  You probably have to guard the custom configuration of etcd.

pkg/acsengine/defaults.go
```diff
@@ -247,6 +247,12 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 				a.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
 			}
 		}
+
+		// default etcd version
```
itowlson: As discussed on Slack, this is not called from some unit tests, and this may be what causes `make test` to fail.  This is a flaw in the tests but would be hard to fix.  As a workaround, it would be good to try adding `"etcdVersion": "2.5.2"` to the `const ExampleAPIModel` JSON at `deploy_test.go` line 18.  This would be quick to try, and if it works it would avoid having failing tests in master which everyone would need to know "oh yeah you can ignore that."

pkg/api/vlabs/validate.go
```diff
@@ -14,15 +14,28 @@ import (
...
+func isValidEtcdVersion(etcdVersion string) error {
+	validVersions := ""
+	for _, ver := range etcdValidVersions {
+		if ver == etcdVersion {
+			return nil
+		}
+		validVersions = fmt.Sprintf("%s %s", validVersions, ver)
+	}
+
+	return fmt.Errorf("Invalid etcd version, valid versions are%s", validVersions)
```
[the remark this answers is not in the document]
khenidak: Added.

pkg/acsengine/engine.go
```diff
@@ -232,7 +232,7 @@ func InitializeTemplateGenerator(ctx Context, classicMode bool) (*TemplateGenera
 
 // GenerateTemplate generates the template from the API Model
 func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerService) (templateRaw string, parametersRaw string, certsGenerated bool, err error) {
-	// named return values are used in order to set err in case of a panic
+	// named return values are used in order to set err in case of a.OrchestratorProfile.panic
```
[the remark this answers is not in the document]
khenidak: No, that is a mistake. I reverted..

pkg/acsengine/defaults.go
khenidak: Thanks for this, I added it. Now tests are went back to normal. On a different note, there has to be away to pickup pkg/acsengine/testdata

JackQuincy:
@khenidak we got master head clean again I'd merge/rebase and see if CI does better now

khenidak:
@JackQuincy the current PR is rebased. The failed tests happening because the test is not using the same defaulting mechanism used for ./acs-engine generate. The property i added gets defaulted there hence tests are failing

JackQuincy:
@khenidak cool just wanted to make sure you had the latest changes as they are more recent then 5 hours ago like your most recent update. But it does look like you are failing on tests not deployments/deep tests.

pkg/api/vlabs/validate.go
```diff
@@ -16,13 +16,26 @@ import (
...
+	etcdValidVersions = [...]string{"2.5.2", "3.2.0"}
 )
 
 func init() {
 	validate = validator.New()
 	keyvaultIDRegex = regexp.MustCompile(`^/subscriptions/\S+/resourceGroups/\S+/providers/Microsoft.KeyVault/vaults/[^/\s]+$`)
 }
 
+func isValidEtcdVersion(etcdVersion string) error {
+	validVersions := ""
```
anhowe: "" is a valid version, because it will be set to default when defaults are set.  Ensure "" is allowed.

pkg/api/vlabs/validate.go
```diff
@@ -16,13 +16,32 @@ import (
 var (
 	validate        *validator.Validate
 	keyvaultIDRegex *regexp.Regexp
+	// Any version has to be mirrored in https://acs-mirror.azureedge.net/github-coreos/etcd-v[Version]-linux-amd64.tar.gz
+	etcdValidVersions = [...]string{"2.5.2", "3.2.0"}
```
lachie83: Any reason why you didn't do 3.2.9?

pkg/api/types.go
```diff
@@ -533,3 +535,18 @@ func (o *OrchestratorProfile) IsVNETIntegrated() bool {
...
+}
+
+// GetAPIServerEtcdAPIVersion Used to set apiserver's etcdapi version
+func (o *OrchestratorProfile) GetAPIServerEtcdAPIVersion() string {
+	// if we are here, version has already been validated..
+	etcdversion, _ := semver.NewVersion(o.KubernetesConfig.EtcdVersion)
+	if 2 == etcdversion.Major() {
+		return "etcd2"
+	}
+	return "etcd3"
```
lachie83: You also need to confirm the kubernetes version. We still support 1.5.x which doesn't support etcd3

parts/kubernetesmastercustomdata.yml
```diff
@@ -1,7 +1,9 @@
 #cloud-config
 
 packages:
+{{if not .OrchestratorProfile.IsCustomEtcdVersion}}
```
lachie83: I would almost argue that we don't use the upstream packages and remove this entirely. The upstream apt packages don't even use the last 2 dot release which contains a bunch of operational fixes. Worth it or do we let etcd2 die on the vine as we will soon deprecate k8s 1.5

parts/kubernetesmastercustomdata.yml
```diff
@@ -335,7 +338,57 @@ write_files:
...
+    [Unit]
+    Description=etcd - highly-available key value store
+    Documentation=https://github.com/coreos/etcd
+    Documentation=man:etcd
+    After=network.target
+    Wants=network-online.target
+    [Service]
+    Environment=DAEMON_ARGS=
+    Environment=ETCD_NAME=%H
+    Environment=ETCD_DATA_DIR=/var/lib/etcd/default
```
lachie83: Are these still valid and recommend for k8s in etcd3

pkg/api/types.go
khenidak: Its an opt in, only. if no version specified it will default to what we have now

pkg/api/vlabs/validate.go
khenidak: Just remaining on a older side of releases. The new versions bring new functions and bugs that takes time to iron out.

parts/kubernetesmastercustomdata.yml
khenidak: All of this gets override-en with env file where data file is actually saved on a mounted data disk

pkg/api/types.go
JackQuincy: We shoudl throw a validation error if they try to use 1.5 and etcd 3

pkg/api/vlabs/validate.go
lachie83: patch releases are more stable. Minors may not be. If you're going for stability then go for the latest 3.1 patch release. You're on the edge with 3.2 anyways. I generally stay a minor behind unless there's something pressing.

parts/kubernetesmastercustomdata.yml
lachie83: Okay let me spin this up and take a look.

pkg/api/types.go
lachie83: Yep. I would otherwise apiserver isn't going to start.

pkg/api/vlabs/validate.go
khenidak: Alright will go for 3.1.

parts/kubernetesmastercustomdata.yml
khenidak: I agree. But should happen outside this PR. This PR is focused on etcd3. The next should focused on using the same binary source for all releases.

pkg/api/types.go
khenidak: Do we still want to do that even with opt-in only?

pkg/acsengine/testdata/etcd-versions/kubernetes.json
```diff
@@ -0,0 +1,53 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "kubernetesConfig": {
+        "etcdVersion": "3.2.0"
```
lachie83: Needs to be 3.1.10

anhowe: This is causing test failure.

lachie83:
I was able to spin up a single and multi-master cluster without issues.

PR #1568, opened 2017-10-10
trying to fix windows build breaks

JackQuincy:
should make windows only clusters behave better. then they do now. Might have other issues as well but this should help.


weinong:
do we need this on 1.7?

jchauncey:
@JackQuincy let me know if you need any help with this. I can show you how to run the tests locally too if you'd like

jchauncey:
wait so we merged thsi but none of the windows tests passed. is that what we expected?

PR #1570, opened 2017-10-10
Use environment variable for passing provision parameters

### LINKED ISSUE #1569: Custom VNET change breaks normal template generation
PR #1150 adds 'VIRTUAL_NETWORK_RESOURCE_GROUP' in parameter. 

When not using custom vnet, the `VIRTUAL_NETWORK_RESOURCE_GROUP` has an empty value, which causes every follow up parameters to shift one position ahead.

Here is the part of the result based on commit `4df8a77c9`:
/etc/kubeletes/azure.json
```
 "cloud":"none", # value for network policy
 "vnetResourceGroup": "k8s-master-20487671-routetable", # value for route table
 "routeTableName": "agentpool1-availabilitySet-20487671", # value for availability set
```

/etc/default/kubelet
```
KUBELET_MAX_PODS=false # value for CLOUDPROVIDER_BACKOFF
```

Here is the process run snapshot from another run:
```
root      2712  0.0  0.0   4528   776 ?        S    03:00   0:00 /bin/sh -c /usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision.sh <guid> <guid> <rg> <location> k8s-subnet k8s-master-16231024-nsg k8s-vnet-16231024  k8s-master-16231024-routetable agentpool1-availabilitySet-16231024
```

Note that there are two spaces between 'k8s-vnet-16231024' and 'k8s-master-16231024-routetable', that's the empty value of 'VIRTUAL_NETWORK_RESOURCE_GROUP'.

karataliu:
Currently kubernetes provision script uses numeric labled parameter in provision script, which is hard to update and debug.

When any of the parameter is empty, the parameter order would be affected.

This is a fix attempt to switch to named parameters by using environment variables.

**Which issue this PR fixes**
Fixes #1569 

**Release note**:


karataliu:
Any feedback? Thanks.

jackfrancis:
I'm very encouraged by this, @anhowe @JackQuincy as long-tenured `acs-engine` curators, what are your thoughts?

parts/kubernetesmastervars.t
```diff
@@ -117,6 +117,8 @@
...
+    "provisionScriptParametersMaster": "[concat(' APISERVER_PRIVATE_KEY=',variables('apiServerPrivateKey'),' CA_CERTIFICATE=',variables('caCertificate'),' CA_PRIVATE_KEY=',variables('caPrivateKey'),' MASTER_FQDN=',variables('masterFqdnPrefix'),' KUBECONFIG_CERTIFICATE=',variables('kubeConfigCertificate'),' KUBECONFIG_KEY=',variables('kubeConfigPrivateKey'),' ADMINUSER=',variables('username'))]",
```
jackfrancis: Let's get rid of the leading ` ` (space) here.

parts/kubernetesmasterresources.t
```diff
@@ -530,7 +530,7 @@
...
+          "commandToExecute": "[concat(variables('provisionScriptParametersCommon'),variables('provisionScriptParametersMaster'),' /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]"
```
jackfrancis: Let's add a ` ` (space) in between `variables('provisionScriptParametersCommon')` and `variables('provisionScriptParametersMaster')`. The current implementation requires `provisionScriptParametersCommon` to be concatenated literally right before `provisionScriptParametersMaster` if we're using them in common, which is a bit brittle.

PR #1150, opened 2017-08-01
Improve Custom VNet support

sylr:
**What this PR does / why we need it**:

This PR implements https://github.com/kubernetes/kubernetes/pull/49725 in acs-engine.

At this moment Custom VNet in a separate RG has one flaw (as far as I'm aware of), Kubernetes is unable to create an `internal load balancer Azure resource` for `internal load balancer k8s service` because the Azure Cloud Provider driver looks for the master subnet in the RG of k8s and this fails (because it's not there).

https://github.com/kubernetes/kubernetes/pull/49725 fixes this by introducing a new configuration in azure.json `vnetResourceGroup`.

This patchsimply adds this new vnetResourceGroup configuration to azure.json based on the resource id of the master subnet.

To work properly is also depends on https://github.com/Azure/acs-engine/pull/1142.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/acs-engine/1150)

PR #1575, opened 2017-10-10
fix regression from 1150

anhowe:
**What this PR does / why we need it**: This PR fixes a regression introduced by #1150 where it causes an offset of provisioning script parameters.


karataliu:
This also fixes #1569 

@anhowe  @JackQuincy  Can you please also take a look at #1570, which introduces an improvement for such cases.

sylr:
Thanks!

PR #1582, opened 2017-10-12
removing unneeded valdiation that is failing update calls. This valid…

JackQuincy:
…ation is done better in the service code and these types can only be created through the service

Service is failing to do a get then update as the serviceprincipal isn't returned fully in get. Allowing it to be empty


JackQuincy:
I was waiting for CI but I realized CI doesn't test this change. So I'll just merge without. Side note do we want to get CI in for agent pools only deployment checking? does that make sense?

PR #1584, opened 2017-10-12
Add cloud-controller-manager support for Kubernetes cluster

karataliu:
Kubernetes v1.8 supports [cloud-controller-manager ](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/) which runs cloud provider related logic, and it is planned in v1.10 that using cloud-controller-manager becomes the only way to interact with clouds.

This PR adds an option 'UseCloudControllerManager' in apimodel for Kubernetes v1.8 version. If turned on, the cluster will be using cloud controller manager mode.

Currently it uses the common image 'cloud-controller-manager-amd64:v1.8.0' which includes all cloud provider implementation. Will later switch to azure dedicated cloud-controller-manager image once ready.

One TODO item is RBAC support, which pends on upstream Kubernetes#53511 .

Please find more details here: https://github.com/kubernetes/kubernetes/issues/50752

**Which issue this PR fixes**

**Release note**:
```release-note
Add cloud-controller-manager support for Kubernetes cluster
```


karataliu:
cc @anhowe @jdumars  This PR is for supporting standalone Azure cloud provider from acs-engine side. Could you please take a look? Thanks!

parts/kubernetesmaster-cloud-controller-manager.yaml
```diff
@@ -0,0 +1,50 @@
...
+      command: 
+        - "cloud-controller-manager"
+        - "--kubeconfig=/var/lib/kubelet/kubeconfig"
+        - "--allocate-node-cidrs=<allocateNodeCidrs>"
+        - "--cluster-cidr=<kubeClusterCidr>"
+        - "--cluster-name=<masterFqdnPrefix>"
+        - "--cloud-provider=azure"
+        - "--cloud-config=/etc/kubernetes/azure.json"
+        - "--leader-elect=true"
+# TODO: RBAC support
```
colemickens: What's needed or outstanding for enabling RBAC?

parts/kubernetesmastercustomdata.yml
```diff
@@ -302,6 +316,16 @@ write_files:
 {{end}}
     sed -i "s|<etcdApiVersion>|{{ .OrchestratorProfile.GetAPIServerEtcdAPIVersion }}|g" "/etc/kubernetes/manifests/kube-apiserver.yaml"
 
+{{if UseCloudControllerManager }}
+    sed -i "s|<kubernetesCcmImageSpec>|{{WrapAsVariable "kubernetesCcmImageSpec"}}|g; s|<masterFqdnPrefix>|{{WrapAsVariable "masterFqdnPrefix"}}|g; s|<allocateNodeCidrs>|{{WrapAsVariable "allocateNodeCidrs"}}|g; s|<kubeClusterCidr>|{{WrapAsVariable "kubeClusterCidr"}}|g; s|<kubernetesCtrlMgrRouteReconciliationPeriod>|{{WrapAsVariable "kubernetesCtrlMgrRouteReconciliationPeriod"}}|g" \
+      /etc/kubernetes/manifests/cloud-controller-manager.yaml
+
+    sed -i "/--\(cloud-config\|cloud-provider\|route-reconciliation-period\)=/d" \
```
colemickens: Is this stripping off `cloud-provider` and `cloud-config` from apiserver? Doesn't that also need to be stripped off of `controller-manager` too?

parts/kubernetesagentcustomdata.yml
```diff
@@ -125,6 +125,11 @@ write_files:
     KUBELET_FIX_43704_3=""
   {{end}}
 {{end}}
+{{if UseCloudControllerManager }}
+    CLOUD_PROVIDER=external
+{{else}}
+    CLOUD_PROVIDER=azure
+{{end}}
 
 - path: "/etc/systemd/system/kubelet.service"
```
colemickens: Is there no change needed for `kubelet.service` to include the `--provider-id` flag?

colemickens:
What's the status of this @karataliu? What level of functionality does the cluster have when this is enabled? I'm guessing volumes don't work, but does networking + load balancing?

I'm trying to understand how many of the items mentioned in the upstream issue need to happen before CCM + Azure is functional for most non-volume use-cases. Thanks.

parts/kubernetesmastercustomdata.yml
karataliu: Yeah, both need to be updated. LIne323~324 is for kube-controller-manager, Line325~326 is for apiserver.

parts/kubernetesmaster-cloud-controller-manager.yaml
karataliu: This is actually meant for enabling 'use-service-account-credentials', which does not work now, pending on kubernetes#53511

parts/kubernetesagentcustomdata.yml
karataliu: provider-id can be detected now:
https://github.com/kubernetes/kubernetes/pull/50730
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/azure/azure_instances.go#L103

karataliu:
@colemickens Yes, volumes feature does not work. I've tried with e2e conformance tests and most cases would pass excepted for the volume ones.

The upstream issue is more about migration work, Azure provider running with cloud-controller-manager could already work on a cluster. But since the whole feature is under beta stage (kubernetes/features#88), it is not recommended for production use.

colemickens:
Thanks for the info. Change seems good to me. I made a similar change in
Tectonic-Installer and everything seemed to work (at least,the sunset I
expected to work). (Granted the network setup in Tectonic-Installer is
Calico and not kubelet...)

karataliu:
Could someone please take a look? Thanks.

colemickens:
(FWIW, this looks good to me. And in a few weeks hopefully, there should be alpha support for CCM under hyperkube as well, so there won't need to be a separate image/process... at least until CCM totally splits out of the main repo and is yanked back out of hyperkube, but that could be a while).

karataliu:
@colemickens Thanks, I saw the ongoing work for hyperkube incorporating CCM.
For the change here, a dedicate parameter for CCM image is still required as it could be set to an out of band CCM release. Also during CI test, a new CCM image is built and tested every time.

jackfrancis:
Thanks for your patience @karataliu, and apologies for the rebase needs. Not a few things have changed in the conflicting surface area, I'd be happy to spawn a new PR with this work if that's easier.

karataliu:
@jackfrancis Updated, please take a look. Thanks.

jackfrancis:
Thanks @karataliu ! Running E2E now.

karataliu:
In fact the default value should be dependent on k8s version, and as planned for 1.10+ release it should only have true value.

Anyway it is good to switch to ternary here, I've updated accordingly.

PR #1600, opened 2017-10-14
use regexp to parse Linux VM names

dmitsh:
**What this PR does / why we need it**: use regexp to parse Linux VM names

**Release note**:

```release-note
```

### DIFF
--- a/pkg/armhelpers/util.go
+++ b/pkg/armhelpers/util.go
@@ -3,12 +3,33 @@ package armhelpers
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/arm/compute"
 	log "github.com/sirupsen/logrus"
 )
+
+const (
+	// TODO: merge with the RP code
+	// here there are 4 capture groups
+	//  #1 is the orchestrator (dcos|k8s|swarm|swarmm)
+	//  #2 is the agent pool name, which can contain -s
+	//  #3 is the Cluster ID for the cluster
+	//  #4 is the index of the vm in the agent pool/master pool
+	k8sLinuxVMNamingFormat         = "^([^-]+)-(.+)-([0-9a-fA-F]{8})-{0,2}([0-9]+)$"
+	k8sLinuxVMOrchestratorIndex    = 1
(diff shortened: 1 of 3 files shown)


pkg/armhelpers/util.go
```diff
@@ -3,13 +3,34 @@ package armhelpers
...
 )
 
+const (
+	// TODO: merge with the RP code
+	// here there are 4 capture groups
+	//  #1 is the orchestrator (dcos|k8s|swarm|swarmm)
+	//  #2 is the agent pool name, which can contain -s
+	//  #3 is the Cluster ID for the cluster
+	//  #4 is the index of the vm in the agent pool/master pool
+	k8sLinuxVMNamingFormat         = "^([^-]+)-(.+)-([0-9a-fA-F]{8})-{0,2}([0-9]+)$"
```
amanohar: If it's k8s we don't need to generalize the first part IMO. We already have an assumption in upgradecluster.go and all the BI.

pkg/armhelpers/util.go
```diff
@@ -39,19 +60,21 @@ func SplitBlobURI(URI string) (string, string, string, error) {
 
 // LinuxVMNameParts returns parts of Linux VM name e.g: k8s-agentpool1-11290731-0
 func LinuxVMNameParts(vmName string) (orchestrator string, poolIdentifier string, nameSuffix string, agentIndex int, err error) {
-	vmNameParts := strings.Split(vmName, "-")
-
-	if len(vmNameParts) != 4 {
+	vmNameParts := vmnameLinuxRegexp.FindStringSubmatch(vmName)
+	if len(vmNameParts) != 5 {
```
amanohar: This needs to remove. The point is it can be more than five.

dmitsh: no, it cannot

pkg/armhelpers/util.go
dmitsh: plz explain

pkg/armhelpers/util.go
amanohar: Discussed offline. This can remain.

pkg/armhelpers/util.go
amanohar: Discussed offline

PR #1607, opened 2017-10-17
Parameterize generator and orchestrator codes

anhowe:
**What this PR does / why we need it**: Parameterize generator and orchestrator codes.


pkg/acsengine/transformtestfiles/k8s_agent_upgrade_template.json
```diff
@@ -785,6 +785,22 @@
...
     },
+    "generatorCode": {
+      "defaultValue": "k8s",
+      "metadata": {
+        "description": "The generator code used to identify the generator"
+      },
+      "type": "string"
+    },
+    "orchestratorName": {
+      "defaultValue": "acsengine",
```
JackQuincy: default values here are swapped

JackQuincy: Same in the rest of these files.

pkg/armhelpers/util.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const (
 	// TODO: merge with the RP code
-	k8sLinuxVMNamingFormat         = "^k8s-(.+)-([0-9a-fA-F]{8})-{0,2}([0-9]+)$"
+	k8sLinuxVMNamingFormat         = "^[0-9a-zA-Z]{3}-(.+)-([0-9a-fA-F]{8})-{0,2}([0-9]+)$"
```
JackQuincy: is a similar change needed for windos?

JackQuincy:
a few small comments.

pkg/armhelpers/util.go
anhowe: fixed

pkg/acsengine/transformtestfiles/k8s_agent_upgrade_template.json
anhowe: fixed

pkg/acsengine/engine.go
```diff
@@ -560,6 +560,8 @@ func getParameters(cs *api.ContainerService, isClassicMode bool) (paramsMap, err
 		addValue(parametersMap, "cloudProviderRatelimitBucket", strconv.Itoa(properties.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket))
 		addValue(parametersMap, "kubeClusterCidr", properties.OrchestratorProfile.KubernetesConfig.ClusterSubnet)
 		addValue(parametersMap, "kubernetesNonMasqueradeCidr", properties.OrchestratorProfile.KubernetesConfig.NonMasqueradeCidr)
+		addValue(parametersMap, "generatorCode", DefaultGeneratorCode)
```
amanohar: The service calls GenerateTemplate() and passes `api.ContainerService` (and gets back template & params) so the info what `generatorCode` and `orchestratorName` can be derived from `api.ContainerService`. That way we prevent 2 code paths how many parameters are set.

pkg/armhelpers/util.go
```diff
@@ -68,10 +68,10 @@ func K8sLinuxVMNameParts(vmName string) (poolIdentifier, nameSuffix string, agen
 	return vmNameParts[k8sLinuxVMAgentPoolNameIndex], vmNameParts[k8sLinuxVMAgentClusterIDIndex], vmNum, nil
 }
 
-// WindowsVMNameParts returns parts of Windows VM name e.g: 50621acs9000
+// WindowsVMNameParts returns parts of Windows VM name e.g: 50621k8s9000
 func WindowsVMNameParts(vmName string) (poolPrefix string, acsStr string, poolIndex int, agentIndex int, err error) {
-	poolPrefix = strings.Split(vmName, "acs")[0]
-	poolInfo := strings.Split(vmName, "acs")[1]
+	poolPrefix = strings.Split(vmName, "k8s")[0]
```
JackQuincy: We can't hardcode this anymore. Either we need to accept the value to split on or we need to accept acs, k8s, or aks.  Honestly to keep existing clusters working I think we have to support handling at least acs/k8s.

anhowe: added a regular expression

Issue #1603, opened 2017-10-16
Dns problems using custom VNET


tesharp:
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: v0.8.0

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes

**What happened**:
I've tried deploying several clusters with v0.8.0, all using custom VNET and different options for networkPolicy (currently using "azure"). They all have dns problems after deployment though. I can see pods running with "kubectl get pods", and nodes show up, but if I try to attach, run og log a pod I only get error message

Error from server: Get https://k8s-master-xxxx:10250/containerLogs/kube-system/kube-apiserver-k8s-master-xxxx/kube-apiserver: dial tcp: lookup k8s-master-xxxx on 10.0.0.5:53: server misbehaving

My custom VNET is running Active Directory with its own dns servers (10.0.0.5), and the vnet has custom dns servers set. This seems to cause problems for dns lookups on linux nodes. On our other linux machines we have followed the guidelines at https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/virtual-network/virtual-networks-name-resolution-ddns.md and created a boot script that register the dns on linux nodes. (After allowing non-secure dns updates in AD DNS). 

Was considering just making this an extension to acs-engine to fix, but maybe there is a better option. Would be nice if it could be set as kubernetesConfig or something. For now I'm just adding this manually to continue testing.

**How to reproduce it** (as minimally and precisely as possible):

Deploy cluster to custom VNET with custom dns servers.

**Anything else we need to know**:

JackQuincy:
This shouldn't be needed. What you should need to do is follow the steps [here ](https://github.com/Azure/ACS/tree/master/docs/VNET )to add the routetable in the custom vnet's subnets so that the internal kubernetes routing functions. This isn't in our custom vnet Documentation though.  I'm not sure why

tesharp:
Using networkPolicy "azure" there is no routetable created, and not sure how that would help. I've tested this by deploying same cluster to another vnet that does not have any dns servers set on vnet. Then the /etc/resolv.conf on master has this content:

nameserver 168.63.129.16
search qddxx3nbbwnebgeeerdhuajxnf.ax.internal.cloudapp.net

[...]

Issue #1595, opened 2017-10-13
Cannot Access Nodes in Custom VNET


lauxtermann-aram:
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: v0.8.0

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes

**What happened**: Everytime I deploy a Custom VNET the entire deployment process goes perfect.

- I can ssh into the master.
- I can 'kubectl describe all', but I only get:

NAME             CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
svc/kubernetes   10.0.0.1     <none>        443/TCP   18m

The PROBLEM is: I cannot see any of my nodes. :(
Even if I 'kubectl get nodes' on the master node itself. I get:
No resources found.

**What you expected to happen**:
I would like to see & access my nodes.

**How to reproduce it** (as minimally and precisely as possible):
Deploy a custom VNET

**Anything else we need to know**:

lauxtermann-aram:
The problem was that I deployed the vnetarmtemplate/azure.kubernetes.json if you use the pre and post deploy scripts everything work fine.

PR #1608, opened 2017-10-17
adding test that cluster ip works which should fill gap in our c…

JackQuincy:
…urrent validation for vnets

Custom vnets are broken and our validation isn't catching it currently. This should catch the gap then I can use it to identify which PR broke us.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: should help with investigations into #1603 #1595


JackQuincy:
I'm expecting failures from anything that runs a custom vnet k8s scenario.

JackQuincy:
hmm... not fail tests anywhere apparently this isn't covering any gaps... I'm going to close for now.

PR #1612, opened 2017-10-18
Convert all Kubernetes VM prefix names to lowercase, and add validation to agent pools only models

anhowe:
**What this PR does / why we need it**:  Kubernetes Agent Pool names cause VM names to be uppercase.  However, kubernetes cannot handle mixed case VM names, because the cloud provider makes them lowercase.  This PR ensure VM names end up as lowercase by adding validation to agent pool only configuration.

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20170831/validate.go
+++ b/pkg/api/agentPoolOnlyApi/v20170831/validate.go
@@ -78,6 +78,12 @@ func (a *Properties) Validate() error {
 		return e
 	}
 
+	for _, agentPoolProfile := range a.AgentPoolProfiles {
+		if e := agentPoolProfile.Validate(); e != nil {
+			return e
+		}
+	}
+
 	if e := a.LinuxProfile.Validate(); e != nil {
 		return e
 	}


parts/kubernetesagentvars.t
```diff
@@ -7,9 +7,9 @@
     "{{.Name}}AvailabilitySet": "[concat('{{.Name}}-availabilitySet-', variables('nameSuffix'))]",
 {{if .IsWindows}}
     "winResourceNamePrefix" : "[substring(variables('nameSuffix'), 0, 5)]",
-    "{{.Name}}VMNamePrefix": "[concat(variables('winResourceNamePrefix'), variables('orchestratorName'), add(900,variables('{{.Name}}Index')))]",
+    "{{.Name}}VMNamePrefix": "[toLower(concat(variables('winResourceNamePrefix'), variables('orchestratorName'), add(900,variables('{{.Name}}Index'))))]",
```
JackQuincy: why?

anhowe: this ensures we never hit this problem again with Kubernetes.

seanknox: @JackQuincy the Azure cloud provider in Kubernetes converts node/VM names to lowercase. This led to a bug with some of our ACS infrastructure. This change ensures VM names are consistent at all places (Azure, Kubernetes, infra).

pkg/api/agentPoolOnlyApi/v20170831/validate.go
```diff
@@ -78,6 +78,12 @@ func (a *Properties) Validate() error {
 		return e
 	}
 
+	for _, agentPoolProfile := range a.AgentPoolProfiles {
+		if e := agentPoolProfile.Validate(); e != nil {
+			return e
+		}
```
seanknox: Not blocking, but does this add missing profile validation to this API?

anhowe: this only adds agent pool length and name validation, I'll follow-up to find out about the other missing validation

parts/kubernetesagentvars.t
JackQuincy: @amanohar @jackfrancis thoughts is this over validation. I feel like this is asking for things to get out of sync in code and us to need to use equalsfold all over the place.

pkg/api/agentPoolOnlyApi/v20170831/validate.go
amanohar: We have the validation in the service.

PR #1627, opened 2017-10-19
Doc update: attach route table to VNET

jackfrancis:
**What this PR does / why we need it**: In our custom VNET docs it doesn't mention the need to attach the created cluster's route table to the custom VNET

**Release note**:

```
NONE
```

### DIFF
--- a/docs/custom-vnet.md
+++ b/docs/custom-vnet.md
@@ -174,7 +174,7 @@ For Kubernetes clusters, we need to update the VNET to attach to the route table
 
 ```
 #!/bin/bash
-rt=$(az network route-table list -g acs-custom-vnet | jq -r '.[].name')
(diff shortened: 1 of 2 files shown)


docs/custom-vnet.md
```diff
@@ -168,5 +168,17 @@ az group deployment create -g acs-custom-vnet --name "ClusterDeployment" --templ
 
 Depending on the number of agent you have asked for the deployment can take a while.
 
+## Post-Deployment: Attach Cluster Route Table to VNET
+
+For Kubernetes clusters, we need to update the VNET to attach to the route table created by the above `az group deployment create` command. An example in bash form:
+
+```
+#!/bin/bash
+rt=$(az network route-table list -g acs-custom-vnet | jq -r '.[].name')
```
JackQuincy: .name only works if they are in the same resource group. . .[].id should work even if the vnet is in a different resource group.

PR #1628, opened 2017-10-19
Fix etcd listening address on master nodes

jchauncey:
This removes the public ip interface from the set of listening addresses for etcd. 

cc @jdumars @anhowe @jackfrancis


examples/kubernetes.json
```diff
@@ -5,7 +5,7 @@
       "orchestratorType": "Kubernetes"
     },
     "masterProfile": {
-      "count": 1,
+      "count": 3,
```
jackfrancis: why this change?

jchauncey: sorry got caught up in the commit didnt mean to push it. just fixed it

PR #1632, opened 2017-10-20
Support for CoreOS as a distro for master and agent nodes

richardjortega:
**What this PR does / why we need it**: This PR enables using CoreOS as a supported distro using the `distro` field within the `masterProfile` and `agentPoolProfiles` api-model sections. An example usage is provided at: `examples/kubernetes-coreos.json` . CoreOS custom files are within: `parts/kubernetesmastercustomdataforcoreos.yml` and `parts/kubernetesagentcustomerdataforcoreos.yml`. 

@anhowe @lachie83 @jackfrancis @uday31in 

Things this commit does NOT have:
- An automated test (would need help with that), right now I've tested it against a rebase of most recent master of acs-engine on Azure.
- acs-engine seems predominantly coupled to Ubuntu distro (e.g. `runcmd` on `cloud-init` isn't supported on CoreOS implementation of cloud-init. This may lead to difficulty of new edits/merges where tests should go across distros (or there may not be feature parity). Right now, I used a pattern I saw in #1408 where another file for custom data was used.  
- This initial implementation uses `cloud-init`/`cloud-config` (specifically CoreOS' implementation of coreos-cloudinit) but CoreOS has stopped supporting coreos-cloudint usage in favor of [Ignition](https://coreos.com/ignition/docs/latest/what-is-ignition.html). 

**Why**: CoreOS was not previously supported (on both Master and Agent nodes).


jackfrancis:
@richardjortega Sorry about the merge conflict, could you kindly fix that? @CecileRobertMichon When Richard has a clean PR, let's run some manual tests w/ this branch and build some coreos clusters.

CecileRobertMichon:
Initial testing of CoreOS cluster looks O.K., going to delay merging this for a couple of weeks since we are waiting on more data overhead on the customData field. This should fix failing E2E tests.

pkg/acsengine/engine.go
```diff
@@ -1811,6 +1834,29 @@ write_files:
...
 
+// Identifies a distro to use for master parameters
+func getDistro(properties *api.Properties) api.Distro {
+	// Set a default distro
+	distro := api.Ubuntu
+
+	if properties.MasterProfile == nil {
+		// Check if AgentPoolProfiles have an assigned distro
+		// Use distro defined by first agent profile
+		if len(properties.AgentPoolProfiles) > 0 {
```
CecileRobertMichon: Please only use the MasterProfile to determine the distro for master parameters. Properties.MasterProfile.Distro should be used to get the Distro. The default distro is set in setMasterNetworkDefaults.

parts/kubernetesmastercustomscript.sh
```diff
@@ -22,6 +22,19 @@
 # APISERVER_PRIVATE_KEY CA_CERTIFICATE CA_PRIVATE_KEY MASTER_FQDN KUBECONFIG_CERTIFICATE
 # KUBECONFIG_KEY ADMINUSER
 
+# Handling for different OSes
+shopt -s nocasematch
+OS=$(cat /etc/*release | grep ^NAME | tr -d 'NAME="')
```
CecileRobertMichon: Would it be possible to do something like this instead?

OS="UBUNTU"
shopt -s noscasematch
if [[ `$(cat /etc/*release| grep ^NAME | tr -d 'NAME="'` ==*"COREOS"* ]]; then
    OS = "COREOS"
 fi
shopt +s noscasematch     // turn off nocasematch

continue on including lines 33-36

This would ensure there are no side effects from doing shopt -s nocasematch

pkg/acsengine/engine.go
richardjortega: So how should I update [setMasterNetworkDefaults](https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults.go#L331) to look at the Agent pool profile? The reason I'm asking is because there's an api-model test related to [agents-only.json](https://github.com/Azure/acs-engine/blob/master/examples/agents-only.json).

CecileRobertMichon: I think if the master is hosted (agent pool only) then the master distro should be Ubuntu by default. In any case, the master defaults should not be set by looking at the agents, @anhowe please correct me if I'm wrong.

CecileRobertMichon:
After secondary inspection, it looks like Ubuntu deployments (and thus e2e tests) are failing due to a breaking change, possibly in kubernetesmastercustomscript.sh. I suspect one of these might be the cause:
https://github.com/richardjortega/acs-engine/blob/4147a644674ffe17d7173fbfd83add827b570161/parts/kubernetesmastercustomscript.sh#L418
https://github.com/richardjortega/acs-engine/blob/4147a644674ffe17d7173fbfd83add827b570161/parts/kubernetesmastercustomscript.sh#L26

If this gets fixed, we should be able to merge without having to wait for customData overhead.

pkg/acsengine/engine.go
richardjortega: Maybe I don't understand the use case of agents only -- is it a special use case, what if a Master was CoreOS? I would assume, we'd want to match agents with the Master's OS?

jackfrancis: The agent-only flow supports AKS (managed kubernetes), and there is currently no support for using acs-engine as a driver for agent-only cluster workflows. So, generally, we will need to avoid modifying those code paths.

richardjortega: @CecileRobertMichon The default distro that is set by  `setMasterNetworkDefaults` is not called by the main engine, it is hardcoded as `api.Ubuntu` in [getParameters](https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/engine.go#L462-L465). I will modify this code to use MasterProfile.Distro.

richardjortega: Because [setMasterNetworkDefaults](https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults.go#L325-L327) returns if the MasterProfile is nil, I will have to set the default distro as Ubuntu within the engine.

parts/kubernetesmastercustomscript.sh
richardjortega: I've added code that doesn't use the case match setting and worrying about side effects make sense.

```bash
OS=$(cat /etc/*-release | grep ^ID= | tr -d 'ID="' | awk '{print toupper($0)}')
UBUNTU_OS_NAME="UBUNTU"
RHEL_OS_NAME="RHEL"
COREOS_OS_NAME="COREOS"
```

docs/clusterdefinition.md
```diff
@@ -55,6 +55,7 @@ Here are the valid values for the orchestrator types:
 |vnetSubnetId|no|specifies the Id of an alternate VNET subnet.  The subnet id must specify a valid VNET ID owned by the same subscription. ([bring your own VNET examples](../examples/vnet))|
 |extensions|no|This is an array of extensions.  This indicates that the extension be run on a single master.  The name in the extensions array must exactly match the extension name in the extensionProfiles.|
 |vnetCidr|no| specifies the vnet cidr when using custom Vnets ([bring your own VNET examples](../examples/vnet))|
+|distro|no| Allows OS selection for Master nodes. Currently supported values are: `ubuntu` and `coreos`. If undefined, `ubuntu` is used as the OS type for Master and Agent nodes. Agent OS is determined by the Master node distro. Supported Orchestrators: Kubernetes |
```
CecileRobertMichon: "Agent OS is determined by the Master node distro" @richardjortega does this mean that the agent profile distro is disregarded no matter what? I think it would be better to take the other approach we discussed: Master profile distro parameter determines master distro and agent profile distro determines agent distro. For both, empty field defaults to ubuntu. What do you think?

richardjortega: Yes, discussed the details with Anthony yesterday. I think I pushed up a WIP and it kicked off a build. They will be independent but will have to do more testing. Also I have merged the files for custom data together.

CecileRobertMichon: Great!

parts/agentparams.t
```diff
@@ -24,6 +24,34 @@
...
+    }, 
+    "{{.Name}}osImagePublisher": {
+      "defaultValue": "Canonical",
+      "metadata": {
+        "description": "OS image publisher."
+      },
+      "type": "string"
+    }, 
+    "{{.Name}}osImageSKU": {
+      "defaultValue": "16.04-DAILY-LTS",
```
CecileRobertMichon: This image was deleted last week so we had to replace it with the monthly version (16.04-LTS). Please see #1755.

richardjortega: Updated.

jackfrancis:
@richardjortega @CecileRobertMichon I moved the examples file into `examples/coreos/` and am running E2E out of band against it now. I'd love to get this merged today!

PR #1636, opened 2017-10-21
adding scale command that can scale up

JackQuincy:
Adds a scale command so that acs-engine users can scale their clusters easier. Another PR will be needed to scale down availability sets.

**Release note**:

```release-note
NONE
```


cmd/scale.go
```diff
@@ -0,0 +1,271 @@
...
+		}
+	}
+}
+
+func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
+	sc.validate(cmd, args)
+
+	if sc.agentPool.IsAvailabilitySets() && sc.agentPool.Count > sc.newDesiredAgentCount {
+		// TODO add scale down of VMS code path
+		log.Fatalln("Scaling down availibility sets is not currently supported")
```
SaaldjorMike: Looks like there is a typo in the word `availability`

cmd/scale.go
```diff
@@ -0,0 +1,271 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return sc.run(cmd, args)
+		},
+	}
+
+	f := scaleCmd.Flags()
+	f.StringVar(&sc.location, "location", "", "location the cluster is deployed in")
+	f.StringVar(&sc.resourceGroupName, "resource-group", "", "the resource group where the cluster is deployed")
+	f.StringVar(&sc.deploymentDirectory, "deployment-dir", "", "the location of the output from `generate`")
+	f.IntVar(&sc.newDesiredAgentCount, "new-agent-count", 0, "desired number of agents")
```
SaaldjorMike: Looks like it is referenced as `new-agent-count` here, but validation of the argument prints a message calling it `agent-count`

cmd/scale.go
JackQuincy: thanks I fixed it

cmd/scale.go
JackQuincy: thanks I fixed it

JackQuincy:
@amanohar @jackfrancis can I get a review here. The code here is ugly. I'm just trying to get it out as a convenience. If this is used a lot I should move a lot of this code into an operation and share it between the RP and acs-engine

JackQuincy:
I can just remove that from the example. It isn't required.

jackfrancis:
@JackQuincy Does `--master-FQDN mycluster.westus2.cloudapp.azure.com` work? Or do you need to do `--master-FQDN https://mycluster.westus2.cloudapp.azure.com`. I'm wondering if we should make it so that only the first usage pattern is valid.

JackQuincy:
both work today. In the RP when I make these calls I use the FQDN.

jackfrancis:
Got it, 👍

PR #1657, opened 2017-10-24
acsengine: fix agent extensions

jessfraz:
Fixes a bug where it was looking at the master profile not the agent.


JackQuincy:
Thanks @jessfraz

PR #1665, opened 2017-10-25
Changed azureclient to allow changes in RP

CecileRobertMichon:
- Exposed ListProviders function for RP
- EnsureProvidersRegistered gets checked outside of getClient --> RP should not run Register if the client is not registered

**Release note**:


pkg/armhelpers/azureclient.go
```diff
@@ -100,7 +100,14 @@ func NewAzureClientWithDeviceAuth(env azure.Environment, subscriptionID string)
 				return nil, err
 			}
 			graphSpt.Refresh()
-			return getClient(env, subscriptionID, tenantID, armSpt, graphSpt)
+
+			c := getClient(env, subscriptionID, tenantID, armSpt, graphSpt)
+			err = c.ensureProvidersRegistered(subscriptionID)
+			if err != nil {
```
weinong: can you move this one level up? The goal is to decouple it from any NewAzureClient* functions.

weinong: By that you will need to expose ensureProvidersRegistered

PR #1685, opened 2017-10-30
fix(test/e2e): Provides a few fixes for various issues around the e2e tests

jchauncey:
* record an error count when we fail around waiting for nodes
* fix dashboard check so that it doesnt keep iterating even after success

feat(test/e2e/config): Add the ability to select a region at random from a preselect list of 5
* eastus, westcentralus, southcentralus, westus2, westcentralus

feat(test/e2e/metrics): Add jenkins metrics
* Also provide tags that specify which ci system ran the tests

### DIFF
--- a/test/e2e/azure/cli.go
+++ b/test/e2e/azure/cli.go
@@ -56,12 +56,13 @@ func NewAccount() (*Account, error) {
 
 // Login will login to a given subscription
 func (a *Account) Login() error {
-	_, err := exec.Command("az", "login",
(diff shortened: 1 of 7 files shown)


jackfrancis:
@jchauncey the tests that ran in this PR didn't have the region randomization, I expected that they would based on the code. Am I missing something?

jchauncey:
basically if a LOCATION env var is set it wont do randomization. Because we have that var set globally they ran all in 1 location

PR #1686, opened 2017-10-30
Switch to RS3 Windows in Kubernetes cluster

JiangtianLi:
This PR switches to use RS3 Windows image in Kubernetes cluster and adjust networking and deployment config accordingly.


jackfrancis:
@JiangtianLi Let me know when this is in good shape for E2E; additionally, @yakman2020 is interested in enabling RS3 support for dc/os clusters as well. If you could scaffold out a PR for that as well that would be wonderful.

JiangtianLi:
@jackfrancis The regression test has also passed.
@dmitsh @rjtsdl there is small part (validation, test) in this PR related to your work. PTAL.

JiangtianLi:
@yakman2020 DC/OS support on windows is being worked on by another team now. I'll let them know.

PR #1688, opened 2017-10-31
moved all kubernetes version definitions to one place

jackfrancis:
**What this PR does / why we need it**: code hygiene: rationalizes k8s versions support

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/api/common/const.go
+++ b/pkg/api/common/const.go
@@ -87,22 +87,40 @@ const (
 	KubernetesDefaultVersion string = KubernetesVersion1Dot7Dot9
 )
 
-// AllKubernetesSupportedVersions maintain a list of available k8s versions in acs-engine
-var AllKubernetesSupportedVersions = []string{
-	KubernetesVersion1Dot8Dot0,
-	KubernetesVersion1Dot8Dot1,
-	KubernetesVersion1Dot8Dot2,
-	KubernetesVersion1Dot7Dot1,
-	KubernetesVersion1Dot7Dot2,
-	KubernetesVersion1Dot7Dot4,
-	KubernetesVersion1Dot7Dot5,
-	KubernetesVersion1Dot7Dot7,
-	KubernetesVersion1Dot7Dot9,
-	KubernetesVersion1Dot6Dot6,
-	KubernetesVersion1Dot6Dot9,
-	KubernetesVersion1Dot6Dot11,
-	KubernetesVersion1Dot5Dot8,
-	KubernetesVersion1Dot5Dot7,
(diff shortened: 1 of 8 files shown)


jackfrancis:
@JackQuincy @amanohar @dmitsh FYI, this PR declares supported acs-engine k8s versions in one place (rather than scattered throughout the codebase).

pkg/api/common/const.go
```diff
@@ -87,22 +87,40 @@ const (
...
+	KubernetesVersion1Dot7Dot5:  true,
+	KubernetesVersion1Dot7Dot7:  true,
+	KubernetesVersion1Dot7Dot9:  true,
+	KubernetesVersion1Dot8Dot0:  true,
+	KubernetesVersion1Dot8Dot1:  true,
+	KubernetesVersion1Dot8Dot2:  true,
+}
+
+// GetSupportedKubernetesVersion verifies that a passed-in version string is supported, or returns a default version string if not
+func GetSupportedKubernetesVersion(version string) string {
```
amanohar: The function name `GetSupportedKubernetesVersion` is a bit confusing in terms of what it does.

Given that this function is a combination of whether a version is supported by ACS E or not and if not it returns the default version, I was wondering if there is value is breaking this function.
 E.g. if someone asks for 1.5.5, they will get 1.7.7. Is this ok/expected?

Should this be left to the caller? i.e. they check if a version is supported. And if it's not supported they explicitly ask for a default version (`GetDefaultVersion`)

And also `GetCompatibleVersionForKubernetesVersion` (bad name I admit) could return the next supported patch if that version is not available instead of default version.

jackfrancis: The purpose of the function is just to literally provide the same value as the multi-case implementation that it replaces.

And yes, the function as currently implemnted is designed to be unbreakable (i.e. something that the caller would invoke post-validation). We could add an error object as a 2nd response if an unknown string is passed it. It's not really required based on what I'm trying to do (capture all of those identical case statements in one place), but it might make the function a little easier to read and understand.

jackfrancis: @amanohar I reread the code and kicked the tires on an alternate implementation, and I think this is is the right one. Basically, in order to rationalize the prior switch/case duplicate code instances, we need a function that does exactly what `GetSupportedKubernetesVersion` does: validates that a release string is supported by returning the same string if validation passes, and if not, permissively return the default release string to deliver a working, supported release version.

The immediate high-level purpose of this PR is to eliminate instances of duplicate code to make version management saner going forward. I'd like to lock this PR in, and would be happy to improve the expressiveness of this going forward!

CecileRobertMichon:
lgtm

PR #1690, opened 2017-10-31
fix(Get-AzureConstants.py): remove the date string from the header

jchauncey:
* add test to circle config
* add make target to run test in test.mk
* add script that executes the test


jchauncey:
dont merge this yet! I have it intentionally failing right now so I could show you how it works

PR #1702, opened 2017-11-02
Fix for NVIDIA driver install

sozercan:
Currently, NVIDIA drivers are automatically installed but needs `nvidia-smi` or system restart to work properly. Also, kubelet does not recognize GPUs (`alpha.kubernetes.io/nvidia-gpu`) unless it is restarted.
With this PR, immediately after driver installation finishes, GPU should be available in Kubernetes.

We didn't need this before since VMs were restarting after deployment, but they are not restarting anymore. Is it intended they are not restarting anymore?

/cc @wbuchwalter


jackfrancis:
@sozercan There were some recent optimizations to post-deployment reboots (i.e., they are required under fewer conditions, I think). @jchauncey or @anhowe can give more detail.

jackfrancis:
Thanks @sozercan !

PR #1703, opened 2017-11-02
Dev environment setup for k8s on Windows machine

JiangtianLi:
This PR adds linux container setup script for develop and build k8s on Windows machine, similar to acs-engine devenv script.

### DIFF
--- a/Dockerfile.k8swin
+++ b/Dockerfile.k8swin
@@ -1,0 +1,22 @@
+FROM buildpack-deps:xenial
+
+RUN apt-get update \
+    && apt-get -y upgrade \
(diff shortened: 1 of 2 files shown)


jackfrancis:
@JiangtianLi does it make sense to incorporate this as a `make` target in `Makefile`?

JiangtianLi:
@jackfrancis  It is the same as devenv.ps1, which is not in Makefile, because to use make on Windows, usually you need to get into Linux container first. And this ps1 is to get into Linux container.

JiangtianLi:
@jackfrancis  I think the failure is not related. The test succeeded last time.

PR #1704, opened 2017-11-02
return correct list of available upgrades

dmitsh:
return correct list of available upgrades

### DIFF
--- a/pkg/api/orchestrators.go
+++ b/pkg/api/orchestrators.go
@@ -157,57 +157,46 @@ func kubernetesInfo(csOrch *OrchestratorProfile) ([]*OrchestratorVersionProfile,
 
 func kubernetesUpgrades(csOrch *OrchestratorProfile) ([]*OrchestratorProfile, error) {
 	ret := []*OrchestratorProfile{}
-	var err error
+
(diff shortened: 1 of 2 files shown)


pkg/api/orchestrators.go
```diff
@@ -157,59 +157,48 @@ func kubernetesInfo(csOrch *OrchestratorProfile) ([]*OrchestratorVersionProfile,
...
 		// add next version
-		ret = append(ret, &OrchestratorProfile{
-			OrchestratorType:    Kubernetes,
-			OrchestratorVersion: common.KubernetesVersion1Dot8Dot2,
-		})
-	case strings.HasPrefix(csOrch.OrchestratorVersion, "1.8"):
-		// check for patch upgrade
-		if ret, err = addPatchUpgrade(ret, csOrch.OrchestratorVersion, common.KubernetesVersion1Dot8Dot2); err != nil {
-			return ret, err
+		if nextVersion.Major() == nextMajor && nextVersion.Minor() == nextMinor {
```
JackQuincy: I don't like how this is custom logic that can get out of sync from our list of all orcherstators. Why not just get the list of all orcherstators/versions and pick the one we are on? That way they never get out of sync.

PR #1705, opened 2017-11-03
Adding missing Gov regions

gsacavdm:

**Release note**:

```release-note
```


gsacavdm:
@jackfrancis , someone is running those commands PS/CLI commands and semi-manually dumping that into the azureconst.go no? I have a hard time imagining this is fully automated because there is no way those commands would return regions from the different sovereign clouds in one single run. Depending on which account you sign in with, these commands will return different regions. So someone needs to be stitching together the results from running these commands across clouds.

If you run these commands yourself, you won't get usgovvirginia nor usgoviowa either, whatever mechanism was used to add those two should be used again.

jackfrancis:
Can confirm that the command (for my az account at least) does indeed generate the file you see in `pkg/acsengine/azureconst.go`:

```
$ rm -f pkg/acsengine/azureconst.go && python pkg/acsengine/Get-AzureConstants.py
$ git diff pkg/acsengine/azureconst.go 
diff --git a/pkg/acsengine/azureconst.go b/pkg/acsengine/azureconst.go
index 84abfa20..d41fb4ef 100644
--- a/pkg/acsengine/azureconst.go
+++ b/pkg/acsengine/azureconst.go
@@ -1,6 +1,6 @@
 package acsengine
 
-// AUTOGENERATED FILE - last generated 2017-10-26 03:04:32
+// AUTOGENERATED FILE - last generated 2017-11-06 23:13:50
 
 // AzureLocations provides all azure regions in prod.
 // Related powershell to refresh this list:
``

(i.e., the only change to the file is dated comment at the top)

You could read the generator script and investigate if add'l foo would help accommodate gov regional coverage.

File: pkg/acsengine/Get-AzureConstants.py [hunk 1]
```diff
@@ -85,6 +85,8 @@ def getFileContents(dcosMasterMap, masterAgentMap, kubernetesAgentMap, sizeMap,
 	"germanynortheast",
 	"usgovvirginia",
 	"usgoviowa",
+	"usgovarizona",
```
gsacavdm: @jackfrancis , turns out the generator scripts have those regions hardcoded, so I just added them there.

jackfrancis:
Feel free to run the script and include an updated version of `pkg/acsengine/azureconst.go` so we can smoke test the script updates. Thanks!

jackfrancis:
(I see that `pkg/acsengine/azureconst.go` doesn't currently have a masthead update, so I assume it was manually updated.)

gsacavdm:
@jackfrancis , done, almost seemed to perfect, came out exactly the same except for the modified header.

pkg/acsengine/Get-AzureConstants.ps1
```diff
@@ -115,6 +115,8 @@ Get-Locations() {
 	$locationList += "germanynortheast"
 	$locationList += "usgoviowa"
 	$locationList += "usgovvirginia"
+	$locationList += "usgovarizona"
```
anhowe: You can undo these changes, this file should actually be deleted, since we only use the python file.

jackfrancis:
@gsacavdm I'll delete the powershell script in a follow-up commit, so no worries there on your account. Kicking off E2E now.

jackfrancis:
@gsacavdm could you kindly rebase? We inadvertently moved an unrelated change to azureconst.go and broke this branch :( Will merge forthwith after things are clean, thanks for your patience.

gsacavdm:
There you go @jackfrancis , rebased :)

jackfrancis:
Thanks again for the contribution @gsacavdm !

PR #1750, opened 2017-11-10
Add k8s node label validation

CecileRobertMichon:
**What this PR does / why we need it**: Currently there is a bug where if an invalid label is entered by the user as "custom label" for a Kubernetes cluster, the deployment succeeds but the nodes never come up with `kubectl get nodes`. This PR adds a validation step and only adds labels if they are valid.

**Release note**:

```release-note
```


CecileRobertMichon:
@jackfrancis there are two things I want to discuss here:
1) Originally we talked about adding a fix to truncate a string if it was too long. After reflection, that would only be a "half" fix because their are also restrictions on the characters used ie. "lab@@@l" is not valid. Also "some-loooooooooooong-string"  might become "some-looooooooooong-" when truncated which is invalid because it ends with a dash. 
2) Is it acceptable to remove an invalid user defined label instead of "fixing it" to make it valid? Potential ways I see of dealing with invalid labels:
- discard them
- fix them
- throw an error at deployment asking the user to input a valid label
I don't think fixing is an option since we'd have to deal with special characters and strings ending with non alphanumeric characters. So the question is, is it worth failing deployment for an invalid label? I want to say no, but I need a second opinion.

jackfrancis:
Because we derive the label name from the resource group string, to do this correctly we'd really have to validate both at once, taking into account both restrictions (Azure resource group string req's, Kubernetes label string req's). That's hard :) So, let's just do two enforce k8s label requirements when we auto-create the label.

I agree that we probably will be chasing edge cases for a long time if we attempt to validate for the user. In terms of the last character, though, it should be easy to do the following:

- lastCharFunc() --> if last char is illegal, truncate it, and continue recursively
- truncateFunc() --> if string is too long, truncate it to a valid size, and then call lastCharFunc() to make sure the last char is valid (if not, lastCharFunc() will truncate further until the trailing char is valid)

CecileRobertMichon:
@jackfrancis this PR is meant to fix the bug where an invalid custom node labels breaks the cluster. Eg. If I input "&&&" as my custom label the deployment succeeds but then `kubectl get nodes` returns "resources not found". The resource name label validation is a separate issue.

jackfrancis:
Sorry for the confusion! Yes, I say discard it *and* fail validation.

PR #1753, opened 2017-11-10
keeping track of provision.sh runs

### LINKED ISSUE #1752: clusters rebooting all nodes at the same time
---
NO

---
ISSUE

---
all versioned versions and most of old git history

kubernetes

**What happened**:
All nodes in the cluster are rebooting at the same time. this is being caused by this.  https://github.com/Azure/custom-script-extension-linux/issues/117 we should probably harden our custom script so this isn't possible anymore

jackfrancis:
and only executing provision.sh if we haven’t already

**What this PR does / why we need it**: ensures that provision script only runs once

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1752 

**Release note**:

```
only execute provision script once
```

### DIFF
--- a/parts/kubernetesmastercustomscript.sh
+++ b/parts/kubernetesmastercustomscript.sh
@@ -428,4 +428,6 @@ if $REBOOTREQUIRED; then
   /bin/bash -c "shutdown -r 1 &"
 fi
 
-echo `date`,`hostname`, endscript>>/opt/m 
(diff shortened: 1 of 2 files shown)


jackfrancis:
@JackQuincy I appreciate your approval even in excess :)

PR #1755, opened 2017-11-11
update Ubuntu image

dmitsh:
update Ubuntu image


dlouvier:
Thanks! This was breaking my deployment today and I had to change the sku by hand today :d

PR #1764, opened 2017-11-13
removing master custom script extension from scaling templates

JackQuincy:
Currently old clusters that are scaled with a new acs-engine will lose their azure.json and parts of their kubelet config. Causing the cluster to stop working. this is caused by the switch from positional arguments to named variables in our cse


JackQuincy:
thanks for quick review and yes lets get E2E

PR #1771, opened 2017-11-14
Add k8s custom label validation

CecileRobertMichon:
**What this PR does / why we need it**: Currently there is a bug where if an invalid label is entered by the user as "custom label" for a Kubernetes cluster, the deployment succeeds but the nodes never come up with kubectl get nodes. This PR adds a validation step and only adds labels if they are valid.

**Release note**:

```release-note
```


pkg/api/vlabs/validate.go
```diff
@@ -666,6 +671,16 @@ func validateUniquePorts(ports []int, name string) error {
 	return nil
 }
 
+func validateKubernetesLabelValue(v string) error {
+	labelValueMaxLength := 63
+	labelIdentifierFmt := "([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]"
+	var re = regexp.MustCompile("^" + labelIdentifierFmt + "$")
```
JackQuincy: look at \pkg\armhelpers\util.go and how we compile our regexes there.

pkg/api/vlabs/validate.go
```diff
@@ -666,6 +671,16 @@ func validateUniquePorts(ports []int, name string) error {
 	return nil
 }
 
+func validateKubernetesLabelValue(v string) error {
+	labelValueMaxLength := 63
```
JackQuincy: this should be a const declared statically

pkg/api/vlabs/validate.go
```diff
@@ -357,6 +357,11 @@ func (a *Properties) Validate(isUpdate bool) error {
 			switch a.OrchestratorProfile.OrchestratorType {
 			case DCOS:
 			case Kubernetes:
+				for _, l := range agentPoolProfile.CustomNodeLabels {
```
amanohar: To confirm is this the label that was added a couple of weeks ago what the label can be the name of resource group? If yes, I would set the limit to 90 character since that's ARM resource group length limit.

CecileRobertMichon: @amanohar no this is for custom node labels

CecileRobertMichon: Also https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set  " Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between."

amanohar: thanks!

pkg/api/vlabs/validate.go
```diff
@@ -666,6 +678,13 @@ func validateUniquePorts(ports []int, name string) error {
 	return nil
 }
 
+func validateKubernetesLabelValue(v string) error {
+	if !(len(v) == 0) && !labelValueRegex.MatchString(v) {
```
JackQuincy: dropped the length check

CecileRobertMichon:
I'm working on adding validation for the key so this PR isn't ready yet

pkg/api/vlabs/validate.go
CecileRobertMichon: Yep I did that on purpose. I added it to the regexp instead, that way i's cleaner. I forgot to delete the length const, I'll push that with my next commit.

PR #1772, opened 2017-11-15
adding 'cordon and drain' to upgrade

dmitsh:
- adding 'cordon and drain' for agent pool upgrades
- creating a new agent node prior to starting deleting and recreating existing agent nodes


pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -38,13 +38,20 @@ type UpgradeAgentNode struct {
...
+		kubeAPIServerURL = kan.UpgradeContainerService.Properties.MasterProfile.FQDN
+	}
 
 	// Currently in a single node cluster the api server will not be running when this point is reached on the first node so it will always fail.
-	// err := operations.SafelyDrainNode(kan.Client, log.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vm.Name)
-	// if err != nil {
-	// 	log.Infoln(fmt.Sprintf("Error draining agent VM: %s", *vm.Name))
-	// 	return err
-	// }
+	err := operations.SafelyDrainNode(kan.Client, logrus.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
```
JackQuincy: can we surround this with if kan.kubeConfig != nil{} So that we can not pass it in if we don't want to do this

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +213,81 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
+			agentIndex := getAvailableIndex(upgradeVMs)
+			ku.logger.Infof("Adding auxiliary agent node with index %d", agentIndex)
 
 			err = upgradeAgentNode.CreateNode(*agentPool.Name, agentIndex)
 			if err != nil {
-				ku.logger.Infof("Error creating upgraded agent VM: %s\n", *vm.Name)
+				ku.logger.Infof("Error creating auxiliary node\n")
 				return err
 			}
+			upgradeVMs[agentIndex] = &UpgradeVM{"", true}
```
JackQuincy: huh, don't we want to store it's name here? we depend on it below.  We could make CreateNode return the name of the node created

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +213,81 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
-			err := upgradeAgentNode.DeleteNode(vm.Name)
-			if err != nil {
-				ku.logger.Infof("Error deleting agent VM: %s\n", *vm.Name)
-				return err
-			}
+		// Create an auxiliary node if hasn't been created yet.
+		// This node is used to take on the load from deleting nodes
+		addedNode := false
+		if agentCount > 0 && len(upgradeVMs) != agentCount {
+			addedNode = true
```
JackQuincy: this shouldn't be needed.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +213,81 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
+			err := upgradeAgentNode.DeleteNode(&vm.Name)
 			if err != nil {
-				ku.logger.Infof("Error validating upgraded agent VM: %s, err: %v\n", *vm.Name, err)
+				ku.logger.Infof("Error deleting agent VM: %s\n", vm.Name)
 				return err
 			}
 
-			upgradedAgentsIndex[agentIndex] = true
+			// do not create last node in favor of auxillary node
+			if addedNode && upgradedCount == toBeUpgraded-1 {
```
JackQuincy: I feel like this logic should be easier and that we should do it in reverse, 
if NotLastVM { 
CreateNode
}
delete from map

pkg/operations/kubernetesupgrade/upgrader.go
JackQuincy: The only case we don't add a node is if we failed part way and we already have an extra node. Right? or did I miss something

pkg/operations/kubernetesupgrade/upgrader.go
JackQuincy: Condition should be  "upgradedCount < toBeUpgraded -1"

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -38,13 +38,20 @@ type UpgradeAgentNode struct {
...
 func (kan *UpgradeAgentNode) DeleteNode(vmName *string) error {
+	var kubeAPIServerURL string
+
+	if kan.UpgradeContainerService.Properties.HostedMasterProfile != nil {
+		kubeAPIServerURL = kan.UpgradeContainerService.Properties.HostedMasterProfile.FQDN
+	} else {
+		kubeAPIServerURL = kan.UpgradeContainerService.Properties.MasterProfile.FQDN
+	}
 
 	// Currently in a single node cluster the api server will not be running when this point is reached on the first node so it will always fail.
```
JackQuincy: because of this can we set kubeconfig to nil in single master upgrade command? or fail with a force option that makes kubeConfig empty/nil?

pkg/operations/kubernetesupgrade/upgrader.go
dmitsh: We don't know the name here, and don't need it as we are not going to upgrade this node.
I will add comments

pkg/operations/kubernetesupgrade/upgrader.go
dmitsh: Yes, correct. I added this to make upgrade idempotent.

pkg/operations/kubernetesupgrade/upgradeagentnode.go
dmitsh: I will add validation for master nodes, so the nodes will be ready.

dmitsh:
Will create a clean PR

PR #1776, opened 2017-11-15
adding 'cordon and drain' for agent pool upgrades

dmitsh:
- adding 'cordon and drain' for agent pool upgrades
- creating a new agent node prior to starting deleting and recreating existing agent nodes
- waiting for master nodes to become ready after upgrade prior to proceeding with agent pool upgrades

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrademasternode.go
+++ b/pkg/operations/kubernetesupgrade/upgrademasternode.go
@@ -87,10 +87,6 @@ func (kmn *UpgradeMasterNode) Validate(vmName *string) error {
 
 	masterURL := kmn.UpgradeContainerService.Properties.MasterProfile.FQDN
 
-	if masterURL == "" {
-		return kmn.Translator.Errorf("Control plane FQDN was not set.")
-	}
-
 	client, err := kmn.Client.GetKubernetesClient(masterURL, kmn.kubeConfig, interval, timeout)
 	if err != nil {
 		return err
--- a/pkg/operations/kubernetesupgrade/upgrademasternode.go
+++ b/pkg/operations/kubernetesupgrade/upgrademasternode.go
@@ -73,22 +73,22 @@ func (kmn *UpgradeMasterNode) CreateNode(poolName string, masterNo int) error {
 	return nil
 }
 
-// Validate will verify the that master/agent node has been upgraded as expected.
+// Validate will verify the that master node has been upgraded as expected.
 func (kmn *UpgradeMasterNode) Validate(vmName *string) error {
 	if vmName == nil || *vmName == "" {
 		kmn.logger.Warningf("VM name was empty. Skipping node condition check")
 		return nil
 	}
 
-	var masterURL string
-	if kmn.UpgradeContainerService.Properties.HostedMasterProfile != nil {
-		masterURL = kmn.UpgradeContainerService.Properties.HostedMasterProfile.FQDN
-	} else {
-		masterURL = kmn.UpgradeContainerService.Properties.MasterProfile.FQDN
+	if kmn.UpgradeContainerService.Properties.MasterProfile == nil {
+		kmn.logger.Warningf("Master profile was empty. Skipping node condition check")
+		return nil
 	}
 
+	masterURL := kmn.UpgradeContainerService.Properties.MasterProfile.FQDN
+
 	if masterURL == "" {
-		kmn.Translator.Errorf("Control plane FQDN was not set.")
+		return kmn.Translator.Errorf("Control plane FQDN was not set.")
 	}
 
 	client, err := kmn.Client.GetKubernetesClient(masterURL, kmn.kubeConfig, interval, timeout)
@@ -99,11 +99,11 @@ func (kmn *UpgradeMasterNode) Validate(vmName *string) error {
 	ch := make(chan struct{}, 1)
 	go func() {
 		for {
-			agentNode, err := client.GetNode(*vmName)
+			masterNode, err := client.GetNode(*vmName)
 			if err != nil {
 				kmn.logger.Infof("Master VM: %s status error: %v\n", *vmName, err)
 				time.Sleep(time.Second * 5)
-			} else if node.IsNodeReady(agentNode) {
+			} else if node.IsNodeReady(masterNode) {
 				kmn.logger.Infof("Master VM: %s is ready", *vmName)
 				ch <- struct{}{}
 			} else {
--- a/pkg/operations/kubernetesupgrade/upgrademasternode.go
+++ b/pkg/operations/kubernetesupgrade/upgrademasternode.go
@@ -5,6 +5,8 @@ import (
 	"math/rand"
 	"time"
 
+	"k8s.io/client-go/pkg/api/v1/node"
+
 	"github.com/Azure/acs-engine/pkg/api"
 	"github.com/Azure/acs-engine/pkg/armhelpers"
 	"github.com/Azure/acs-engine/pkg/i18n"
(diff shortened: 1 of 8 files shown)


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +215,86 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
+		}
 
-			err := upgradeAgentNode.DeleteNode(vm.Name)
-			if err != nil {
-				ku.logger.Infof("Error deleting agent VM: %s\n", *vm.Name)
-				return err
-			}
+		// Create an auxiliary node if hasn't been created yet.
+		// This node is used to take on the load from deleting nodes
+		addedNode := false
```
JackQuincy: you missed my point on the last PR. We shouldn't need this bool. we should be able to just use the current count and desired count to figure this one out.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +215,86 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
+		// In a normal mode of operation the actual number of VMs in the pool [len(upgradeVMs)] is equal to agentCount.
+		// However, if the upgrade failed in the middle, the actual number of VMs might be agentCount+1 to include auxiliary node.
+		// In the later case we don't create an extra node.
+		if agentCount > 0 && len(upgradeVMs) == agentCount {
```
JackQuincy: not sure we really need to be concerned about being asked to upgrade 0 node agent pools, but I guess it doesn't hurt.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +215,86 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
-				ku.logger.Infof("Error validating upgraded agent VM: %s, err: %v\n", *vm.Name, err)
+				ku.logger.Infof("Error deleting agent VM: %s\n", vm.Name)
 				return err
 			}
 
-			upgradedAgentsIndex[agentIndex] = true
+			// do not create last node in favor of auxiliary node
+			if addedNode && upgradedCount == toBeUpgraded-1 {
+				ku.logger.Infof("Skipping creation of VM %s (indx %d) in favor of auxiliary node", vm.Name, agentIndex)
+				delete(upgradeVMs, agentIndex)
```
JackQuincy: be careful with this. Concurrent modification while looping through a map makes it very easy to skip nodes etc. Though in this case it seems like you only do it once you've determined you have already upgraded all the nodes you need to upgrade

dmitsh: deleting map element while looping through it is safe.

pkg/operations/kubernetesupgrade/upgrader.go
dmitsh: I'm using 'addedNode' in the loop below. IMO it make the code more clean and readable.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -216,51 +215,83 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
-		agentsToCreate := agentCount - len(upgradedAgentsIndex)
+		agentsToCreate := agentCount - upgradedCount
 		ku.logger.Infof("Expected agent count in the pool: %d, Creating %d more agents\n", agentCount, agentsToCreate)
 
 		// NOTE: this is NOT completely idempotent because it assumes that
 		// the OS disk has been deleted
 		for i := 0; i < agentsToCreate; i++ {
 			agentIndexToCreate := 0
-			for upgradedAgentsIndex[agentIndexToCreate] == true {
+			for upgradeVMs[agentIndexToCreate].Upgraded == true {
```
JackQuincy: part of me wonders if we have a bug here. Where if we are short vms we might not have a vm at an index. which would cause an index/key not found error and we aren't doing the safe way to check the map.  But I don't think this is going to change with this PR. @amanohar agree?

pkg/operations/kubernetesupgrade/upgrademasternode.go
```diff
@@ -72,5 +75,51 @@ func (kmn *UpgradeMasterNode) CreateNode(poolName string, masterNo int) error {
...
+
+	client, err := kmn.Client.GetKubernetesClient(masterURL, kmn.kubeConfig, interval, timeout)
+	if err != nil {
+		return err
+	}
+
+	ch := make(chan struct{}, 1)
+	go func() {
+		for {
+			agentNode, err := client.GetNode(*vmName)
```
amanohar: Rename the var

pkg/operations/kubernetesupgrade/upgrademasternode.go
```diff
@@ -72,5 +75,51 @@ func (kmn *UpgradeMasterNode) CreateNode(poolName string, masterNo int) error {
...
 func (kmn *UpgradeMasterNode) Validate(vmName *string) error {
-	return nil
+	if vmName == nil || *vmName == "" {
+		kmn.logger.Warningf("VM name was empty. Skipping node condition check")
+		return nil
+	}
+
+	var masterURL string
+	if kmn.UpgradeContainerService.Properties.HostedMasterProfile != nil {
+		masterURL = kmn.UpgradeContainerService.Properties.HostedMasterProfile.FQDN
```
amanohar: If this is hostedmasterprofile then there is no node to validate or this code path won't even be it. Can you validate?

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -19,6 +19,12 @@ type Upgrader struct {
 	kubeConfig string
 }
 
+// UpgradeVM holds VM name and upgrade status
+type UpgradeVM struct {
```
amanohar: I am not clear about the utility of this extra struct. Name can already be derived from: `upgradedAgentsIndex`. If you want change it I would recommend replacing `upgradedAgentsIndex` with this struct.

amanohar: Or to put it another way what is the utility of using a combination of vm name and index to derive information?

pkg/operations/kubernetesupgrade/upgrader.go
amanohar: I don't think this should be merged its written now and corner cases could easily result in error. Plus see my question above. I am not sure if adding upgradeVMs adds value.

pkg/operations/kubernetesupgrade/upgrader.go
dmitsh: I need an array of all VMs, both upgraded or to be upgraded. UpgradeVM serves an element in that array. Index is used to create/delete the VMs. Let's discuss offline if the purpose in not clear.

JackQuincy: Side note the change to naming vms makes it slightly harder to generate vm name based on index, but we must have the data since we are making templates to deploy new vms.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -197,10 +196,15 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
-		for _, app := range ku.ClusterTopology.DataModel.Properties.AgentPoolProfiles {
+		var agentCount, agentPoolIndex int
+		var agentOsType api.OSType
+		var agentPoolName string
+		for indx, app := range ku.ClusterTopology.DataModel.Properties.AgentPoolProfiles {
 			if app.Name == *agentPool.Name {
 				agentCount = app.Count
+				agentOsType = app.OSType
+				agentPoolName = app.Name
+				agentPoolIndex = indx
```
amanohar: Where is this being used?

dmitsh: vmName, err := armhelpers.GetK8sVMName(agentOsType, ku.DataModel.Properties.HostedMasterProfile != nil,
				ku.NameSuffix, agentPoolName, agentPoolIndex, agentIndex)

PR #1793, opened 2017-11-18
Fix rescheduler test

CecileRobertMichon:
**What this PR does / why we need it**: fix failing regression test because expecting rescheduler addon which is disabled by default

**Release note**:


test/cluster-tests/kubernetes/test.sh
```diff
@@ -102,7 +103,7 @@ fi
 ###### Check existence and status of essential pods
 
 # we test other essential pods (kube-dns, kube-proxy) separately
-pods="heapster rescheduler kube-addon-manager kube-apiserver kube-controller-manager kube-scheduler tiller"
+pods="heapster kube-addon-manager kube-apiserver kube-controller-manager kube-scheduler tiller"
```
jackfrancis: you could just do the following after the original `pods` variable assignment above:

```
if (( ${EXPECTED_RESCHEDULER} != 0 )); then
    pods="$pods rescheduler"
fi
```

PR #1801, opened 2017-11-20
making the error body be returned on deployment errors from arm in pr…

JackQuincy:
…eflight

makes preflight errors in the deployment be returned from the arm helper so above code can return them/act on them

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -292,13 +292,14 @@ func (dc *deployCmd) run() error {
...
-	_, err = dc.client.DeployTemplate(
+	if res, err := dc.client.DeployTemplate(
 		dc.resourceGroup,
 		fmt.Sprintf("%s-%d", dc.resourceGroup, deploymentSuffix),
 		templateJSON,
 		parametersJSON,
-		nil)
-	if err != nil {
+		nil); err != nil {
+		b, _ := json.Marshal(res)
```
weinong: does it provide value? `json.Marshal()` does not serialize the response body...

JackQuincy:
@weinong this time I actually caused a preflight valdiation error and valdiated the response body gets printed. Mostly is redundant with the error message but could technically be different depending on the message

PR #1816, opened 2017-11-22
make 'cordon and drain' conditional in upgrade

dmitsh:
During upgrade operation do not invoke 'cordon and drain' for unhealthy nodes.

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -50,8 +50,8 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 
 		err := operations.SafelyDrainNode(kan.Client, logrus.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
 		if err != nil {
-			kan.logger.Errorf(fmt.Sprintf("Error draining agent VM %s: %v", *vmName, err))
-			return err
+			kan.logger.Warningf("Error draining agent VM %s. Proceeding with deletion. Error: %v", *vmName, err)
+			// Proceed with deletion anyways
 		}
 	}
 	if err := operations.CleanDeleteVirtualMachine(kan.Client, kan.logger, kan.ResourceGroup, *vmName); err != nil {
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -48,7 +48,7 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 			kubeAPIServerURL = kan.UpgradeContainerService.Properties.MasterProfile.FQDN
 		}
 
-		err := operations.SafelyDrainNode(kan.Client, logrus.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
+		err := operations.SafelyDrainNode(kan.Client, kan.logger, kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
 		if err != nil {
 			kan.logger.Warningf("Error draining agent VM %s. Proceeding with deletion. Error: %v", *vmName, err)
 			// Proceed with deletion anyways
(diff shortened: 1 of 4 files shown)


pkg/operations/kubernetesupgrade/upgrademasternode.go
```diff
@@ -32,7 +32,7 @@ type UpgradeMasterNode struct {
 // DeleteNode takes state/resources of the master/agent node from ListNodeResources
 // backs up/preserves state as needed by a specific version of Kubernetes and then deletes
 // the node
-func (kmn *UpgradeMasterNode) DeleteNode(vmName *string) error {
+func (kmn *UpgradeMasterNode) DeleteNode(vmName *string, dummy bool) error {
```
weinong: `dummy`?

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -37,7 +37,7 @@ type UpgradeAgentNode struct {
 // DeleteNode takes state/resources of the master/agent node from ListNodeResources
 // backs up/preserves state as needed by a specific version of Kubernetes and then deletes
 // the node
-func (kan *UpgradeAgentNode) DeleteNode(vmName *string) error {
+func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
```
weinong: where is it being used?

JackQuincy: might want to use this value you are plumbing through

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -234,7 +234,7 @@ func (ku *Upgrader) upgradeAgentPools() error {
 
 			if vmProvisioningState != string(api.Succeeded) {
 				ku.logger.Infof("Deleting agent VM %s in provisioning state %s\n", *vm.Name, vmProvisioningState)
```
JackQuincy: do we have this log? I'm wondering if the failure cases are caused because of the node just not being in K8s at which point it should be if empty don't drain, else do. Thoughts?

dmitsh: If the VM's provisioning state is not succeeded, then it really doesn't matter whether the node in K8S or not. K8S is installed as part of CSE. If that failed, then the node is not healthy from Kubernetes perspective, therefore no need to drain it. If the provisioning has failed before CSE, then K8S is not installed at all, hence again, no need to drain.

JackQuincy: There are other reasons a node might be unhealthy in k8s. Some of those cases there might be work loads running on the node. I agree in the case cse fails then we should just skip draining. But in that case I'm guessing it just won't have a provision state/k8s won't know it exists since the node will never report in.

dmitsh: Sure, the node might become unhealthy (from K8S perspective), but the provisioning state will still be 'Succeeded' (from ARM perspective), right? The provisioning state doesn't change once the VM has been provisioned.

JackQuincy: ah this is the vm in arm perspective. Sorry I'm getting confused. Yeah then this should make sense.

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -37,21 +37,23 @@ type UpgradeAgentNode struct {
...
+			kubeAPIServerURL = kan.UpgradeContainerService.Properties.HostedMasterProfile.FQDN
+		} else {
+			kubeAPIServerURL = kan.UpgradeContainerService.Properties.MasterProfile.FQDN
+		}
 
-	err := operations.SafelyDrainNode(kan.Client, logrus.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
-	if err != nil {
-		kan.logger.Errorf(fmt.Sprintf("Error draining agent VM: %s", *vmName))
-		return err
+		err := operations.SafelyDrainNode(kan.Client, logrus.New().WithField("operation", "upgrade"), kubeAPIServerURL, kan.kubeConfig, *vmName, time.Minute)
```
JackQuincy: can we switch this to use kan.logger?

PR #1817, opened 2017-11-22
change default azure disk storage class for managed disk cluster

### LINKED ISSUE #1736: The default storage class is unmanaged even create a managed disk k8s cluster

---

ISSUE
---

v0.9.1
---

kubernetes.json settings:

"masterProfile": {                           
    "count": 1,                              
    "dnsPrefix": "k8s-master",        
    "vmSize": "Standard_D2_v2",              
    "storageProfile" : "ManagedDisks"        
},                                           
"agentPoolProfiles": [{                      
    "name": "agentpool1",                    
    "count": 3,                              
    "vmSize": "Standard_D2_v2",              
    "storageProfile" : "ManagedDisks"        
}],                                          

Kubernetes v1.7.9

**What happened**:

I provisioned a kubernetes cluster via acs-engine with managed disks storage profile, while trying to install a wordpress helm chart via helm install command, found the pods start failed with following error:

AttachVolume.Attach failed for volume "pvc-e0606d76-c445-11e7-a3a0-0017fa00247a" : Attach volume "spz-k8s-t3-master-dynamic-pvc-e0606d76-c445-11e7-a3a0-0017fa00247a" to instance "k8s-agentpool1-20711447-2" failed with compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="OperationNotAllowed" Message="Addition of a blob based disk to VM with managed disks is not supported."

It seems the root cause is the acs-engine didn't set the default storage class to managed disk, so while the helm chart installing, it will create a unmanaged storage account, then the mount pvc volume will failed.

In most of cases, the helm charts should be developed by 3rd party and they commonly use the default storage class to provision pv, the acs-engine should automatically set the default storage class to managed disk if the k8s cluster is managed disk.

1. Add storageProfile" : "ManagedDisks" field in the masterProfile and agentPoolProfiles sections in kubernetes.json file.
2. acs-engine generate kubernetes.json to generate the ARM template and parameter files.
3. create an ARM deployment based on generated files
4. install helm in k8s master node 
5. use helm install stable/wordpress
6. use kubectl get events to see the volume mount failed message

Some helm chart developers allow to receive the storage class settings from input configuration file, but the helm charts end user shouldn't have to solve this issue by themselves, they are deserved to avoid the differences between acs-engine provisioned k8s cluster and the "standard" k8s cluster (since the managed disk issue is Azure specific)

andyzhangx:
Usually customer would use default azure disk to create k8s objects, e.g. helm char, when user create a k8s cluster with managed disk agent pool, the default azure disk storage class should also be set as managed_standard. 

I have done test for deploying both managed and unmanaged k8s cluster, it works well.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1736

BTW, there is condition that customer would deploy two agent pool, one is unmanged, another is managed, then `default` storage class won't work for both. In that situation, we would suggest customer not use default azure disk storage class.

**Release note**:

```
change default azure disk storage class for managed disk cluster
```


parts/kubernetesmasteraddons-managed-azure-storage-classes.yaml
```diff
@@ -0,0 +1,35 @@
...
+    kubernetes.io/cluster-service: "true"
+provisioner: kubernetes.io/azure-disk
+parameters:
+  kind: Managed
+  storageaccounttype: Premium_LRS
+---
+apiVersion: storage.k8s.io/v1beta1
+kind: StorageClass
+metadata:
+  name: unmanaged-standard
```
jackfrancis: @andyzhangx I notice that the unmanaged yaml file sets this value to `managed-standard`. Is it intentional that the two `...-azure-storage-classes.yaml` manifests have these somewhat non-intuitive distinctions?

andyzhangx: @jackfrancis good catch, because the default storage class of `kubernetesmasteraddons-managed-azure-storage-classes.yaml` is already `managed-standard`. Anyway, the `unmanaged-standard` storage class is useless in managed cluster and `managed-standard` storage class is useless in unmanaged cluster.  I will change this storage class to `managed-standard`

jackfrancis:
@andyzhangx You're going to have to account for the delivery of the file to the master at `/etc/kubernetes/addons/azure-storage-classes.yaml`. See here:

https://github.com/Azure/acs-engine/blob/master/parts/kubernetesmastercustomdata.yml#L165

andyzhangx:
@jackfrancis yes, `MASTER_ADDON_AZURE_STORAGE_CLASSES_B64_GZIP_STR` will be changed to managed storage class if it's a agent pool with managed disk. And I have tested it works well.

jackfrancis:
Thanks @andyzhangx !

PR #1819, opened 2017-11-22
Add Azure file test for windows

JiangtianLi:
Add Azure file test for windows


jackfrancis:
@JiangtianLi There are two files which need a gofmt cleansing:

```
test/e2e/kubernetes/persistentvolumeclaims/persistentvolumeclaims.go:1::warning: file is not gofmted with -s (gofmt)
test/e2e/kubernetes/storageclass/storageclass.go:1::warning: file is not gofmted with -s (gofmt)
```

Let me know when this PR is ready for an E2E run!

JiangtianLi:
@jackfrancis Thanks. Will fix the format.

JiangtianLi:
@jackfrancis @jchauncey Is it possible to run only this azure file test on 1.8.2? Due to [the fix in controller-manager in 1.8.2](https://github.com/kubernetes/kubernetes/blob/v1.8.2/pkg/volume/azure_file/azure_provision.go#L140) and [the bug in 1.7.9](https://github.com/kubernetes/kubernetes/blob/v1.7.9/pkg/volume/azure_file/azure_provision.go#L141), and because of [storage name limit](https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-shares--directories--files--and-metadata), this test may fail due to long cluster name.

PR #1829, opened 2017-11-27
 change k8s specific config except gcr.io mirror for azure china to default value

pengzhisun:

Since some of the binary files in acsengine.blob.core.chinacloudapi.cn are out-of-dated and that storage account is not maintained anymore, change to the default download source acs-mirror.azureedge.net, the download speed should be acceptable from azure china network environment.

**Release note**:

```release-note
```
NONE


pkg/acsengine/defaults.go
```diff
@@ -130,12 +130,8 @@ var (
...
 		KubernetesSpecConfig: KubernetesSpecConfig{
-			KubernetesImageBase:              "crproxy.trafficmanager.net:6000/google_containers/",
-			TillerImageBase:                  "crproxy.trafficmanager.net:6000/kubernetes-helm/",
-			EtcdDownloadURLBase:              "https://acsengine.blob.core.chinacloudapi.cn/github-coreos",
-			CNIPluginsDownloadURL:            "https://acsengine.blob.core.chinacloudapi.cn/cni/cni-plugins-amd64-latest.tgz",
-			VnetCNILinuxPluginsDownloadURL:   "https://acsengine.blob.core.chinacloudapi.cn/cni/azure-vnet-cni-linux-amd64-latest.tgz",
-			VnetCNIWindowsPluginsDownloadURL: "https://acsengine.blob.core.chinacloudapi.cn/cni/azure-vnet-cni-windows-amd64-latest.zip",
+			KubernetesImageBase: "crproxy.trafficmanager.net:6000/google_containers/",
+			TillerImageBase:     "crproxy.trafficmanager.net:6000/kubernetes-helm/",
 		},
```
jackfrancis: What are we doing about the other 4 config properties that are deleted as a part of this change?

sharmasushant: Why are we removing the download URLs for Vnet CNI plugin?

pengzhisun: In previous version, the azure-vnet-cni-linux-amd64-v0.9.tar.gz was downloaded from  github directly, but due to the latency and GFW issue the download speed is not acceptable from Azure China network, so we added these 3 cni related config properties to download binaries from a local storage account, the original commit is: https://github.com/Azure/acs-engine/commit/678223cd1e0221148d580348693f66a817da88b9#diff-8edb2653431a5da9ad1f4286097702a7

But in release v0.9.4, we found the EtcdDownloadURLBase property with value              "https://acsengine.blob.core.chinacloudapi.cn/github-coreos" is not correct, because we don't have the github-coreos mirror in acsengine blob storage, which will cause the k8s cluster provision failed in Azure China, the original commit for this change is: https://github.com/Azure/acs-engine/commit/14fbc90bdca2a3e5fd0f354373b254be077225f8#diff-8edb2653431a5da9ad1f4286097702a7

To resolve this issue, we have temporarily added the missed binary to that blob storage, but we have noticed the cni related binaries in that storage account are out-of-dated and wasn't maintained anymore, then we tried to install them directly from the official mirror repo acs-mirror.edge.net in Azure China network and found the download speed is more acceptable than github.

This PR is not removing the download URLs for the cni related properties at all, we are just removing the obsolete mirror and leveraging the official acs-mirror just like other national clouds.

jackfrancis: @sharmasushant @pengzhisun We'll defer to your real world usage patterns on the other side of the Wall, thanks for continuing to maintain this vector for us!

PR #1837, opened 2017-11-28
Adding Prometheus and Grafana

ritazh:
Adding Prometheus and Grafana to acs-engine to enable out of the box k8s monitoring experience

We have decided to create a new extension for this purpose to fully utilize the existing Prometheus and Grafana helm charts to manage releases and updates of existing charts. The new `prometheus-grafana-k8s` extension installs the helm cli and installs the Prometheus and Grafana charts.

cc @tstringer

### DIFF
--- a/extensions/prometheus-grafana-k8s/README.md
+++ b/extensions/prometheus-grafana-k8s/README.md
@@ -1,15 +1,83 @@
 # prometheus-grafana Extension
 
 
-Sample prometheus-grafana extension.  Runs the following on the master:
+This is the prometheus-grafana extension.  Add this extension to your json file as shown below to automatically enable prometheus and grafana in your new Kubernetes cluster.
 
 ```
- 
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes"
+    },
+    "masterProfile": {
+      "count": 1,
+      "dnsPrefix": "",
+      "vmSize": "Standard_DS2_v2",
+      "extensions": [
+        { 
+          "name": "prometheus-grafana-k8s"
+        }
+      ]
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "agentpool1",
+        "count": 3,
+        "vmSize": "Standard_DS2_v2",
+        "availabilityProfile": "AvailabilitySet"
+      }
+    ],
+    "linuxProfile": {
+      "adminUsername": "azureuser",
+      "ssh": {
+        "publicKeys": [
+          {
+            "keyData": ""
+          }
+        ]
+      }
+    },
+    "extensionProfiles": [
+      { 
+        "name": "prometheus-grafana-k8s", 
+        "version": "v1",
+        "rootURL": "https://raw.githubusercontent.com/ritazh/acs-engine/feat-monitor/"
+      }
+    ],
+    "servicePrincipalProfile": {
+      "clientId": "",
+      "secret": ""
+    }
+  }
+}
 ```
 
-You can validate that the extension was run by running:
+
+The following script will be executed on the master:
+
 ```
-kubectl get pods --show-all
+$ prometheus-grafana-k8s.sh
+```
+
+You can validate that the extension is running as expected with the following commands:
+
+```
+$ kubectl get pods --show-all
+
+$ NAMESPACE=default
+$ K8S_SECRET_NAME=dashboard-grafana
+
+# Get user name and password for the grafana dashboard
+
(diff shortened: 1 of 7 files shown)


jackfrancis:
Thanks @ritazh! Could you describe the UX for opting into having this extension installed on your acs-engine deployed cluster?

extensions/prometheus-grafana-k8s/README.md
```diff
@@ -0,0 +1,98 @@
+# prometheus-grafana Extension
+
+
+This is the prometheus-grafana extension.  Add this extension to your json file as shown below to automatically enable prometheus and grafana in your new Kubernetes cluster.
```
jackfrancis: Could you kindly replace "your json file" with "the api model you pass as input into `acs-engine`"

jackfrancis:
@ritazh Ignore that, I see the api model pattern documented in the README!

extensions/prometheus-grafana-k8s/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+$ prometheus-grafana-k8s.sh
+```
+
+You can validate that the extension is running as expected with the following commands:
+
+```
+$ kubectl get pods --show-all
+
+$ NAMESPACE=default
```
jchauncey: Is this configurable? If not it probably should be.

ritazh:
@jackfrancis To opt out of this feature, you simple do not include `extensions` and `extensionProfiles` in the json.

extensions/prometheus-grafana-k8s/README.md
tstringer: Currently it is not configurable. Definitely agreed that the end user should be able to specify a namespace.

jackfrancis:
@tstringer Would you add an example `.json` api model to the `examples/extensions/` directory so that we can easily run E2E tests using this cluster configuration? Here's something to use as a reference, an example api model that uses the OMS extension:

https://github.com/Azure/acs-engine/blob/master/examples/extensions/kubernetes.oms.json

extensions/prometheus-grafana-k8s/README.md
```diff
@@ -0,0 +1,98 @@
...
+            "keyData": ""
+          }
+        ]
+      }
+    },
+    "extensionProfiles": [
+      { 
+        "name": "prometheus-grafana-k8s", 
+        "version": "v1",
+        "rootURL": "https://raw.githubusercontent.com/ritazh/acs-engine/feat-monitor/"
```
jackfrancis: What's the purpose of the `rootURL` property? We probably want it to refer to something a little more persistent.

tstringer: That's where it'll pull the extension itself from. And absolutely that should be pointing to the upstream repo, but we were using @ritazh's fork to test this out.

This is a chicken and the egg thing. What is usually done to test out an extension in a fork before being merged upstream?

tstringer:
@jackfrancis absolutely, we can add a sample api model config to the repo. That'll be done shortly.

extensions/prometheus-grafana-k8s/README.md
jackfrancis: Understood. Because this is an optional property, can we remove it from the examples/ api model, or is there value in configuring your cluster to include this extension with the `rootURL` property properly defined?

ritazh: `rootURL` is used by `dta = strings.Replace(dta, "EXTENSION_URL_REPLACE", extensionProfile.RootURL, -1)` in `engine.go` and `EXTENSION_URL_REPLACE` is used in `template-link.json` as:
```
"templateLink": {
            "uri": "EXTENSION_URL_REPLACEextensions/prometheus-grafana-k8s/v1/template.json",
            "contentVersion": "1.0.0.0"
        },
```

tstringer:
@jchauncey this pr now includes support to specify a non-default namespace to put these components in. Thanks for pointing that out!

cosmincojocar:
@jackfrancis Please could you also review this PR which adds Prometheus instrumentation to Azure API calls https://github.com/kubernetes/kubernetes/pull/56286?

jchauncey:
@cosmincojocar do metrics like this - `mc := newMetricContext("public_ip_addresses", "list", az.ResourceGroup, az.SubscriptionID)` capture the time it takes for the public ip address to be provisioned?

cosmincojocar:
The metric will record the duration of the Azure API call (in this case list public IP addresses) and also if the call was successful or an error. The metrics are grouped per subscription and resource group.

PR #1838, opened 2017-11-28
normalizing the location in agent pool only clusters

JackQuincy:
We are doing this for ACS objects and need to for AKS so that we can't get issues if an incorrect value is set.


jackfrancis:
Thanks @amanohar

Issue #1826, opened 2017-11-24
Make out of resource handling flags configurable


aku105:
**Is this a request for help?**:  
No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
FEATURE

---

**What version of acs-engine?**:
Latest

---

Kubernetes has configurations available for handling Out Of Resource scenarios as described [here](https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/).
This request is to allow eviction threshold policies to be configured in apimodel.json itself.

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes

**What happened**:
N/A

**What you expected to happen**:
N/A

**How to reproduce it** (as minimally and precisely as possible):
N/A

**Anything else we need to know**:
N/A

PR #1843, opened 2017-11-28
add configurable hard eviction threshold option

pidah:
This PR adds an option for configuring the hard eviction threshold. The default value in kubernetes is 100MB which is too small. When the node is rapidly under memory pressure, kubelet may not be able to kick-in the pod eviction policy before the node itself gets OOM'd.  A more detailed discussion on the issue is here https://github.com/kubernetes/kops/pull/2982 and recommendations/guidleines here: https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#Known-issues

 This PR also proposes that the default value of hard eviction threshold on the nodes be set to 10% of system memory.

fixes Azure/acs-engine#1826

### DIFF
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -40,8 +40,8 @@ const (
 	DefaultNetworkPolicy = "none"
 	// DefaultKubernetesNodeStatusUpdateFrequency is 10s, see --node-status-update-frequency at https://kubernetes.io/docs/admin/kubelet/
 	DefaultKubernetesNodeStatusUpdateFrequency = "10s"
-	// DefaultKubernetesHardEvictionThreshold is "memory.available<10%, see --eviction-hard at https://kubernetes.io/docs/admin/kubelet/
-	DefaultKubernetesHardEvictionThreshold = "memory.available<10%"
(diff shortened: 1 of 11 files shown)


jackfrancis:
Thanks @pidah !

Run `go fmt -s pkg/api/types.go && go fmt -s pkg/api/vlabs/types.go` and then commit the changes to fix unit tests.

pkg/acsengine/const.go
```diff
@@ -40,6 +40,8 @@ const (
 	DefaultNetworkPolicy = "none"
 	// DefaultKubernetesNodeStatusUpdateFrequency is 10s, see --node-status-update-frequency at https://kubernetes.io/docs/admin/kubelet/
 	DefaultKubernetesNodeStatusUpdateFrequency = "10s"
+	// DefaultKubernetesHardEvictionThreshold is "memory.available<10%, see --eviction-hard at https://kubernetes.io/docs/admin/kubelet/
+	DefaultKubernetesHardEvictionThreshold = "memory.available<10%"
```
jackfrancis: According to the k8s docs, this default string should be:

`"memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%"`

pidah: @jackfrancis thanks! I have updated the default string. We were proposing memory.available<10% as a default here, but now I think its better raised in the upstream k8s repo

PR #1844, opened 2017-11-29
Add ACI Connector

sozercan:
Adds [ACI Connector](https://github.com/Azure/aci-connector-k8s) as a configurable add-on in ACS-engine


pkg/acsengine/defaults.go
```diff
@@ -130,8 +131,13 @@ var (
 		},
 		//KubernetesSpecConfig - Due to Chinese firewall issue, the default containers from google is blocked, use the Chinese local mirror instead
 		KubernetesSpecConfig: KubernetesSpecConfig{
-			KubernetesImageBase: "crproxy.trafficmanager.net:6000/google_containers/",
-			TillerImageBase:     "crproxy.trafficmanager.net:6000/kubernetes-helm/",
+			KubernetesImageBase:              "crproxy.trafficmanager.net:6000/google_containers/",
```
jackfrancis: Let's punt on modifying the China cloud spec, we usually defer to the China team to manage this. If you'd like, you could open up an issue to add ACI connector addon capability to China cloud and we can work with them on it.

pkg/api/vlabs/types.go
```diff
@@ -184,11 +184,22 @@ type KubernetesContainerSpec struct {
...
+
 // KubernetesAddon defines a list of addons w/ configuration to include with the cluster deployment
 type KubernetesAddon struct {
-	Name       string                    `json:"name,omitempty"`
-	Enabled    *bool                     `json:"enabled,omitempty"`
-	Containers []KubernetesContainerSpec `json:"containers,omitempty"`
+	Name        string                      `json:"name,omitempty"`
+	Enabled     *bool                       `json:"enabled,omitempty"`
+	Containers  []KubernetesContainerSpec   `json:"containers,omitempty"`
+	Environment []KubernetesEnvironmentSpec `json:"environment,omitempty"`
```
jackfrancis: It would be great to make this a `map[string]string` so it has general purpose value. `KubernetesAddon` is meant to be a generic interface for configuring an addon.

jackfrancis: How about `Config` map[string]string` `json:"config,omitempty"`

pkg/acsengine/engine.go
```diff
@@ -136,6 +137,7 @@ var kubernetesAddonYamls15 = map[string]string{
 	"MASTER_ADDON_KUBERNETES_DASHBOARD_DEPLOYMENT_B64_GZIP_STR": "kubernetesmasteraddons-kubernetes-dashboard-deployment1.5.yaml",
 	masterAddonAzureStorageClasses:                              masterAddonAzureStorageClassesFileUnmanaged,
 	"MASTER_ADDON_TILLER_DEPLOYMENT_B64_GZIP_STR":               "kubernetesmasteraddons-tiller-deployment1.5.yaml",
+	"MASTER_ADDON_ACI_CONNECTOR_DEPLOYMENT_B64_GZIP_STR":        "kubernetesmasteraddons-aci-connector-deployment1.5.yaml",
```
dtzar: Do we really need this separate file?  I don't think we are supporting K8s 1.5 and your yaml is the same for the 1.5 and the non-1.5 version, yes?

pkg/acsengine/defaults.go
```diff
@@ -169,6 +174,22 @@ var (
 		},
 	}
 
+	// DefaultACIConnectorAddonsConfig is the default ACI Connector Kubernetes addon Config
+	DefaultACIConnectorAddonsConfig = api.KubernetesAddon{
+		Name:    DefaultACIConnectorAddonName,
+		Enabled: pointerToBool(api.DefaultACIConnectorAddonEnabled),
+		Config:  map[string]string{},
```
jackfrancis: Add the default vals to `Config` here as well. E.g.,

```Config: map[string]string{
    "clientId": <defaultVal>,
    "clientKey": <defaultVal>,
   <etc...>
}```

File: pkg/api/convertertoapi.go [hunk 5]
```diff
@@ -635,6 +635,10 @@ func convertAddonsToAPI(v *vlabs.KubernetesConfig, a *KubernetesConfig) {
 				MemoryLimits:   v.Addons[i].Containers[j].MemoryLimits,
 			})
 		}
+
```
jackfrancis: @sozercan I added the below lines to adapt the KubernetesAddons default assignment convenience function to include `Config` key/vals as well

PR #1848, opened 2017-11-29
pin docker version swarm

CecileRobertMichon:
Add back swarm e2e

**What this PR does / why we need it**: Swarm clusters are broken because of new Docker version incompatibility. This pins the Docker version for Swarm clusters.

**Release note**:

```release-note
```

### DIFF
--- a/parts/configure-swarm-cluster.sh
+++ b/parts/configure-swarm-cluster.sh
@@ -20,6 +20,7 @@ POSTINSTALLSCRIPTURI=${7}
 BASESUBNET=${8}
 DOCKERENGINEDOWNLOADREPO=${9}
 DOCKERCOMPOSEDOWNLOADURL=${10}
+DOCKER_CE_VERSION=17.03.*
 VMNAME=`hostname`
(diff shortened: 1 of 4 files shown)


parts/configure-swarm-cluster.sh
```diff
@@ -167,7 +168,9 @@ retrycmd_if_failure() { for i in 1 2 3 4 5; do $@; [ $? -eq 0  ] && break || sle
 installDocker()
 {
   for i in {1..10}; do
-    wget --tries 4 --retry-connrefused --waitretry=15 -qO- https://get.docker.com | sh
+    add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
```
weinong: I think you need to install the gpg key.
```sh
apt-get install -y apt-transport-https ca-certificates curl 
curl --max-time 60 -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - 
```

CecileRobertMichon: Yes you're right thank you!

PR #1855, opened 2017-11-29
setting default etcd version to v2.3.8

jackfrancis:
**What this PR does / why we need it**: upgrade default etcd version to v2.3.8

**Special notes for your reviewer**: This is the latest v2 release of etcd:

https://github.com/coreos/etcd/releases/tag/v2.3.8

**Release note**:

```
set default etcd version to v2.3.8
```

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -349,11 +349,6 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 			}
 		}
 
-		// default etcd version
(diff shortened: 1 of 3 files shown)


pkg/acsengine/defaults.go
```diff
@@ -349,11 +349,6 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 			}
 		}
 
-		// default etcd version
```
jackfrancis: These lines are unreachable code.

PR #1858, opened 2017-11-30
fixing agent pool upgrade logic

dmitsh:
fixing agent pool upgrade logic
 - checking provisioning state of existing VMs
 - do not create extra VM if all nodes have been already upgraded
 - adding more logs


weinong:
can we add tests?

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -221,98 +233,126 @@ func (ku *Upgrader) upgradeAgentPools() error {
...
 				vmProvisioningState = *vm.VirtualMachineProperties.ProvisioningState
 			}
+			agentIndex, _ := armhelpers.GetVMNameIndex(vm.StorageProfile.OsDisk.OsType, *vm.Name)
 
-			if vmProvisioningState != string(api.Succeeded) {
-				ku.logger.Infof("Deleting agent VM %s in provisioning state %s\n", *vm.Name, vmProvisioningState)
+			switch vmProvisioningState {
+			case "Creating":
+				fallthrough
+			case "Updating":
```
JackQuincy: these can be written more compactly:
 case "Creating", "Updating", "Succeeded": 
Really a matter of choice though. @jackfrancis @amanohar any opinion on which style we want?

jackfrancis: +1 for the succinct style. See the "Multiple Cases" here:

https://github.com/golang/go/wiki/Switch

dmitsh: sure, I will change that.

dmitsh:
@weinong We would need to overhaul current ARM simulation framework. I will create a task and we will do that in a separate PR.

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -112,30 +113,27 @@ func (kan *UpgradeAgentNode) Validate(vmName *string) error {
...
-	go func() {
-		for {
+	retryTimer := time.NewTimer(time.Millisecond)
+	timeoutTimer := time.NewTimer(timeout)
+	for {
+		select {
+		case <-timeoutTimer.C:
+			retryTimer.Stop()
+			kan.logger.Errorf("Node was not ready within %v", timeout)
+			return fmt.Errorf("Node was not ready within %v", timeout)
```
JackQuincy: [nit] err := fmt.Errorf("Node was not ready within %v", timeout")
kan.logger.Error(err.Error())
return err

PR #1873, opened 2017-12-01
Update k8s GPU doc

wbuchwalter:
Fix another issue with K8s GPU documentation.  
Modifying `LD_LIBRARY_PATH` is not ideal as the image we are running might also have made some changes to this env variable that we would be overwriting.  
This leads to issues with CUDA in the latest tensorflow images for example.

By instead mounting the drivers directly in `/usr/local/nvidia/lib64` we don't need to update `LD_LIBRARY_PATH` anymore and can avoid some potential issues.


jackfrancis:
Thanks @wbuchwalter !

PR #1875, opened 2017-12-02
added scaling test for azure CNI

dmitsh:
added scaling test for azure CNI


examples/azure-cni/k8s-scaledown.json.env
```diff
@@ -0,0 +1,4 @@
+ACSE_POSTDEPLOY=examples/azure-cni/k8s-scale.sh
+
```
JackQuincy: Based on the error I saw. it makes me think we might need to sleep a minute or 2 before scaling down.

examples/azure-cni/k8s-scale.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ -e "${ENV_FILE}" ]; then
+  source "${ENV_FILE}"
+fi
+
+[[ ! -z "${NEW_AGENT_NODE_COUNT:-}" ]] || (echo "Must specify NEW_AGENT_NODE_COUNT" && exit 1)
+
+OUTPUT="_output/${INSTANCE_NAME}"
+
+# allow nodes to run for a while before scaling
+sleep 180
```
JackQuincy: after I merge this shouldn't be needed anymore which will help with these tests duration a little.

examples/azure-cni/k8s-scaledown.json.env
```diff
@@ -0,0 +1,4 @@
...
+NEW_AGENT_NODE_COUNT=2
+EXPECTED_NODE_COUNT=3
```
tamilmani1989: what's this EXPECTED_NODE_COUNT is for? Where we are using it?

PR #1879, opened 2017-12-04
Cordon retry

JackQuincy:
This PR makes the Cordoning of a node retry in cases where there was a concurrent modification between the get and PUT up to 5 times.
We have seen this in some tests.

**Release note**:

```release-note
Retry Cordon in cases of a concurrent modification.
```

### DIFF
--- a/pkg/operations/cordondrainvm.go
+++ b/pkg/operations/cordondrainvm.go
@@ -9,7 +9,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apiserver/pkg/registry/generic/registry"
 	"k8s.io/client-go/pkg/api/v1"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/operations/cordondrainvm.go
```diff
@@ -34,14 +39,24 @@ func SafelyDrainNode(az armhelpers.ACSEngineClient, logger *log.Entry, masterURL
...
+			return err
+		}
+		node.Spec.Unschedulable = true
+		node, err = client.UpdateNode(node)
+		if err != nil {
+			// If this error is because of a concurrent modification get the update
+			// and then apply the change
+			if strings.Contains(err.Error(), kubernetesOptimisticLockErrorMsg) {
+				logger.Infof("Node %s got an error suggesting a concurrent modification. Will retry to cordon", nodeName)
+				continue
```
dmitsh: Just curious whether we need to wait before retrying?

pkg/operations/cordondrainvm.go
```diff
@@ -14,6 +15,10 @@ import (
 const (
 	interval            = time.Second * 1
 	mirrorPodAnnotation = "kubernetes.io/config.mirror"
+
+	// This is checked into K8s code but I was getting into vendoring issues so I copied it here instead
+	kubernetesOptimisticLockErrorMsg = "the object has been modified; please apply your changes to the latest version and try again"
```
jackfrancis: This seems like a brittle pattern match. (especially the `;`) Is just `"the object has been modified"` unique enough?

(Not to mention that this will need to be massaged over time as error messages change.)

JackQuincy: This is a constant in k8s code that we aren't currently vendoring when I did I ran into dependency conflicts. there wasn't a code I could of queued off of other wise I would of

pkg/operations/cordondrainvm.go
JackQuincy: I decided not to since this is a concurrent modification and I figured the other guy was done. We could wait a little in case they are making q bunch of small changes instead of a big one but I'm inclined to think it will usually be a batch update.

jackfrancis:
Happy to watch this and add a delay if needed. Thanks @JackQuincy !

PR #1884, opened 2017-12-05
 add hard eviction threshold docs

pidah:
This PR adds docs for the --eviction-threshold flag


jackfrancis:
Thanks @pidah !

PR #1893, opened 2017-12-06
Kubernetes v1.9.0 support

jackfrancis:
**What this PR does / why we need it**: brings Kubernetes v1.9.0 support

**Release note**:

```
Kubernetes v1.9.0 support
```

### DIFF
--- a/pkg/acsengine/k8s_versions.go
+++ b/pkg/acsengine/k8s_versions.go
@@ -12,7 +12,7 @@ var KubeConfigs = map[string]map[string]string{
 		"hyperkube": "hyperkube-amd64:v1.9.0",
 		"ccm":       "cloud-controller-manager-amd64:v1.9.0",
 		"dockerEngineVersion":       "1.12.*",
-		DefaultDashboardAddonName:   "kubernetes-dashboard-amd64:v1.9.0",
+		DefaultDashboardAddonName:   "kubernetes-dashboard-amd64:v1.9.0", // TODO fix missing gcrio.azureedge.net/google_containers/kubernetes-dashboard-amd64:v1.9.0
 		"exechealthz":               "exechealthz-amd64:1.2",
 		"addonresizer":              "addon-resizer:1.7",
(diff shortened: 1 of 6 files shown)


pkg/acsengine/k8s_versions.go
```diff
@@ -8,6 +8,33 @@ import (
 
 // KubeConfigs represents Docker images used for Kubernetes components based on Kubernetes versions (major.minor.patch)
 var KubeConfigs = map[string]map[string]string{
+	common.KubernetesVersion1Dot9Beta1: {
+		"hyperkube": "hyperkube-amd64:v1.9.0-beta.1",
+		"ccm":       "cloud-controller-manager-amd64:v1.9.0",
```
karataliu: Here should be 'cloud-controller-manager-amd64:v1.9.0-beta.1'.

Another thing to note is azure's 'ccm' image release is planned to be shipped via public dockerhub, which does not share 'gcrio.azureedge.net' prefix then. This could be addressed later when we move to that image.

pkg/acsengine/k8s_versions.go
```diff
@@ -8,6 +8,33 @@ import (
...
+		"exechealthz":               "exechealthz-amd64:1.2",
+		"addonresizer":              "addon-resizer:1.7",
+		"heapster":                  "heapster-amd64:v1.4.2",
+		"dns":                       "k8s-dns-kube-dns-amd64:1.14.5",
+		"addonmanager":              "kube-addon-manager-amd64:v6.4-beta.2",
+		"dnsmasq":                   "k8s-dns-dnsmasq-nanny-amd64:1.14.5",
+		"pause":                     "pause-amd64:3.0",
+		DefaultTillerAddonName:      DefaultTillerImage,
+		DefaultReschedulerAddonName: DefaultReschedulerImage,
+		"windowszip":                "v1.8.4-1int.zip", // TODO add Windows image
```
JiangtianLi: You can change to `v1.9.0-1int.zip`

PR #1897, opened 2017-12-07
setting coreos api models to use kubenet

jackfrancis:
**What this PR does / why we need it**: CoreOS Kubernetes clusters don't currently work w/ Azure CNI networking. Setting api model examples to explicitly use kubenet, now that Azure CNI is default networking implementation.

**Release note**:

```
NONE
```

### DIFF
--- a/examples/coreos/kubernetes-coreos-hybrid.json
+++ b/examples/coreos/kubernetes-coreos-hybrid.json
@@ -2,7 +2,10 @@
   "apiVersion": "vlabs",
   "properties": {
     "orchestratorProfile": {
-      "orchestratorType": "Kubernetes"
+      "orchestratorType": "Kubernetes",
+      "kubernetesConfig": {
+        "networkPolicy": "none"
+      }
     },
     "masterProfile": {
       "count": 1,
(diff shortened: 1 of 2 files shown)


jackfrancis:
@richardjortega FYI, with the next release of acs-engine (anything after `v0.10.0`) you'll need to update your CoreOS api models to reflect the changes in this PR. We turned on Azure CNI by default for Kubernetes clusters, which doesn't currently work with CoreOS.

Feel free to iterate over this if you're interested in having your CoreOS clusters be backed by Azure CNI, or continue to use kubenet with the modifications to the api model. To clarify: this is not a functional change in behavior, simply a change to default configuration. Formerly, `"networkPolicy": "none"` was the default, which meant that CoreOS clusters deployed successfully without having to explicitly set this property. The new default is now `"networkPolicy": "azure"` if nothing is user-defined; so you'll have to explicitly define it as `"networkPolicy": "none"` going forward. (Until we make Azure CNI work w/ CoreOS.)

richardjortega:
Will have to look into this and understand it.

PR #1864, opened 2017-12-01
WIP Etcd TLS encryption

CecileRobertMichon:
**What this PR does / why we need it**: This PR adds TLS encryption to etcd for client/server communication (peer to peer to follow).

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #681 

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #1929, opened 2017-12-14
Etcd TLS (CSE edition)

### LINKED ISSUE #681: etcd communication encrypted with TLS
All communication with etcd should be encrypted with self-signed TLS certificates.

CecileRobertMichon:
**What this PR does / why we need it**: This PR adds TLS encryption to client-server and peer-to-peer etcd communications. This ensures that no external pod can install etcd and access the values stored in the etcd server.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #681 

**Special notes for your reviewer**: This PR is a second implementation attempt, the first being #1864. The first implementation did not take upgrade into account: a new node would be locked out of the cluster since it wouldn't have the right certs. This PR generates the certs in the 'generate' step and passes them to the masters using CSE.

**Release note**:

```release-note
```


parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -484,19 +485,20 @@ coreos:
...
 - /opt/azure/containers/mountetcd.sh
 - sudo /bin/chown -R etcd:etcd /var/lib/etcddisk
 - systemctl stop etcd
 - sudo -u etcd rm -rf /var/lib/etcd/default
-- sudo /bin/sed -i s/Restart=on-abnormal/Restart=always/g /lib/systemd/system/etcd.service
```
CecileRobertMichon: This is unrelated to TLS but removes an unnecessary line since /lib/systemd/system/etcd.service no longer exists

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -31,6 +32,10 @@ COREOS_OS_NAME="COREOS"
 # Set default kubectl
 KUBECTL=/usr/local/bin/kubectl
 
+MASTER_OFFSET=$(echo `hostname` | cut -d'-' -f 4)
```
CecileRobertMichon: If anyone knows better way to get the master number from the custom script please let me know... I tried passing in copyIndex(masterOffset) in provisionScriptParametersMaster but I can't use copyIndex() in that context

pkg/acsengine/pki.go
```diff
@@ -87,22 +92,74 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
+		return nil, nil, nil, nil, nil, nil, e4
+	}
+	if e5 != nil {
+		return nil, nil, nil, nil, nil, nil, e5
 	}
+	if e6 != nil {
+		return nil, nil, nil, nil, nil, nil, e5
+	}
+
+	time.Sleep(10 * time.Second)
```
CecileRobertMichon: I had to add this line here otherwise the go routines didn't have time to finish executing before return and the certs wouldn't get completely generated. Looking for suggestions to improve. cc @anhowe since you're the one who added the go func()'s in this file.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -484,19 +485,20 @@ coreos:
...
 - systemctl daemon-reload
+- retrycmd_if_failure cat /opt/azure/containers/etcdcerts.complete
+- sudo /bin/chown etcd:etcd /etc/kubernetes/certs/etcd*.key
 - systemctl restart etcd
-- for i in $(seq 1 20); do curl --max-time 60 http://127.0.0.1:2379/v2/machines; [ $? -eq 0 ] && break || sleep 5; done
-- retrycmd_if_failure() { for i in 1 2 3 4 5; do $@; [ $? -eq 0  ] && break || sleep 5; done ; }
+- curl --cacert /etc/kubernetes/certs/ca.crt --cert /etc/kubernetes/certs/etcdclient.crt --key /etc/kubernetes/certs/etcdclient.key --max-time 60 "{{WrapAsVerbatim "variables('masterEtcdClientURLs')[copyIndex(variables('masterOffset'))]"}}"/v2/machines >> /opt/etcdtls
```
weinong: seems we already have a loop curling below this line.

parts/k8s/kubernetesmastercustomscript.sh
anhowe: take a look at configure-swarm-cluster.sh where ```MASTERCOUNT=${3}```.  This can be set on the incoming env variables.

parts/k8s/kubernetesmastercustomdata.yml
CecileRobertMichon: You're right I forgot to clean this up after doing some testing, good catch!

parts/k8s/kubernetesmastervars.t
```diff
@@ -152,7 +177,7 @@
...
-    "provisionScriptParametersMaster": "[concat('APISERVER_PRIVATE_KEY=',variables('apiServerPrivateKey'),' CA_CERTIFICATE=',variables('caCertificate'),' CA_PRIVATE_KEY=',variables('caPrivateKey'),' MASTER_FQDN=',variables('masterFqdnPrefix'),' KUBECONFIG_CERTIFICATE=',variables('kubeConfigCertificate'),' KUBECONFIG_KEY=',variables('kubeConfigPrivateKey'),' ADMINUSER=',variables('username'))]",
+    "provisionScriptParametersMaster": "[concat('APISERVER_PRIVATE_KEY=',variables('apiServerPrivateKey'),' CA_CERTIFICATE=',variables('caCertificate'),' CA_PRIVATE_KEY=',variables('caPrivateKey'),' MASTER_FQDN=',variables('masterFqdnPrefix'),' KUBECONFIG_CERTIFICATE=',variables('kubeConfigCertificate'),' KUBECONFIG_KEY=',variables('kubeConfigPrivateKey'),' ETCD_SERVER_CERTIFICATE=',variables('etcdServerCertificate'),' ETCD_CLIENT_CERTIFICATE=',variables('etcdClientCertificate'),' ETCD_SERVER_PRIVATE_KEY=',variables('etcdServerPrivateKey'),' ETCD_CLIENT_PRIVATE_KEY=',variables('etcdClientPrivateKey'),' ETCD_PEER_CERTIFICATES=',string(variables('etcdPeerCertificates')),' ETCD_PEER_PRIVATE_KEYS=',string(variables('etcdPeerPrivateKeys')),' ADMINUSER=',variables('username'))]",
```
anhowe: you can set MASTER_COUNT here.

CecileRobertMichon: MASTERCOUNT is the number of master nodes if I'm not mistaken, I need the index of the current master. I thought of using masterOffset but I can't use the copyIndex() function from here

pkg/acsengine/pki.go
```diff
@@ -87,22 +92,74 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
 		errors <- err
 	}()
 
+	go func() {
+		var err error
+		organization := make([]string, 1)
+		organization[0] = "system:masters"
+		ip := net.ParseIP("127.0.0.1").To4()
+		peerIPs := append(extraIPs, ip)
+		etcdServerCertificate, etcdServerPrivateKey, err = createCertificate("etcd", caCertificate, caPrivateKey, false, nil, peerIPs, organization)
```
JunSun17: Use different common names for etcd server/client/peer.

Should the isServer field be true?

Not sure whether extraIPs is needed here.

pkg/acsengine/pki.go
```diff
@@ -87,22 +92,74 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
+		go func() {
+			var err error
+			organization := make([]string, 1)
+			organization[0] = "system:masters"
+			ip := net.ParseIP("127.0.0.1").To4()
+			peerIPs := append(extraIPs, ip)
+			etcdPeerCertificate := new(x509.Certificate)
+			etcdPeerPrivateKey := new(rsa.PrivateKey)
+			etcdPeerCertificate, etcdPeerPrivateKey, err = createCertificate("etcd", caCertificate, caPrivateKey, false, nil, peerIPs, organization)
+			etcdPeerCertPairs = append(etcdPeerCertPairs, &PkiKeyCertPair{CertificatePem: string(certificateToPem(etcdPeerCertificate.Raw)), PrivateKeyPem: string(privateKeyToPem(etcdPeerPrivateKey))})
```
JunSun17: Will this cause race condition since func() are run in parallel?

pkg/acsengine/pki.go
```diff
@@ -87,22 +92,74 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
+		return nil, nil, nil, nil, nil, nil, e3
+	}
+	if e4 != nil {
+		return nil, nil, nil, nil, nil, nil, e4
+	}
+	if e5 != nil {
+		return nil, nil, nil, nil, nil, nil, e5
 	}
+	if e6 != nil {
+		return nil, nil, nil, nil, nil, nil, e5
```
JunSun17: Change e5 to e6.

pkg/acsengine/pki.go
```diff
@@ -133,6 +190,10 @@ func createCertificate(commonName string, caCertificate *x509.Certificate, caPri
 		template.DNSNames = extraFQDNs
 		template.IPAddresses = extraIPs
 		template.ExtKeyUsage = append(template.ExtKeyUsage, x509.ExtKeyUsageServerAuth)
+	} else if commonName == "etcd" {
+		template.IPAddresses = extraIPs
+		template.ExtKeyUsage = append(template.ExtKeyUsage, x509.ExtKeyUsageClientAuth)
```
JunSun17: Should server cert support x509.ExtKeyUsageServerAuth only and client cert support x509.ExtKeyUsageClientAuth only? Peer cert maybe support both.

pkg/acsengine/engine.go
```diff
@@ -811,6 +818,27 @@ func addSecret(m paramsMap, k string, v interface{}, encode bool) {
...
+		parts := keyvaultSecretPathRe.FindStringSubmatch(str)
+		if parts == nil || len(parts) != 5 {
+			if encode {
+				values[i] = base64.StdEncoding.EncodeToString([]byte(str))
+			} else {
+				values[i] = str
+			}
+		}
+	}
+	addValue(m, k, values)
```
JunSun17: Do you need to call addKeyvaultReference function as in the above addSecret function?

parts/k8s/kubernetesmasterresources.t
```diff
@@ -530,7 +530,7 @@
         "autoUpgradeMinorVersion": true,
         "settings": {},
         "protectedSettings": {
-          "commandToExecute": "[concat(variables('provisionScriptParametersCommon'),' ',variables('provisionScriptParametersMaster'),' /usr/bin/nohup /bin/bash -c \"stat /opt/azure/containers/provision.complete || /bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]"
+          "commandToExecute": "[concat(variables('provisionScriptParametersCommon'),' ',variables('provisionScriptParametersMaster'), ' MASTER_INDEX=',copyIndex(variables('masterOffset')),' /usr/bin/nohup /bin/bash -c \"stat /opt/azure/containers/provision.complete || /bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]"
```
JunSun17: Should MASTER_INDEX be passed in through provisionScriptParametersMaster?

pkg/acsengine/pki.go
CecileRobertMichon: Good catch, thanks!

parts/k8s/kubernetesparams.t
```diff
@@ -25,6 +25,42 @@
...
+      "metadata": {
+        "description": "The base 64 server certificates used on the master"
+      }, 
+      "type": "array"
+    }, 
+    "etcdPeerPrivateKeys": {
+      "metadata": {
+        "description": "The base 64 server private keys used on the master."
+      }, 
+      "type": "array"
```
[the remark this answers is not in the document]
CecileRobertMichon: No there isn't unfortunately (https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-deployment#parameters) but each string inside the array gets encoded.

parts/k8s/kubernetesmasterresources.t
CecileRobertMichon: That's not possible because copyIndex() needs the copy object which is part of the resource. Putting in in provisionScriptParametersMaster was my first intention too. This is the solution @anhowe and I came up with instead.

pkg/acsengine/pki.go
CecileRobertMichon: ExtraIPs is needed because the certs need the peer IPs as SANs in order for tls to work. 

I think you're right that I should differentiate server/client. This was "hack" to get it to work by giving all the etcd certs clientAuth, servAuth and peer IPs. If you look at the if loop isServer needs to be false so that it goes to `else if CN == "etcd"` but this can probably be done more elegantly.

pkg/acsengine/pki.go
CecileRobertMichon: Yes and no. It does mean that the order of the certs is non deterministic. However, I don't think that that's an issue since they all have the same properties and in the end we want 1/3/5 certs signed by the same CA. At this stage it doesn't matter which cert gets assigned to which peer. Please let me know if I am missing something or something is wrong with that logic.

pkg/acsengine/pki.go
CecileRobertMichon: For simplicity I just added both for now since one way or the other didn't seem to work but that's on my todo list of small improvements :)

pkg/acsengine/engine.go
CecileRobertMichon: I will look into this, not sure

pkg/acsengine/pki.go
JunSun17: In the append function, if two threads try to modify etcdpeerCertPairs at the same time, would one overwrite another?

CecileRobertMichon: I think I fixed it, thank you for your comment 🙂

parts/k8s/kubernetesmasterresources.t
anhowe: thanks @CecileRobertMichon , the solution above looks good

pkg/acsengine/pki.go
```diff
@@ -52,61 +52,119 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
 	go func() {
 		var err error
 		organization := make([]string, 1)
 		organization[0] = "system:masters"
-		kubeConfigCertificate, kubeConfigPrivateKey, err = createCertificate("client", caCertificate, caPrivateKey, false, nil, nil, organization)
+		kubeConfigCertificate, kubeConfigPrivateKey, err = createCertificate("clientclient", caCertificate, caPrivateKey, false, false, nil, nil, organization)
 		errors <- err
 	}()
 
+	go func() {
```
anhowe: How much extra time does this add to the generation?

pkg/acsengine/pki.go
```diff
@@ -52,61 +52,119 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
+		errors <- err
+	}()
+
+	etcdPeerCertPairs = make([]*PkiKeyCertPair, masterCount)
+	for i := 0; i < masterCount; i++ {
+		go func(i int) {
+			var err error
+			organization := make([]string, 1)
+			organization[0] = "system:masters"
+			ip := net.ParseIP("127.0.0.1").To4()
```
JunSun17: It is a bit strange that the IPs for client, server and peers are all 127.0.0.1. In a deployment environment, do all master nodes use the same 127.0.0.1? It may worth some investigation later for this.

It might also worth (?) to wrap these cert generation logic in another helper function since they are duplicated for 3 times here.

pkg/acsengine/pki.go
```diff
@@ -52,61 +52,119 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
...
+		return nil, nil, nil, nil, nil, nil, e4
+	}
+	if e5 != nil {
+		return nil, nil, nil, nil, nil, nil, e5
 	}
+	if e6 != nil {
+		return nil, nil, nil, nil, nil, nil, e6
+	}
+
+	time.Sleep(10 * time.Second)
```
JunSun17: I searched and looks golang can use sync.WaitGroup. Like this: https://stackoverflow.com/questions/46906647/wait-result-of-multiple-goroutines.

Sorry I am still new to golang and I may be wrong here. Just FYI.

CecileRobertMichon:
@anhowe I observed the following performance:

- around 5 seconds more in the generate step to generate the additional certs
- no significant difference in deployment time

pkg/acsengine/pki.go
CecileRobertMichon: I actually had to add 2 error variables since each go routine is a separate error. Checking all the errors at the end waits for all the routines to finish. Thanks for the suggestion!

pkg/acsengine/pki.go
CecileRobertMichon: I'm confused, what do you mean by "It is a bit strange that the IPs for client, server and peers are all 127.0.0.1"?

127.0.0.1 is the loopback address so every machine has the same one. Here I'm adding it to the list of peer IP addresses since etcd tls does client-server and peer-to-peer authentication.

JunSun17: Sorry for the confusion, I though this IP is used for the identity in the cert, it looks it is used in SAN, so it should be good.

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -89,6 +93,45 @@ else
...
 
+ETCD_SERVER_PRIVATE_KEY_PATH="/etc/kubernetes/certs/etcdserver.key"
+touch "${ETCD_SERVER_PRIVATE_KEY_PATH}"
+chmod 0600 "${ETCD_SERVER_PRIVATE_KEY_PATH}"
+chown root:root "${ETCD_SERVER_PRIVATE_KEY_PATH}"
+echo "${ETCD_SERVER_PRIVATE_KEY}" | base64 --decode > "${ETCD_SERVER_PRIVATE_KEY_PATH}"
+
+ETCD_CLIENT_PRIVATE_KEY_PATH="/etc/kubernetes/certs/etcdclient.key"
+touch "${ETCD_CLIENT_PRIVATE_KEY_PATH}"
+chmod 0644 "${ETCD_CLIENT_PRIVATE_KEY_PATH}"
```
CecileRobertMichon: Is this safe? Either we leave it as 0600 and then etcdctl doesn't work since the key needs permission (so sudo etcdctl is required instead) or use 0644 to be able to read the key for etcdctl commands. @jackfrancis

jackfrancis: What user context does etcdctl run under? If we assign file ownership of the private key to that user, then `0600` should work.

CecileRobertMichon: True, right now I'm assigning it to etcd in custom data but I think it should work if I leave it as root. I'll test + push the change.

PR #1939, opened 2017-12-15
Update some flags in the apiserver.

brendandburns:
Disables serving on `--insecure-port`
Disables golang profiling.

@jackfrancis


parts/k8s/kubernetesmaster-kube-apiserver.yaml
```diff
@@ -17,7 +17,8 @@ spec:
         - "--admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota"
         - "--address=0.0.0.0"
         - "--allow-privileged"
-        - "--insecure-port=8080"
+        - "--insecure-port=0"
```
slack: This interacted strangely with `/heatlhz` in k8s 1.8 on AKS. A random listen port was assigned and would later fail health /cc @khenidak 

We instead kept the port as `8080` and added `--insecure-bind-address=127.0.0.1`

jackfrancis:
@JunSun17 is working on CIS-related remediations

PR #1942, opened 2017-12-15
Updates to kubelet flags.

brendandburns:
* disable read-only port
   * enforce kernel settings
   * don't keep terminated pod volumes


jackfrancis:
@brendanburns `--protect-kernel-defaults` option enforcement yields this message on a 1.7-built cluster:

```
kubelet.go:1323] Failed to start ContainerManager [Invalid kernel flag: vm/overcommit_memory, expected value: 1, actual value: 0, Invalid kernel flag: kernel/panic, expected value: 10, actual value: 0, Invalid kernel flag: kernel/panic_on_oops, expected value: 1, actual value: 0]
```

I'm going to triage that option out of this PR and create a backlog issue to look into kernel defaults as a separate work item.

PR #1946, opened 2017-12-19
Cleanup addons logic.

JunSun17:
**What this PR does / why we need it**: Centralize the addon related code, simplify the code change needed on dev side to add a new addon.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # Part of #1869 

**Special notes for your reviewer**:

Centralize the addon related code in addons.go. To add a new addon:
1) an addon config yaml file needs to be created and being put under k8s/addon directory.
2) add the new addon to the list kubernetesAddonSettings in addons.go. The list entry is like this:

```
kubernetesAddonSetting{
    "kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml", // addon yaml file created
    profile.OrchestratorProfile.KubernetesConfig.IsDashboardEnabled(), // toggle logic to enable addon
    "kubernetes-dashboard-deployment.yaml", // the file name which will be copied to master nodes
},
```

Then code should pickup the addon and write proper statements into the kubernetesmastercustomerdata.yaml.

Also moved all 1.5 version addon yamls to k8s/addon/1.5 directory. The code will pick the proper versioned addon files based on the input OrchestratorVersion. No hard-coding of version number in the code.

manifests and artifacts can be treated the same way if needed.

**Release note**:

```release-note
```

PR #1947, opened 2017-12-19
Support https for kubernetes-dashboard

karataliu:
For kubernetes v1.9, default dashboard was turn on with https: https://github.com/kubernetes/kubernetes/pull/53046

Thus default dashboard location will turn to:
http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/

This will fail with `tls: oversized record received with length 20527` if dashboard does not use https

**Which issue this PR fixes**
Related:
- https://github.com/kubernetes/dashboard/wiki/FAQ#im-accessing-dashboard-over-https
- https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md
```
Updated Dashboard add-on to version 1.8.0: The Dashboard add-on now deploys with https enabled. The Dashboard can be accessed via kubectl proxy at http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/. The /ui redirect is deprecated and will be removed in 1.10. 
```
The old `kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml` was copied to `kubernetesmasteraddons-kubernetes-dashboard-http-deployment.yaml`.

https related changes were applied to `kubernetesmasteraddons-kubernetes-dashboard-http-deployment.yaml` directly.

**Release note**:
Support https kubernetes dashboard

### DIFF
--- a/parts/k8s/addons/1.9/kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml
+++ b/parts/k8s/addons/1.9/kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml
@@ -1,0 +1,95 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    k8s-app: kubernetes-dashboard
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
+  name: kubernetes-dashboard
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: kubernetes-dashboard
+  labels:
+    k8s-app: kubernetes-dashboard
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: kubernetes-dashboard
+  namespace: kube-system
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    kubernetes.io/cluster-service: "true"
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard
+  namespace: kube-system
+spec:
+  ports:
+  - port: 443
+    targetPort: 8443
+  selector:
+    k8s-app: kubernetes-dashboard
+  type: NodePort
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      k8s-app: kubernetes-dashboard
+  template:
+    metadata:
+      labels:
+        k8s-app: kubernetes-dashboard
+    spec:
+      containers:
+      - args: 
+        - --auto-generate-certificates
+        - --heapster-host=http://heapster.kube-system:80
+        image: <kubernetesDashboardSpec>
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml
```diff
@@ -62,18 +62,20 @@ spec:
     spec:
       containers:
       - args: 
+        - --auto-generate-certificates
```
karataliu: note: this option only available in 1.8.0+ Dashboard image

pkg/acsengine/k8s_versions.go
```diff
@@ -521,3 +521,26 @@ var KubeConfigs = map[string]map[string]string{
...
+	common.KubernetesVersion1Dot7Dot1:  true,
+	common.KubernetesVersion1Dot7Dot0:  true,
+	common.KubernetesVersion1Dot6Dot13: true,
+	common.KubernetesVersion1Dot6Dot12: true,
+	common.KubernetesVersion1Dot6Dot11: true,
+	common.KubernetesVersion1Dot6Dot9:  true,
+	common.KubernetesVersion1Dot6Dot6:  true,
+}
+
+func kubernetesDashboardUsesHTTP(version string) bool {
```
brendandburns: Shouldn't this be based on the dashboard spec instead of the Kubernetes version?  The kubernetes versions will support the new dashboard, right?

brendandburns:
One quick comment.  I'd like to see us deploy the new dashboard even for older 1.7/1.8 clusters assuming that works...

pkg/acsengine/k8s_versions.go
karataliu: (Changes here will be updated since https://github.com/Azure/acs-engine/pull/1946 introduces new approach for locating addons manifest.)

Yes, actually it should be decided by dashboard image version. 
One thing to note is in https://github.com/kubernetes/kubernetes/pull/53046, it changes default redirect of `kubectl proxy` '/ui' to use https. And this change now applies for v1.9.0+ .

That means if we're using v1.8 k8s with dashboard configured to HTTPS. Navigating to '/ui' would still redirect to the HTTP dashboard, which is not available. This break existing user experience who is using '/ui' for redirecting.

Thus I'd propose enabling it for v1.9 only. See also below versioning test comment.

karataliu:
A detailed versioning matrix could be found here: https://github.com/kubernetes/dashboard/wiki/Compatibility-matrix

It mentions that 1.8 dashboard will only fully support 1.8+ cluster.

I've tested with acs-engine generate cluster of following versions to use the 1.8 dashboard with HTTPS:

v1.6.13
v1.7.10
- `http://localhost:8001/ui` will redirect to http location:`http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/`
- Manually navigating to `http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/` will load the ui, but the page reports an error:
`the server could not find the requested resource`

v1.8.4
- `http://localhost:8001/ui` will redirect to http location:`http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/`
- Manually navigating to `http://localhost:8001/api/v1/namespaces/kube-system/services/kubernetes-dashboard/proxy/` works

v1.9.0
v1.10.0-alpha.1
- Works well

jackfrancis:
@karataliu could you kindly rebase from master? We just merged in v1.9.0 support, so getting that change in your branch should make it easier to test the effects of this change in 1.9

karataliu:
@jackfrancis sure. done rebasing.

jackfrancis:
I'm going to merge this as-is so we can get some bake time. @karataliu do you want submit a follow-up PR w/ this change for older k8s versions? @brendanburns FYI

karataliu:
@jackfrancis 

I would propose enabling https dashboard for 1.9+ only, see comment https://github.com/Azure/acs-engine/pull/1947#issuecomment-353542307
elder versions either does not work, or has an incorrect '/ui' redirection.

What do you think?

PR #1952, opened 2017-12-19
Add linux side custom bootstrap url to vlabs api

yakman2020:
Allows use of custom bootstrap URLs for both windows and linux. This is needed in order to deploy with acs-engine for Continuous intergration testing for windows dcos development. We need to be able to install a current built image rather than a boiler plate standard image.

Currently we can do that in windows, but not linux.


jackfrancis:
@yakman2020 thanks! this looks good, let's get rid of the `pkg/acsengine/.gdb_history` addition (I assume it snuck in there)

yakman2020:
@jackfrancis Yup. Sorry. Removed

PR #1954, opened 2017-12-20
master and agent kubelet config

### LINKED ISSUE #1917: Option to disable cloud provider on kubelet

---

---

Summary: When creating large clusters that do not use PV's, it can be beneficial to reduce ARM QoS load from the Agent kubelets.  The option here would disable the cloud provider on the agent kubelets.

CecileRobertMichon:
**What this PR does / why we need it**: Adds ability to override kubelet config with agent and master specific kubelet configs. This is needed to be able to disable cloud provider on agent kubelets.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1917

**Release note**:

```release-note
```

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -873,10 +873,10 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			}
 			return buf.String()
 		},
-		"GetKubeletConfigKeyVals": func(isMaster bool, kc *api.KubernetesConfig) string {
+		"GetKubeletConfigKeyVals": func(profile *api.AgentPoolProfile) string {
 			kubeletConfig := cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig
-			if !isMaster {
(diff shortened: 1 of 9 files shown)


pkg/acsengine/engine.go
```diff
@@ -873,8 +873,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			}
 			return buf.String()
 		},
-		"GetKubeletConfigKeyVals": func() string {
+		"GetKubeletConfigKeyVals": func(profile *api.AgentPoolProfile) string {
```
jackfrancis: I think we should change the function arg here to a `*api.KubernetesConfig`, and just make this a simple string convenience function (look for a `KubeletConfig` of type `map[string]string`, and then concatenate all the key/vals together).

We then move all the "what type of KubernetesConfig are you?" logic to the caller, and not inside the function.

What am I missing that would suggest we should do it this way instead?

PR #1958, opened 2017-12-20
Only generate certs not provided by the user

CecileRobertMichon:
**What this PR does / why we need it**: Checks if the user has provided certs and only generate certs if some certs or all certs are missing. Only use the generated cert/key pair for certs not provided by the user.

**Release note**:

``` 
MERGE THIS WITH PR #1929 (they are complimentary)
```


jackfrancis:
This PR depends upon #1929

PR #1966, opened 2017-12-23
Improving IP address assignment for master nodes with Azure CNI.

### LINKED ISSUE #1916: Error created Azure master nodes using azure CNI
Yes
---

Issue
---

0.10.0
---

Kubernetes 1.8

**What happened**:
When trying to deploy for a small Azure based docker config something strange is occurring with master nodes network interfaces.
The errors I am encountering are (subscription data stripped)

:

> IP configuration /subscriptions/xxx/resourceGroups/rg-test/providers/Microsoft.Network/networkInterfaces/k8s-master-15426332-nic-2/ipConfigurations/ipconfig1 is using the private IP address 10.50.1.6 which is already allocated to resource /subscriptions/xxx/resourceGroups/rg-test/providers/Microsoft.Network/networkInterfaces/k8s-master-15426332-nic-1.
and
> IP configuration /subscriptions/xxx/resourceGroups/rg-test/providers/Microsoft.Network/networkInterfaces/k8s-master-15426332-nic-0/ipConfigurations/ipconfig1 is using the private IP address 10.50.1.4 which is already allocated to resource /subscriptions/xxx/resourceGroups/rg-test/providers/Microsoft.Network/networkInterfaces/k8s-master-15426332-nic-1.

I have no idea why multiple interfaces are attempting to take the same IP. I would think they should take unique IP's

I'm not sure, I have attached part of my model for reference.
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
	  "orchestratorRelease": "1.8",
	  "kubernetesConfig": {
        "networkPolicy": "azure"
		}
    },
    "masterProfile": {
      "count": 3,
      "dnsPrefix": "syn-prod",
      "vmSize": "Standard_D2_v2",
      "vnetSubnetId": "/subscriptions/xxx/resourceGroups/Labs-Prod-k8s/providers/Microsoft.Network/virtualNetworks/Labs-Prod/subnets/k8s-Master-Net",
      "firstConsecutiveStaticIP": "10.50.1.4",
	  "ipAddressCount": 5,
	  "distro": "coreos"
    },
    "agentPoolProfiles": [
      {
        "name": "publicpool",
        "count": 3,
        "vmSize": "Standard_D2_v2",
        "OSDiskSizeGB": 200,
        "storageProfile" : "ManagedDisks",
        "vnetSubnetId": "/subscriptions/xxx/resourceGroups/Labs-Prod-k8s/providers/Microsoft.Network/virtualNetworks/Labs-Prod/subnets/PublicAgentPool",
		"ipAddressCount": 7,
		"distro": "coreos"
      },
      {
        "name": "privatepool",
        "count": 10,
        "vmSize": "Standard_D2_v2",
        "OSDiskSizeGB": 200,
        "storageProfile" : "ManagedDisks",
        "vnetSubnetId": "/subscriptions/xxx/resourceGroups/Labs-Prod-k8s/providers/Microsoft.Network/virtualNetworks/Labs-Prod/subnets/PrivateAgentPool",
		"ipAddressCount": 10,
		"distro": "coreos"
      }
    ],
}

tamilmani1989:

This PR is applicable in the scenario where customer tries to deploy kubernetes cluster in an existing Vnet/subnet and uses Azure-CNI (the default choice). Existing implementation requires customers to pick the firstConsecutiveStaticIP from the IP addresses towards the end of the subnet (so that dynamic ipconfig's don't select ips that are statically chosen). This PR removes this restriction by doing the following:
1. It makes all IP configs assigned to master nodes as static.
2. It  makes NICs of agent nodes dependent on NICs of master nodes. This avoids the conflicting IP assignments in case master and agent nodes are in same subnet.

**Which issue this PR fixes**

fixes #1916

### DIFF
--- a/docs/kubernetes/features.md
+++ b/docs/kubernetes/features.md
@@ -166,29 +166,15 @@ Per default Calico still allows all communication within the cluster. Using Kube
 
 ## Custom VNET
 
-ACS Engine supports deploying into an existing VNET. Operators must specify the ARM path/id of Subnets for the `masterProfile` and  any `agentPoolProfiles`, as well as the first IP address to use for IP static IP allocation in `firstConsecutiveStaticIP`. Additionally, to prevent source address NAT'ing within the VNET, we assign to the `vnetCidr` property in `masterProfile` the CIDR block that represents the usable address space in the existing VNET.
+ACS Engine supports deploying into an existing VNET. Operators must specify the ARM path/id of Subnets for the `masterProfile` and  any `agentPoolProfiles`, as well as the first IP address to use for static IP allocation in `firstConsecutiveStaticIP`. Please note that in any azure subnet, the first four and the last ip address is reserved and can not be used. Additionally, to prevent source address NAT'ing within the VNET, we assign to the `vnetCidr` property in `masterProfile` the CIDR block that represents the usable address space in the existing VNET.
 
-Depending upon the size of the VNET address space, during deployment, it is possible to experience IP address assignment collision between the required Kubernetes static IPs (one each per master and one for the API server load balancer, if more than one masters) and Azure CNI-assigned dynamic IPs (one for each NIC on the agent nodes). In practice, the larger the VNET the less likely this is to happen; some detail, and then a guideline.
-
-First, the detail:
-
-* Azure CNI assigns dynamic IP addresses from the "beginning" of the subnet IP address space (specifically, it looks for available addresses starting at ".4" ["10.0.0.4" in a "10.0.0.0/24" network])
-* acs-engine will require a range of up to 16 unused IP addresses in multi-master scenarios (1 per master for up to 5 masters, and then the next 10 IP addresses immediately following the "last" master for headroom reservation, and finally 1 more for the load balancer immediately adjacent to the afore-described _n_ masters+10 sequence) to successfully scaffold the network stack for your cluster
-
-A guideline that will remove the danger of IP address allocation collision during deployment:
-
-* If possible, assign to the `firstConsecutiveStaticIP` configuration property an IP address that is near the "end" of the available IP address space in the desired  subnet.
-  * For example, if the desired subnet is a `/24`, choose the "239" address in that network space
-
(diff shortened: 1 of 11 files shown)


jackfrancis:
@tamilmani1989 @sharmasushant  FYI I rebased this from master so I could run more recent tests against it. Please re-fetch this from your fork if we need to do further dev!

sharmasushant:
@jackfrancis gentle ping...

JunSun17:
@tamilmani1989 @sharmasushant 

Thanks for working on this! As far as I know, this static IP assignments on masters should resolve issues for master NICs competing for IP resources. However, there are still other network resources are trying to acquire IPs in parallel, and I do see you are enforcing NICs of agent nodes dependent on NICs of master nodes to resolve this issue on agent nodes side.

The addition of dependencies on our network resources seems should work but I feel it complicates our code logic and make it hard to maintain and evolve. Whenever we introduce a new network resource, we will need to add additional dependencies, and in the end, everything should be organized in a serial fashion to avoid IPs conflict.

My knowledge about Azure CNI is quite limited, so please correct me if my understanding is not correct. My question is whether there is some atomic IPs reservation logic on the Azure CNI side, when two requests comes in parallel, CNI should not allow an IP to be given to two quests, right?

sharmasushant:
@JunSun17 can you please elaborate on what the below means? Who is trying to acquire IPs in parallel?

_"However, there are still other network resources are trying to acquire IPs in parallel"_

JunSun17:
@sharmasushant I though the other network resource like load balancers will try to acquire IPs in parallel, but after checking the code, I see some dependency rules has already been set up to avoid parallel requests. But still, I feel to manage such dependencies on the client side really complicates things, say if we want to add another network resource, we will have to setup dependency rules explicitly, and our upgrade story also becomes unnecessarily complicated due to the imposed CNI dependency logic. So I am wondering if Azure CNI can add some internal logic to handle this issue.

sharmasushant:
@JunSun17 One of the challenge here is that acs-engine cannot know the current state of ip-allocations by communicating with Azure (e.g., to determine which IP is available or which one should be reserved).
As a result, we rely on user input (starting ip value) to generate the template. The current solution put the burden of making sure that consecutive IPs are available on the user.

The other option could be to let Azure handle ip address allocation by making all IPs dynamic (both master and agent nodes). However, I think this solution is not what majority of our customers want. @jackfrancis

PR #1973, opened 2017-12-29
 add etcd encryption at rest

pidah:
This PR adds an option to enable kubernetes data encryption at rest (disabled by default as it is currently an alpha feature). This complements https://github.com/Azure/acs-engine/pull/1929 which encrypts etcd communication on the wire. More information is available here: https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -86,7 +86,7 @@ write_files:
       name: localclustercontext
     current-context: localclustercontext
 
-{{if .OrchestratorProfile.KubernetesConfig.EnableEtcdEncryption}}
+{{if EnableDataEncryptionAtRest}}
 - path: "/etc/kubernetes/encryption-config.yaml"
   permissions: "0600"
   owner: "root"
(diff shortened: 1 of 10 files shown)


pidah:
Nice @jackfrancis 😊

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -86,6 +86,26 @@ write_files:
...
     current-context: localclustercontext
 
+{{if EnableDataEncryptionAtRest}}
+- path: "/etc/kubernetes/encryption-config.yaml"
+  permissions: "0600"
+  owner: "root"
+  content: |
+    apiVersion: v1
+    kind: Config
+    kind: EncryptionConfig
```
mrosterm: 2 kind fields?

jackfrancis: I think @mrosterm is right. @pidah could you submit a follow-up PR to remove the `kind: Config` line? Based on the docs you provided that's the resource type we want.

Thanks both!

pidah: @jackfrancis sorry just saw this, will raise a fix PR.

PR #1978, opened 2018-01-02
secure communciation from kubernetes API to kubelet API endpoint

pidah:
Communication from kubelet -> API server is currently secure, but not API server -> kubelet. The  kubelet exposes an unauthenticated endpoint on port 10250 which can be exploited from any other node on the cluster.

This PR adds:
- client cert authentication for the kubernetes API against the kubelet API
- Disables anonymous access to the kubelet API
- Restricts access to the kubelet API to only the kubernetes API

More info is available here:

https://kubernetes.io/docs/admin/kubelet-authentication-authorization/
https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/
https://github.com/kubernetes/kubernetes/issues/52184


jackfrancis:
@pidah Thanks!

`gofmt -s pkg/acsengine/defaults-kubelet.go` to fix unit test lint failure, btw.

jackfrancis:
@pidah nevermind I gofmt'd myself, gonna run E2E now

PR #1989, opened 2018-01-04
Addressing CIS issues reported by Kubernetes Auto Analyzer.

### LINKED ISSUE #1828: ACS-Engine CIS compliance issues
No

---

ISSUE

---

Latest

---

The configuration of the ACS-Engine Kubernetes cluster and it's components does not conform to CIS benchmarking standards 1.8

Kubernetes

**What happened**:
There are a high number of failures reported which include multiple critical issues.

Reduce the number of critical issues reported (at the very least).

This article has the information on how to run the bench mark https://github.com/nccgroup/kube-auto-analyzer which produces CIS report being referred to in the issue.

N/A

JunSun17:
**What this PR does / why we need it**: Multiple issues are reported by running Kubernetes Auto Analyzer (https://github.com/nccgroup/kube-auto-analyzer). This PR tries to address those issues that can be easily fixed.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1828

**Special notes for your reviewer**: No need to review yet, will go through test first to see if it breaks things. 

**Release note**:

```release-note
```

### DIFF
--- a/docs/clusterdefinition.md
+++ b/docs/clusterdefinition.md
@@ -280,17 +280,25 @@ Below is a list of apiserver options that are *not* currently user-configurable,
 
 |apiserver option|default value|
 |---|---|
-|"--admission-control"|"NamespaceLifecycle, LimitRanger, ServiceAccount, DefaultStorageClass, ResourceQuota, DenyEscalatingExec"|
+|"--admission-control"|"NamespaceLifecycle, LimitRanger, ServiceAccount, DefaultStorageClass, ResourceQuota, DenyEscalatingExec, AlwaysPullImages, SecurityContextDeny"|
 |"--address"|"0.0.0.0"|
 |"--advertise-address"|*calculated value that represents listening URI for API server*|
 |"--allow-privileged"|"true"|
+|"--anonymous-auth"|"false|
+|"--audit-log-maxage"|"30"|
+|"--audit-log-maxbackup"|"10"|
+|"--audit-log-maxsize"|"100"|
+|"--audit-log-path"|"/var/log/apiserver/audit.log"|
 |"--insecure-port"|"8080"|
 |"--secure-port"|"443"|
+|"--service-account-lookup"|"true"|
 |"--etcd-cafile"|"/etc/kubernetes/certs/ca.crt"|
 |"--etcd-certfile"|"/etc/kubernetes/certs/etcdclient.crt"|
 |"--etcd-keyfile"|"/etc/kubernetes/certs/etcdclient.key"|
 |"--etcd-servers"|*calculated value that represents etcd servers*|
 |"--etcd-quorum-read"|"true"|
+|"--profiling"|"false"|
+|"--repair-malformed-updates"|"false"|
 |"--tls-cert-file"|"/etc/kubernetes/certs/apiserver.crt"|
 |"--tls-private-key-file"|"/etc/kubernetes/certs/apiserver.key"|
 |"--client-ca-file"|"/etc/kubernetes/certs/ca.crt"|
@@ -300,7 +308,7 @@ Below is a list of apiserver options that are *not* currently user-configurable,
 |"--service-cluster-ip-range"|*see serviceCIDR*|
 |"--storage-backend"|*calculated value that represents etcd version*|
 |"--v"|"4"|
-|"--authorization-mode"|"RBAC" (*if enabledRbac is true*)|
+|"--authorization-mode"|"Node", and "RBAC" (*if enabledRbac is true*)|
 |"--experimental-encryption-provider-config"|"/etc/kubernetes/encryption-config.yaml" (*if enableDataEncryptionAtRest is true*)|
 |"--requestheader-client-ca-file"|"/etc/kubernetes/certs/proxy-ca.crt" (*if enableAggregatedAPIs is true*)|
 |"--proxy-client-cert-file"|"/etc/kubernetes/certs/proxy.crt" (*if enableAggregatedAPIs is true*)|
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -10,12 +10,19 @@ import (
 func setAPIServerConfig(cs *api.ContainerService) {
 	o := cs.Properties.OrchestratorProfile
 	staticLinuxAPIServerConfig := map[string]string{
-		"--admission-control":          "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec",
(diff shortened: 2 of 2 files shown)


parts/k8s/manifests/kubernetesmaster-kube-apiserver.yaml
```diff
@@ -48,6 +48,15 @@ spec:
         - "--requestheader-extra-headers-prefix=X-Remote-Extra-"
         - "--requestheader-group-headers=X-Remote-Group"
         - "--requestheader-username-headers=X-Remote-User"
+        - "--anonymous-auth=false"
+        - "--profiling=false"
```
anhowe: how does this affect each existing kubernetes version including those created by acs-engine, and those created by services that use acs-engine?

JunSun17: These changes are based on CIS benchmark, for example, "--anonymous-auth=false" is recommended here:
https://workbench.cisecurity.org/sections/35552/recommendations/98632

I think ideally, each version of Kubernetes should have their own security guidelines, which translates into a bunch of recommended argument settings in yaml files.

For now, the CIS benchmark I used is for Kubernetes 1.8, the recommendations looks pretty generic across multiple versions of Kubernetes to me. So I have not see the needs to do some further versioning yet. Plus, these changes will need to pass multiple version e2e tests before they will hit prod.

Yes, it should affect both acs-engine and the RPs that consumes it.

karataliu:
This breaks acs-engine generated cluster for existing versions.

For the following pod, it would succeed with `acs-engine v1.10` generated `k8s v1.8` cluster, but fail on `acs-engine v1.12` generated `k8s v1.8` cluster. Due to `SecurityContextDeny` option introduced.

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    a: ss
spec:
  containers:
  - name: nginx
    image: nginx
    securityContext:
      seLinuxOptions:
        level: s0
    ports:
    - containerPort: 80
```


Suppose we should support configurable 'admission-control': provide a recommended value as default, and allow user to override.

jackfrancis:
@karataliu Thanks for reporting, could you give an example workflow that will fail without the manual config override? (And can you clarify which config override is now necessary for older clusters?)

JunSun17:
@karataliu Thank you for reporting the issue! Actually the "admission-control" parameter has been made user configurable after this PR. I will investigate the issue you reported to see if the "SecurityContextDeny" should be removed from the default config.

karataliu:
@jackfrancis @JunSun17 
To be clear, there are 2 things here:
- Since default config changes, using **default** config in v0.10.0 and v0.12.0 to create v1.8 cluster will result out in cluster with different behaviour
- If user wants features blocked by SecurityContextDeny/DenyEscalatingExec, manually overriding is required for v1.8 cluster with v0.12 acs-engine, 

Test yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    a: ss
spec:
  containers:
  - name: nginx
    image: nginx
    securityContext:
      seLinuxOptions:
        level: s0
    ports:
    - containerPort: 80
```

Three combinations:
1. v0.10.0 acs-engine 
apimodel:
```
"orchestratorProfile": {
      "orchestratorRelease": "1.8",
      "kubernetesConfig": {}
    },
```

Result: Pod run successfully
`pod "nginx" created`

2. v0.12.0 acs-engine
apimodel:
```
"orchestratorProfile": {
      "orchestratorRelease": "1.8",
      "kubernetesConfig": {}
    },
```
Result: Failure
```
Error from server (Forbidden): error when creating "ng.yaml": pods "nginx" is forbidden: SecurityContext.SELinuxOptions is forbidden
```

3. v0.12.0 acs-engine with config override
apimodel:
```
"orchestratorProfile": {
      "orchestratorRelease": "1.8",
      "kubernetesConfig": {
        "apiServerConfig": {
          "--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota"
        }
      }
    },
```
Result: Pod run successfully
`pod "nginx" created`

JunSun17:
@karataliu Thanks for providing the details on this issue.

I checked the Kubernetes document on this section: https://kubernetes.io/docs/admin/admission-controllers/#securitycontextdeny, and it says

> SecurityContextDeny
> This admission controller will deny any pod that attempts to set certain escalating SecurityContext fields. This should be enabled if a cluster doesn’t utilize pod security policies to restrict the set of values a security context can take.

Because we have not implement Pod Security Policy, my understanding is the above statement asks to set the SecurityContextDeny flag in admission control.

I do not know how many contains are actively setting their own pod level security context, and thus are impacted by this default setting. If it is a common case, I think it makes sense to remove this flag for now and probably implement Pod Security Policy to mediate this issue. Feel free to share your opinion on this.

jackfrancis:
@JunSun17 Pod Security Policy as an option went in yesterday in fact:

https://github.com/Azure/acs-engine/pull/2048

Take a look at that and see if there's a path forward.

JunSun17:
@jackfrancis @karataliu Great, #2048 just come in in time. I see the seLinux policy is set as:

```
+  seLinux:
+    # This policy assumes the nodes are using AppArmor rather than SELinux.
+    rule: 'RunAsAny'
```

@karataliu do we need to set nginx security context? Is it come in by default or by customization?

karataliu:
Since acs-engine by default uses ubuntu as host vm image, selinux is thus unavailable in this case.

I found those when running k8s e2e tests, some of which would fail on a v0.12.0 acs-engine generated cluster.

The selinux configuration comes with following:
https://github.com/kubernetes/kubernetes/blob/c1cd70ad160395160ed64b78ffe8aafdab042352/test/e2e/common/empty_dir.go#L450 

Other failures are also related to the admission-control parameter change:

```
Message: "pods \"pod-34652c7c-fc27-11e7-b1ff-0242ac110003\" is forbidden: pod.Spec.SecurityContext.SELinuxOptions is forbidden",
Message: "pods \"privileged-pod\" is forbidden: cannot exec into or attach to a privileged container",
Message: "pods \"pod-configmaps-bd1a308f-fc28-11e7-bbc0-0242ac110003\" is forbidden: pod.Spec.SecurityContext.RunAsUser is forbidden",
```

As we can see, `SecurityContextDeny` blocks `pod.Spec.SecurityContext.SELinuxOptions` and `pod.Spec.SecurityContext.RunAsUser`, and `DenyEscalatingExec` blocks `privileged-pod`.

JunSun17:
@karataliu I did some testing and verified by making some changes in PR #2048, using a proper PSP file, the nginx deployment works. It should be the long term solution and documented in Issue #2123.

For now, to reduce impact of this setting, I will remove SecurityContextDeny, as in PR #2125.

PR #1994, opened 2018-01-04
Modify build script for 1.7 due to extra cherry-pick in upstream

JiangtianLi:

Modify build script for 1.7 due to extra cherry-pick in upstream


JiangtianLi:
@andyzhangx

PR #1995, opened 2018-01-04
retrieving deployment error details during upgrade

dmitsh:
- retrieving deployment error details during upgrade


cmd/upgrade.go
```diff
@@ -195,7 +195,7 @@ func (uc *upgradeCmd) run(cmd *cobra.Command, args []string) error {
 		Client:      uc.client,
 		StepTimeout: uc.timeout,
 	}
-	log.Fatalf("%v", upgradeCluster.StepTimeout)
```
JackQuincy: opps thanks for removing my debugging line.

pkg/api/orchestrators.go
```diff
@@ -181,6 +181,9 @@ func kubernetesUpgrades(csOrch *OrchestratorProfile) ([]*OrchestratorProfile, er
 	case currentMajor == 1 && currentMinor == 7:
 		nextMajor = 1
 		nextMinor = 8
+	case currentMajor == 1 && currentMinor == 8:
+		nextMajor = 1
```
JackQuincy: shouldn't this be a seperate PR as this feels unrelated and should be easy to push through on it's own.

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -87,11 +87,19 @@ func (kan *UpgradeAgentNode) CreateNode(poolName string, agentNo int) error {
...
-	if err != nil {
-		return err
+	if err == nil {
+		return nil
 	}
 
-	return nil
+	kan.logger.Errorf("Deployment %s failed with error %v", deploymentName, err)
+	// Get deployment error details
+	errResp, e := armhelpers.GetDeploymentError(kan.Client, kan.logger, kan.ResourceGroup, deploymentName)
```
JackQuincy: if I remember right there are two types of errors here. One is a problem during the deployment and the other is during validation. This assumes all are the former. To figure out which kind it is we should need to look at the response etc as well as the error.

pkg/api/orchestrators.go
dmitsh: agreed

pkg/armhelpers/deploymentError.go
```diff
@@ -0,0 +1,150 @@
...
+		// DeploymentOperationProperties contains error from one specific resource type so the
+		// chances of multiple deployment errors being returned for a single resource type is slim
+		// (but possible) based on current error/QoS analysis. In those cases where multiple errors
+		// are returned ACS will pick the first error code for determining whether this is an internal
+		// or a client error. This can be reevaluated later based on practical experience.
+		// However, note that customer will be returned the entire contents of "StatusMessage" object
+		// (like before) so they have access to all the errors returned by ARM.
+		logger.Infof("Found ResourceDeploymentFailure error code - error response = '%v'", *errresp)
+		details := errresp.Body.Details
+		if len(details) > 0 {
```
weinong: why do we want to lose the rest of error?

pkg/armhelpers/deploymentError.go
```diff
@@ -0,0 +1,150 @@
...
+
+func toArmErrors(logger *logrus.Entry, deploymentName string, operationsList resources.DeploymentOperationsListResult) ([]*ErrorResponse, error) {
+	ret := []*ErrorResponse{}
+
+	if operationsList.Value == nil {
+		return ret, nil
+	}
+
+	for _, operation := range *operationsList.Value {
+		if operation.Properties == nil || *operation.Properties.ProvisioningState != string(api.Failed) {
```
weinong: can `ProvisioningState` be nil?

pkg/armhelpers/deploymentError.go
```diff
@@ -0,0 +1,150 @@
...
+		errList = append(errList, eList...)
+	}
+	switch len(errList) {
+	case 0:
+		return nil, fmt.Errorf("unable to extract deployment error for %s", deploymentName)
+	case 1:
+		return errList[0], nil
+	default:
+		// combine all errors
+		for _, e := range errList[1:] {
```
weinong: why cant we just return an array of `*ErrorResponse`? Why do we want to change it?

pkg/armhelpers/deploymentError.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// ErrorResponse  defines Resource Provider API 2.0 Error Response Content structure
+type ErrorResponse struct {
+	Body Error `json:"error"`
+}
+
+// Error implements error interface to return error in json
+func (e *ErrorResponse) Error() string {
+	return e.Body.Error()
```
anhowe: is this recursive at all?  I remember in C# there were errors stuffed inside other errors, and turtles all the way down.

dmitsh: yes, Error.Error() is recursive.

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -87,11 +88,23 @@ func (kan *UpgradeAgentNode) CreateNode(poolName string, agentNo int) error {
...
+	kan.logger.Errorf("Deployment %s failed with error %v", deploymentName, err)
+	// Get deployment error details
+	errRespList, e := armhelpers.GetDeploymentError(kan.Client, kan.logger, kan.ResourceGroup, deploymentName)
+	if e != nil || len(errRespList) == 0 {
+		kan.logger.Errorf("Failed to get error details for deployment %s: %v", deploymentName, e)
+		// return original deployment error
+		return err
+	}
+	errStrList := make([]string, len(errRespList))
+	for i, errResp := range errRespList {
```
weinong: I thought you are changing it to return an array. why changed your mind?

dmitsh: I returned an array, but eventually have to turn it into an error

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -80,18 +80,26 @@ func (kan *UpgradeAgentNode) CreateNode(poolName string, agentNo int) error {
 	deploymentSuffix := random.Int31()
 	deploymentName := fmt.Sprintf("agent-%s-%d", time.Now().Format("06-01-02T15.04.05"), deploymentSuffix)
 
-	_, err := kan.Client.DeployTemplate(
+	depExt, err := kan.Client.DeployTemplate(
```
JackQuincy: there is another possibility here. you are assuming it will work or there will be a list of deployment errors.  If the deployment fails early on, due to something like a quota issue, it will return the err in the response and won't have started any deployment operations. In the RP I have an example in arminteractor where we called deploytemplate there on how to find that error and parse it out.

pkg/apierror/const.go
```diff
@@ -0,0 +1,63 @@
...
+	NetworkingInternalOperationError      ErrorCode = "NetworkingInternalOperationError"
+	QuotaExceeded                         ErrorCode = "QuotaExceeded"
+	Unauthorized                          ErrorCode = "Unauthorized"
+	ResourcesOverConstrained              ErrorCode = "ResourcesOverConstrained"
+	ControlPlaneProvisioningInternalError ErrorCode = "ControlPlaneProvisioningInternalError"
+	ControlPlaneProvisioningSyncError     ErrorCode = "ControlPlaneProvisioningSyncError"
+	ControlPlaneInternalError             ErrorCode = "ControlPlaneInternalError"
+	ControlPlaneCloudProviderNotSet       ErrorCode = "ControlPlaneCloudProviderNotSet"
+
+	// From Microsoft.WindowsAzure.ContainerService.API.AcsErrorCode
```
JackQuincy: the errors from here down appear to be RP errors. I'm not sure any of them apply in the ACS-Engine scenario. We should have a const file in the RP with these values and make sure that the datamodel down here just stores a string and matches these that way we can set values not defined here.

amanohar:
My feedback and concerns about this PR remains the same as what we discussed in person:
1. The error codes are _very_ specific to what the RP and do not belong in ACS Engine at all
2. The error categories are not even an ARM concept and are internal to the RP. So while there is nothing confidential about it, it breaks abstraction and does not belong in ACS Engine
3. This change significantly impacts agility (and development cycle of the RP) of the RP OR we will end up with error codes in 2 places because a dev will just add to the RP to save time. 
4. My recommendations to you was to return ARM’s deployment results to the upgrade cmd and to the RP and RP can do the parsing.

PR #2000, opened 2018-01-05
Consistent naming between linux and windows agents

mellenburg:
The windows agent VMs were provisioned with a name that is the shorted suffix + acs + 900
E.G.
Given a suffix of 40448065
linux agents with the name linpri are given a VMSS labeled like so: dcos-linpri-40448065--vmss
before this change, windows agents with the name wpub were given a VMSS named 40448acs901-vmss

With this change, this VMSS would be named:  dcos-wpub-40448065--vmss

**Which issue this PR fixes**:
Lack of consistent naming for this one resource makes it much harder to easily identify from the Resource Manager

**Release note**:
`NONE`


yakman2020:
There is a 14 char limit to windows host names due to a .net limit. Note that isn't 14 char unique, but rather 14 char total.  So the windows host name is synthesized to fit in that limit.

PR #2004, opened 2018-01-05
Added Windows Custom Image. Only works in dcos, but can be added to k8s

yakman2020:
Allows the use of custom windows system disk images in the VLabs API deployments. This in turn allows us to use confidential daily build, or specially provisioned images.  This is implemented completely for DC/OS. Someone in the k8s team will probably want to apply the changes the k8s parts.

While this is implemented completely for DC/OS, it modifies the WindowsProfile so should be usable in any of the different code legs, such as k8s or docker swarm with fairly simple mods.

**Release note**:

### DIFF
--- a/parts/windowsparams.t
+++ b/parts/windowsparams.t
@@ -16,4 +16,11 @@
         "description": "Version of the Windows Server 2016 OS image to use for the agent virtual machines."
       },
       "type": "string"
+    },
+    "agentWindowsSourceUrl": {
+      "defaultValue": "",
+      "metadata": {
+        "description": "The source of the generalized blob which will be used to create a custom windows image for the agent virtual machines."
+      },
+      "type": "string"
     }
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -123,10 +123,11 @@ type PublicKey struct {
 
 // WindowsProfile represents the windows parameters passed to the cluster
 type WindowsProfile struct {
-	AdminUsername string            `json:"adminUsername"`
-	AdminPassword string            `json:"adminPassword"`
-	ImageVersion  string            `json:"imageVersion"`
-	Secrets       []KeyVaultSecrets `json:"secrets,omitempty"`
+	AdminUsername         string            `json:"adminUsername"`
+	AdminPassword         string            `json:"adminPassword"`
+	ImageVersion          string            `json:"imageVersion"`
+	WindowsImageSourceURL string            `json:"windowsImageSourceURL"`
+	Secrets               []KeyVaultSecrets `json:"secrets,omitempty"`
 }
 
 // ProvisioningState represents the current state of container service resource.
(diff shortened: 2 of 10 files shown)


parts/agentparams.t
```diff
@@ -52,6 +52,13 @@
       },
       "type": "string"
     },
+    "{{.Name}}WindowsImageSourceUrl": {
+      "defaultValue": "",
```
anhowe: reference the assigned value.  Looks at examples from Subnet or VMSize

parts/windowsparams.t
```diff
@@ -16,4 +16,11 @@
         "description": "Version of the Windows Server 2016 OS image to use for the agent virtual machines."
       },
       "type": "string"
-    }
\ No newline at end of file
+    },
+    "agentWindowsSourceUrl": {
```
anhowe: How is this different from WindowsImageSourceUrl?

pkg/api/types.go
```diff
@@ -123,10 +123,11 @@ type PublicKey struct {
...
 // WindowsProfile represents the windows parameters passed to the cluster
 type WindowsProfile struct {
-	AdminUsername string            `json:"adminUsername"`
-	AdminPassword string            `json:"adminPassword"`
-	ImageVersion  string            `json:"imageVersion"`
-	Secrets       []KeyVaultSecrets `json:"secrets,omitempty"`
+	AdminUsername         string            `json:"adminUsername"`
+	AdminPassword         string            `json:"adminPassword"`
+	ImageVersion          string            `json:"imageVersion"`
+	WindowsImageSourceURL string            `json:"windowsImageSourceURL"`
```
anhowe: I would recommend apply to whole cluster (all agent pools). or per agent pool, but not both.  Which matches your scenario better?

parts/windowsparams.t
yakman2020: For now, it isn't except its in variables() rather than parameters().

pkg/api/types.go
yakman2020: Whole cluster.  I beleive that is doing that. Am I missing something?

parts/agentparams.t
yakman2020: done.

pkg/api/types.go
anhowe: If "whole cluster" is your scenario, I think you can completely remove the per node pool change you made in parts/agentparams.t.

jackfrancis:
@yakman2020 this looks good to go except for the cruft in `pkg/acsengine/json.go`.

jackfrancis:
Also, to address @anhowe's comment, please add something like this to `pkg/api/vlabs/validate.go` under the `func (a *Properties) Validate(isUpdate bool) error {` function:

```
if a.OrchestratorProfile.OrchestratorType != DCOS {
	if a.WindowsProfile.HasCustomImage() {
		return fmt.Errorf("Windows custom image is only supported for DC/OS orchestrator type")
	}
}
```

yakman2020:
validation added.

yakman2020:
@anhowe  Added validation, removed per host custom image name.

PR #2012, opened 2018-01-08
generic Kubernetes API server config interface

jackfrancis:
**What this PR does / why we need it**: exposes a generic key/val interface for configuring apiserver, using the configuration keys documented here:

https://kubernetes.io/docs/reference/generated/kube-apiserver/

API model usage patterns will look like this:

```
"kubernetesConfig": {
    <...>
    "apiServerConfig": {
          "--request-timeout": "30s"
    }
    <...>
}
```

**Release note**:

```
add a generic Kubernetes API Server config interface
```


docs/clusterdefinition.md
```diff
@@ -209,7 +209,7 @@ Below is a list of controller-manager options that acs-engine will configure by
 |"--route-reconciliation-period"|"10s"|
 
 
-Below is a list of kubelet options that are *not* currently user-configurable, either because a higher order configuration vector is available that enforces kubelet configuration, or because a static configuration is required to build a functional cluster:
+Below is a list of controller-manager options that are *not* currently user-configurable, either because a higher order configuration vector is available that enforces controller-manager configuration, or because a static configuration is required to build a functional cluster:
```
jackfrancis: Correcting copy/paste errors from a prior, related PR.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -182,23 +182,13 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
...
-    AAD_ISSUER_HOST="sts.windows.net"
-    if [ "$VAR_TARGET_ENV" = "AzureChinaCloud" ]; then
-        AAD_ISSUER_HOST="sts.chinacloudapi.cn"
-    fi
-
-    OIDC_ISSUER_URL="https://$AAD_ISSUER_HOST/$AAD_TENANT_ID/"
-    perl -pi -e "s|--oidc-client-id=\K(?=\")|$OIDC_CLIENT_ID| || s|--oidc-issuer-url=\K(?=\")|$OIDC_ISSUER_URL|" "/etc/kubernetes/manifests/kube-apiserver.yaml"
-{{else}}
-    sed -i "/--oidc-client-id\|--oidc-issuer-url\|--oidc-username-claim/d" "/etc/kubernetes/manifests/kube-apiserver.yaml"
+    sed -i "/--oidc-issuer-url/s/$/$AAD_TENANT_ID/" "/etc/kubernetes/manifests/kube-apiserver.yaml"
```
jackfrancis: We still need to do this post-deployment transformation because I'm not aware of any other way to implement the fallback tenantID assignment. The one here uses this `kubernetesmastervars.t` statement:

`"tenantId": "[subscription().tenantId]",`

Which gives us our fallback `tenantId` value (see `AAD_TENANT_ID=${VAR_AAD_TENANT_ID:-$VAR_TENANT_ID}` above). My understanding is that `subscription()` is an ARM function that we won't have access to anywhere else.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -245,47 +235,22 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
...
-    sed -i "/requestheader-group-headers/d" "/etc/kubernetes/manifests/kube-apiserver.yaml"
-    sed -i "/requestheader-username-headers/d" "/etc/kubernetes/manifests/kube-apiserver.yaml"
-{{end}}
-    sed -i "s|<etcdApiVersion>|{{ .OrchestratorProfile.GetAPIServerEtcdAPIVersion }}|g" "/etc/kubernetes/manifests/kube-apiserver.yaml"
-
 {{if UseCloudControllerManager }}
     sed -i "s|<kubernetesCcmImageSpec>|{{WrapAsVariable "kubernetesCcmImageSpec"}}|g; s|<masterFqdnPrefix>|{{WrapAsVariable "masterFqdnPrefix"}}|g; s|<allocateNodeCidrs>|{{WrapAsVariable "allocateNodeCidrs"}}|g; s|<kubeClusterCidr>|{{WrapAsVariable "kubeClusterCidr"}}|g; s|<kubernetesCtrlMgrRouteReconciliationPeriod>|{{GetCloudControllerManagerRouteReconciliationPeriod .OrchestratorProfile.KubernetesConfig}}|g" \
       /etc/kubernetes/manifests/cloud-controller-manager.yaml
-
-    sed -i "/--\(cloud-config\|cloud-provider\|route-reconciliation-period\)=/d" \
```
jackfrancis: This should have been removed in #1960, doing so now.

pkg/acsengine/defaults-controller-manager.go
```diff
@@ -12,8 +13,6 @@ func setControllerManagerConfig(cs *api.ContainerService) {
 		"--kubeconfig":                       "/var/lib/kubelet/kubeconfig",
 		"--allocate-node-cidrs":              strconv.FormatBool(!o.IsAzureCNI()),
 		"--cluster-cidr":                     o.KubernetesConfig.ClusterSubnet,
-		"--cloud-provider":                   "azure",
```
jackfrancis: This should have been removed in #1960, doing so now.

pkg/acsengine/defaults-controller-manager.go
```diff
@@ -30,6 +29,12 @@ func setControllerManagerConfig(cs *api.ContainerService) {
 		staticLinuxControllerManagerConfig["--cluster-name"] = cs.Properties.HostedMasterProfile.DNSPrefix
 	}
 
+	// Enable cloudprovider if we're not using cloud controller manager
```
jackfrancis: This should have been done in #1960, doing so now.

parts/k8s/kubernetesmastervars.t
```diff
@@ -267,8 +266,8 @@
       "[concat(variables('masterFirstAddrPrefix'), add(3, int(variables('masterFirstAddrOctet4'))))]",
       "[concat(variables('masterFirstAddrPrefix'), add(4, int(variables('masterFirstAddrOctet4'))))]"
     ],
-    "masterEtcdServerPort": 2380,
-    "masterEtcdClientPort": 2379,
+    "masterEtcdServerPort": "[parameters('masterEtcdServerPort')]",
```
jackfrancis: Moved these vals to `pkg/acsengine/const.go` (and added appropriate changes elswhere).

pkg/acsengine/engine.go
```diff
@@ -873,6 +872,20 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
...
+			apiServerConfig := kc.APIServerConfig
+			// Order by key for consistency
+			keys := []string{}
+			for key := range apiServerConfig {
+				keys = append(keys, key)
+			}
+			sort.Strings(keys)
+			var buf bytes.Buffer
+			for _, key := range keys {
+				buf.WriteString(fmt.Sprintf("\\\"%s=%s\\\", ", key, apiServerConfig[key]))
```
jackfrancis: The escaped slashes `\\` in front of the escaped quote `\"` is needed in order to produce a literal quote char in the resulting yaml file.

This should be done in the above getters as well, will do so in a follow-up PR.

pkg/acsengine/defaults-apiserver.go
```diff
@@ -0,0 +1,107 @@
...
+	// We don't support user-configurable values for the following,
+	// so any of the value assignments below will override user-provided values
+	var overrideAPIServerConfig map[string]string
+	if cs.Properties.HasWindows() {
+		overrideAPIServerConfig = staticWindowsAPIServerConfig
+	} else {
+		overrideAPIServerConfig = staticLinuxAPIServerConfig
+	}
+	for key, val := range overrideAPIServerConfig {
+		o.KubernetesConfig.APIServerConfig[key] = val
```
JunSun17: Some argument like "--admission-control" and "--authorization-mode" takes a list of features, and probably should allow feature addition (as long as static config is a subset of the user input config), instead of override. It can be done in a later change if needed.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -245,47 +228,29 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
...
-      /etc/kubernetes/manifests/kube-controller-manager.yaml
-    sed -i "/--\(cloud-config\|cloud-provider\)=/d" \
-      /etc/kubernetes/manifests/kube-apiserver.yaml
 {{end}}
     sed -i "s|<kubernetesControllerManagerConfig>|{{GetControllerManagerConfigKeyVals .OrchestratorProfile.KubernetesConfig}}|g" "/etc/kubernetes/manifests/kube-controller-manager.yaml"
+    sed -i "s|<kubernetesAPIServerConfig>|{{GetAPIServerConfigKeyVals .OrchestratorProfile.KubernetesConfig}}|g" "/etc/kubernetes/manifests/kube-apiserver.yaml"
```
JunSun17: It is probably better to take APIServerConfig (instead of KubernetesConfig) as argument for GetAPIServerConfigKeyVals.

The original code logic is pretty tangled here and I have not dig very deep, I assume GetAPIServerConfigKeyVals will return the already processed argument map. Please just confirm it.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -245,47 +228,29 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
...
+    sed -i "s|<kubernetesAPIServerIP>|{{WrapAsVariable "kubernetesAPIServerIP"}}|g" "/etc/kubernetes/manifests/kube-apiserver.yaml"
+{{ if .HasAadProfile }}
+    VAR_AAD_TENANT_ID={{WrapAsVariable "aadTenantId"}}
+    VAR_TENANT_ID={{WrapAsVariable "tenantId"}}
+    AAD_TENANT_ID=${VAR_AAD_TENANT_ID:-$VAR_TENANT_ID}
+    sed -i "/--oidc-issuer-url/s/$/$AAD_TENANT_ID/" "/etc/kubernetes/manifests/kube-apiserver.yaml"
```
JunSun17: If the above GetAPIServerConfigKeyVals returns already processed map, why are we overriding "--oidc-issuer-url" here?

jackfrancis: Good question. It's because the only way (that I know of) to assign the right tenantId value is in the flow of this template: we depend upon `{{WrapAsVariable "tenantId"}}` as a fallback value, and that fallback value is not available in the flow of the map-producing code. See the `subscription()` method which the `tenantId` variable depends upon in `kubernetesmastervars.t`.

pkg/acsengine/defaults-apiserver.go
jackfrancis: I agree this would be a useful addition to this surface area. Feature gates also are special, in that they essentially contain key/vals themselves. Yeah, I'd prefer a follow-up PR for that.

parts/k8s/kubernetesmastercustomdata.yml
jackfrancis: This is good feedback. Let's do this: after all of these config refactors are done, we can optimize the number of key/val getter functions in `engine.go` (we have two types, I think, which is better than having one function per config file). A part of that improvement would be passing in the specific config as the arg instead of kubernetesConfig.

PR #2017, opened 2018-01-09
Generic cloud-controller-manager config interface

jackfrancis:
**What this PR does / why we need it**: exposes a generic key/val interface for configuring cloud-controller-manager, using the configuration keys documented here:

https://kubernetes.io/docs/reference/generated/cloud-controller-manager/

API model usage patterns will look like this:

```
"kubernetesConfig": {
    <...>
    "cloudControllerManagerConfig": {
          "--route-reconciliation-period": "1m"
    }
    <...>
}
```

```
Generic controller-manager config interface
```


jackfrancis:
@JunSun17 Kindly requesting review. :) This is the final generic k8s runtime config to expose!

pkg/acsengine/defaults-cloud-controller-manager.go
```diff
@@ -0,0 +1,69 @@
...
+	staticWindowsCloudControllerManagerConfig := make(map[string]string)
+	for key, val := range staticLinuxCloudControllerManagerConfig {
+		staticWindowsCloudControllerManagerConfig[key] = val
+	}
+	// Windows cloud-controller-manager config overrides
+	// TODO placeholder for specific config overrides for Windows clusters
+
+	// Default cloud-controller-manager config
+	defaultCloudControllerManagerConfig := map[string]string{
+		"--node-monitor-grace-period": DefaultKubernetesCtrlMgrNodeMonitorGracePeriod,
```
JunSun17: From the md file, the default is "--route-reconciliation-period", a typo?

pkg/acsengine/engine.go
```diff
@@ -872,6 +872,20 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			}
 			return strings.TrimSuffix(buf.String(), ", ")
 		},
+		"GetCloudControllerManagerConfigKeyVals": func(kc *api.KubernetesConfig) string {
```
CecileRobertMichon: Is it necessary to pass in the KubernetesConfig here? For distinct agent/master configs it needs to be passed in but since this looks like it's using the orchestratorProfile's kubernetesConfig here you could also use cs.Properties.OrchestratorProfile.KubernetesConfig since the getTemplateFuncMap gets passed in a ContainerService

pkg/acsengine/engine.go
```diff
@@ -872,6 +872,20 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
...
+			cloudControllerManagerConfig := kc.CloudControllerManagerConfig
+			// Order by key for consistency
```
CecileRobertMichon: if this starts getting repeated in more configs (it's already in GetControllerManagerConfigKeyVals, GetKubeletConfigKeyVals and GetApiServerConfigKeyVals) maybe it's time to extract it to a function

examples/kubernetes-config/kubernetes-cloud-controller-manager.json
```diff
@@ -0,0 +1,39 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "orchestratorRelease": "1.8",
+      "kubernetesConfig": {
+        "useCloudControllerManager": true
```
CecileRobertMichon: Might be beneficial to put a CloudControllerManagerConfig field here for reference

jackfrancis: I'm not sure I want to encourage someone to use this yet until we spend some time testing various custom configurations that actually work! :)

pkg/acsengine/defaults-cloud-controller-manager.go
jackfrancis: Good catch! Actually, the doc is correct, it's an error in the `defaultCloudControllerManagerConfig` object! (see the changes to `parts/k8s/manifests/kubernetesmaster-cloud-controller-manager.yaml`). I'll fix.

pkg/acsengine/engine.go
jackfrancis: I'm actually planning to submit a follow-up PR that generalizes the getter function so that it works on `map[string]string`s generically.

pkg/acsengine/engine.go
jackfrancis: Oh, yes! What I said above.

PR #2028, opened 2018-01-10
Jenkins soak tests

CecileRobertMichon:
**What this PR does / why we need it**: Adds support for Jenkins soak tests in e2e suite

**Release note**:

```release-note
```


test/e2e/runner.go
```diff
@@ -51,6 +51,11 @@ func main() {
 
 	// Only provision a cluster if there isnt a name present
 	if cfg.Name == "" {
+		if cfg.SoakClusterName != "" {
```
JunSun17: Should this belong to some cleanup step (maybe the teardown function at the bottom of this file)? The comments on this section of code says it is for provisioning a cluster.

test/e2e/runner/cli_provisioner.go
```diff
@@ -76,17 +76,22 @@ func (cli *CLIProvisioner) Run() error {
...
+	if cli.Config.SoakClusterName == "" {
+		out, err := exec.Command("ssh-keygen", "-f", cli.Config.GetSSHKeyPath(), "-q", "-N", "", "-b", "2048", "-t", "rsa").CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("Error while trying to generate ssh key:%s\nOutput:%s", err, out)
+		}
+		exec.Command("chmod", "0600", cli.Config.GetSSHKeyPath()+"*")
 	}
-	exec.Command("chmod", "0600", cli.Config.GetSSHKeyPath()+"*")
 
 	publicSSHKey, err := cli.Config.ReadPublicSSHKey()
```
JunSun17: if the above keygen is skipped when soak cluster name is provided, should the publicSSHKey part be skipped too?

CecileRobertMichon: The keygen is skipped because for Jenkins soak test the tests use the Jenkins key so it is not necessary to generate a new key. However, the cli provisioner still needs to read the public key that was provided.

test/e2e/runner.go
CecileRobertMichon: The reason why I put this here is that when generating a new soak cluster it is necessary to delete the previously existing soak cluster before deploying a new one since they use the same resource group (and we want to make sure we don't keep them after we're done testing on them).

CecileRobertMichon:
@JunSun17 to answer your question, yes that's exactly right. The goal of this PR is to facilitate the Jenkins soak tests such as https://jenkins.azure-containers.io/view/ACSE-Ginkgo/job/ginkgo-soak-test-k8s-1.7/ by allowing us to run E2E on a pre-existing cluster. The SoakClusterName env var is provided only for soak tests so if it is empty then E2E will run as usual.

test/e2e/runner.go
JunSun17: I see, thanks for the explanation.

test/e2e/runner/cli_provisioner.go
JunSun17: Just to confirm, so the public key is not the one generated by keygen, right?

CecileRobertMichon: No, it's the one in _output/cluster-name-ssh which is created before the test is ran

JunSun17: Thanks for the clarification.

PR #2032, opened 2018-01-10
Add --feature-gates handling for kubelet and api server

### LINKED ISSUE #2021: Add support for kubernetes --feature-gates in template

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): FEATURE REQUEST

---

**What version of acs-engine?**: latest master (8fd4ac4)

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**: Kubernetes - all versions

**What you expected to happen**: Various behaviours (and preview features) in Kubernetes are controlled by `--feature-gates`. As per the [docs](https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md#kubeletconfig), specifying `--feature-gates` is currently not permitted.

Proposal:

1. Allow `--feature-gates` to be specified in the `kubeletConfig`. 
Looking at the code, the agents currently include `Accelerators=true` in `--feature-gates` for kubernetes version 1.6.0 and above, so this behaviour should be preserved. 
2. Add an `apiserverConfig` section to `kubernetesConfig` as per the `kubeletConfig` to allow apiserver specific configuration (including `--feature-gates`). This was discussed in a comment [here](https://github.com/Azure/acs-engine/pull/1854#issuecomment-348030399)

Sample config:

```json

"kubernetesConfig": {
    "kubeletConfig": {
        "--cluster-dns": "172.17.1.11",
        "--max-pods": "30",
        "--feature-gates": {
            "MountPropagation": "true",
            "TaintNodesByCondition": "true"
        },
        <etc...>
    },
    "apiserverConfig": {
        "--storage-backend": "etcd3",
        "--feature-gates": {
            "MountPropagation": "true"
        },
        <etc...>
    }
}
```

**Update**: 
As per the [comments](https://github.com/Azure/acs-engine/pull/2032#issuecomment-357813879) from @paulbouwer we should ensure that this covers settings `--feature-gates` for:

- [x] kubelet
- [x] api-server
- [x] kube-controller-manager
- [ ] kube-proxy
- [ ] kube-scheduler

stuartleeks:
Current work in progress, for feedback

Add handling of `--feature-gates` for kubernetes. The `--feature-gates` value is a comma separated string, e.g. `key1=value1,key2=value2`
For `kubeletConfig`, this preserves the existing behaviour of adding `Accelerators=true` for the agent config (i.e. not master) for kubernetes 1.6.0 and later

**What this PR does / why we need it**: As per #2021, adding support for `--feature-gates` will enable easier testing of kubernetes features (especially preview features)

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2021 

* I think that this fits with the discussion in [issue #2021](#2021) - let me know if you want any further changes

**Release note**:

```release-note

```

### DIFF
--- a/pkg/acsengine/defaults_test.go
+++ b/pkg/acsengine/defaults_test.go
@@ -377,10 +377,6 @@ func TestKubeletFeatureGatesEnsureMasterAndAgentConfigUsedFor1_6_0(t *testing.T)
 	}
 }
 
-// TODO - tests:
-//  - check merging
-//  - check that master/agent overrides are used correctly
-
 func getMockAddon(name string) api.KubernetesAddon {
 	return api.KubernetesAddon{
 		Name:    name,
--- a/pkg/acsengine/defaults_test.go
+++ b/pkg/acsengine/defaults_test.go
@@ -287,6 +287,100 @@ func TestPointerToBool(t *testing.T) {
 	}
 }
 
+func TestKubeletFeatureGatesEnsureAcceleratorsNotSetFor1_5_0(t *testing.T) {
+	mockCS := getMockBaseContainerService("1.5.0")
+	properties := mockCS.Properties
+
+	// No KubernetesConfig.KubeletConfig set for MasterProfile or AgentProfile
+	// so they will inherit the top-level config
+	properties.OrchestratorProfile.KubernetesConfig = getKubernetesConfigWithFeatureGates("TopLevel=true")
+
+	setKubeletConfig(&mockCS)
+
+	// Verify that the Accelerators feature gate has not been applied to master or agents
+	agentFeatureGates := properties.AgentPoolProfiles[0].KubernetesConfig.KubeletConfig["--feature-gates"]
+	if agentFeatureGates != "TopLevel=true" {
+		t.Fatalf("setKubeletConfig modified the agent profile (version 1.5.0): expected 'TopLevel=true' got '%s'", agentFeatureGates)
+	}
+	masterFeatureFates := properties.MasterProfile.KubernetesConfig.KubeletConfig["--feature-gates"]
+	if masterFeatureFates != "TopLevel=true" {
+		t.Fatalf("setKubeletConfig modified feature gates for master profile: 'TopLevel=true' got '%s'", agentFeatureGates)
+	}
+}
+func TestKubeletFeatureGatesEnsureAcceleratorsOnAgentsFor1_6_0(t *testing.T) {
+	mockCS := getMockBaseContainerService("1.6.0")
+	properties := mockCS.Properties
+
+	// No KubernetesConfig.KubeletConfig set for MasterProfile or AgentProfile
+	// so they will inherit the top-level config
+	properties.OrchestratorProfile.KubernetesConfig = getKubernetesConfigWithFeatureGates("TopLevel=true")
+
+	setKubeletConfig(&mockCS)
+
+	agentFeatureGates := properties.AgentPoolProfiles[0].KubernetesConfig.KubeletConfig["--feature-gates"]
+	if agentFeatureGates != "Accelerators=true,TopLevel=true" {
+		t.Fatalf("setKubeletConfig did not add 'Accelerators=true' for agent profile: expected 'Accelerators=true;TopLevel=true' got '%s'", agentFeatureGates)
+	}
+
+	// Verify that the Accelerators feature gate override has only been applied to the agents
+	masterFeatureFates := properties.MasterProfile.KubernetesConfig.KubeletConfig["--feature-gates"]
+	if masterFeatureFates != "TopLevel=true" {
+		t.Fatalf("setKubeletConfig modified feature gates for master profile: expected 'TopLevel=true' got '%s'", agentFeatureGates)
+	}
+}
+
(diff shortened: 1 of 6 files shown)


croomes:
@stuartleeks this looks great and seems like it would solve our use case.  Is there anything I can help with?

stuartleeks:
@jackfrancis - as far as I could tell, in the original code `Accelerators=true` was only added to the kubelet for the agent config. In the changes you've pushed it looks like it is now added to the master config as well. Was that intended?

jackfrancis:
@stuartleeks I hope you don't mind I took a crack at this as well (see my 3 commits). This is meant to accomplish the following:

- set `"Accelerators=true"` for all > 1.5 clusters by default
- apply that default setting to orchestrator-wide, master-specific, and per-agent-pool kubelet configs
- allow user-configurable feature gates across all three kubelet configs, and enforcing

We aren't allowing users to bypass `--feature-gates` entirely (i.e., even an explicit value of `""` will yield `"Accelerators=true"`), but I think that's not a current user requirement.

jackfrancis:
@stuartleeks Let me audit the old code, thanks in any event for paying close attention!

stuartleeks:
Here's where `Accelerators=true` was previously set: https://github.com/Azure/acs-engine/pull/2032/commits/31f5c1b979027b43acea780b4c520de925dccbd4#diff-7d5f50f227edfc447ff04539035aa614L113

jackfrancis:
Yep, you're right! I also rationalized the minimum version requirement a bit in my latest commit.

I'm going to run E2E against this, of course feel free to continue to move this forward if we're not yet where we need to be functionality-wise. Thanks again for encouraging use in this domain!

Also let's add a description to the PR. Looks like we're shipping with a single comma-delimited string that we'll deserialize, e.g.:

`"--feature-gates": "foo=bar,twoplustwo=five,that=this"`

Is that correct? I'm happy to ship with that, we can worry about making that more resilient (for example, is it possible for a val to contain a comma?) later.

stuartleeks:
@jackfrancis I've made a few more changes, including adding tests for the target behaviour based on my understanding of our discussions.

I've added a note at the top of the PR on the format of the `--feature-gates` property as suggested. Looking at the docs, all keys for feature gates are alpha chars (no commas), and values all look to be true/false. The format is based on the format that the `--feature-gates` switch accepts

paulbouwer:
@stuartleeks @jackfrancis Feature gates are exposed as options on:
- [kubelet](https://kubernetes.io/docs/reference/generated/kubelet/)
- [api-server](https://kubernetes.io/docs/reference/generated/kube-apiserver/)
- [kube-controller-manager](https://kubernetes.io/docs/reference/generated/kube-controller-manager/)
- [kube-proxy](https://kubernetes.io/docs/reference/generated/kube-proxy/)
- [kube-scheduler](https://kubernetes.io/docs/reference/generated/kube-scheduler/)

This PR only seems to add them to the **kubelet**. Is that going to be an issue for other use cases / scenarios?

Another thought (i'll raise a separate issue) I had around comma separated lists was a way to add or remove a value from a default list or simply replace the entire list. I've been looking into this as part of adding `Initializers` to the `--admission-control` option in **api-server**. If we come up with a re-usable approach, it can be leveraged for any option that requires a list.

stuartleeks:
@paulbouwer - good point!

I've updated the title of the PR to reflect kubelet and apiserver (I tested both and they are working). I'll check back in the code and see which other components have configuration elements. (The configuration is something that @jackfrancis has been working on recently)

I've also made a note in the associated issue (#2021) that we should support the areas you listed.

stuartleeks:
@jackfrancis - I just tested and specifying `--feature-gates` on the `kube-controller-manager` config results in them being passed through in the generated template.

That leave `kube-proxy` and `kube-scheduler`, neither of which currently have config sections.

Are you ok to merge this PR and track `--feature-gate` support for those two in the linked issue? (I suspect that they will just work like `api-server` and `kube-controller-manager` based on the config pattern you've been using).

pkg/acsengine/defaults_test.go
```diff
@@ -287,6 +287,100 @@ func TestPointerToBool(t *testing.T) {
...
+	}
+
+	// Verify that the Accelerators feature gate override has only been applied to the agents
+	masterFeatureFates := properties.MasterProfile.KubernetesConfig.KubeletConfig["--feature-gates"]
+	if masterFeatureFates != "MasterLevel=true" {
+		t.Fatalf("setKubeletConfig master profile: expected 'MasterLevel=true' got '%s'", agentFeatureGates)
+	}
+}
+
+// TODO - tests:
```
jackfrancis: Are these TODOs still operative? (we can merge with them in there if they mark gaps in test surface area that we want to fill in the future)

jackfrancis:
@stuartleeks your plan sounds good, we should have cycles in the upcoming release cycle to genericize kube-proxy and kube-scheduler

pkg/acsengine/defaults_test.go
stuartleeks: Oops - no, I've added the tests now. Just pushed a commit to remove the comments

stuartleeks:
@jackfrancis I've pushed a commit to address your review comment. Let me know if there is anything else you'd like updated before you merge

PR #2039, opened 2018-01-11
Use Api Certs for kubelet client authentication

pidah:
Use k8s API certs for kubelet client authentication and not the node client certs. Fixes an error in my earlier PR Azure/acs-engine#1978

### DIFF
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -32,8 +32,8 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		"--profiling":                  "false",
 		"--repair-malformed-updates":   "false",
 		"--service-account-key-file":   "/etc/kubernetes/certs/apiserver.key",
-		"--kubelet-client-certificate": "/etc/kubernetes/certs/client.crt",
-		"--kubelet-client-key":         "/etc/kubernetes/certs/client.key",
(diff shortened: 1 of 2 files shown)


pkg/acsengine/defaults-apiserver.go
```diff
@@ -81,7 +81,7 @@ func setAPIServerConfig(cs *api.ContainerService) {
 
 	// Default apiserver config
 	defaultAPIServerConfig := map[string]string{
-		"--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny",
+		"--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny,PodSecurityPolicy",
```
JunSun17: Should PodSecurityPolicy be optional instead of default, I thought it will need a policy file around to work?

pidah: @JunSun17 yes you are right, will try and update PR soon

jackfrancis:
Closing this as it's probably a little stale at this point. @pidah let me know if we want an issue to track the work we still need to do here.

PR #2040, opened 2018-01-11
Fix import of satori/go.uuid

ultimateboy:
1. Pin `satori/go.uuid` to 1.x
2. Update `satori/uuid` import to `satori/go.uuid` in code
3. `make build-vendor` after the glide changes

Currently, in glide.lock, we see the same package listed twice with different versions. `satori/uuid` and `satori/go.uuid`. The former is a redirect to the latter. While this change does not completely remove the duplicate reference, it's a step in the right direction. As these changes propagate to other dependencies, the duplicate reference will be removed.

With this change, however, at least the same version of the package is imported, instead of two different versions.


jackfrancis:
Thanks @ultimateboy !

PR #2048, opened 2018-01-12
Add Enable Pod Security Option

pidah:
This PR adds an option to enable Pod Security Policy.
It applies a privileged policy to the kube-system namespace to enable successful bootstrapping of the cluster components, and a restricted policy elsewhere by default. Both policies are sourced from the docs here:  https://kubernetes.io/docs/concepts/policy/pod-security-policy/#

### DIFF
--- a/pkg/acsengine/addons.go
+++ b/pkg/acsengine/addons.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/Azure/acs-engine/pkg/api"
+	"github.com/Azure/acs-engine/pkg/helpers"
 )
 
 type kubernetesFeatureSetting struct {
@@ -88,7 +89,7 @@ func kubernetesManifestSettingsInit(profile *api.Properties) []kubernetesFeature
 		{
 			"kubernetesmaster-pod-security-policy.yaml",
 			"pod-security-policy.yaml",
-			profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy != nil && *profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy,
+			helpers.IsTrueBoolPointer(profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy),
 		},
 		{
 			"kubernetesmaster-kube-apiserver.yaml",
--- a/pkg/acsengine/addons.go
+++ b/pkg/acsengine/addons.go
@@ -86,6 +86,11 @@ func kubernetesManifestSettingsInit(profile *api.Properties) []kubernetesFeature
 			profile.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager != nil && *profile.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager,
 		},
 		{
+			"kubernetesmaster-pod-security-policy.yaml",
+			"pod-security-policy.yaml",
+			profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy != nil && *profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy,
+		},
+		{
 			"kubernetesmaster-kube-apiserver.yaml",
 			"kube-apiserver.yaml",
 			true,
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -101,7 +101,7 @@ func setAPIServerConfig(cs *api.ContainerService) {
 
 	// Pod Security Policy configuration
 	if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnablePodSecurityPolicy) {
-		defaultAPIServerConfig["--admission-control"] = "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny,PodSecurityPolicy"
+		defaultAPIServerConfig["--admission-control"] = defaultAPIServerConfig["--admission-control"] + ",PodSecurityPolicy"
 	}
 
 	// If no user-configurable apiserver config values exists, use the defaults
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -99,6 +99,11 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		}
 	}
 
+	// Pod Security Policy configuration
+	if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnablePodSecurityPolicy) {
+		defaultAPIServerConfig["--admission-control"] = "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny,PodSecurityPolicy"
+	}
+
 	// If no user-configurable apiserver config values exists, use the defaults
 	if o.KubernetesConfig.APIServerConfig == nil {
 		o.KubernetesConfig.APIServerConfig = defaultAPIServerConfig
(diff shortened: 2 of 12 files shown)


jackfrancis:
FYI @pidah I rebased from master to resolve merge conflicts. You'll want to re-pull down this branch from your remote!

pkg/acsengine/defaults-apiserver.go
```diff
@@ -99,6 +99,11 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		}
 	}
 
+	// Pod Security Policy configuration
+	if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnablePodSecurityPolicy) {
+		defaultAPIServerConfig["--admission-control"] = "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny,PodSecurityPolicy"
```
jackfrancis: To make `--admission-control` more maintainable going forward, let's just add `",PodSecurityPolicy"` to the end of the existing string at this point in the code (after default assignment) if `EnablePodSecurityPolicy` is true, rather than replacing the entire string.

pkg/api/vlabs/validate.go
```diff
@@ -111,6 +111,13 @@ func (o *OrchestratorProfile) Validate(isUpdate bool) error {
 						}
 					}
 				}
+				if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnablePodSecurityPolicy) {
+					if o.KubernetesConfig.EnableRbac != nil {
```
jackfrancis: We could use `helpers.IsTrueBoolPointer` again here (in this case we'd use it to determine not true). E.g.:

`if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnablePodSecurityPolicy) && !helpers.IsTrueBoolPointer(o.KubernetesConfig.EnableRbac)`

pkg/acsengine/engine.go
```diff
@@ -1564,6 +1564,10 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"EnableDataEncryptionAtRest": func() bool {
 			return helpers.IsTrueBoolPointer(cs.Properties.OrchestratorProfile.KubernetesConfig.EnableDataEncryptionAtRest)
 		},
+		"EnablePodSecurityPolicy": func() bool {
```
jackfrancis: It doesn't look like this function is ever used.

pkg/acsengine/addons.go
```diff
@@ -85,6 +85,11 @@ func kubernetesManifestSettingsInit(profile *api.Properties) []kubernetesFeature
 			"cloud-controller-manager.yaml",
 			profile.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager != nil && *profile.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager,
 		},
+		{
+			"kubernetesmaster-pod-security-policy.yaml",
+			"pod-security-policy.yaml",
+			profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy != nil && *profile.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy,
```
jackfrancis: `helpers.IsTrueBoolPointer` is better here (we'll do a follow-up PR to clean up other usage patterns before that convenience function was introduced, for example in the `kubernetesmaster-cloud-controller-manager.yaml` entry above

pkg/acsengine/defaults-apiserver.go
pidah: updated. I was going to suggest we refactor the admission-control flags into a []string or a map instead of a string.

pkg/api/vlabs/validate.go
pidah: updated

pkg/acsengine/engine.go
pidah: removed, thanks !

pkg/acsengine/addons.go
pidah: updated.

pkg/acsengine/defaults-controller-manager.go
```diff
@@ -44,10 +44,11 @@ func setControllerManagerConfig(cs *api.ContainerService) {
...
 	defaultControllerManagerConfig := map[string]string{
-		"--node-monitor-grace-period":   DefaultKubernetesCtrlMgrNodeMonitorGracePeriod,
-		"--pod-eviction-timeout":        DefaultKubernetesCtrlMgrPodEvictionTimeout,
-		"--route-reconciliation-period": DefaultKubernetesCtrlMgrRouteReconciliationPeriod,
-		"--terminated-pod-gc-threshold": DefaultKubernetesCtrlMgrTerminatedPodGcThreshold,
+		"--node-monitor-grace-period":       DefaultKubernetesCtrlMgrNodeMonitorGracePeriod,
+		"--pod-eviction-timeout":            DefaultKubernetesCtrlMgrPodEvictionTimeout,
+		"--route-reconciliation-period":     DefaultKubernetesCtrlMgrRouteReconciliationPeriod,
+		"--terminated-pod-gc-threshold":     DefaultKubernetesCtrlMgrTerminatedPodGcThreshold,
+		"--use-service-account-credentials": DefaultKubernetesCtrlMgrUseSvcAccountCreds,
```
JunSun17: I think this should be removed since it is controlled by RBAC at the botton of this file.

```
 	if *o.KubernetesConfig.EnableRbac {
 		o.KubernetesConfig.ControllerManagerConfig["--use-service-account-credentials"] = "true"		 		o.KubernetesConfig.ControllerManagerConfig["--use-service-account-credentials"] = "true"
 	}
```

JunSun17: My concern is that it changes the default behavior of "--use-service-account-credentials", it used to be false when RBAC is off, but now it becomes true.

jackfrancis: @pidah Can you explain the use-case where a cluster would want `--use-service-account-credentials` without RBAC enabled? In other words, what is the justification for disentangling this controller-manager runtime option from higher order acs-engine RBAC enforcement?

JunSun17:
@pidah 

I tried the latest code with your changes and add config as:

```
      "kubernetesConfig": {
        "enableRbac": true,
        "enablePodSecurityPolicy": true
      }
```
However, it does not start PSP.  So I logged in and tried to manually start PSP by:

`kubectl create -f /etc/kubernetes/manifests/pod-security-policy.yaml     `                                                  

I got the following error, 

`error: error validating "/etc/kubernetes/manifests/pod-security-policy.yaml": error validating data: found invalid field allowPrivilegeEscalation for v1beta1.PodSecurityPolicySpec; if you choose to ignore these errors, turn validation off with --validate=false`

So it looks the PSP config file needs to be updated to make it work right out of box.

pidah:
@JunSun17 @jackfrancis  ah right, the `allowPrivilegeEscalation` field was added in kubernetes 1.8. I presume you were testing on 1.7 ? Will raise a PR to remove this field.

pkg/acsengine/defaults-controller-manager.go
pidah: @jackfrancis @JunSun17 No strong opinion – as I understand it, the `--use-service-account-credentials` flag ensures the kubernetes controller manager uses the individual designated service accounts for each control loop instead of a single service account credential for all control loops. This segregation/subdivision seemed like a good default and shouldn't impact the cluster when RBAC is not enabled.

jackfrancis:
@pidah Let's do the latter for simplicity (for now). We'll just fail validation for `enablePodSecurityPolicy` for 1.7 and below.

JunSun17:
@pidah @jackfrancis Yes, I believe my test is with 1.7. I think another way to work around this is to apply versioning on the PSP configure file, i.e. create 1.7 and 1.8 folder under the parts/k8s/manifests folder, like what has been done for addons and artifacts. I guess some work will be needed to generate PSP configure files for 1.7 and 1.8 separately.

pidah:
@JunSun17 yes that is another way. As @jackfrancis said, I will raise a PR to fail validation on 1.7 and below for now. If there is demand for PSP in 1.7 we can raise another PR to support it.

PR #2049, opened 2018-01-12
Fix k8s 1.6 regression

CecileRobertMichon:
**What this PR does / why we need it**: Only add Node authorization-mode to default api-server config if the k8s version is greater or equal to 1.7.0

**Release note**:

```release-note
```


pkg/acsengine/defaults-apiserver.go
```diff
@@ -88,7 +88,10 @@ func setAPIServerConfig(cs *api.ContainerService) {
 
 	// RBAC configuration
 	if helpers.IsTrueBoolPointer(o.KubernetesConfig.EnableRbac) {
-		defaultAPIServerConfig["--authorization-mode"] = "Node,RBAC"
+		defaultAPIServerConfig["--authorization-mode"] = "RBAC"
+		if isKubernetesVersionGe(o.OrchestratorVersion, "1.7.0") {
```
jackfrancis: We need this same version guard in the `defaultAPIServerConfig` assignment above as well.

brendandburns:
Given 1.9.x support, I think it might be time to deprecate 1.6?



________________________________________

PR #2050, opened 2018-01-12
enableSecureKubelet and unit tests

jackfrancis:
**What this PR does / why we need it**: Adds a `enableSecureKubelet` option to `kubernetesConfig` for opting out of securt communications between API server and kubelet.

Also added unit tests for default kubelet and apiserver config enforcement.

**Release note**:

```
enableSecureKubelet
```


pkg/api/types.go
```diff
@@ -616,12 +617,15 @@ func (p *Properties) HasAadProfile() bool {
 
 // GetAPIServerEtcdAPIVersion Used to set apiserver's etcdapi version
 func (o *OrchestratorProfile) GetAPIServerEtcdAPIVersion() string {
-	// if we are here, version has already been validated..
```
jackfrancis: These modifications prevent nil dereferences (handy when scaffolding out test fixtures)

Issue #2047, opened 2018-01-12
master doesn't work with AAD


tesharp:
**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

**What version of acs-engine?**: master (pulled today)

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**: Kubernetes

**What happened**:

Deploying a new cluster I just got an error:

VM has reported a failure when processing extension 'cse0'. Error message: \\\"Enable failed: failed to execute command: command terminated with exit status=3\\n[stdout]\\n\\n[stderr]\\nstat: cannot stat '/opt/azure/containers/provision.complete': No such file or directory\\n\\\"

Tried to troubleshoot and found out the apiserver never started. In the end I noticed that /etc/kubernetes/manifests/kube-apiserver.yaml had some strange error. In the args list it seemed like it had appended the AAD Tenant Id after all arguments. 

args: ["--address=0.0.0.0", "--admission-control=NamespaceLifecucle,........, "--v=4"]fa5c22d4-xxxx-xxxx-xxxx-4ffdc39axxxxfa5c22d4-xxxx-xxxx-xxxx-4ffdc39axxxx

It had even done it twice. It appears a change 3 days ago might have caused this (https://github.com/Azure/acs-engine/commit/d216b9438773b007e6e7dfb152ed120ce994bd76). Changing how it generates this, no longer one line per argument which would have made it worked with oidc-issuer-url.. Need to have more fun with sed here it seems to create correct oidc-issuer-url, and not add tenant id at end of line. 

**What you expected to happen**:

Api server to successfully start when deploying any clusters.

**How to reproduce it** (as minimally and precisely as possible):

Deploy a cluster with AAD Tenant set. Failes in deployment of 1.9 clusters. For 1.8 it didn't fail when running acs-engine, but api-server was not running.. so thats another issue again.

**Anything else we need to know**:

PR #2055, opened 2018-01-15
Fix issue with apiserver when using AADProfile (#2047)

tesharp:
Fixes issue mentioned in https://github.com/Azure/acs-engine/issues/2047 where kube-apiserver.yaml file is incorrect when setting AADProfile in api-model. Moved issuer url generation to go file instead of using sed.


jackfrancis:
@tesharp Thanks for this contribution! Did my PR here break AAD as reported in #2047?:

https://github.com/Azure/acs-engine/commit/d216b9438773b007e6e7dfb152ed120ce994bd76#diff-91eb315fd90ae6b7177cd6290a360600

The intention there was to transfer the logic intact, which I would distill as "if I have an AAD tenant ID, use that to construct the oidc issuer URL, if not, use the default cluster tenant ID to construct the oid issuer URL".

What did my change break? Did I mess up the implementation, or did I misinterpret the inention?

tesharp:
It seemed to break now because in the kube-apiserver.yaml the arguments are now on one line, instead of one line per arguments. Before change the --oidc-issuer-url was on a separate line, so when the sed ran it just appended the tenant id to the end of that line. After the commit however all arguments are on the same line, so the same sed didn't work appropriately. Appending to the end of the line just added tenant id in wrong place, not after definition of --oidc-issuer-url argument. 

Intention was good, just different implementation details messing it up :) I moved the logic to apiserver-defaults.go file now as some of the other arguments had been. Seemed clearer that way.

jackfrancis:
Thanks for the clear explanation, and for doing the tedious work of moving this into a better place (+1 to less `sed` in this codebase)!

PR #2056, opened 2018-01-15
Remove agent windows version from vlabs API

itowlson:
**What this PR does / why we need it**: This reverts https://github.com/Azure/acs-engine/pull/1438.  In subsequent discussions on Slack, @anhowe identified that the Windows OS disk image version should _not_ be part of the API because of long-term maintenance considerations.  It should be configurable via `parameters.json` but not the API.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: None

**Special notes for your reviewer**: None

**Release note**:

```release-note
NONE
```


jackfrancis:
@yakman2020 thoughts? Does this configuration property have regular usage in your experience?

jackfrancis:
Based on the nature of the broad usage patterns + the possibility that changing the data model in a destructive way could break back-compat scenarios, let's abandon this one.

Issue #1972, opened 2017-12-28
Cannot scale Windows agent pool (Kubernetes Cluster)


bhope:
**Is this a request for help?**:
Yes

**Is this an ISSUE or FEATURE REQUEST?**
 Issue

**What version of acs-engine?**:
v0.11.0

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes 1.7.7

**What happened**

After I deployed a Hybrid Kubernetes cluster with Windows and Linux agent pool, 
I tried to scale out the Windows agents pool using the acs-engine scale command.
After executing the command, I get the following error:

```
INFO[0000] validating...
INFO[0003] Name suffix: 25663911
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/Azure/acs-engine/cmd.(*scaleCmd).run(0xc42078a000, 0xc42058cb40, 0xc420164540, 0x0, 0xc, 0x0, 0x0)
	/Users/jackfrancis/work/src/github.com/Azure/acs-engine/cmd/scale.go:209 +0x219f
github.com/Azure/acs-engine/cmd.newScaleCmd.func1(0xc42058cb40, 0xc420164540, 0x0, 0xc, 0x0, 0x0)
	/Users/jackfrancis/work/src/github.com/Azure/acs-engine/cmd/scale.go:64 +0x52
github.com/Azure/acs-engine/vendor/github.com/spf13/cobra.(*Command).execute(0xc42058cb40, 0xc420164480, 0xc, 0xc, 0xc42058cb40, 0xc420164480)
	/Users/jackfrancis/work/src/github.com/Azure/acs-engine/vendor/github.com/spf13/cobra/command.go:647 +0x3f8
github.com/Azure/acs-engine/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc4206aa6c0, 0xc42058c480, 0xc42058c240, 0xc42058c000)
	/Users/jackfrancis/work/src/github.com/Azure/acs-engine/vendor/github.com/spf13/cobra/command.go:726 +0x339
github.com/Azure/acs-engine/vendor/github.com/spf13/cobra.(*Command).Execute(0xc4206aa6c0, 0xc420000448, 0x0)
	/Users/jackfrancis/work/src/github.com/Azure/acs-engine/vendor/github.com/spf13/cobra/command.go:685 +0x2b
main.main()
[...]

PR #2059, opened 2018-01-16
Scale windows VM bug: catch error for len(vms) < 1

CecileRobertMichon:
**What this PR does / why we need it**: This is a temporary fix which catches the error where ListVirtualMachines returns 0 for a resource group. This causes scale operations to fail for clusters with windows nodes as reported by #1972.

**Release note**:

```release-note
```


cmd/scale.go
```diff
@@ -188,6 +188,9 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 		vms, err := sc.client.ListVirtualMachines(sc.resourceGroupName)
 		if err != nil {
 			log.Fatalln("failed to get vms in the resource group. Error: %s", err.Error())
+		} else if len(*vms.Value) < 1 {
+			// TODO fix bug: clusters with windows vms always fail here
```
jackfrancis: Let's remove this TODO line. I think checking for the length of the response is general, and any existing bugs that expose this symptom are coincidental.

PR #2067, opened 2018-01-17
k8s/script: allow parallelizing custom script without clear-containers

jessfraz:
cc @CecileRobertMichon :)

might be prettier with `w=1` in the url: https://github.com/Azure/acs-engine/pull/2067/files?w=1

### DIFF
--- a/parts/k8s/kubernetesmastercustomscript.sh
+++ b/parts/k8s/kubernetesmastercustomscript.sh
@@ -20,7 +20,7 @@
 
 # Master only secrets
 # APISERVER_PRIVATE_KEY CA_CERTIFICATE CA_PRIVATE_KEY MASTER_FQDN KUBECONFIG_CERTIFICATE
-# KUBECONFIG_KEY ETCD_SERVER_CERTIFICATE ETCD_SERVER_PRIVATE_KEY ETCD_CLIENT_CERTIFICATE ETCD_CLIENT_PRIVATE_KEY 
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -644,18 +644,23 @@ users:
...
-echo `date`,`hostname`, RunCmdCompleted>>/opt/m
+if [[ "$CONTAINER_RUNTIME" == "clear-containers" ]]; then
+	# If the container runtime is "clear-containers" we need to ensure the
+	# run command is completed _before_ we start installing all the dependencies
+	# for clear-containers to make sure there is not a dpkg lock.
+	ensureRunCommandCompleted
+	echo `date`,`hostname`, RunCmdCompleted>>/opt/m
+fi
 
 if [[ $OS == $UBUNTU_OS_NAME ]]; then
```
CecileRobertMichon: Is there a reason why this got moved up in your previous PR? if not it can probably go back to right after ```echo `date`,`hostname`,` RunCmdCompleted>>/opt/m```

jessfraz: I moved it up since we are installing things for clearcontainers i didn't want it to upgrade... but can move inside the `if` if that hellps

CecileRobertMichon: Sounds good. I think it's needed for all runtimes, not just clearcontainers so let's leave it as is

PR #2074, opened 2018-01-17
update pause image for v1.8 and v1.9 k8s clusters

jackfrancis:
**What this PR does / why we need it**: Updates pause to pause-amd64:3.1 for all v1.8 and v1.9 Kubernetes clusters

**Release note**:

```
pause-amd64:3.1 for v1.8 and v1.9 clusters
```


pkg/acsengine/k8s_versions.go
```diff
@@ -47,7 +47,7 @@ var KubeConfigs = map[string]map[string]string{
 		"dns":                       "k8s-dns-kube-dns-amd64:1.14.5",
 		"addonmanager":              "kube-addon-manager-amd64:v6.5",
 		"dnsmasq":                   "k8s-dns-dnsmasq-nanny-amd64:1.14.5",
-		"pause":                     "pause-amd64:3.0",
+		"pause":                     "pause-amd64:3.1",
```
JunSun17: I did a search on "pause-amd64:3.0" in the code base and it is also showed up in common.KubernetesVersion1Dot9Dot1 section, should that use the new version too? Or maybe they are targets in a different PR for 1.9.1?

jackfrancis: Oversight, will fix, thanks!

JunSun17: Actually also checked the other PRs, since the package with versions are all over the code, can we extract them to some constant string to avoid human errors?

jackfrancis: I'm happy to do that in a follow-up PR.

JunSun17: Great. :)

PR #2086, opened 2018-01-18
Add documentation to update Route Table when Custom VNET is in different RG

### LINKED ISSUE #2085: Adding Route Table to Custom VNET in different Resource Group fails
**Is this an ISSUE or FEATURE REQUEST?**: ISSUE
**What version of acs-engine?**: v0.11.0 cda96312, clean

Kubernetes 1.8.4

**What happened**:
I followed the instructions to add the Route Table into the custom VNET, no errors were received, but when running
```
kubectl proxy
```
Then attempt to open the dashboard at http://localhost:8001/ui
The following was received
```
Starting to serve on 127.0.0.1:8001
I0118 13:53:24.007134   61824 logs.go:41] http: proxy error: context canceled
```

This is the same behaviour as before adding the Route Table into the custom VNET.
Note this works as expected if the Custom VNET is in the same Resource Group as the Kubernetes Cluster.

The browser to open the Kubernetes Dashboard, and calls be proxied through to the dashboard POD

Follow the instructions on the Custom VNET page, apart from creating another Resource Group for the Custom VNET.

Have found the answer [Impossible to add route-table to subnet on different resource groups](https://github.com/Azure/azure-cli/issues/4817)
This states that rather than specifying subnet name etc. that if you use the full canonical name option it does work. I have tested this, and can confirm that it is correct. 
PR will be coming shortly

davesykeselateral:
Updated Custom VNET docs to add instructions for route table when the… Custom VNET is in a different Resource Group to the Kubernetes cluster. 
From my experiments, although the existing example bash script does not error, it does not successfully update the Route Table in the VNET when the VNET is in a different Resource Group to the Kubernetes cluster. From https://github.com/Azure/azure-cli/issues/4817 it appears that this works when you use the full canonical ids for the subnet, this PR adds an example that does this

**Which issue this PR fixes**
`fixes #2085`

**Release note**:

```release-note
Add to Custom VNet documentation to show how to update Route Table when Custom VNET is in different Resource Group
```

### DIFF
--- a/docs/custom-vnet.md
+++ b/docs/custom-vnet.md
@@ -170,7 +170,7 @@ Depending on the number of agent you have asked for the deployment can take a wh
 
 ## Post-Deployment: Attach Cluster Route Table to VNET
 
-For Kubernetes clusters, we need to update the VNET to attach to the route table created by the above `az group deployment create` command. An example in bash form:
+For Kubernetes clusters, we need to update the VNET to attach to the route table created by the above `az group deployment create` command. An example in bash form for if the VNET is in the same ResourceGroup as the Kubernetes Cluster:
 
 ```
 #!/bin/bash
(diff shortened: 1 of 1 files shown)


jackfrancis:
@davesykeselateral Thanks for your contribution, I'll get this in this week.

jackfrancis (approved):
Thanks @davesykeselateral!

PR #2095, opened 2018-01-18
Preserve subnets during scale up

### LINKED ISSUE #2063: Scale up deletes subnets
**Is this a request for help?**: Yes

---

---

**What version of acs-engine?**: Whatever is currently used by the Azure CLI (2.0.23) / REST API

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**:  Kubernetes, default version

**What happened**: When I scaled up the agent pool, it deleted a subnet I had added to the cluster vnet.  (Scaling _down_ the agent pool did _not_ affect the subnet,)  If it can't delete the subnet, the scale operation fails.

**What you expected to happen**: That my subnet would not be affected by agent pool scaling

1. Create an ACS Kubernetes cluster using Azure CLI 2.0.23.
2. View the resource group in the Azure portal.
3. Select the vnet created for the cluster (k8s-vnet-xxxxxxxx).
4. Add a subnet named `test-subnet` to this vnet.
5. Return to the resource group and select the container service.
6. Select `Agents` and increase the agent pool size from 3 to 4.
7. Wait for the resize to complete.
8. Select the vnet again and view its subnets.
9. Note that `test-subnet` no longer exists.

**Anything else we need to know**: You can also try adding a public IP to the custom subnet.  Your scale up operation will then fail because it can't delete the subnet while it's in use.

itowlson:
**What this PR does / why we need it**: When you scale up an acs-engine cluster (using `az acs scale` or `acs-engine scale`), any subnets of the cluster vnet other than the automatic `k8s-master` subnet are deleted.  This is unexpected, and if the subnet is in use (for example because it contains a gateway or IP address) then the subnet cannot be deleted and this blocks the scaling operation.

The underlying problem is that scale up recreates the ARM template for the resource group and re-applies it.  Because the ARM template includes only the original automatic subnet, ARM assumes that the desired state of the vnet is to contain only the automatic subnet, and therefore attempts to delete any other subnets.

This PR excludes the vnet from the ARM template generated for scaling, using the same techniques already used to exclude the route table and NSG.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2063 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/acsengine/transform/transform.go
+++ b/pkg/acsengine/transform/transform.go
@@ -36,10 +36,12 @@ const (
 	vmssResourceType = "Microsoft.Compute/virtualMachineScaleSets"
 	vmExtensionType  = "Microsoft.Compute/virtualMachines/extensions"
 	nicResourceType  = "Microsoft.Network/networkInterfaces"
+	vnetResourceType = "Microsoft.Network/virtualNetworks"
 
 	// resource ids
-	nsgID = "nsgID"
-	rtID  = "routeTableID"
+	nsgID  = "nsgID"
+	rtID   = "routeTableID"
+	vnetID = "vnetID"
 )
 
 // Translator defines all required interfaces for i18n.Translator.
@@ -104,6 +106,7 @@ func (t *Transformer) NormalizeForK8sVMASScalingUp(logger *logrus.Entry, templat
 	}
 	rtIndex := -1
 	nsgIndex := -1
+	vnetIndex := -1;
 	resources := templateMap[resourcesFieldName].([]interface{})
 	for index, resource := range resources {
 		resourceMap, ok := resource.(map[string]interface{})
@@ -129,6 +132,14 @@ func (t *Transformer) NormalizeForK8sVMASScalingUp(logger *logrus.Entry, templat
 			}
 			rtIndex = index
 		}
+		if ok && resourceType == vnetResourceType {
+			if vnetIndex != -1 {
+				err := t.Translator.Errorf("Found 2 resources with type %s in the template. There should only be 1", vnetResourceType)
+				logger.Warnf(err.Error())
+				return err
+			}
+			vnetIndex = index
+		}
 
 		dependencies, ok := resourceMap[dependsOnFieldName].([]interface{})
 		if !ok {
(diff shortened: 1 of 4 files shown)


feiskyer:
@itowlson Is this ready for merge? It seems the change make sense

itowlson:
@feiskyer I'd value a review before merge.  I _think_ it's a safe change, and that it's the desired behaviour for all scenarios that I know of -- but others on the acs-engine team will have a better sense of possible customer networking scenarios and whether the current behaviour is intentional.  Would you or @jackfrancis be a suitable reviewer perhaps?

feiskyer:
@itowlson Thanks. LGTM in my review.

dmitsh:
LGTM. Could you test upgrade scenario for 3 masters with and without VNET?

itowlson:
Upgrade seems to work, but destroys subnets (recreates vnet), so I guess we should fix it to preserve the vnet in that scenario as well.  Thanks for catching that - I'd only considered the scale scenario, not upgrade.  _returns to the code mines_

itowlson:
I've applied a fix so the `upgrade` command preserves the vnet, but it needs re-review because the way the upgrade path strips things out of the base templates is very different from the way the scale path does it!  Can you take a look please @feiskyer @dmitsh?  Thanks!

rocketraman:
Looking forward to this one as we use custom VNETs in our acs-engine cluster...

jackfrancis:
@itowlson and @JunSun17, could you work together on reconciling the changes in this PR and in this PR that merged recently:

https://github.com/Azure/acs-engine/commit/93c99d1b0bbdaf6776d38c2f42f1e7b91abe5971#diff-d11913d43be1f6b820fa831e9216a6cc

It looks like @JunSun17 explicitly removed the master NIC dependency stuff, while @itowlson's change in this PR refined it.

stephenlawrence:
Version: canary
GitCommit: 7c1d1699
GitTreeState: dirty

I still am seeing this issue when testing with this branch. I utilize a RedisCache in the resource group and create a redis-subnet for it (required). 

"Subnet redis-subnet is in use by /subscriptions/SUBID/resourceGroups/RGNAME/providers/Microsoft.Cache/Redis/MyRedisCache and cannot be deleted"

CecileRobertMichon:
I think we can close this PR since #2994 was merged? @jackfrancis

jackfrancis:
I'd defer to @itowlson, who would probably be happy to discover that (if?) we've fixed this 5 months later ;)

itowlson:
@CecileRobertMichon @jackfrancis Yep, #2994 looks pretty similar to this, except that I don't think it addresses the upgrade scenario (unless that code path has changed) which was requested on this PR.  I agree we should close this but perhaps we should keep upgrade on the radar, as again we don't want e.g. Redis or App Gateway subnets to block an upgrade.  Other than that, YAY!

jackfrancis:
Thanks for checking in @itowlson, I agree of course that upgrade/scale should share the same underlying functional assumptions

Issue #2015, opened 2018-01-09
[Windows Nodes] "There is no enough space on the disk"


Stygy:
**Is this a request for help?**:

YES

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

ISSUE

**What version of acs-engine?**:

v0.11.0

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes 1.7.9

**What happened**:
Cannot pull images because of a "There is not enough space on the disk" error:
![image](https://user-images.githubusercontent.com/6739343/34728649-769b5672-f528-11e7-822a-9595e4acaa9a.png)
 
K8s shows that there is no disk space issue: 
![image](https://user-images.githubusercontent.com/6739343/34728535-235de56a-f528-11e7-910d-ff370d7427be.png)

**What you expected to happen**:
Pod created with no issue

**How to reproduce it** (as minimally and precisely as possible):
1) Create a Hybrid Cluster with this json: https://goo.gl/9zA9wF
2) Create 6 deployments with microsoft/iis (version 1709) as base image. These will work
3) Create an additional deployment with the same image, the error will start to appear. 

**Anything else we need to know**:
I've upgraded the master machine to D12_V2 but that didn't fix the error. https://goo.gl/g7FjuQ

andyzhangx:
you need to assign a bigger OS disk size, see example:
https://github.com/Azure/acs-engine/blob/master/examples/windows/kubernetes-manageddisks.json#L20

Stygy:
I've setup a 200GB disk (https://goo.gl/9zA9wF). What would be a recommended size for the OS disk? Can I resize the disk without rebuilding the cluster?

andyzhangx:
One possible solution is use the resource disk which has 100GB to download image, need to find a way to change docker settings?

robinsondotnet:
I've got the same error. Have anyone solved? 
I am using a image based on microsoft/dotnet-framework:4.7 . This image is around 12GB. I think that it  shouldn't be a problem.

andyzhangx:
@JiangtianLi do you know anyway to let docker use resource disk in Windows, I only know how to do it in Linux. Resource disk has 100GB size, it would good place for temp dir.

Stygy:
[...]

PR #2097, opened 2018-01-18
Extend windows os drive size when customized OSDiskSizeGB is used

JiangtianLi:
Fix https://github.com/Azure/acs-engine/issues/2015, extend windows os drive size when customized OSDiskSizeGB is used


JiangtianLi:
@anhowe PTAL

PR #2098, opened 2018-01-19
Adopt CIS Kubernetes Benchmark, Part 3: Kubelet

JunSun17:
**What this PR does / why we need it**: Multiple issues are found by checking against CIS Kubernetes Benchmark (https://workbench.cisecurity.org/benchmarks/581/sections/35559). This PR tries to address those issues that can be easily fixed in Kubelet.

**Which issue this PR fixes**: #1828

### DIFF
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -115,6 +115,10 @@ const (
 	DefaultMasterEtcdServerPort = 2380
 	// DefaultMasterEtcdClientPort is the default etcd client port for Kubernetes master nodes
 	DefaultMasterEtcdClientPort = 2379
+	// DefaultKubeletEventQPS is 0, see --event-qps at https://kubernetes.io/docs/reference/generated/kubelet/
+	DefaultKubeletEventQPS = "0"
+	// DefaultKubeletCadvisorPort is 0, see --cadvisor-port at https://kubernetes.io/docs/reference/generated/kubelet/
+	DefaultKubeletCadvisorPort = "0"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


jackfrancis:
Two questions:

1) are we confident we aren't breaking any existing usage patterns by disabling cAdvisor by default?
2) how does granting unlimited event creation velocity address security concerns?

JunSun17:
@jackfrancis 

> 1. are we confident we aren't breaking any existing usage patterns by disabling cAdvisor by default?

I do not see we are using cAdvisor, please correct me if I am wrong. Set the port to 0 is recommended by CIS, and since it is in the default settings, user can still override them if issue arises.

So far, it passed the CircleCI tests.

> 2. how does granting unlimited event creation velocity address security concerns?

It is recommended by CIS (https://workbench.cisecurity.org/sections/35559/recommendations/98713), I believe the purpose is to not limit queries to have a better monitoring of the k8s environment.

karataliu:
To keep a note here for E2E test run:

This change would break following e2e test case:
https://github.com/kubernetes/kubernetes/blob/v1.10.0-beta.0/test/e2e/network/proxy.go#L78

The case depends on cadvisor to listen on 4194, setting 0 to the port will disable cadvisor listening.

Log:
```
ErrStatus: {
TypeMeta: {Kind: "", APIVersion: ""},
ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
Status: "Failure",
Message: "the server is currently unable to handle the request",
Reason: "ServiceUnavailable",
Details: {
Name: "",
Group: "",
Kind: "",
UID: "",
Causes: [
{
Type: "UnexpectedServerResponse",
Message: "Error: 'dial tcp 10.240.0.6:4194: getsockopt: connection refused'\nTrying to reach: 'http://k8s-agentpool1-21491089-0:4194/containers/'",
Field: "",
},
],
RetryAfterSeconds: 0,
},
Code: 503,
},
}
the server is currently unable to handle the request
```

jackfrancis:
@karataliu To pass the above E2E test, we would build a cluster with the following `kubernetesConfig` configuration:

```
<...>
    "kubernetesConfig": {
        <...>
        "kubeletConfig": {
            <...>
            "--cadvisor-port": "4194",
            <...>
        }
        <...>
    }
<...>
```

PR #2104, opened 2018-01-19
Kind should be only EncryptionConfig for encyption-config.yaml

jackfrancis:
**What this PR does / why we need it**: Removes the unnecessary Config `Kind` value in `/etc/kubernetes/encryption-config.yaml`.

See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

**Release note**:

```
EncryptionConfig for encryption at rest implementation
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -91,7 +91,6 @@ write_files:
   permissions: "0600"
   owner: "root"
   content: |
-    apiVersion: v1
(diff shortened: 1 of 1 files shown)


jackfrancis:
@pidah fyi

SaaldjorMike:
Ah yes, apiVersion is also defined twice. Didn't even see that first time around :-)

jackfrancis:
@pidah No problem, should have caught that myself, thanks for weighing in!

PR #2120, opened 2018-01-23
only create cert files on master

CecileRobertMichon:
**What this PR does / why we need it**: Only create cert files for etcd if on master

**Release note**:

```release-note
```


jackfrancis:
Can we add a bash "constant" to the cse so we know more concretely that we're a master? Rather than the stuff like this:

`echo "APISERVER_PRIVATE_KEY is non-empty, assuming master node"`

Let's just pass a `MASTER_NODE="true"`, var or something more explicit like that. After that we can decompose the script a bit to make "what executes on master" and "what executes on nodes" easier to read/maintain.

Issue #2123, opened 2018-01-23
Setup a default allow-all Pod Security Policy in Kubernetes API server


JunSun17:
**Is this a request for help?**: No

**Is this an ISSUE or FEATURE REQUEST?** (choose one): Feature

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**: Kubernetes

**What you expected to happen**: For security concerns and auditing requirement, we should setup a Pod Security Policy for Kubernetes API server, and users should be able to modify it to apply their customization.

Issue #2109, opened 2018-01-19
Document how to run pods as non-root now SecurityContextDeny is a default admission controller


yuvipanda:
**Is this a request for help?**: No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: v0.12

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**

Kubernetes

**What happened**:

I tried to install a helm chart (z2jh.jupyter.org) that has been working fine on same k8s version created with older acs-engine versions, and I expected it to work. It did not, and when more digging was performed, it turned out that:

1. The chart contained a deployment that explicitly specified a non-0 uid (1000) to run
2. However, per this patch (https://github.com/Azure/acs-engine/pull/1989), the SecurityContextDeny admission controller was enabled by default. This means that pods that set *any* uid will fail, and by default this means that you can *only* run pods that run all containers as root.

SecurityContextDeny seems to operate outside the scope of PodSecurityPolicies as well, so am actually unsure how to get around this? I've temporarily edited /etc/kubernetes/manifests to remove SecurityContextDeny from the list of admission controllers.

**What you expected to happen**:

One of:

1. It continues to work as it did before
2. A detailed note in the release notes saying SecurityContextDeny is now a default admission controller, the ramifications for this, how to disable it if necessary, and how to allow specific UIDs to run.

**How to reproduce it** (as minimally and precisely as possible):

Run a pod that explicitly specifies a non 0 uid to run.

**Anything else we need to know**:

IMO, PodSecurityPolicies are a *much* better security mechanism than SecurityContextDeny, and SecurityContextDeny should not be a default. The same effect can be achieved with a restrictive PSP by default instead, if desired.

JunSun17:
@yuvipanda Please refer to this comment (https://github.com/Azure/acs-engine/pull/1989#issuecomment-358526373) as a way to disable SecurityContextDeny setting for now.

I am reading into documents about PSP to see if removing SecurityContextDeny would provide better customer experience. Will update shortly on this. Thanks!

PR #2125, opened 2018-01-23
Remove SecurityContextDeny setting in API server admission control.

JunSun17:
**What this PR does / why we need it**: Remove SecurityContextDeny setting in API server admission control since it introduces friction for users to apply workloads with security requirement. This issue should be addressed later by adopting a default allow-all Pod Security Policy, see issue #2123.

**Which issue this PR fixes**: #2109

### DIFF
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -82,7 +82,7 @@ func setAPIServerConfig(cs *api.ContainerService) {
 
 	// Default apiserver config
 	defaultAPIServerConfig := map[string]string{
-		"--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages,SecurityContextDeny",
+		"--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages",
 		"--authorization-mode": "Node",
 	}
 


jackfrancis:
@pidah Should we simply move this to be another `--admission-control` addition when `EnablePodSecurityPolicy` is set to true?

Currently when `EnablePodSecurityPolicy` is set to true we add `",PodSecurityPolicy"` to the list of `--admission-control` options.

JunSun17:
@jackfrancis Ideally, we should use PSP, and I have tested it working at some level. This work will require some time to investigate and test, and I do not see it is very urgent. The proposed plan is documented in Issue #2123.

For now, I think the quickest way to unblock this issue is to remove the SecurityContextDeny setting.

jackfrancis:
Fair enough @JunSun17. @pidah FYI we're removing `,SecurityContextDeny` as a default option for `--admission-control` to address a larger set of default cluster cases. For your own clusters if you need this option make sure to include it in the `apiServerConfig` for your api model definitions.

pidah:
@jackfrancis @JunSun17 PodSecurityPolicy and SecurityContextDeny effectively do the same thing and were not intended to be used together. However, PSP is easier to work with, as it offers greater flexibility when defining varying security policy defaults.

campbelldgunn:
@jackfrancis @pidah I got caught by this when trying to install jenkins into its own namespace. The pod was basically locked out. I removed the entry `SecurityContextDeny` from the apiserver and that did not work. it required me to add the entry in the template of
```
"orchestratorProfile": {
      "orchestratorRelease": "1.8",
      "kubernetesConfig": {
        "apiServerConfig": {
          "--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota"
        }
      }
    },
```
And rebuild the cluster. Now everything is working fine.

jackfrancis:
@shaharyakir At present, acs-engine does not support configuration-level operations on an existing cluster. The way to do this is to ssh into the nodes and manually change / restart daemons.

@campbelldgunn Thanks for adding your experience. Can you recommend any changes in the way any of acs-engine's configuration opinions are implemented, or is that just an FYI?

PR #2138, opened 2018-01-24
--authorization-mode=Node only if secure kubelet

jackfrancis:
**What this PR does / why we need it**: If we turn off secure kublet communications, we need to not set `--authorization-mode` in the API server.

Also cleaned up related unit tests

**Release note**:

```
--authorization-mode=Node only if secure kubelet
```


jackfrancis:
@CecileRobertMichon @JunSun17 Kindly review and, if lgtm, merge this today.

PR #2144, opened 2018-01-24
Upgrade docker-engine to 1.13.* for all Kubernetes clusters >= v1.7

jackfrancis:
**What this PR does / why we need it**: Update to 1.13.* docker-engine release channel by default for k8s versions >= 1.7.0

**Release note**:

```
Update to 1.13.* docker-engine release channel by default for k8s versions >= 1.7.0
```


JunSun17:
@jackfrancis I noticed you changed docker version for only one minor version in each 1.7, 1.8 and 1.9 release? Is that the minor version we used in the E2E tests? Most tests are already passed, and I just want to make sure they are really testing with the new docker 1.3.

jackfrancis:
That’s right. The tests should print to stdout the exact ver they are testing.

JunSun17:
Got it, looks all good to me. Will sign off after you update all docker versions for 1.7, 1.8 and 1.9.

PR #2155, opened 2018-01-25
Keyvault etcd certs

### LINKED ISSUE #2102: Unable to reference 'etcdPeerCertificates' and 'etcdPeerPrivateKeys' from KeyVault in apimodel.json
Yes

---

ISSUE

---

v0.12.1

---

Kubernetes

**What happened**:
If I reference the etcd peer certs and keys in apimodel.json (at `.properties.certificateProfile.etcdPeerCertificates` and `.properties.certificateProfile.etcdPeerPrivateKeys`) using a KeyVault, the certificate values are not fetched from KeyVault. Instead the path to the secret in KeyVault is passed on to the masters.

This means that the etcd on master doesn't start properly and hence the cluster doesn't work.

It looks like the issue lies somewhere in acs-engine because the parameters file generated after `acs-engine generate` doesn't have the KeyVault references, instead they are the base64 encoded paths to secret in KeyVault.

Secrets referred via KeyVault should work for `etcdPeerCertificates` and `etcdPeerPrivateKeys` as well.

In the attached [apimodel.zip](https://github.com/Azure/acs-engine/files/1646963/apimodel.zip), replace the `--REDACTED--` and KeyVault paths in certificateProfile section to valid values and try to deploy.
Once master VMs are created check for content of `/etc/kubernetes/certs/etcdpeer0.crt` or `/etc/kubernetes/certs/etcdpeer0.key`. It will be the KeyVault path to secret instead of secret's value from KeyVault. Etcd service should be failed and error can be seen in `journalctl -xe`.

CecileRobertMichon:
**What this PR does / why we need it**: etcd tls added new certs to acs-engine. Changing some of the design to allow use of Keyvault certs.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2102 

**Release note**:

```release-note
```


pkg/acsengine/engine.go
```diff
@@ -506,8 +506,12 @@ func getParameters(cs *api.ContainerService, isClassicMode bool, generatorCode s
 			addSecret(parametersMap, "etcdServerPrivateKey", properties.CertificateProfile.EtcdServerPrivateKey, true)
 			addSecret(parametersMap, "etcdClientCertificate", properties.CertificateProfile.EtcdClientCertificate, true)
 			addSecret(parametersMap, "etcdClientPrivateKey", properties.CertificateProfile.EtcdClientPrivateKey, true)
-			addArraySecret(parametersMap, "etcdPeerCertificates", properties.CertificateProfile.EtcdPeerCertificates, true)
-			addArraySecret(parametersMap, "etcdPeerPrivateKeys", properties.CertificateProfile.EtcdPeerPrivateKeys, true)
+			for i, pc := range properties.CertificateProfile.EtcdPeerCertificates {
```
CecileRobertMichon: Since this parameter is only relevant for clusters with 1+ masters, would this need some kind of "!HostedMaster" guard?

jackfrancis: how about `properties.MasterProfile != nil`?

PR #2160, opened 2018-01-26
Protect etcd tls from race conditions

### LINKED ISSUE #2162: Race condition in master setup
No
---

Issue
---

master

---

Kubernetes 1.9.2

**What happened**:
Cluster with three masters fails to deploy. Some masters (different number each deploy) fail with 

```
statusMessage:{"status":"Failed","error":{"code":"ResourceOperationFailure","message":"The resource operation completed with terminal provisioning state 'Failed'.","details":[{"code":"VMExtensionProvisioningError","message":"VM has reported a failure when processing extension 'cse2'. Error message: \"Enable failed: failed to execute command: command terminated with exit status=3\n[stdout]\n\n[stderr]\n\"."}]}}
```

Looking at the logs in /var/log/azure/cluster-provision.log, the masters that fail to come up drop out of the script [here](https://github.com/Azure/acs-engine/blob/9d01adcb14aa20f0b5c72703ccf980b0ad436164/parts/k8s/kubernetesmastercustomscript.sh#L562). However, this does not seem to be the root cause.

Investigating further, etcd never starts up, which turns out to be because /etc/kubernetes/certs/etcdserver.key and etcdpeer(number).key are owned by root. This is set [here](https://github.com/Azure/acs-engine/blob/9d01adcb14aa20f0b5c72703ccf980b0ad436164/parts/k8s/kubernetesmastercustomscript.sh#L100). On the machines that successfully start, the keys are owned by etcd, so there seems to be some race condition here. It seems the correct permissions are set by [cloud-init](https://github.com/Azure/acs-engine/blob/9d01adcb14aa20f0b5c72703ccf980b0ad436164/parts/k8s/kubernetesmastercustomdata.yml#L364)? I can't easily see how these two interact, so the race condition is a guess from available data.

Anyway, is there a good reason for not setting etcd as the owner in the customscript.sh also?

Successful deploy

This took a bit longer to diagnose than necessary since the previously run function [ensureEtcd](https://github.com/Azure/acs-engine/blob/9d01adcb14aa20f0b5c72703ccf980b0ad436164/parts/k8s/kubernetesmastercustomscript.sh#L569) does not exit the script if it fails. Shouldn't it?

CecileRobertMichon:
**What this PR does / why we need it**: In some rare cases, etcd certs were not generated by the time custom data was changing the owner of the key files needed by etcd to start. This sets the owner for the appropriate keys in the custom script when key files are created.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2162 

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -165,24 +165,24 @@ coreos:
         ExecStart=/opt/azure/containers/provision-setup.sh
 {{else}}
 runcmd:
+- retrycmd_if_failure() { for i in $(seq 1 36); do $@; [ $? -eq 0  ] && break || sleep 5; done ; }
 - echo `date`,`hostname`, startruncmd>>/opt/m
 - apt-mark hold walinuxagent{{GetKubernetesAgentPreprovisionYaml .}}
 - echo `date`,`hostname`, preaptupdate>>/opt/m
-- apt-get update
+- retrycmd_if_failure apt-get update
 - echo `date`,`hostname`, postaptupdate>>/opt/m
-- apt-get install -y apt-transport-https ca-certificates nfs-common
+- retrycmd_if_failure apt-get install -y apt-transport-https ca-certificates nfs-common
 - echo `date`,`hostname`, aptinstall>>/opt/m
 - systemctl enable rpcbind
 - systemctl enable rpc-statd
 - systemctl start rpcbind
 - systemctl start rpc-statd
 - echo `date`,`hostname`, predockerinstall>>/opt/m
-- for i in 1 2 3 4 5; do curl --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -; [ $? -eq 0 ] && break || sleep 5; done
+- for i in 1 2 3 4 5; do curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -; [ $? -eq 0 ] && break || sleep 5; done
 - echo "deb {{WrapAsVariable "dockerEngineDownloadRepo"}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
 - "echo \"Package: docker-engine\nPin: version {{WrapAsVariable "dockerEngineVersion"}}\nPin-Priority: 550\n\" > /etc/apt/preferences.d/docker.pref"
-- apt-get update
-- apt-get install -y ebtables
-- apt-get install -y docker-engine
+- retrycmd_if_failure apt-get update
+- retrycmd_if_failure apt-get install -y ebtables docker-engine
 - echo "ExecStartPost=/sbin/iptables -P FORWARD ACCEPT" >> /etc/systemd/system/docker.service.d/exec_start.conf
 - systemctl daemon-reload
 - echo `date`,`hostname`, postdockerinstall>>/opt/m
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -178,7 +178,7 @@ runcmd:
 - systemctl start rpcbind
 - systemctl start rpc-statd
 - echo `date`,`hostname`, predockerinstall>>/opt/m
-- for i in 1 2 3 4 5; do curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -; [ $? -eq 0 ] && break || sleep 5; done
+- retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
 - echo "deb {{WrapAsVariable "dockerEngineDownloadRepo"}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
 - "echo \"Package: docker-engine\nPin: version {{WrapAsVariable "dockerEngineVersion"}}\nPin-Priority: 550\n\" > /etc/apt/preferences.d/docker.pref"
 - retrycmd_if_failure apt-get update
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -345,7 +345,8 @@ coreos:
         ExecStart=/opt/azure/containers/provision-setup.sh
 {{else}}
 runcmd:
-- retrycmd_if_failure() { for i in {1..5}; do $@; [ $? -eq 0  ] && break || sleep 5; done ; }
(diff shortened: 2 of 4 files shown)


gmagniez:
Hi,
At line 302, " passwd -u "$etcd" > /dev/null" is also incorrect and generates an error, should be " passwd -u "etcd" > /dev/null"

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -165,24 +165,24 @@ coreos:
...
-- apt-get install -y apt-transport-https ca-certificates nfs-common
+- retrycmd_if_failure apt-get install -y apt-transport-https ca-certificates nfs-common
 - echo `date`,`hostname`, aptinstall>>/opt/m
 - systemctl enable rpcbind
 - systemctl enable rpc-statd
 - systemctl start rpcbind
 - systemctl start rpc-statd
 - echo `date`,`hostname`, predockerinstall>>/opt/m
-- for i in 1 2 3 4 5; do curl --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -; [ $? -eq 0 ] && break || sleep 5; done
+- retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
```
JunSun17: There are two retry loops, do we need both? Also applies to similar code below.

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -165,24 +165,24 @@ coreos:
         ExecStart=/opt/azure/containers/provision-setup.sh
 {{else}}
 runcmd:
+- retrycmd_if_failure() { for i in $(seq 1 36); do $@; [ $? -eq 0  ] && break || sleep 5; done ; }
```
JunSun17: Just curious, why we pick 36?

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -165,24 +165,24 @@ coreos:
...
 - echo `date`,`hostname`, predockerinstall>>/opt/m
-- for i in 1 2 3 4 5; do curl --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -; [ $? -eq 0 ] && break || sleep 5; done
+- retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
 - echo "deb {{WrapAsVariable "dockerEngineDownloadRepo"}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
 - "echo \"Package: docker-engine\nPin: version {{WrapAsVariable "dockerEngineVersion"}}\nPin-Priority: 550\n\" > /etc/apt/preferences.d/docker.pref"
-- apt-get update
-- apt-get install -y ebtables
-- apt-get install -y docker-engine
+- retrycmd_if_failure apt-get update
+- retrycmd_if_failure apt-get install -y ebtables docker-engine
```
JunSun17: What will happen if ebtables is installed but docker-engine failed. Just wonder if we should put them to two separate retries...

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -292,22 +292,17 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
...
     set -x
+    retrycmd_if_failure() { for i in $(seq 1 36); do $@; [ $? -eq 0  ] && break || sleep 5; done ; }
     ETCD_VER=v{{WrapAsVariable "etcdVersion"}}
     DOWNLOAD_URL={{WrapAsVariable "etcdDownloadURLBase"}}
     mkdir -p /tmp/etcd-download
-    curl -L ${DOWNLOAD_URL}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
+    retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 -L ${DOWNLOAD_URL}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
     tar xzvf /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz -C /usr/bin/ --strip-components=1
-    useradd -U "etcd"
+    useradd -U "etcd" > /dev/null 2>&1
```
JunSun17: Why do we want to discard all errors and outputs?

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -272,7 +273,7 @@ function configNetworkPolicy() {
 function installClearContainersRuntime() {
 	# Add Clear Containers repository key
 	echo "Adding Clear Containers repository key..."
-	curl -sSL "https://download.opensuse.org/repositories/home:clearcontainers:clear-containers-3/xUbuntu_16.04/Release.key" | apt-key add -
+	curl -sSL --retry 5 --retry-delay 10 --retry-max-time 30 "https://download.opensuse.org/repositories/home:clearcontainers:clear-containers-3/xUbuntu_16.04/Release.key" | apt-key add -
```
JunSun17: not related to this PR, but are we ok to pull data from open internet? I kind of get the impression that we put a copy in Azure and download from there.

parts/k8s/kubernetesagentcustomdata.yml
CecileRobertMichon: 5 * 36 = 180s = 3 minutes, this is a generous timeout to ensure custom script had time to run. I'll have a follow up PR to parametrize this and clean up the runcmd in general.

parts/k8s/kubernetesmastercustomdata.yml
CecileRobertMichon: Because the etcd user gets created in either custom script or custom data, whichever gets there first and we want to ignore errors if the user already exists

parts/k8s/kubernetesmastercustomscript.sh
CecileRobertMichon: @jessfraz might be able to answer, but this is out of scope for this PR since I'm only adding retries to all the curl commands.

parts/k8s/kubernetesagentcustomdata.yml
CecileRobertMichon: Good question, this was taken directly from https://github.com/Azure/acs-engine/pull/2182, I'm merging the two since we had overlap. From what I understand (just tested this), apt-get ignores already installed packages so if one gets installed but not the other the retry will only try to install the one missing.

parts/k8s/kubernetesagentcustomdata.yml
CecileRobertMichon: Again, taken from #2182 but I assume the two retries are complementary. @jackfrancis can you confirm / correct me?

jackfrancis: There is only one loop, but we're additionally using curl's retry option. I think this is O.K. for now.

PR #2163, opened 2018-01-26
Increase timeout for etcd certificates to become available

### LINKED ISSUE #2162: Race condition in master setup

duga3:
**What this PR does / why we need it**: Increases timeout when waiting for etcd certificates to become available.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2162

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -306,7 +306,8 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
         if [ -e /opt/azure/containers/etcdcerts.complete ]; then
             break
         fi
-        sleep 1
+        echo "Waiting for etcdcerts to be completed...($i)"
+        sleep 5
     done
     sudo sed -i "1iETCDCTL_ENDPOINTS=https://127.0.0.1:2379" /etc/environment
     sudo sed -i "1iETCDCTL_CA_FILE={{WrapAsVariable "etcdCaFilepath"}}" /etc/environment
--- a/parts/k8s/kubernetesmastercustomscript.sh
+++ b/parts/k8s/kubernetesmastercustomscript.sh
@@ -565,15 +565,22 @@ function ensureApiserver() {
 }
 
 function ensureEtcd() {
+    etcdStarted=1
     for i in {1..600}; do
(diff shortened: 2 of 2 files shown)


duga3:
Closing in favor of PR #2160

PR #2164, opened 2018-01-26
Adding francecentral to the list of certificate fqdn

serbrech:
Add francecentral to the list of fqdn handled by the generated certs.
France Central region is in preview right now.

I had to add `.decode('utf-8')` to get the string to be handled properly by the json library. this might be because I ran on bash.exe. I'm hoping that it will not cause problems in other environment, and will be compatible.

**Release note**:

```release-note
Add Central France region to the list of fqdn handled by the generated certificate.
```

### DIFF
--- a/pkg/acsengine/Get-AzureConstants.py
+++ b/pkg/acsengine/Get-AzureConstants.py
@@ -7,11 +7,11 @@ import json
 time = strftime("%Y-%m-%d %H:%M:%S", gmtime())
 
 def getAllSizes():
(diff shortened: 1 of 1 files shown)


jackfrancis:
Thanks for this @serbrech! As a rule we like to have only project admins run the const generation script and make changes to `pkg/acsengine/azureconst.go`. Would you mind cleaning up that file so it doesn't carry any changes in this PR, then we'll test the changes to the script, and if it looks good, we'll file a follow-up PR immediately after merging this one with the applied changes to `pkg/acsengine/azureconst.go`.

Let me know if that doesn't make sense!

PR #2165, opened 2018-01-26
Mount in /var/lib/cni from the host.

brendandburns:
Without this PR on kubelet restart /var/lib/cni is recreated, this causes kubenet to lose its database and may cause it to hand out duplicate IP addresses.


rjtsdl:
LGTM

PR #2173, opened 2018-01-30
Allow 1 core master node VM sizes

dtzar:
This allows acs-engine to provision D(S)1 series VM sizes for Kubernetes master nodes.  Per the [Kubernetes cluster size guide](https://kubernetes.io/docs/admin/cluster-large/), ~ 1 core + 3.5 GB of RAM is the recommended sizing for K8s clusters with 1-5 agent nodes.  I successfully deployed a 3-node master K8s cluster with the `Standard_DS1_v2` size using acs-engine and it worked with no problems.  I didn't test the other two D1 sizes, but assume it should work also considering those are almost identical in hardware specs.

```
Allow DS1 series VMs for Kubernetes master nodes
```


jackfrancis:
@dtzar Thanks for this! We actually generate `pkg/acsengine/azureconst.go` programmatically using the `pkg/acsengine/Get-AzureConstants.py` script. Do you wanna dig into that script to see why the three SKUs you're looking to add are not getting whitelisted?

dtzar:
@jackfrancis please go ahead and double check me, but pretty sure that last commit should fix the auto-generation.  I suspect that moving to 1 core will add additional values than I did on the first commit.

jackfrancis:
This looks good. Let's get rid of the changes to the const file itself for this PR, and then I'll submit a follow-up PR with the const changes.

PR #2180, opened 2018-01-31
Adding ServiceNodeExclusion as a default flag for Controller Manager

bnookala:

Enables `ServiceNodeExclusion` as a default on controller manager to allow virtual kubelet nodes (ie ACI Connector) to play nice with Load Balancers on Kubernetes.

See: https://github.com/virtual-kubelet/virtual-kubelet/issues/41 for more details. 

@jackfrancis - this is the change I had mentioned I was going to make via slack. 

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/acsengine/defaults-controller-manager.go
+++ b/pkg/acsengine/defaults-controller-manager.go
@@ -64,6 +64,9 @@ func setControllerManagerConfig(cs *api.ContainerService) {
 		}
 	}
 
+	// Enables Node Exclusion from Services (toggled on agent nodes by the alpha.service-controller.kubernetes.io/exclude-balancer label).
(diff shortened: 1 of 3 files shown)


jackfrancis:
@bnookala you can run gofmt -s against the following files to fix linting errors:

- `pkg/acsengine/defaults-controller-manager.go`
- `pkg/acsengine/defaults-kubelet.go`
- `pkg/acsengine/defaults.go`

bnookala:
@jackfrancis oops - should be fixed now.

jackfrancis:
@bnookala Just to confirm the functionality before I merge this: the intention here is to add this flag only for >= 1.9 clusters, correct?

bnookala:
@jackfrancis Yes. As far as the version requirement, I noted that a function, `addDefaultFeatureGates` has a minimum k8s cluster version requirement - so I'm using that to add the flag here: https://github.com/Azure/acs-engine/pull/2180/files#diff-009cf3e2e37f5d4cce79e3d32311670aR68

PR #2183, opened 2018-02-01
update prometheus-grafana addon

martell:

Updates the prometheus-grafana-addon

Some parts of the addon install already gets the latest images which means they depend on prometheus 2.x.
Currently it is in a broken state because of this with the following error
`level=error msg="flag provided but not defined: -storage.tsdb.path" source="main.go:74"`
So lets just bump to 2.x and have this working again.

To support 1.9.0 we need a new version fo kube-state-metrics. (v1.2.0)

### DIFF
--- a/extensions/prometheus-grafana-k8s/v1/prometheus_values.yaml
+++ b/extensions/prometheus-grafana-k8s/v1/prometheus_values.yaml
@@ -17,7 +17,7 @@ alertmanager:
   ##
   image:
     repository: prom/alertmanager
-    tag: v0.9.1
+    tag: v0.13.0
     pullPolicy: IfNotPresent
 
   ## Additional alertmanager container arguments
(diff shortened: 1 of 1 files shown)


trstringer:
@martell were your tests successful? I'm seeing some regressions here that appear like they could be related to bumping the major version but keeping the pre-existing config:

```
level=error ts=2018-02-01T20:42:57.401831538Z caller=main.go:579 err="Error loading config couldn't load configuration (--config.file=/etc/config/prometheus.yml): parsing YAML file /etc/config/prometheus.yml: yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `rule_fi...` into config.plain"
level=info ts=2018-02-01T20:42:57.401880038Z caller=main.go:581 msg="See you next time!"
```

cc // @ritazh

martell:
@trstringer It was in a broken state already so I just bumped the version in my initial commit to see if it fixed the issue, It seems we do not have real ci tests for the extensions.
I wasn't expecting it to be merged right away :)

It seems that since 2.1 `|-` is no longer liked in the yaml file.
I updated my branch to resolve this.
If someone could confirm that would be great.

PR #2184, opened 2018-02-01
Add additional v3 vm sizes permitted for dcos master and agent

yakman2020:
Current DCOS templates are not allowed to use the VM sizes like D2s_v3. We very much need those.

**Release note**:

```release-note
```


jackfrancis:
@yakman2020 These const values are auto-generated, so we'll have to update the script at `pkg/acsengine/Get-AzureConstants.py`. The reason why `Standard_D2s_v3` isn't available is because the `resourceDiskSizeInMb` value for that SKU doesn't meet the minimum requirement defined in the script above. Here's the output for that SKU from `az vm list-sizes -l eastus` (using eastus as an example):

```
{
    "maxDataDiskCount": 4,
    "memoryInMb": 8192,
    "name": "Standard_D2s_v3",
    "numberOfCores": 2,
    "osDiskSizeInMb": 1047552,
    "resourceDiskSizeInMb": 16384
  },
```

Do we want to to decrease the value of this constant here?:

https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/Get-AzureConstants.py#L23

(That will inevitably bring in a bunch more master SKU sizes into the whitelist, I assume that's O.K.)

yakman2020:
Yup. We really only need 30-40Gb for basic DCOS, though the users may disagree. I would try 80GB initially

jackfrancis:
Standard_D2s_v3 looks to only expose 16GB, so changing the minimum from ~100GB to ~80GB will still exclude Standard_D2s_v3.

yakman2020:
So, I don't see how that could be correct. I'm deploying on D2s_v3 and they have 32GB main memory and a 128 gb system disk. Something is wrong with that data, albeit I'm specifying the size of the OS disk.

yakman2020:
not correct. OK. My D3s instance is 40GB  system disk. That is fine. Presumably I can specify larger in the deployment.

PR #2189, opened 2018-02-02
add default audit policy

pidah:
This PR adds a default audit policy to enable advanced audit logging as described here: 
https://kubernetes.io/docs/tasks/debug-application-cluster/audit/

### DIFF
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -14,9 +14,6 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		"--advertise-address":          "<kubernetesAPIServerIP>",
 		"--allow-privileged":           "true",
 		"--anonymous-auth":             "false",
-		"--audit-log-maxage":           "30",
-		"--audit-log-maxbackup":        "10",
-		"--audit-log-maxsize":          "100",
 		"--audit-log-path":             "/var/log/audit.log",
 		"--insecure-port":              "8080",
 		"--secure-port":                "443",
@@ -87,8 +84,11 @@ func setAPIServerConfig(cs *api.ContainerService) {
 
 	// Default apiserver config
 	defaultAPIServerConfig := map[string]string{
-		"--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages",
-		"--authorization-mode": "Node",
+		"--admission-control":   "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,DenyEscalatingExec,AlwaysPullImages",
+		"--authorization-mode":  "Node",
+		"--audit-log-maxage":    "30",
+		"--audit-log-maxbackup": "10",
+		"--audit-log-maxsize":   "100",
 	}
 
 	// RBAC configuration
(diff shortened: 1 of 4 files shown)


pkg/acsengine/defaults-apiserver.go
```diff
@@ -17,7 +17,7 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		"--audit-log-maxage":           "30",
 		"--audit-log-maxbackup":        "10",
 		"--audit-log-maxsize":          "100",
-		"--audit-log-path":             "/var/log/apiserver/audit.log",
+		"--audit-log-path":             "/var/log/audit.log",
```
jackfrancis: @pidah what's the reason for this change?

pidah: @jackfrancis /var/log/apiserver directory does not exist

jackfrancis: oy

jackfrancis:
@lachie83 @brendanburns Any thoughts on anticipated negative side-effects of turning this on automatically for 1.8 and newer clusters? Should we make this an opt-in feature (default: off)?

lachie83:
I would like to watch the size of this file over time. How verbose is it? Given our stance to enable RBAC by default I'm inclined to say that we should enable it by default so long as RBAC is enabled and we rotate/compress the audit log.

pidah:
@jackfrancis 👍

yastij:
@pidah @jackfrancis - is there currently any possiblity to override the proposed policy ? Also the audit file contains some ununsed rules (as some higher order rules are used instead) cc @feiskyer

pidah:
@yastij atm you can't override the proposed policy. Maybe you can recommend a better default policy ? cc @jackfrancis

arnaudmz:
@pidah What about having a host mount to externalize logs?

They currently seem stuck into the APIServer container.

jackfrancis:
@yastij We could move the `--audit-policy-file` apiserver config to be user-overridable. Are you interested in simply turning it off? Could you test on acluster that `--audit-policy-file=""` (empty string) runtime apiserver config is a valid way to turn this off?

yastij:
@jackfrancis - we are interested in overriding the provided rules, currently there's no known workaround for this ?

jackfrancis:
Current (like right now) workaround would be to look for `--audit-policy-file` in your generated templates and replace the value "/etc/kubernetes/manifests/audit-policy.yaml" with the value "".

If you can wait until this afternoon, we should be able to merge a fix by then. @pidah Are there any other considerations that I'm not taking into account? Would simply removing the `--audit-policy-file` reference have negative side-effects without also doing other configuration hygiene?

yastij:
@jackfrancis - we want `--audit-policy-file` to be user configurable, overriding the generated value with "" won't work as it won't capture our audit-policy file.

jackfrancis:
Understood @yastij. What's your current process of delivering your custom audit policy `.yaml` to the cluster hosts? Are you doing that manually? (or programmatically, but outside of the scope of acs-engine?)

yastij:
currently we've just started on it, to me the right thing would be to provide an acs-flag to point the .yaml and let acs do the work.

jackfrancis:
I agree @yastij that a next step would be moving beyond the static template that audit policy currently uses. That's a non-trivial amount of work, but I've created an issue placeholder here:

https://github.com/Azure/acs-engine/issues/2298

In the meantime/short term we could evolve the existing static policy file if there is consensus as to an improved "default"; and if we need true customization we'll have to do that as a manual, post-deployment step.

yastij:
sure we can tweak the the audit-policy rules.

cc @arnaudmz @khaldoune

jackfrancis:
@pidah I'm just observing now this on a 1.8.9 and 1.9.4 cluster in journalctl kubelet:

`file.go:149] Can't process manifest file "/etc/kubernetes/manifests/audit-policy.yaml": /etc/kubernetes/manifests/audit-policy.yaml: couldn't parse as pod(no kind "Policy" is registered for version "audit.k8s.io/v1beta1"), please check manifest file.`

PR #2190, opened 2018-02-02
pulling down provision logs during e2e runs

jackfrancis:
This PR pulls down provision-related log files in to `_logs/`, and disables deployment retries by default.

While running E2E on this PR, because of the logs, I was able to identify a bug, the fix is included in this PR as a demonstration of the power of logs!

**Release note**:

```
NONE
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -363,7 +363,7 @@ runcmd:
 - retrycmd_if_failure curl --cacert /etc/kubernetes/certs/ca.crt --cert /etc/kubernetes/certs/etcdclient.crt --key /etc/kubernetes/certs/etcdclient.key --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 "{{WrapAsVerbatim "variables('masterEtcdClientURLs')[copyIndex(variables('masterOffset'))]"}}"/v2/machines
 - retrycmd_if_failure apt-get update
 - retrycmd_if_failure apt-get install -y apt-transport-https ca-certificates
-- retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
+- curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
 - echo "deb {{WrapAsVariable "dockerEngineDownloadRepo"}} ubuntu-xenial main" | sudo tee /etc/apt/sources.list.d/docker.list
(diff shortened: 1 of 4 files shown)


test/e2e/runner.go
```diff
@@ -84,6 +88,7 @@ func main() {
 			Config:            engCfg,
 			ClusterDefinition: cs,
 		}
+		cliProvisioner.Engine = eng
```
CecileRobertMichon: What is this doing? Does it assign it to null if the name of the cluster was provided?

CecileRobertMichon: Ah nevermind I just realized this was in the else block

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -719,3 +719,4 @@ fi
 echo `date`,`hostname`, endscript>>/opt/m
 
 mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
+ps auxfww > /opt/azure/provision-ps.log
```
CecileRobertMichon: Is there a reason why this is in /opt/azure and not /opt/azure/containers?

test/e2e/runner.go
jackfrancis: It adds the ClusterDefinition (parsed api model) to the cliProvisioner instance, which is used to form ssh connection config to the master. And all of this convoluted stuff was done so that in cases where there is no provision (i.e., you're testing an existing cluster) we still have all the cluster metadata in that instance object during teardown (in this case not tearing down the cluster but grabbing logs).

parts/k8s/kubernetesmastercustomscript.sh
jackfrancis: Not really, no!

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -363,7 +363,7 @@ runcmd:
 - retrycmd_if_failure curl --cacert /etc/kubernetes/certs/ca.crt --cert /etc/kubernetes/certs/etcdclient.crt --key /etc/kubernetes/certs/etcdclient.key --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 "{{WrapAsVerbatim "variables('masterEtcdClientURLs')[copyIndex(variables('masterOffset'))]"}}"/v2/machines
 - retrycmd_if_failure apt-get update
 - retrycmd_if_failure apt-get install -y apt-transport-https ca-certificates
-- retrycmd_if_failure curl --retry 5 --retry-delay 10 --retry-max-time 30 --max-time 60 -fsSL https://aptdocker.azureedge.net/gpg | apt-key add -
```
jackfrancis: This was never retrying, as the curl error output was being redirected to `apt-key add -`, which was in turn adding that text as a key (oy!)

PR #2196, opened 2018-02-03
Network validation checks during provision

jackfrancis:
**What this PR does / why we need it**: Add DNS + HTTPS checks, capture DNS packets, rationalize som retry. E2E changes to enable log fetching on all master/agent nodes.

**Release note**:

```
Network validation checks during provision
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomscript.sh
+++ b/parts/k8s/kubernetesmastercustomscript.sh
@@ -595,14 +595,16 @@ function ensureEtcdDataDir() {
         return
     else
         echo "/var/lib/etcddisk was not found at /dev/sdc1. Trying to mount all devices."
-        for i in {1..300}; do
+        s = 5
+        for i in {1..60}; do
             sudo mount -a && mount | grep /dev/sdc1 | grep /var/lib/etcddisk;
             if [ "$?" = "0" ]
             then
-                echo "/var/lib/etcddisk mounted at: /dev/sdc1, took $i seconds"
+                (( t = ${i} * ${s} ))
+                echo "/var/lib/etcddisk mounted at: /dev/sdc1, took $t seconds"
                 return
             fi
-            sleep 1
+            sleep $s
         done
     fi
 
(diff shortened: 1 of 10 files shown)


parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -23,6 +23,13 @@
 # KUBECONFIG_KEY ETCD_SERVER_CERTIFICATE ETCD_SERVER_PRIVATE_KEY ETCD_CLIENT_CERTIFICATE ETCD_CLIENT_PRIVATE_KEY
 # ETCD_PEER_CERTIFICATES ETCD_PEER_PRIVATE_KEYS ADMINUSER MASTER_INDEX
 
+# Capture Interesting Network Stuffs during provision
+packetCaptureProvision() {
+    tcpdump -G 600 -W 1 -n -vv -w /var/log/azure/dnsdump.pcap -Z root -i eth0 udp port 53 > /dev/null 2>&1 &
```
amanohar: I would also recommend adding this during cloudinit if possible just to see what the differences are between when cloudinit runs and when CSE runs.

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -23,6 +23,13 @@
...
+}
+
+packetCaptureProvision
```
JunSun17: Not sure how it will impact on the performance side, should it be put behind some debug flag?

CecileRobertMichon: +1, does this significantly increase deployment duration?

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -585,14 +595,14 @@ function ensureEtcdDataDir() {
...
+        for i in {1..300}; do
             sudo mount -a && mount | grep /dev/sdc1 | grep /var/lib/etcddisk;
             if [ "$?" = "0" ]
             then
-                echo "/var/lib/etcddisk mounted at: /dev/sdc1"
+                echo "/var/lib/etcddisk mounted at: /dev/sdc1, took $i seconds"
                 return
             fi
-            sleep 5
+            sleep 1
```
CecileRobertMichon: What's the reason for reducing sleep to 1 here (trying 300 times in 300s vs trying 60 times in 300 seconds) ?

jackfrancis: Lazier way for me to communicate seconds.

parts/k8s/kubernetesmastercustomscript.sh
jackfrancis: This will not have an observable impact on deployment time (unless there is a DNS storm, in which case we might want that data :))

parts/k8s/kubernetesmastercustomscript.sh
jackfrancis: Will push an improvement.

PR #2204, opened 2018-02-05
E2E nginx outbound access test: simplify port test

jackfrancis:
**Release note**:

```
NONE
```


jackfrancis:
@CecileRobertMichon @JunSun17 This PR simplifies the "outbound internet access from nginx pod" test: it will succeed on any non-nil err response from curl. Additionally, it has a fallback test in case bing.com is non-responsive.

PR #2205, opened 2018-02-05
replaced apierror with armerror

dmitsh:
Addressing comments on https://github.com/Azure/acs-engine/commit/a19784e6f1cb8e5883c0cbc8a646464315cf3152

Replaced apierror with armerror as apierror contains RP-related info


pkg/armhelpers/deploymentError.go
```diff
@@ -85,45 +72,36 @@ func DeployTemplateSync(az ACSEngineClient, logger *logrus.Entry, resourceGroupN
...
+	// Check that ARM returned an error
 	if armErr == nil || len(armErr.Message) == 0 || len(armErr.Code) == 0 {
-		logger.Warn("Not an ARM Response")
-		return &apierror.Error{
-			Code:     apierror.InternalOperationError,
-			Message:  depErr.Error(),
-			Category: apierror.InternalError}
+		logger.Warn("Not an ARM Error")
+		return &ArmError{
+			Code:    "DeploymentFailed",
```
amanohar: What is the difference between `ProvisioningFailed` and `DeploymentFailed`. Do we really care/need this distinction? Everything should be in `depErr.Error()`.

pkg/armhelpers/deploymentError.go
```diff
@@ -7,12 +7,39 @@ import (
...
 	"github.com/sirupsen/logrus"
 )
 
-func parseDeploymentOperation(logger *logrus.Entry, operation resources.DeploymentOperation) (*apierror.Error, error) {
+// ArmError is the OData v4 format, used by the RPC and
+// will go into the v2.2 Azure REST API guidelines
+type ArmError struct {
+	Code    string     `json:"code"`
+	Message string     `json:"message"`
+	Target  string     `json:"target,omitempty"`
```
amanohar: I don't think these and `Code`` are needed because `depErr.Error()` should have everything that's required. RP can deserialize it and ACS E can just return array to upgrade cmd that can then print it to stdout.

dmitsh: No, this is not the case. ArmError is parsing ARM response. depErr.Error is a separate object

pkg/armhelpers/deploymentError.go
dmitsh: Sure, we can use same error

PR #2217, opened 2018-02-06
Kubernetes Tiller Addon: configuration to set max-history

ultimateboy:
Adds a new addon configuration for the kubernetes tiller addon to set the max-history environment variable

Without this configuration, tiller will store an infinite number of revisions for each release. As of Helm 2.7, setting this configuration to a non 0 value will force tiller to clean up old revisions, thus limiting the number of configmaps under its control. This environment variable has no impact on previous versions of Helm.

I also modified the default e2e test to set this configuration and then verify that the correct environment variable ends up on the tiller pod.

### DIFF
--- a/examples/e2e-tests/kubernetes/release/default/definition.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition.json
@@ -4,7 +4,16 @@
     "orchestratorProfile": {
       "orchestratorType": "Kubernetes",
       "kubernetesConfig": {
-        "clusterSubnet": "10.239.0.0/16"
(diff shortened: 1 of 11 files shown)


ultimateboy:
Currently trying to debug the windows test by modifying the test to fail so that I could see the output. It does indeed appear that the environment variable is not being correctly set in the windows tiller manifest. More investigation is needed.

ultimateboy:
Managed to figure out the windows tests... sadly, it was very simple... I had forgotten the `kubernetesConfig` object in the windows definition...

PR #2219, opened 2018-02-06
Updates NVIDIA drivers installation

wbuchwalter:
This PR updates NVIDIA drivers installation.
* Instead of installing the drivers through apt, directly download the official *.run drivers from Nvidia websites. Using the *.run file considerably speeds up the installation of the drivers (from 12 to about 2 minutes)
* The reason we used apt before was because the `.run` file doesn't allow to control where the libraries end up getting installed, they always end up in `/usr/lib/x86_64-linux-gnu`, which cause conflict issues when we then need to mount this folder into a pod.
By using Overlayfs we are able to bypass this issue and have the libraries installed in `/usr/local/nvidia/lib64` instead.
* Currently, with apt, drivers are located in `/usr/lib/nvidia-384`. This is not good because it means the `hostpath` will change depending on the version of the driver installed on the host, potentially breaking templates that used to work before. This PR solves that as well since everything is now installed under `/usr/local/nvidia`

Left to do:
* [x] Update documentation
* [x] Do some more testing
* [x] Lint

Cc @lachie83 @sauryadas

### DIFF
--- a/docs/kubernetes/gpu.md
+++ b/docs/kubernetes/gpu.md
@@ -19,7 +19,7 @@ If `alpha.kubernetes.io/nvidia-gpu` is `0` and you just created the cluster, you
 When running a GPU container, you will need to specify how many GPU you want to use. If you don't specify a GPU count, kubernetes will asumme you don't require any, and will not map the device into the container.
 You will also need to mount the drivers from the host (the kubernetes agent) into the container.
 
-On the host, the drivers are installed under `/usr/lib/nvidia-384`.
+On the host, the drivers are installed under `/usr/local/nvidia`.
 
 Here is an example template running TensorFlow: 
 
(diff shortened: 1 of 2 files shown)


jackfrancis:
@wbuchwalter `gofmt -w pkg/acsengine/engine.go` should fix linting error

jackfrancis:
@wbuchwalter @sozercan Could you recommend some practical E2E strategies for testing GPU clusters?

- recommended SKUs to use for build-outs
- any GPU-dependent workloads that will prove out basic functionality?

wbuchwalter:
@jackfrancis 
> recommended SKUs to use for build-outs

NC6 and NV6

> any GPU-dependent workloads that will prove out basic functionality?

Something like this:
https://gist.github.com/wbuchwalter/c69ebba322781e8882f424e52833418c

This template need to be updated to reflect the changes of this PR w.r.t drivers location. But if a GPU is not detected or drivers are not succesfully installed `nvidia-smi` will fail, so it should be a good start.

paulbouwer:
This is a great update and makes gpu support really clean - awesome work @wbuchwalter !

I have a question though. It looks from the comments above, like the contents of this PR got merged into master. But I can't see the changes in either:
- [master - pkg/acsengine/engine.go#L1508-L1556](https://github.com/wbuchwalter/acs-engine/blob/master/pkg/acsengine/engine.go#L1508-L1556)
- [0.13.1 - pkg/acsengine/engine.go#L1681-L1729](https://github.com/Azure/acs-engine/blob/v0.13.1/pkg/acsengine/engine.go#L1681-L1729)

The doc changes however do seem to be merged which leads to some confusion when following them with the 0.13.1 release. The drivers are still being placed into `/usr/lib/nvidia-384` and `/usr/lib/x86_64-linux-gnu`.

/cc @jackfrancis @sozercan @wbuchwalter

wbuchwalter:
@paulbouwer 
You are looking at the master from my fork, which is outdated.
master from acs-engine does have the changes:
https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/engine.go#L1754

jackfrancis:
+1 on awesome work @wbuchwalter :)

paulbouwer:
@wbuchwalter - excellent. Apologies for the master branch mixup. So the updates were all merged 14 days ago.

Still not sure why I'm not seeing those changes when I create a GPU pool using the 0.13.1 acs-engine release.

jackfrancis:
@paulbouwer `v0.13.1` did not include @wbuchwalter's GPU changes. You'll see them in `v0.14.0`, which we're baking now (ETA mid-next week at the latest).

PR #2225, opened 2018-02-07
blocking cse on cluster nodes ready

jackfrancis:
**What this PR does / why we need it**: Adds k8s cluster desired state guarantees into the provision script (and thus cse). The end result is that when ARM returns a successful deployment we have already verified that all expected nodes are ready.

**Release note**:

```
NONE
```


parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -1,28 +1,5 @@
 #!/bin/bash
 
-###########################################################
```
jackfrancis: This mast header text no longer relevant

jackfrancis:
@brendanburns @amanohar RFC

Any concerns with this strategic approach? Is there something similar we can do for agent nodes (these additions will only run on masters, and so will not bring any add'l desired state enhancements to agent pool-only clusters).

parts/k8s/kubernetesmastercustomscript.sh
CecileRobertMichon: 💥

parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -566,35 +536,63 @@ function ensureJournal(){
...
-            if [ "$?" = "0" ]
+        sleep 1
+    done
+    if [ $k8sHealthy -ne 0 ]
+    then
+        echo "k8s cluster is not healthy after $i seconds"
+        exit 3
+    fi
+    for i in {1..1800}; do
+        nodes=$(${KUBECTL} get nodes 2>/dev/null | grep 'Ready' | wc -l)
```
CecileRobertMichon: Since this is grepping for "Ready", is it possible that `nodesActive` would be set to 0 but that in the next step `nodesReady` wouldn't be 0?

jackfrancis: Yes, because `grep 'Ready'` will match both `Ready` and `NotReady`.

The purpose of `nodesActive` is to tag when all nodes have checked in with the api server (either `Ready` or `NotReady`). The purpose of `nodesReady` is to tag when all nodes are ready to receive workloads (i.e., `Ready`).

PR #2237, opened 2018-02-09
Fix azure cni service ip

saiyan86:

This PR fixes service IP connectivity issue when using Azure CNI for Windows nodes. Without this PR, the service IP can't be accessed from public network.


parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -356,6 +359,13 @@ Update-CNIConfig(`$podCIDR, `$masterSubnetGW)
 
 try
 {
+    
+    if (`$global:NetworkPolicy -eq "azure") {
```
JiangtianLi: Why is this needed?

PR #2241, opened 2018-02-09
enforce apt-get update warnings/errors retries

jackfrancis:
Here is a reference:

https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1693900

`apt-get update` returns a `0` exit code even if it experiences symptoms of network issues. It considers these normal operational behavior. It outputs to stdout and stderr, however, so we can grep for these messages and retry accordingly.

This PR introduces a `apt_get_update` func in cloud-init for masters and agents which retries `apt-get update` under known warning or error conditions. Also adds `set -x` to cloud-init runcmd array.

**Release note**:

```
NONE
```

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -167,6 +167,7 @@ coreos:
 runcmd:
 - echo `date`,`hostname`, startruncmd>>/opt/m
 # the first arg is the number of retries, the second arg is the wait duration between two retries and the rest of the args are the cmd to run
+- set -x
 - retrycmd_if_failure() { retries=$1; wait=$2; shift && shift; for i in $(seq 1 $retries); do ${@}; [ $? -eq 0  ] && break || sleep $wait; done; echo Executed \"$@\" $i times; }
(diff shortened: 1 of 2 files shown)


parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -168,12 +168,13 @@ runcmd:
 - echo `date`,`hostname`, startruncmd>>/opt/m
 # the first arg is the number of retries, the second arg is the wait duration between two retries and the rest of the args are the cmd to run
 - retrycmd_if_failure() { retries=$1; wait=$2; shift && shift; for i in $(seq 1 $retries); do ${@}; [ $? -eq 0  ] && break || sleep $wait; done; echo Executed \"$@\" $i times; }
+- retrycmd_if_success() { retries=$1; wait=$2; shift && shift; for i in $(seq 1 $retries); do ${@}; [ $? -ne 0  ] && break || sleep $wait; done; echo Executed \"$@\" $i times; }
```
CecileRobertMichon: So this would only retry if the command succeeds? What about if it fails? Why not just run the command several times?

jackfrancis: This only retries if the command succeeds (such as grep'ing for a failure string).

If the command exits non-zero it should `break` the for loop.

Just running the command several times as well would work assuming the command is idempotent and the most recent run reflects enforced state (in which case always retrying n times where the nth attempt fails would not get you where you want to go). Also, if the command takes time always retrying n times adds T * n to every run, even ones that succeed on the first attempt.

Issue #2246, opened 2018-02-10
label for kubernetes component pods is inconsistent


diwakar-s-maurya:
**Is this a request for help?**:

Yes

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

ISSUE

**What version of acs-engine?**:

v0.12.5

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
1.9.1

**What happened**:
The labels assigned to kube-scheduler, kube-controller-manager is `component: kube-scheduler` and `component: kube-controller-manager` respectively. The label for kubelet and kube-dns is `k8s-app: kubelet ` and `k8s-app: kube-dns` respectively.

**What you expected to happen**:
All the above component should have consistent key name. They should share common key name for the label i.e. `k8s-app`.
Therefore the label assigned to kube-scheduler to be `k8s-app: kube-scheduler` and for kube-controller-manager to be `k8s-app: kube-controller-manager` respectively.
A cluster created using kops on aws assigned common key for label: `k8s-app` to all the components. The helm charts for prometheus by coreos given [here](https://github.com/coreos/prometheus-operator/tree/master/helm) also expect the same key.

**How to reproduce it** (as minimally and precisely as possible):

Create a kubernetes cluster 1.9.1 with acs-engine and check the labels assigned to the pods of kube-scheduler and kube-controller-manager.

**Anything else we need to know**:

PR #2249, opened 2018-02-12
Fix inconsistent labels for kubernetes components (#2246)

diwakar-s-maurya:
* change label key component to k8s-app for cloud-controller-manager, kube-apiserver, kube-controller-manager, kube-scheduler, kube-proxy

It makes the labels put on kubernetes components like scheduler, kube-dns, controller-manager consistent. Before this PR some components use `component` key and some use `k8s-app`.

#2246 

**Release note**:

```release-note
```


jackfrancis:
@lachie83 can you comment on the effects of this change, especially on a long-lived cluster (built w/ acs-engine before this change) that is scaled out w/ a cluster deployment including this change? I.e., for changes that will effect pods scheduled on worker nodes, there would be a heterogeneous label config across the cluster? Thanks!

diwakar-s-maurya:
I noticed that when I create cluster 1.9.4, this PR change is there in the cluster. Closing

PR #2252, opened 2018-02-12
update generateproxycertscript.sh to use secure etcd endpoint/certs

pidah:
When Aggregated API's are enabled with _"enableAggregatedAPIs" : true_ ,  generate-proxy-certs.sh on master performs an etcd healthcheck before creating proxy certs. This etcd healthcheck fails because its calling the pre-0.12.* unsecured etcd endpoint. This PR adds the secure etcd endpoint/cert credentials.


CecileRobertMichon:
`retrycmd_if_failure etcdctl cluster-health` might need to become `retrycmd_if_failure sudo etcdctl cluster-health` because etcdclient.key is protected... Unless this whole script is ran from root.
https://github.com/DigitalInnovation/acs-engine/blob/34bd8c7ff550ef7876d2861b0c60b356e807f5d3/parts/k8s/kubernetesmastergenerateproxycertscript.sh#L48

pidah:
@CecileRobertMichon I validated that it runs as root and picks up the credentials.

PR #2253, opened 2018-02-12
--use-service-account-credentials=false if no rbac

jackfrancis:
**What this PR does / why we need it**: If RBAC is disabled `--use-service-account-credentials` needs to be false for controller-manager

**Release note**:

```
--use-service-account-credentials=false if no rbac
```


jackfrancis:
@pidah Could you sanity check this configuration change from your perspective? Is there any other reason besides RBAC why --use-service-account-credentials would be true? We're doing that enforcement here:

https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults-controller-manager.go#L82

pidah:
@jackfrancis lgtm

PR #2187, opened 2018-02-02
update prometheus config for 2.1

martell:
**What this PR does / why we need it**:
fixes #2142, as a follow up to #2183

PR #2257, opened 2018-02-12
Fix monitoring extension

trstringer:
**What this PR does / why we need it**: fixes the broken prometheus-grafana extension

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2183, fixes #2187

**Special notes for your reviewer**: this extension wasn't functioning due to an oversight of sliding helm charts. This PR not only includes the fix for this particular problem, but also adds support for Prometheus v2. The second functionality that this adds is the ability to change (using `extensionParameters`) which chart config is used when installing Prometheus (this is a nice-to-have for end users, but basically a necessity for contributors to this extension)

cc // @ritazh 

**Release note**:

```release-note
Fix monitoring extension, and add support for Prometheus v2
```

### DIFF
--- a/extensions/prometheus-grafana-k8s/v1/prometheus_values.yaml
+++ b/extensions/prometheus-grafana-k8s/v1/prometheus_values.yaml
@@ -27,7 +27,11 @@ alertmanager:
   ## The URL prefix at which the container can be accessed. Useful in the case the '-web.external-url' includes a slug
   ## so that the various internal URLs are still able to access as they are in the default case.
   ## (Optional)
-  baseURL: ""
+  prefixURL: ""
+
+  ## External URL which can access alertmanager
+  ## Maybe same with Ingress host name
+  baseURL: "/"
 
   ## Additional alertmanager container environment variable
   ## For instance to add a http_proxy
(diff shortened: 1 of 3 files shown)


jackfrancis:
@trstringer Since you're the maintainer of this extension you probably have better test coverage than we do. How do your tests look against the current state of this PR?

trstringer:
@jackfrancis this extension is working with my tests!

martell:
awesome, thanks @trstringer.
work had me tied down for at least another week, so glad someone else could tackle it.

extensions/prometheus-grafana-k8s/v1/prometheus_values.yaml
```diff
@@ -163,7 +171,7 @@ configmapReload:
   ##
   image:
     repository: jimmidyson/configmap-reload
-    tag: v0.2.2
+    tag: v0.1
```
martell: Is there a reason not to use 0.2.2?
0.1 is from some time in 2016

trstringer: That's the default in the Prometheus helm chart v5.1.3. In my opinion, if we want to see this changed then it should be reflected on the charts repo, and then we can PR here to bump up the version of the chart itself. At this point, we want to stay as stable as possible by sticking with default chart values where appropriate.

PR #2260, opened 2018-02-13
E2E test - 50 nodes

jackfrancis:
- Adds a 50 node API model for testing
- Generalizes the max-history tiller test
- Move vm list gathering to just after deployment (the prior method was not working in the jenkins context)

**Release note**:

```
NONE
```

### DIFF
--- a/test/e2e/engine/template.go
+++ b/test/e2e/engine/template.go
@@ -169,25 +169,13 @@ func (e *Engine) HasDashboard() bool {
 }
 
 // HasTiller will return true if tiller addon is enabled
-func (e *Engine) HasTiller() bool {
+func (e *Engine) HasTiller() (bool, api.KubernetesAddon) {
 	for _, addon := range e.ExpandedDefinition.Properties.OrchestratorProfile.KubernetesConfig.Addons {
 		if addon.Name == "tiller" {
-			return *addon.Enabled
+			return *addon.Enabled, addon
 		}
 	}
-	return false
-}
-
-// TillerMaxHistory will return true if tiller addon is enabled
-func (e *Engine) TillerMaxHistory() string {
-	for _, addon := range e.ExpandedDefinition.Properties.OrchestratorProfile.KubernetesConfig.Addons {
-		if addon.Name == "tiller" {
-			if addon.Config != nil {
-				return addon.Config["max-history"]
-			}
-		}
-	}
-	return "5"
+	return false, api.KubernetesAddon{}
 }
 
 // HasACIConnector will return true if aci-connector addon is enabled
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -125,23 +125,20 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 		})
 
 		It("should have tiller running", func() {
-			if eng.HasTiller() {
+			if hasTiller, tillerAddon := eng.HasTiller(); hasTiller {
 				running, err := pod.WaitOnReady("tiller", "kube-system", 3, 30*time.Second, cfg.Timeout)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(running).To(Equal(true))
-			} else {
-				Skip("tiller disabled for this cluster, will not test")
-			}
-		})
-
-		It("should have the expected tiller max-history", func() {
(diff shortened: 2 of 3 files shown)


examples/e2e-tests/kubernetes/node-count/50-nodes/definition.json
```diff
@@ -0,0 +1,39 @@
...
+    "masterProfile": {
+      "count": 5,
+      "dnsPrefix": "",
+      "vmSize": "Standard_D2_v2",
+      "OSDiskSizeGB": 200
+    },
+    "agentPoolProfiles": [
+      {
+        "name": "agentpool1",
+        "count": 50,
```
CecileRobertMichon: In the docs (https://github.com/Azure/acs-engine/blob/master/docs/kubernetes-large-clusters.md) ` We recommend the use of smaller pools (e.g., count of 20) over larger pools (e.g., count of 100); produce your desired total node count with lots of pools, as opposed to as few as possible.`  Would it make sense to change this to several smaller pools since people are likely to look at example files and to copy them for their own implementation?

jackfrancis: @amanohar @khenidak Is the above guidance still valid? (i.e., don't use node pools > 20 count)

test/e2e/engine/template.go
```diff
@@ -178,6 +178,18 @@ func (e *Engine) HasTiller() bool {
 	return false
 }
 
+// TillerMaxHistory will return true if tiller addon is enabled
```
CecileRobertMichon: The comment isn't consistent with the code here

test/e2e/engine/template.go
```diff
@@ -178,6 +178,18 @@ func (e *Engine) HasTiller() bool {
...
+func (e *Engine) TillerMaxHistory() string {
+	for _, addon := range e.ExpandedDefinition.Properties.OrchestratorProfile.KubernetesConfig.Addons {
+		if addon.Name == "tiller" {
+			if addon.Config != nil {
+				return addon.Config["max-history"]
+			}
+		}
+	}
+	return "5"
```
CecileRobertMichon: Why return 5 here? Why not "-1" or nil?

CecileRobertMichon: Actually just realized this is the default max history. In that case it should be inside the for loop (return 5 if addon tiller was found but config is nil) and then return -1 is no tiller addon was found.

test/e2e/remote/ssh.go
```diff
@@ -30,7 +30,8 @@ type Connection struct {
 func NewConnection(host, port, user, keyPath string) (*Connection, error) {
 	conn, err := net.Dial("unix", os.Getenv("SSH_AUTH_SOCK"))
 	if err != nil {
-		log.Fatal(err)
+		log.Println("unable to establish net connection, probably due to $SSH_AUTH_SOCK not being set")
```
CecileRobertMichon: Here is it possible to add the value of SSH_AUTH_SOCK? Ie. ("unable to establish net connection, $SSH_AUTH_SOCK has value %s", os.Getenv("SSH_AUTH_SOCK"))

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -125,23 +125,20 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
-
-		It("should have a tiller max-history of 5", func() {
-			if eng.HasTiller() {
-				pods, err := pod.GetAllByPrefix("tiller-deploy", "kube-system")
-				Expect(err).NotTo(HaveOccurred())
-				// There is only one tiller pod and one container in that pod.
-				actualTillerMaxHistory, err := pods[0].Spec.Containers[0].GetEnvironmentVariable("TILLER_HISTORY_MAX")
-				Expect(err).NotTo(HaveOccurred())
-				Expect(actualTillerMaxHistory).To(Equal("5"))
+				if tillerAddon.Config != nil {
```
CecileRobertMichon: Would this ever be nil? What happens if it is?

jackfrancis: The current configuration enforces always non-nil:

https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults.go#L752

So the above is just for good hygiene.

PR #2270, opened 2018-02-14
Clarify that only Calico supports Kubernetes network policies

dennis-benzinger-hybris:

Both acs-engine and Kubernetes use the term "network policy" but with a different meaning. This PR clarifies this a bit.


jackfrancis:
Thanks @dennis-benzinger-hybris, this is an improvement. After signing the CLA I'll merge these changes.

DanielWangZhanggui:
just want to confirm if using azure vnet plugin on kubernetes, does it support Kubernetes Network Policy? If not, do we have a plan to support it ?

All the best,
Daniel

jackfrancis:
@DanielWangZhanggui by Azure VNET do you mean Azure CNI? At the moment (1) Azure CNI is IPAM only and (2) it doesn't play well w/ calico, which is the best way to get NetworkPolicy implemented in an acs-engine-generated kubernetes cluster.

So, unfortunately, it's one or the other. If NetworkPolicy is primarily important, you may use calico with kubenet.

We are working on this, yes, and hope to have a 1st class solution in the next few release cycles!

sharmasushant:
Yes, We do have plans to support policies in Azure CNI. It’s in progress and we will provide an update soon. @saiyan86

PR #2277, opened 2018-02-14
More e2e tests

CecileRobertMichon:
**What this PR does / why we need it**: adds cluster definitions for Jenkins ginkgo tests to test disabling specific features

**Release note**:

```release-note
```


examples/e2e-tests/kubernetes/release/features-off/addons-disabled.json
```diff
@@ -0,0 +1,56 @@
...
+          },
+          {
+            "name": "rescheduler",
+            "enabled": false
+          }
+        ]
+      }
+    },
+    "masterProfile": {
+      "count": 5,
```
jackfrancis: For these tests let's build 1-master clusters to optimize test time and infra overhead.

examples/e2e-tests/kubernetes/kubernetes-config/addons-disabled.json
```diff
@@ -0,0 +1,56 @@
...
+          },
+          {
+            "name": "rescheduler",
+            "enabled": false
+          }
+        ]
+      }
+    },
+    "masterProfile": {
+      "count": 5,
```
jackfrancis: I still think we should do single masters for these tests, thoughts?

jackfrancis:
So, I wonder if we should just get rid of `examples/e2e-tests/kubernetes/kubernetes-config/addons-enabled.json` and move that coverage to the standard test (i.e., all addons are enabled for our standard version tests)

examples/e2e-tests/kubernetes/kubernetes-config/addons-disabled.json
CecileRobertMichon: Yes that's a miss, I'll fix it! same for addons-enabled

CecileRobertMichon:
My reasoning on this was that the standard test should test the default behavior with no optional features enabled/disabled and then we would turn on/off specific features like addons in separate tests. Otherwise what justifies having addons enabled in the standard tests but not other things like extensions and other non-default features?

jackfrancis:
I think "other things like extensions and other non-default features" should also be turned on on the standard tests as you point out. Thoughts?

CecileRobertMichon:
I like that idea but what about testing a "vanilla" cluster with default values for everything? Because the standard test definition is the one used for PR e2e with every k8s version. Also adds the question of, is everything supported in every version?

jackfrancis:
How about this:

- a "vanilla" cluster for each version (different versions will have different default configurations when not user-provided)
- a "kitchen sink" cluster that includes everything that is supported by all versions (configuration options that are only compatible for particular versions get their own test tied to the most recent compatible version)

In this case "kitchen sink" is the definition that we currently use, that we can easily substitute a k8s version into.

PR #2291, opened 2018-02-15
E2E: HPA as a property of Deployments

jackfrancis:
Couple HPA with deployments in E2E, which makes them easier to create/delete.

**Release note**:

```
NONE
```


jackfrancis:
@CecileRobertMichon the key functional part here is the cleanup of the hpa spec. See `kubectl delete hpa -n default php-apache-kubernetes-southcentralus-26346-66050` in the linux 1.8 test.

The re-arranging of things is an implementation opinion.

PR #2297, opened 2018-02-16
cmd/deploy: Handle error due to missing permissions during deploy

### LINKED ISSUE #2157: acs-engine should check that provided service principal (if any) has enough rights in the target subscription to create Azure resources
NO

---

FEATURE REQUEST

---

any

---

Kubernetes, but I assume it also applies to DC/OS and Swarm

**What happened**:
While acs-engine is able to deploy Azure resources in the target subscription, it never checks/warns that the service principal provided (if any) in the api-model file has the right to create resources in that subscription, like route tables, load-balancers, volumes, etc.
This can lead to non-working clusters like in #2100 (shame on me) if the service principal does not have proper rights in the target subscription.

Source of the confusion is that acs-engine does not rely on service principal to create a Kubernetes cluster and since ARM template is spotlessly processed, you have the feeling that you don't have any right issue, but the cluster won't be able to add any Azure resource and will fail to work properly.

In case a service principal is provided in the api-model file, acs-engine should check that it has the right to spawn resources in the target subscription.

vvvrrooomm:
If the user calling acs-engine deploy does not have the authorization to perform action 'Microsoft.Authorization/roleAssignments/write'  acs-engine loops forever. This was fixed:

* CreateRoleAssignmentSimple could already return an error. This error is now used if arm returns status 403 (not enough permissions).
  This is opposed to status 400 that seems to be issued to signal work in progress during service principal generation (by arm).
* autoFillApimodel did duplicate the retry logic of CreateRoleAssignmentSimple, the duplciated code was removed here. This also allows to properly fail if CreateRoleAssignmentSimple returns an error

**Which issue this PR fixes**
might fix #2157 
related to #2281

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -215,14 +215,11 @@ func autofillApimodel(dc *deployCmd) {
 			log.Warnf("created application with applicationID (%s) and servicePrincipalObjectID (%s).", applicationID, servicePrincipalObjectID)
 
 			log.Warnln("apimodel: ServicePrincipalProfile was empty, assigning role to application...")
-			for {
-				err = dc.client.CreateRoleAssignmentSimple(dc.resourceGroup, servicePrincipalObjectID)
-				if err != nil {
-					log.Debugf("Failed to create role assignment (will retry): %q", err)
-					time.Sleep(3 * time.Second)
-					continue
-				}
-				break
+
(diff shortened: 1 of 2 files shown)


jackfrancis:
Thanks @vvvrrooomm ! FYI `go fmt -w cmd/deploy.go` and `go fmt -w pkg/armhelpers/graph.go` (and follow-up commit) will fix linting errors in test job.

PR #2304, opened 2018-02-16
Replace Glide with Dep

juan-lee:
Glide (https://github.com/Masterminds/glide) suggests moving to Golang Dep. Here you go!

I didn't include it in the PR, but I'd recommend removing most of what is in Gopkg.toml (imported from glide.yaml) as I believe it was misused previously where the lockfile should have sufficed. At the very least I'd recommend tweaking each constraint to snap to a major version.


mboersma:
> I'd recommend removing most of what is in Gopkg.toml 

The guidance @carolynvs gave me here differs if the Go package is intended to be imported as a library, which acs-engine is. In that case, it's preferable to keep Gopkg.toml authoritative and detailed as you have here.

carolynvs:
I just took a look at the Gopkg.toml and would recommend that you keep it as-is for a couple reasons:

1. Now when a downstream consumer of your project uses dep, the constraints in your Gopkg.toml will help ensure that they end up with compatible versions of your dependencies.
2. In the future, this will allow you to run `dep ensure -update github.com/pkg/example` and dep will update the lock and vendor to use the latest release of the specified package, without going past major version boundaries.

The second point is a nice to have, while the first point is required if you are a library. 😀

jackfrancis:
@mboersma @carolynvs I'm inclined to postpone this change (replace glide w/ dep) for a few months to not introduce add'l change into the project. Are we able to identify that by delaying this change we inflict pain on projects that import acs-engine (especially projects that are themselves dep-enabled for dependency management)?

mboersma:
@jackfrancis I don't think sticking with `glide` vendoring for now hampers other projects that import acs-engine: no rush. We were just clarifying the point that when acs-engine does move to `dep`, the detailed Gopkg.toml in this PR is the way to go.

jackfrancis:
@mboersma Thanks for confirming, and thanks @juan-lee for giving us the right dep approach :)

I'm going to close this PR, and we'll re-open it as a reference when we're ready to make this transition.

Issue #2263, opened 2018-02-13
FailedNodeAllocatableEnforcement on cgroups-per-qos false


jcharlytown:
**Is this a request for help?**:
Yes
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
Issue
---

**What version of acs-engine?**:
As of now all shipping with PR #1960 
---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Tested with Kubernetes 1.8.4, 1.8.6, 1.8.7

**What happened**:
Kubelets constantly file events like this:
```
Warning  FailedNodeAllocatableEnforcement  49s (x3 over 2m)  kubelet, k8s-master-56038831-0  Failed to update Node Allocatable Limits "": failed to set supported cgroup subsystems for cgroup : Failed to set config for supported subsystems : failed to write 8342011904 to memory.limit_in_bytes: write /var/lib/docker/overlay2/05c98c88dcf284951773f777b09d37101b1eed0eb43b7e5295ee0600f4390f1e/merged/sys/fs/cgroup/memory/memory.limit_in_bytes: invalid argument
```

**What you expected to happen**:
This warning not to pop up, i.e., `Normal   NodeAllocatableEnforced           4s                kubelet, k8s-master-56038831-0  Updated Node Allocatable limit across pods` instead

**How to reproduce it** (as minimally and precisely as possible):
Deploying cluster using this model (I removed some secrets and ssh key):
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorRelease": "1.8",
      "orchestratorVersion": "1.8.6",
      "kubernetesConfig": {
        "kubernetesImageBase": "gcrio.azureedge.net/google_containers/",
        "DockerBridgeSubnet": "172.17.0.1/16",
        "kubeletConfig": {
          "--eviction-hard": "memory.available<500Mi,nodefs.available<10%,nodefs.inodesFree<5%",
          "--system-reserved": "memory=1.5Gi"
        }
      }
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "qos-cgroups-test",
      "vmSize": "Standard_D2_v3",
      "storageProfile": "ManagedDisks",
      "oauthEnabled": false,
      "distro": "ubuntu",
      "vnetCidr": "10.244.0.0/16"
    },
    "agentPoolProfiles": [
      {
        "name": "agents",
[...]

PR #2310, opened 2018-02-19
set kubelet defaults for --cgroups-per-qos &  --enforce-node-allocatable

pidah:
This PR fixes the following kubelet error discussed here: https://github.com/Azure/acs-engine/issues/2263
```
Warning  FailedNodeAllocatableEnforcement  34s (x5787 over 4d)  kubelet, k8s-agent-41758893-0  Failed to update Node Allocatable Limits "": failed to set supported cgroup subsystems for cgroup : Failed to set config for supported subsystems : failed to write 135184371712 to memory.limit_in_bytes: write /var/lib/docker/overlay2/fe259a690a2ff4ec750072fe76a64af2c85acfeb2ff4ac1752a245e2503896af/merged/sys/fs/cgroup/memory/memory.limit_in_bytes: invalid argument
```


jackfrancis:
@lachie83 are you able to weigh in on any gotchas w/ statically setting `--cgroups-per-qos=true` and `--enforce-node-allocatable=pods` for all built clusters?

jackfrancis:
After doing some further reading, this change looks to be idiomatic with current defaults (1.8 and 1.9). I'm 👍 with this change.

jackfrancis:
@yastij Are you speaking to the Windows concern? It looks to me like our Windows build script is producing images that explicitly override the defaults for Linux in this PR. @JiangtianLi to confirm

JiangtianLi:
@jackfrancis cgroups-per-qos doesn't apply to windows kubelet.

JiangtianLi:
enforce-node-allocatable is associated with cgroups-per-qos.

jackfrancis:
Thanks @JiangtianLi, we'll proceed with this PR w/ the understanding that there is no Windows impact. 

Though it would be nice to figure out why those references are here:

https://github.com/Azure/acs-engine/blob/master/parts/k8s/kuberneteswindowssetup.ps1#L288

JiangtianLi:
@jackfrancis --cgroups-per-qos & --enforce-node-allocatable is by default true in 1.6 (maybe 1.7).

jackfrancis:
@JiangtianLi that's true, but the line in the windows powershell script is explicitly setting it to `false`

JiangtianLi:
@jackfrancis Correct. For windows kubelet, if we don't explicitly set to false, for some k8s version, it will be true and it doesn't apply to windows.

yastij:
@jackfrancis - I was speaking about this change, not the script

PR #2321, opened 2018-02-22
Update docs to reflect current kubelet --read-only-port config

jackfrancis:
we are not actually setting —read-only-port=0 for kubelet

**What this PR does / why we need it**: Update docs to reflect current kubelet --read-only-port config

**Release note**:

```
NONE
```


jackfrancis:
@CecileRobertMichon FYI

PR #2326, opened 2018-02-22
Private clusters

### LINKED ISSUE #221: Clusters without public IPs
In some cases customers would like to use Azure Express Route without public IP addresses assigned.  This issue tracks the need an popularity of this request.

CecileRobertMichon:
**What this PR does / why we need it**: The goal of this PR is to allow building a cluster without public IPs by:
1. building a cluster without a publically facing load balancer
2. generating kubeconfig artifacts that point to the API server via the internal load balancer IP address
The option can be enabled with the flag `enablePrivateCluster` in `KubernetesConfig` in the apimodel. By default `enablePrivateCluster` is false.
A user would then be able to move the appropriate kubeconfig artifact onto a host with kubectl installed in the same k8s VNET (or onto a peer VNET that routes to the k8s VNET) and access the cluster from that host.

What is left TODO:

- [x] add docs
- [x] add example 
- [x] remove publicIP resource (right now it is still being created but the DNS name doesn't resolve to an actual IP) and master FQDN output

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #221 

What do people think about calling the flag `enablePrivateCluster` and placing it inside the `KubernetesConfig`? Another option would be to place it in the `MasterProfile` but I thought this would add some complexity since it only applies to k8s clusters. Thoughts?

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -68,11 +68,11 @@
         "name": "loop"
       },
       {{if not IsHostedMaster}}
-      {{if not IsPrivateCluster}}
-      "dependsOn": [
-        "[concat('Microsoft.Network/publicIPAddresses/', variables('masterPublicIPAddressName'))]"
-      ],
-      {{end}}
+        {{if not IsPrivateCluster}}
+          "dependsOn": [
+            "[concat('Microsoft.Network/publicIPAddresses/', variables('masterPublicIPAddressName'))]"
+          ],
+        {{end}}
       {{end}}
       "location": "[variables('location')]",
       "name": "[concat(variables('storageAccountPrefixes')[mod(add(copyIndex(),variables('{{.Name}}StorageAccountOffset')),variables('storageAccountPrefixesCount'))],variables('storageAccountPrefixes')[div(add(copyIndex(),variables('{{.Name}}StorageAccountOffset')),variables('storageAccountPrefixesCount'))],variables('{{.Name}}AccountName'))]",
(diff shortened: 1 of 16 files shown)


parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -68,10 +68,12 @@
         "name": "loop"
       },
       {{if not IsHostedMaster}}
+      {{if not IsPrivateCluster}}
```
jackfrancis: Is there no way to do an `and` to the above required `if` conditions?

yfried:
Hi,
How do I create the jumpbox? the doc just tells me to create a VM.
Do I need the jumpbox if I have a VPN connection to the cluster?

CecileRobertMichon:
@yfried

CecileRobertMichon:
@yfried if you have a vnet peered with another or a VPN to the clusters' vnet then you don't need a jumpbox and you can use the other vnet or your VPN to access the cluster.  to create a jumpbox you can use the Azure Portal to create a VM in the same

CecileRobertMichon:
@yfried if you have a vnet peered with another or a VPN connection to the cluster then you don't need a jumpbox and you can use the other vnet or your VPN to access the cluster. 

If you don't have a peer vnet nor a VPN you will need to create a jumpbox. After deploying your cluster with no public IPs, you can use the Azure Portal to create a VM in the same resource group and vnet as your cluster. You should then be able to ssh into the VM and access your private cluster (after following the instructions in the docs to install kubectl and copy over the kubeconfig).

The next step to this process is to automate the jumpbox creation as part of deployment but the first step is to get feedback on the initial implementation and make sure we're meeting the functionality + security requirements.

SilverGaucho:
Hi

@CecileRobertMichon the flag _enablePrivateCluster_ is not a recognized Flag using --api-model by the acs-engine

CecileRobertMichon:
@SilverGaucho are you building from the master branch? This PR has not yet been included in an acs-engine release

sylus:
I was curious if I have already deployed  a cluster is it possible to change my apimodel and then do another `az group deployment` or will I need a fresh cluster?

Thanks for the amazing work!

CecileRobertMichon:
@sylus the simple answer is no. Right now there is no "official" solution for modifying running clusters with acs-engine. Note that if you can, it is better hygiene to build a fresh cluster whenever possible. That being said, there is a way to manually modify a running cluster to delete all public IP resources and make it private. I can walk you through it if you're interested.

sylus:
Ah thanks you so much for your response, I am using Ark so in theory it should be easy to bring up another cluster. We had a lot of security policies re AD and a custom vnet so have mostly just being doing az group deployment updates and have been lucky so far that nothing has seemed to have any issues.

Thanks for the heads up that this won't be possible and what the best practice is, I greatly appreciate it.

I definitely would be interested if you get the chance and will happily submit a P.R. based on that if helps. Thanks again! :)

yfried:
Hi, when will this be released? I see it's not in 1.13.

sylus:
@CecileRobertMichon if get a chance would mind pinging me on k8s slack channel so can run through private ip? Thanks for any help you can provide :) Username: willh

CecileRobertMichon:
@yfried this will be released very shortly in v0.14.0 along with PR #2401. @sylus I will ping you when I get the chance, might not be today though, we're trying to get k8s 1.9.4 merged and v0.14.0 out...

jackfrancis:
This is in https://github.com/Azure/acs-engine/releases/tag/v0.14.0

PR #2327, opened 2018-02-23
Use original FirstConsecutiveStaticIP during upgrade.

JunSun17:

Allow upgrade work for cluster deployed with v0.13.0.

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -274,7 +274,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 
 	sc.containerService.Properties.AgentPoolProfiles = []*api.AgentPoolProfile{sc.agentPool}
 
-	template, parameters, _, err := templateGenerator.GenerateTemplate(sc.containerService, acsengine.DefaultGeneratorCode)
+	template, parameters, _, err := templateGenerator.GenerateTemplate(sc.containerService, acsengine.DefaultGeneratorCode, false)
 	if err != nil {
 		log.Fatalf("error generating template %s: %s", sc.apiModelPath, err.Error())
 		os.Exit(1)
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -239,12 +239,12 @@ var (
 )
 
 // SetPropertiesDefaults for the container Properties, returns true if certs are generated
-func SetPropertiesDefaults(cs *api.ContainerService) (bool, error) {
+func SetPropertiesDefaults(cs *api.ContainerService, isUpgrade bool) (bool, error) {
 	properties := cs.Properties
 
 	setOrchestratorDefaults(cs)
 
-	setMasterNetworkDefaults(properties)
+	setMasterNetworkDefaults(properties, isUpgrade)
 
 	setHostedMasterNetworkDefaults(properties)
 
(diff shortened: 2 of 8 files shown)


pkg/acsengine/defaults.go
```diff
@@ -471,7 +471,9 @@ func setMasterNetworkDefaults(a *api.Properties) {
 			if a.OrchestratorProfile.IsAzureCNI() {
 				// When VNET integration is enabled, all masters, agents and pods share the same large subnet.
 				a.MasterProfile.Subnet = a.OrchestratorProfile.KubernetesConfig.ClusterSubnet
-				a.MasterProfile.FirstConsecutiveStaticIP = getFirstConsecutiveStaticIPAddress(a.MasterProfile.Subnet)
+				if len(a.MasterProfile.FirstConsecutiveStaticIP) == 0 {
```
CecileRobertMichon: What happens if the user provides a value for FirstConsecutiveStaticIP but not for Subnet?

JunSun17: The current code will honor the FirstConsecutiveStaticIP if it is given by user. I think it is ok and if this value does not work, user can always remove it in their model and let it to be calculated automatically by the getFirstConsecutiveStaticIPAddress method.

CecileRobertMichon: The current code overrides the FirstConsecutiveStaticIP if the subnet is invalid or nil. Adding that if statement without checking the subnet would mean a user could potentially have an apimodel with an a FirstConsecutiveStaticIP but no subnet which means that the IP and the subnet would not correspond since the subnet would still be the default one. Or provide both a FirstConsecutiveStaticIP and a ClusterSubnet and only ClusterSubnet would get overridden? (All of this is looking at it from the perspective of a first deployment, not upgrade). I don't think " if this value does not work, user can always remove it in their model" is good UX since we wouldn't be failing the apimodel validation.

pkg/acsengine/defaults.go
```diff
@@ -471,7 +471,9 @@ func setMasterNetworkDefaults(a *api.Properties) {
...
+					a.MasterProfile.FirstConsecutiveStaticIP = getFirstConsecutiveStaticIPAddress(a.MasterProfile.Subnet)
+				}
 			} else {
 				a.MasterProfile.Subnet = DefaultKubernetesMasterSubnet
 				a.MasterProfile.FirstConsecutiveStaticIP = DefaultFirstConsecutiveKubernetesStaticIP
```
CecileRobertMichon: What about this case? Since DefaultFirstConsecutiveKubernetesStaticIP should we also keep FirstConsecutiveStaticIP for kubenet clusters?

JunSun17: To simply the logic in the initial deployment and upgrade workflow, I introduce an isUpgrade flag which can be used to fine control features we want to pick in the upgrade route. This should address your concern.

cmd/scale.go
```diff
@@ -274,7 +274,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 
 	sc.containerService.Properties.AgentPoolProfiles = []*api.AgentPoolProfile{sc.agentPool}
 
-	template, parameters, _, err := templateGenerator.GenerateTemplate(sc.containerService, acsengine.DefaultGeneratorCode)
+	template, parameters, _, err := templateGenerator.GenerateTemplate(sc.containerService, acsengine.DefaultGeneratorCode, false)
```
CecileRobertMichon: It might be interesting to check the behavior of scale / backward compatibility of scale after reverting the Azure CNI static IPs PR... Do we know that scale wasn't affected?

JunSun17: Good point. My understanding is that we are only scaling agent nodes but not master nodes, so hopefully it should not be affected. I will check if we have testing coverage on this.

JunSun17: Validated with scaling tests and passed.

PR #2354, opened 2018-02-28
Private clusters iteration 2: change the server for the cluster kubeconfig

CecileRobertMichon:
**What this PR does / why we need it**: Changes the server IP for the kubeconfig in /var/lib/kubelet/ and the one in .kube/ generated by the custom script

**Release note**:

```release-note
```


parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -73,7 +73,11 @@ write_files:
     - name: localcluster
       cluster:
         certificate-authority: /etc/kubernetes/certs/ca.crt
-        server: {{WrapAsVerbatim "concat('https://', variables('masterPrivateIpAddrs')[copyIndex(variables('masterOffset'))], ':443')"}}
+        {{- if IsPrivateCluster}}
+            server: {{WrapAsVariable "kubeconfigServer"}}
+        {{- else}}
+            server: {{WrapAsVerbatim "concat('https://', variables('masterPrivateIpAddrs')[copyIndex(variables('masterOffset'))], ':443')"}}
```
CecileRobertMichon: @lachie83 now I'm wondering if we need this if statement... Since the server IP was already the master's IP address and it's different for each master I don't think changing it to the ILB IP for >1 master / the first master IP for 1 master makes that much sense. What do you think?

lachie83: Yeah I agree. This does however raise a question around the failure-modes if kubelet can only communicate with the local api-server in a multi-master cluster.

PR #2365, opened 2018-03-01
Remove deprecated '--require-kubeconfig' for k8s

karataliu:

The option '--require-kubeconfig' option was no longer required and removed in
https://github.com/kubernetes/kubernetes/pull/59306/files

With v1.10.0-beta.0, kubelet will fail to run when call with he option
```
F0301 03:10:48.067923      19 server.go:142] unknown flag: --require-kubeconfig
```

**Which issue this PR fixes**

**Release note**:

```release-note
```


feiskyer:
+1. I need to remove this flag manually for new 1.10 images.

jackfrancis:
@feiskyer @karataliu looks like 1.7 still needs this, see my commit; will re-test

PR #2366, opened 2018-03-01
Change type of EtcdDiskSizeGB from string to int

### LINKED ISSUE #2231: nit: etcd diskSize is string
Just a little nit - 
https://github.com/Azure/acs-engine/blob/master/pkg/api/types.go#L245 

size should be an int not a string

shubheksha:
This PR changes the data type of EtcdDiskSizeGB from `string` to `int`.

Fixes #2231 

NA

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -1548,7 +1548,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 				case "etcdVersion":
 					val = cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdVersion
 				case "etcdDiskSizeGB":
-					val = cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB
+					val = string(cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB)
 				default:
 					val = ""
 				}
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -1548,7 +1548,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 				case "etcdVersion":
 					val = cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdVersion
 				case "etcdDiskSizeGB":
-					val = string(cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB)
+					val = strconv.Itoa(cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB)
 				default:
 					val = ""
 				}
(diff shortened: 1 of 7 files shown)


pkg/acsengine/engine.go
```diff
@@ -1548,7 +1548,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 				case "etcdVersion":
 					val = cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdVersion
 				case "etcdDiskSizeGB":
-					val = cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB
+					val = string(cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB)
 				default:
```
shubheksha: @CecileRobertMichon, not sure if this is the right thing to do here. What do you think?

pkg/api/vlabs/types.go
```diff
@@ -250,7 +250,7 @@ type KubernetesConfig struct {
 	GCHighThreshold              int               `json:"gchighthreshold,omitempty"`
 	GCLowThreshold               int               `json:"gclowthreshold,omitempty"`
 	EtcdVersion                  string            `json:"etcdVersion,omitempty"`
-	EtcdDiskSizeGB               string            `json:"etcdDiskSizeGB,omitempty"`
+	EtcdDiskSizeGB               int               `json:"etcdDiskSizeGB,omitempty"`
```
shubheksha: Found the same problem here, changed the type there as well

pkg/acsengine/engine.go
CecileRobertMichon: You might want to use `strconv.Itoa(cs.Properties.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB) ` like it's being done for `gchighthreshold` and `gclowthreshold` a few lines above for consistency

shubheksha: Ah, I missed that! Thanks for pointing it out, I'll make the change

CecileRobertMichon:
You'll want to change the parameter type [here](https://github.com/Azure/acs-engine/blob/ec456b1637e1bd94bde0255e881f336a81bbc4ac/parts/k8s/kubernetesparams.t#L661)

shubheksha:
Yep, identified it. It needs to be changed in the part file.

jackfrancis:
As I feared, if we upgrade a cluster previously deployed using the prior (string) type spec:

```
time="2018-03-05T20:20:47Z" level=info msg=validating...
time="2018-03-05T20:20:50Z" level=info msg="Error returned by LoadContainerService: json: cannot unmarshal string into Go struct field Properties.orchestratorProfile of type int"
time="2018-03-05T20:20:50Z" level=fatal msg="error parsing the api model: json: cannot unmarshal string into Go struct field Properties.orchestratorProfile of type int"
```

shubheksha:
@jackfrancis, ah right - this will break previously deployed clusters using the `string` type. How can we get around this? Is there any way we can use the string type for backward compatibility but use int going forward?

jackfrancis:
@shubheksha Firstly, thanks for tackling this! I think we'll want to address this more generally in terms of "how do we adapt upgrade implementation so that it doesn't depend literally on the type specification of the template at the time of initial deployment". That will unblock lots of future issues like this one.

jackfrancis:
Closing this PR and the referring issue as well.

PR #2376, opened 2018-03-02
Reuse GetCloudTargetEnv in FormatAzureProdFQDN

gsacavdm:
**What this PR does / why we need it**: It reuses the GetCloudTargetEnv function in the FormatAzureProdFQDN function to avoid having duplicate logic that converts the region into the Cloud environment

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -368,7 +368,7 @@ func FormatAzureProdFQDNs(fqdnPrefix string) []string {
 
 // FormatAzureProdFQDN constructs an Azure prod fqdn
 func FormatAzureProdFQDN(fqdnPrefix string, location string) string {
-	FQDNFormat := ""
+	var FQDNFormat string
 	switch GetCloudTargetEnv(location) {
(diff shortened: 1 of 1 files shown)


pkg/acsengine/engine.go
```diff
@@ -368,13 +368,16 @@ func FormatAzureProdFQDNs(fqdnPrefix string) []string {
 
 // FormatAzureProdFQDN constructs an Azure prod fqdn
 func FormatAzureProdFQDN(fqdnPrefix string, location string) string {
-	FQDNFormat := AzureCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix
-	if location == "chinaeast" || location == "chinanorth" {
+	FQDNFormat := ""
```
jackfrancis: change this to `var FQDNFormat string` to fix linting error

gsacavdm:
@jackfrancis , done!

pkg/acsengine/engine.go
gsacavdm: Done!

PR #2381, opened 2018-03-02
Pass --location to containerService

### LINKED ISSUE #2380: acs-engine deploy fails for Azure US Government due to template parameters for AzurePublicCloud
---
**Is this an ISSUE or FEATURE REQUEST?**: Issue
---
**What version of acs-engine?**: v0.13.1
---

**What happened**: Running `acs-engine deploy --api-model apimodel.json --location usgovvirginia --azure-env AzureUSGovernmentCloud` results in a deployment that does not have the appropriate template parameters for AzureUSGovernment, rather it has the parameters for AzurePublic:
* OSImageVersion: 16.04.201801260
* fqdnEndpointSuffix: cloudapp.azure.com
* targetEnvironment: AzurePublicCloud

You can check these in the Azure portal, navigate to the resource group created by acs-engine deploy, click on Deployments, select the deployment and look at the parameters.

**What you expected to happen**: The resulting Azure deployment should have the appropriate AzureUSGovernment template parameters set:
* OSImageVersion: latest
* fqdnEndpointSuffix: cloudapp.usgovcloudapi.net
* targetEnvironment: AzureUSGovernmentCloud

Run the command noted above.

gsacavdm:
**What this PR does / why we need it**: This PR is needed to get `deploy`, `scale` and `upgrade` working with sovereign clouds without having to add the location to the apimodel.json file. 

**Which issue this PR fixes**: fixes #2380 

Before this PR, the `--location` wasn't being passed down to `deploy.go:templateGenerator.GenerateTemplate`, so  the resulting template would not have the sovereign cloud parameters (e.g. `osImageVersion`, `fqdnEndpointSuffix`, `targetEnvironment`) unless you ALSO specified `"location": "usgovvirginia"` in the `apimodel.json` file.

To fix this, we should pass the value from `--location` into the `containerService` object. There are different ways / places to do this. I'm not a big fan of replicating the check however there's already similar replication across these three commands (deploy, scale and upgrade) already.

I opted for only populating the `containerService.Location` if it didn't have one already, and if it does and it doesn't match the `--location`, error out. I think this is the safest approach.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -111,14 +111,20 @@ func (dc *deployCmd) validate(cmd *cobra.Command, args []string) error {
 			Locale: dc.locale,
 		},
 	}
+
+	if dc.location == "" {
+		return fmt.Errorf(fmt.Sprintf("--location must be specified"))
+	}
 	// skip validating the model fields for now
 	dc.containerService, dc.apiVersion, err = apiloader.LoadContainerServiceFromFile(dc.apimodelPath, false, false, nil)
 	if err != nil {
 		return fmt.Errorf(fmt.Sprintf("error parsing the api model: %s", err.Error()))
 	}
 
-	if dc.location == "" {
-		return fmt.Errorf(fmt.Sprintf("--location must be specified"))
+	if dc.containerService.Location == "" {
+		dc.containerService.Location = dc.location
+	} else if dc.containerService.Location != dc.location {
+		return fmt.Errorf(fmt.Sprintf("--location does not match api model location"))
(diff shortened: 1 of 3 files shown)


jackfrancis:
@dmitsh Could you kindly review the upgrade/scale changes?

jackfrancis:
@dmitsh not currently, but I infer from your remark that we should run against this PR (which we can do) :)

andyzhangx:
@jackfrancis so how many acs-engine does not support sovereign cloud?

PR #2409, opened 2018-03-07
[WIP] Install NVIDIA drivers for all GPU skus

wbuchwalter:
This PR installs the NVIDIA drivers for every N series VM now that we got the agreement signed with NVIDIA.

Also updates the drivers.

[WIP] because I am still making sure the updated drivers didn't break anything.

cc @lachie83 @jackfrancis

### DIFF
--- a/pkg/acsengine/engine.go
+++ b/pkg/acsengine/engine.go
@@ -1754,7 +1754,7 @@ func isNSeriesSKU(profile *api.AgentPoolProfile) bool {
 func getGPUDriversInstallScript(profile *api.AgentPoolProfile) string {
 
 	// latest version of the drivers. Later this parameter could be bubbled up so that users can choose specific driver versions.
-	dv := "384.111"
+	dv := "390.30"
 	dest := "/usr/local/nvidia"
 
 	/*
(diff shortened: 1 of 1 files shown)


jackfrancis:
@wbuchwalter See this lint complaint:

`pkg/acsengine/engine.go:1796:2:warning: ineffectual assignment to na (ineffassign)`

What would you like to do with that, and the associated `getGPUDriversNotInstalledWarningMessage` func, which is called, but whose return value (`na`) is never used?

jackfrancis:
(also FYI I pushed up a commit to fix another linting error, so you'll have to pull :))

wbuchwalter:
@jackfrancis Sorry, you were faster than me :)
I also removed the test testing for SKUs without licences since there aren't any left.

I basically removed all the code handling "no-licence" SKUs since there aren't any at this time and there are no new commercial GPUs announced by NVIDIA, so it might be several months before we encouter this issue again.

The code could be simplified a bit more, since all GPUs have the same script now, but we might need different scripts for different GPUs later (like RDMA support for the N*r VMs), so I kept it that way.

It that fine for you?

PR #2410, opened 2018-03-08
fix disk unavailable issue when mounting multiple azure disks due to dev name change

### LINKED ISSUE #1918: unable to use azure disk in StatefulSet since /dev/sd* changed after detach/attach disk
yes
---

ISSUE
---

v0.9.1
---

Kubernetes v1.7.9

**What happened**:
We create a storageClass
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: uat
provisioner: kubernetes.io/azure-disk
parameters:
  skuName: Standard_LRS
  location: chinaeast
  storageAccount: X
```
then we create statefulset to use it

```
volumeMounts:
        - name: admin-persistent-storage
          mountPath: /u01
  volumeClaimTemplates:
  - metadata:
      name: admin-persistent-storage
    spec:
      storageClassName: uat
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 50Gi
```
At first, we could use this disk normally, but after a period of time, we found that the state of the disk changed to read-only. After the deletion of pod, the newly generated pod could use the same disk, but after a period of time, the disk became read-only again.
there are some error message in pod,u01 is our pv disk.

```
[root@admin-0 /]# cd u01
[root@admin-0 u01]# ls
ls: reading directory .: Input/output error
```

```
dmesg|grep error
[852311.079420] EXT4-fs error (device sdf): ext4_find_entry:1465: inode #1310721: comm java: reading directory lblock 0
[852311.283939] EXT4-fs error (device sdf): ext4_wait_block_bitmap:503: comm java: Cannot read block bitmap - block_group = 5, block_bitmap = 1030
[852311.292376] EXT4-fs error (device sdf): ext4_discard_preallocations:4058: comm java: Error -5 reading block bitmap for 5
[852311.301009] EXT4-fs error (device sdf): ext4_wait_block_bitmap:503: comm java: Cannot read block bitmap - block_group = 5, block_bitmap = 1030
```
Can normal use of dynamic disk

andyzhangx:
fix disk unavailable issue when mounting multiple azure disks due to dev name change

When User mount the 6th azure data disk on an D2_V2 VM, other azure disk would be unavailable due to device name change. This PR set the default `cachingmode` as `None` to fix this issue.
You could find more details in this upstream PR: https://github.com/kubernetes/kubernetes/pull/60346

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1918

@jackfrancis @JiangtianLi This is a critical bug fix, we have changed the default `cachingmode` value as `None` in upstream code, while as a simple workaround, we could change the  `cachingmode` value to `None` in azure disk storage class. Could you fix it in AKS also? Thanks.

**Release note**:

```
fix disk unavailable issue when mounting multiple azure disks
```

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-managed-azure-storage-classes.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-managed-azure-storage-classes.yaml
@@ -10,6 +10,7 @@ provisioner: kubernetes.io/azure-disk
 parameters:
   kind: Managed
   storageaccounttype: Standard_LRS
+  cachingmode: None
 ---
 apiVersion: storage.k8s.io/v1beta1
 kind: StorageClass
@@ -22,6 +23,7 @@ provisioner: kubernetes.io/azure-disk
 parameters:
   kind: Managed
   storageaccounttype: Premium_LRS
+  cachingmode: None
 ---
 apiVersion: storage.k8s.io/v1beta1
 kind: StorageClass
@@ -34,6 +36,7 @@ provisioner: kubernetes.io/azure-disk
 parameters:
   kind: Managed
   storageaccounttype: Standard_LRS
+  cachingmode: None
 ---
 kind: StorageClass
 apiVersion: storage.k8s.io/v1
--- a/parts/k8s/addons/kubernetesmasteraddons-unmanaged-azure-storage-classes.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-unmanaged-azure-storage-classes.yaml
@@ -7,6 +7,8 @@ metadata:
   labels:
     kubernetes.io/cluster-service: "true"
 provisioner: kubernetes.io/azure-disk
+parameters:
+  cachingmode: None
 ---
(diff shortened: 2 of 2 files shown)


parts/k8s/addons/kubernetesmasteraddons-managed-azure-storage-classes.yaml
```diff
@@ -10,6 +10,7 @@ provisioner: kubernetes.io/azure-disk
 parameters:
   kind: Managed
   storageaccounttype: Standard_LRS
+  cachingmode: None
```
JiangtianLi: What is the default cachingmode if it is not set?

andyzhangx: default is `ReadWrite`, and I have set to `None` in current k8s upstream code from v1.7-v1.10, all minor version should fix this issue.

PR #2413, opened 2018-03-08
Add Kubernetes 1.10-beta.2 support

### LINKED ISSUE #2404: k8s 1.10 support note: Use kube-addon-manager-amd64:8.5+ image for addonmanager
To keep a note here for k8s 1.10 support.

Current addonmanager image used by k8s 1.9 does not suit k8s 1.10
```
"addonmanager":              "kube-addon-manager-amd64:v6.5",
```
The behavior is addon-manager won't start.
For example, with single mater note, the master won't treat it self as leader, thus does not start working:
```
INFO: Not elected leader, going back to sleep.
INFO: Leader is k8s-master-40459714-0_e8338f59-0802-11e8-9439-000d3a107392
INFO: Not elected leader, going back to sleep.
INFO: Leader is k8s-master-40459714-0_e8338f59-0802-11e8-9439-000d3a107392
INFO: Not elected leader, going back to sleep.
INFO: Leader is k8s-master-40459714-0_e8338f59-0802-11e8-9439-000d3a107392
```

Root cause is identity naming rule change introduced by:
https://github.com/kubernetes/kubernetes/pull/58302
https://github.com/kubernetes/kubernetes/pull/58386

Updating to `kube-addon-manager-amd64:v8.5` image will do the fix.

jackfrancis:
**What this PR does / why we need it**: Enables folks to start building out v1.10 clusters.

Fixes #2404

**Release note**:

```
Add Kubernetes 1.10-beta.2 support
```


pkg/acsengine/k8s_versions.go
```diff
@@ -8,6 +8,32 @@ import (
 )
 
 var k8sComponentVersions = map[string]map[string]string{
+	"1.10": {
```
jackfrancis: @feiskyer Can you comment on the image references below for v1.10? Especially if there are additional k8s componentry (manifests/addons) that we want to build into the foundations for these clusters.

jackfrancis:
//TODO

metrics-server pod is failing with the following error:

`F0309 04:15:59.001244       1 heapster.go:97] Could not create the API server: cluster doesn't provide requestheader-client-ca-file`

pkg/acsengine/k8s_versions.go
feiskyer: @jackfrancis The versions are still not clear now. We may need to wait for the formal 1.10 release.

jackfrancis:
@feiskyer Thanks. I'm going to merge this so we can begin experimenting.

PR #2414, opened 2018-03-09
Jumpbox kubeconfig

CecileRobertMichon:
**What this PR does / why we need it**:  installs kubectl + kubeconfig on a private cluster jumpbox

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesjumpboxcustomdata.yml
+++ b/parts/k8s/kubernetesjumpboxcustomdata.yml
@@ -1,0 +1,10 @@
+#cloud-config
+
+write_files:
+
+# TODO: Add kubeconfig
+
+runcmd:
+- curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{.OrchestratorProfile.OrchestratorVersion}}/bin/linux/amd64/kubectl
+- chmod +x ./kubectl
+- sudo mv ./kubectl /usr/local/bin/kubectl
--- a/parts/k8s/kubernetesjumpboxcustomdata.yml
+++ b/parts/k8s/kubernetesjumpboxcustomdata.yml
@@ -3,13 +3,14 @@
 write_files:
 
 - path: "/home/{{WrapAsVariable "jumpboxUsername"}}/.kube/config"
-  permissions: "0744"
+  permissions: "0644"
   encoding: "gzip"
   owner: "root"
   content: |
     {{GetKubeConfig}}
 
 runcmd:
-- curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{.OrchestratorProfile.OrchestratorVersion}}/bin/linux/amd64/kubectl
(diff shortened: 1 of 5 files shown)


parts/k8s/kubernetesjumpboxcustomdata.yml
```diff
@@ -0,0 +1,15 @@
+#cloud-config
+
+write_files:
+
+- path: "/home/{{WrapAsVariable "jumpboxUsername"}}/.kube/config"
```
jackfrancis: Is there a possible race condition between cloud-init and user `<jumpBoxUsername>` creation on the host OS?

CecileRobertMichon: I don't think so jumpboxUsername is a variable just like we use variables in master custom data, it stays as a variable up until deployment and jumpboxUsername is passed in from the apimodel

parts/k8s/kubernetesjumpboxcustomdata.yml
```diff
@@ -0,0 +1,15 @@
...
+write_files:
+
+- path: "/home/{{WrapAsVariable "jumpboxUsername"}}/.kube/config"
+  permissions: "0744"
+  encoding: "gzip"
+  owner: "root"
+  content: |
+    {{GetKubeConfig}}
+
+runcmd:
```
jackfrancis: Add this beauty:

`retrycmd_if_failure() { retries=$1; wait=$2; shift && shift; for i in $(seq 1 $retries); do ${@}; [ $? -eq 0  ] && break || sleep $wait; done; echo Executed \"$@\" $i times; }`

And then change the curl statement accordingly, e.g.:

`retrycmd_if_failure 10 5 curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{.OrchestratorProfile.OrchestratorVersion}}/bin/linux/amd64/kubectl`

parts/k8s/kubernetesjumpboxcustomdata.yml
```diff
@@ -0,0 +1,15 @@
+#cloud-config
+
+write_files:
+
+- path: "/home/{{WrapAsVariable "jumpboxUsername"}}/.kube/config"
+  permissions: "0744"
```
jackfrancis: I think this should be `"0644"`

parts/k8s/kubernetesjumpboxcustomdata.yml
jackfrancis: Sorry if I wasn't clear: `/home/<jumpboxUsername>/` won't exist until the user `<jumpboxUsername>` is created. What process creates that user, and does it happen before cloud-init?

CecileRobertMichon: I checked and it's not the case

PR #2423, opened 2018-03-09
Sync flags with azure cli shorthands

kargakis:

Sync shorthand flags between `acs-engine` and `az`. Convenient when jumping from one tool to the other. I may have missed other flags that are the same between the commands but these two (`location`, `resource-group`) seem to be used most often (at least by myself so far). 

**Release note**:

```release-note
NONE
```


kargakis:
@jackfrancis sorry I repushed my commit to update my gpg signature. It seems that I have messed with your review?

PR #2432, opened 2018-03-12
clear containers: add to e2e test

jessfraz:
Why? To make sure it doesn't break :)


jessfraz:
Also if this one is already in the e2e then we don't need this: https://github.com/Azure/acs-engine/blob/master/examples/e2e-tests/kubernetes/kubernetes-config/clear-containers.json

Just making sure.

PR #2433, opened 2018-03-12
simplify code and turn on more linters

jessfraz:
Things that were simplified:

- a lot of returns can be simplified to return the error early
- checking of bools can be done with a bang (!) instead of == false, etc
- remove some dead code

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -322,7 +322,6 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 		if sc.agentPool.IsAvailabilitySets() {
 			addValue(parametersJSON, fmt.Sprintf("%sOffset", sc.agentPool.Name), highestUsedIndex+1)
 		}
-		break
 	case api.Swarm:
 	case api.SwarmMode:
 	case api.DCOS:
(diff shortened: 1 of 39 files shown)


cmd/scale.go
```diff
@@ -353,12 +352,15 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	}
 	var apiVersion string
 	sc.containerService, apiVersion, err = apiloader.LoadContainerServiceFromFile(sc.apiModelPath, false, true, nil)
+	if err != nil {
```
jackfrancis: So actually doing something with this `err` makes perfect sense. @amanohar @dmitsh could you take a look at this code and make sure that there isn't a reason why we actually *want* to absorb this error no matter what? (in which case we should throw away the `err` reference and replace with `_`).

I mean, we definitely *want* to not ignore errors, but if doing so in this scale code actually breaks something holistically, we want to know about it so that we can fix the entire stack.

pkg/acsengine/defaults.go
```diff
@@ -824,7 +824,7 @@ func assignDefaultAddonVals(addon, defaults api.KubernetesAddon) api.KubernetesA
 	}
 	for key, val := range defaults.Config {
 		if addon.Config == nil {
-			addon.Config = make(map[string]string, 0)
+			addon.Config = make(map[string]string)
```
jackfrancis: ❤️

pkg/acsengine/defaults.go
```diff
@@ -854,8 +854,7 @@ func addDefaultFeatureGates(m map[string]string, version string, minVersion stri
 }
 
 func combineValues(inputs ...string) string {
-	var valueMap map[string]string
-	valueMap = make(map[string]string)
+	valueMap := make(map[string]string)
```
jackfrancis: ❤️ ❤️

PR #2435, opened 2018-03-12
Specifying 256GB instead of 128 for etcd disk

### LINKED ISSUE #2510: High etcd 2.2.5 IO writes and CPU IOwait on really small Kubernetes cluster (3 nodes)
**ISSUE/HELP**

**Orchestrator Kubernetes 1.7.10 and etcd 2.2.5**

**What happened**:
We have a few really small Kubernetes clusters hosted on azure virtual machines (IaaS) and deployed with acs-engine.
The SSD storage hosting etcdDisk is providing 500 IOPS and a throughput limit of 100 MB/sec which seemed to be more than enough for this really small cluster.

On all clusters, we are recording a really high amount of CPU IOwait on all master nodes (~40%) and a constant 300 IOPS (100% writes) and 1MB/s writen. All of this comes from the etcd process and is located on etcd own disk.

As far as we can tell, the azure SSD disk hosting etcd working directory is not up to its limit (we made some tests and got to the advertised 500 with both big and small writes). The performance throughput doesn't seem to be the issue also as we are very far from the 100 MB/s limit.

I'm really skeptical about this intensive write use of the SSD provided to etcd and I feel that this is not a normal behaviour, especially for such a small cluster (3 workers only !).

Can you provide me with some input regarding whether or not this is a normal behaviour ?

sylr:
128 is the limit of the P10 pricing tier which allocates 500 io/s per disk.

With disk like this the master nodes have a constant 30% IOWait load.

Etcd being overly needy in iops we need to change tier and 256 is the limit
of the next tier (P15) which allocates 1000 io/s.

See: https://azure.microsoft.com/en-us/pricing/details/managed-disks/

Fixes #2510

### DIFF
--- a/pkg/acsengine/const.go
+++ b/pkg/acsengine/const.go
@@ -106,12 +106,12 @@ const (
 	DefaultEtcdVersion = "3.2.16"
 	// DefaultEtcdDiskSize specifies the default size for Kubernetes master etcd disk volumes in GB
 	DefaultEtcdDiskSize = "256"
-	// EtcdDiskSizeGT3Nodes = size for Kubernetes master etcd disk volumes in GB if > 3 nodes
-	EtcdDiskSizeGT3Nodes = "512"
-	// EtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
-	EtcdDiskSizeGT10Nodes = "1024"
-	// EtcdDiskSizeGT20Nodes = size for Kubernetes master etcd disk volumes in GB if > 20 nodes
-	EtcdDiskSizeGT20Nodes = "2048"
+	// DefaultEtcdDiskSizeGT3Nodes = size for Kubernetes master etcd disk volumes in GB if > 3 nodes
+	DefaultEtcdDiskSizeGT3Nodes = "512"
+	// DefaultEtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
+	DefaultEtcdDiskSizeGT10Nodes = "1024"
+	// DefaultEtcdDiskSizeGT20Nodes = size for Kubernetes master etcd disk volumes in GB if > 20 nodes
+	DefaultEtcdDiskSizeGT20Nodes = "2048"
 	// DefaultReschedulerAddonName is the name of the rescheduler addon deployment
 	DefaultReschedulerAddonName = "rescheduler"
 	// DefaultMetricsServerAddonName is the name of the kubernetes Metrics server addon deployment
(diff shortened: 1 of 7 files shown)


khenidak:
Yes. We should up this default value. Id argue to go for 1023GB to get highest I/o specially for large clusters

jackfrancis:
Is 1023GB going to be compatible with all VM SKU sizes? And is it equally compatible with StorageAccount or ManagedDisk?

khenidak:
yes - disk size is not checked for premium/standard. only VM sku

also - I made a mistake. 2TB is the highest IO (7500 iops). 1TB(5000 iops)

khenidak:
this are not related values AFIAK. one is data disk size and one ephemeral disk size (AKA resource disk). The resource disk size has direct relation to VM sku. however the data disk size has no relation to VM sku. as in i can have 2 core machine with 4 TB data disk size

jackfrancis:
Got it. So the actual change is more like:

before: "We sell you Kubernetes for cheap!"
after: "We sell you Kubernetes for not so cheap!"

khenidak:
while there is a difference in cost, i doubt that it is that high. storage is dirt cheap. However to your point maybe we should keep it at 256GB and have a section for high `perf clusters tuning`

jackfrancis:
What I'd like to do is deliver a simple "default threshold" beyond which we bump to 1TB. Do you have a gut feel for the node count after which we should bump a cluster up to 1TB?

jackfrancis:
@khenidak @sylr See my recent commit to this PR. What do we think about that implementation? If we like it, I'll add test coverage.

sylr:
Am I right to understand that the more agent the cluster has, the more ETCD will be solicited ?

jackfrancis:
@sylr Roughly, yes. Enough correlation that it makes sense to nudge defaults upward if users don't provide a value.

CecileRobertMichon:
We should document this default pattern in the docs especially if it's likely to change pricing... Also I realized etcdDiskSize is missing from https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md#kubernetesconfig

pkg/acsengine/const.go
```diff
@@ -105,7 +105,13 @@ const (
...
 	DefaultEtcdVersion = "3.2.16"
 	// DefaultEtcdDiskSize specifies the default size for Kubernetes master etcd disk volumes in GB
-	DefaultEtcdDiskSize = "128"
+	DefaultEtcdDiskSize = "256"
+	// EtcdDiskSizeGT3Nodes = size for Kubernetes master etcd disk volumes in GB if > 3 nodes
+	EtcdDiskSizeGT3Nodes = "512"
+	// EtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
+	EtcdDiskSizeGT10Nodes = "1024"
+	// EtcdDiskSizeGT20Nodes = size for Kubernetes master etcd disk volumes in GB if > 20 nodes
+	EtcdDiskSizeGT20Nodes = "2048"
```
CecileRobertMichon: Here I think it would make sense to change naming to be `DefaultEtcdDiskSize<Insert_Number_of_Nodes>` since all of them are defaults that can be overridden by the user if I understand correctly. The name EtcdDiskSizeGT20Nodes + the comment makes it seem like it's static

PR #2437, opened 2018-03-12
Deregister node in api server during cordon-and-drain

dmitsh:
After upgrade operation, the api server sometimes lists already deleted nodes.
Adding explicit call to deregister nodes during cordon-and-drain.

### DIFF
--- a/pkg/operations/cordondrainvm.go
+++ b/pkg/operations/cordondrainvm.go
@@ -31,15 +31,20 @@ type drainOperation struct {
 type podFilter func(v1.Pod) bool
 
 // SafelyDrainNode safely drains a node so that it can be deleted from the cluster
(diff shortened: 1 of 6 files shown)


pkg/operations/cordondrainvm.go
```diff
@@ -62,7 +62,16 @@ func SafelyDrainNode(az armhelpers.ACSEngineClient, logger *log.Entry, masterURL
 
 	//Evict pods in node
 	drainOp := &drainOperation{client: client, node: node, logger: logger, timeout: timeout}
-	return drainOp.deleteOrEvictPodsSimple()
+	if err = drainOp.deleteOrEvictPodsSimple(); err != nil {
```
jackfrancis: The original usage of `return drainOp.deleteOrEvictPodsSimple()` is preferred here.

PR #2443, opened 2018-03-12
Adding Support for DC/OS 1.11.0

bernadinm:
This PR adds support for the latest DC/OS version 1.11.0

**Release note**:

```release-note
* Added DC/OS 1.11.0 version
* Tested and confirmed that the cluster launches and runs successfully.
* Updated documentation to include newer internal changes that are required for anyone who wants to update to a newer version in the future. 
```

### DIFF
--- a/docs/contributing-dcos.md
+++ b/docs/contributing-dcos.md
@@ -15,6 +15,9 @@ We now need to find the package GUID of each variant.
 In each template you should find a string that looks like: `dcos-config--setup_<Some GUID>`, this GUID is what we are looking for.
 Extract the GUIDs from the 3 differents templates, and them in `engine.go/getPackageGUID` for your specific DC/OS version.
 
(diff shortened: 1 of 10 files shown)


ionutbalutoiu:
Hello, Miguel.

I would like to make a suggestion. Could you please add a condition in the code [here](https://github.com/Azure/acs-engine/blob/master/parts/dcos/dcosmastervars.t#L113-L146) to allow the port 22 for the SSH into the first master node ?

Right now this is only done for DCOS 1.9 and I don't see any reason why to not enable it for 1.11 as well. 

This helps tools like `dcos-launch` to utilize that port to run integration tests.

Regards,
Ionut Balutoiu

bernadinm:
@ionutbalutoiu, I removed the port condition as discussed. It looks like there will be more work to enable and it may not be required if we have another solution in place for the time being. We can revisit this for another PR.

dmitsh:
Closed in favor of  https://github.com/Azure/acs-engine/pull/2609

PR #2457, opened 2018-03-13
add timeout to restart docker

CecileRobertMichon:
**What this PR does / why we need it**: add a retry to `systemctl restart docker` command to guard against stuck command

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomscript.sh
+++ b/parts/k8s/kubernetesmastercustomscript.sh
@@ -204,6 +204,8 @@ function ensureFilepath() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesmastercustomscript.sh
```diff
@@ -202,6 +202,8 @@ function ensureFilepath() {
     fi
 }
 
+function retrycmd_if_failure() { retries=$1; wait=$2; shift && shift; for i in $(seq 1 $retries); do ${@}; [ $? -eq 0  ] && break || sleep $wait; done; echo Executed \"$@\" $i times; }
```
jackfrancis: FYI it must be this line that is breaking things, should we revert to the bash-compatible version and retry?

PR #2405, opened 2018-03-07
Kubernetes v1.9.4 support

jackfrancis:
**What this PR does / why we need it**: Adds support for Kubernetes v1.9.4

https://github.com/kubernetes/kubernetes/releases/tag/v1.9.4

Fixes #2392 
Fixes #2151

**Release note**:

```
adds Kubernetes v1.9.4 support
```

PR #2458, opened 2018-03-13
orchestratorRelease by platform in vlabs

jackfrancis:
**What this PR does / why we need it**: Allows us to support `orchestratorRelease` for Windows and Linux while delivering distinct "latest" versions (e.g., `v1.9.4` for Linux and `v1.9.3` for Windows)

The current use-case requirement we have is: `"orchestratorRelease": "1.9"` should yield a `v1.9.4` cluster for Linux and `v1.9.3` for Windows. This PR will allow us to fulfill that in PR #2405 

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/api/common/helper_test.go
+++ b/pkg/api/common/helper_test.go
@@ -35,52 +35,52 @@ func Test_GetValidPatchVersion(t *testing.T) {
 }
 
 func Test_RationalizeReleaseAndVersion(t *testing.T) {
-	version := RationalizeReleaseAndVersion(Kubernetes, "", "")
+	version := RationalizeReleaseAndVersion(Kubernetes, "", "", false)
 	if version != KubernetesDefaultVersion {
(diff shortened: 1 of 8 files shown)


pkg/api/common/helper_test.go
```diff
@@ -32,56 +32,99 @@ func Test_GetValidPatchVersion(t *testing.T) {
...
+	}
+
+	expected := "99.1.2"
+	version = GetLatestPatchVersion("99.1", []string{"99.1.1", "99.1.2-beta.5", expected})
+	if version != expected {
+		t.Errorf("GetLatestPatchVersion returned the wrong latest version, expected %s", expected)
+	}
 }
 
 func Test_RationalizeReleaseAndVersion(t *testing.T) {
```
qike-ms: Add test cases to test for windows?

PR #2485, opened 2018-03-19
WIP: Renaming resources to something more meaningful 

### LINKED ISSUE #2367: Allow custom resource name prefixes
Hi,

Currently acs-engine generates name prefixes like this :

```
"masterLbName": "[concat(variables('orchestratorName'), '-master-lb-', variables('nameSuffix'))]",
"masterVMNamePrefix": "[concat(variables('orchestratorName'), '-master-', variables('nameSuffix'), '-')]",
```

Which ends up having resources named :

  - k8s-master-89789797-0
  - k8s-master-lb-89789797-0
  - ...

This is fine when you are managing **1 cluster** but as soon as you have at least 2 **it becomes a nightmare** as you are absolutely not able to tell to which cluster your resource belongs to by looking at its name.

This can be really dangerous because you might end up doing things on a cluster which it was not intended for.

Also if you are monitoring things, an alert telling you k8s-master-89789797-0 is down does not tell you if we are talking about a development server or a production one.

We should be able to configure those prefixes globally, for instances I'd like my resources to be named as followed, `<cluster-name>-<resource-type>-<
#>` :

 - k8s1-us-dev-master-[0-9]+
 - k8s1-us-dev-agent-[0-9]+
 - etc ...

Regards.

sylr:

Rename resources
from: `<orchestrator>-<resource_type>-<########>[-<index>]`
to: `<cluster_name>-<resource_type>[-<index>]`

Why? Because current naming convention makes managing **several** acs-engine generated clusters a nightmare due to the lack of hints in the resources name about which cluster the resource belong to.

**Which issue this PR fixes** : fixes #2367

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -12,7 +12,7 @@
 {{end}}
       ],
       "location": "[variables('location')]",
-      "name": "[concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic1')]",
+      "name": "[concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic0')]",
       "properties": {
 {{if .IsCustomVNET}}
         "networkSecurityGroup": {
@@ -126,7 +126,7 @@
         "[concat('Microsoft.Storage/storageAccounts/',variables('storageAccountPrefixes')[mod(add(add(div(copyIndex(variables('{{.Name}}Offset')),variables('maxVMsPerStorageAccount')),variables('{{.Name}}StorageAccountOffset')),variables('dataStorageAccountPrefixSeed')),variables('storageAccountPrefixesCount'))],variables('storageAccountPrefixes')[div(add(add(div(copyIndex(variables('{{.Name}}Offset')),variables('maxVMsPerStorageAccount')),variables('{{.Name}}StorageAccountOffset')),variables('dataStorageAccountPrefixSeed')),variables('storageAccountPrefixesCount'))],variables('{{.Name}}DataAccountName'))]",
   {{end}}
 {{end}}
-        "[concat('Microsoft.Network/networkInterfaces/', variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic1')]",
+        "[concat('Microsoft.Network/networkInterfaces/', variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic0')]",
         "[concat('Microsoft.Compute/availabilitySets/', variables('{{.Name}}AvailabilitySet'))]"
       ],
       "tags":
@@ -153,7 +153,7 @@
         "networkProfile": {
           "networkInterfaces": [
             {
-              "id": "[resourceId('Microsoft.Network/networkInterfaces',concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic1'))]"
+              "id": "[resourceId('Microsoft.Network/networkInterfaces',concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic0'))]"
             }
           ]
         },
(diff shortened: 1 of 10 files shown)


sylr:
@CecileRobertMichon @jackfrancis Do you have some feedbacks about this ?

jackfrancis:
I like where this is going. Realistically, we'll have to put this through a lot of backwards compatibility baking, so please be patient with us on this. To explain why it may sit in the queue for a while. :)

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -12,7 +12,7 @@
 {{end}}
       ],
       "location": "[variables('location')]",
-      "name": "[concat(variables('{{.Name}}VMNamePrefix'), 'nic-', copyIndex(variables('{{.Name}}Offset')))]",
+      "name": "[concat(variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-nic1')]",
```
CecileRobertMichon: Why the numbering here? Is this addressing a use case of several NICs?

pkg/api/vlabs/types.go
```diff
@@ -30,6 +30,7 @@ type ContainerService struct {
 
 // Properties represents the ACS cluster definition
 type Properties struct {
+	Name                    string                   `json:"clusterName,omitempty"`
```
CecileRobertMichon: What happens to the resource names if the user doesn't provide a cluster name?

CecileRobertMichon:
Looks great so far @sylr, I added a few comments. Have you thought of making the DNS prefix dependent on the cluster name? Or do we want to keep those as two independent fields?

parts/k8s/kubernetesagentresourcesvmas.t
sylr: Not really, I thought if one day a one present itself you will already have a numbered nic.

pkg/api/vlabs/types.go
sylr: Good question, Curretnly It should fail I guess because I did implement but I guess it should default to the old `<orchestrator>-<########>-...` name.

Do you agree ?

sylr: It won't really be like before cause you'll have `<orchestrator>-<########>-master-1` instead of `<orchestrator>-master-<########>--1` but it's pretty close.

CecileRobertMichon: It'd be good to keep the exact same name as before for users who don't provide a cluster name I think. That way we don't introduce change for people who have in place tools that rely on the existing name and using the new naming is optional.

sylr: Seems difficult as I'm modifying the templates to be like `<cluster_name>-<type>-<#>[-...]`.

What I can probably do is to have `<cluster_name>` default to `<orchestrator>-<########>` if not provided.

CecileRobertMichon: Let's make sure that naming resources stays the same if clusterName is not provided by the user to ensure backward compatibility

parts/k8s/kubernetesagentresourcesvmas.t
CecileRobertMichon: I think for now it's fine to leave it without numbering since there is only one per VM. If you do see value in numbering, it should start from 0 to be consistent with numbering of other resources.

jackfrancis:
@sylr We don't have inclusion of a change like this on the near term roadmap. Let's close this and re-open later (or refer to it) when we have cycles for this type of improvement

jackfrancis:
@sylr Understood. Generally, I think it's healthy project hygiene to communicate bad news. Not at all suggesting that folks shouldn't advocate for their changes even with resistance, just that: "based on the low likelihood of this being addressed in the near future, it makes the most sense for this PR to be closed".

parts/k8s/kubernetesmasterresources.t
```diff
@@ -657,7 +657,7 @@
               "createOption": "Empty"
               ,"diskSizeGB": "[variables('etcdDiskSizeGB')]"
               ,"lun": 0
-              ,"name": "[concat(variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')),'-etcddisk')]"
+              ,"name": "[concat(variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')),'_EtcdDisk')]"
```
pascalgn: I think all other resource names are lower case (e.g. _clustername-master-availabilityset_). Are you sure about deviating from that convention here?

PR #2494, opened 2018-03-20
[RFC] Add support for customizing nsg's for inbound traffic to k8s master node

vosst:
This PR adds support for customizing the two network security rules `allow_kube_ssh` and `allow_kube_tls` by means of altering the `apimodel` that is passed to `acs-engine deploy`. Being able to alter those rules at cluster creation time (as opposed to using `az network` after the fact) increases our efficiency quite significantly. More to this, it allows us to specify multiple source address prefixes.

We would like to gather early feedback for our changes and ask for guidance whether the approach taken here is reasonable. More specifically:
 * we (somewhat naively :)) adjusted api versions to ensure that `sourceAddressPrefixes` and `destinationAddressPrefixes` are supported. How would the API versioning be handled in production code?
 * Is the abstraction around `NetworkAccessProfile` and `MasterNetworkAccessProfile` reasonable or should we rather resort to using the respective types from the Azure golang SDK?
 * Is the integration with the api loading infrastructure and selection of defaults reasonable?

Looking forward to your feedback.


pkg/api/types.go
```diff
@@ -301,24 +301,43 @@ type DcosConfig struct {
...
+	OSDiskSizeGB             int                         `json:"osDiskSizeGB,omitempty"`
+	VnetSubnetID             string                      `json:"vnetSubnetID,omitempty"`
+	VnetCidr                 string                      `json:"vnetCidr,omitempty"`
+	FirstConsecutiveStaticIP string                      `json:"firstConsecutiveStaticIP,omitempty"`
+	Subnet                   string                      `json:"subnet"`
+	IPAddressCount           int                         `json:"ipAddressCount,omitempty"`
+	StorageProfile           string                      `json:"storageProfile,omitempty"`
+	HTTPSourceAddressPrefix  string                      `json:"HTTPSourceAddressPrefix,omitempty"`
+	OAuthEnabled             bool                        `json:"oauthEnabled"`
+	NetworkAccessProfile     *MasterNetworkAccessProfile `json:"networkAccessProfile,omitEmpty"`
```
sylr: I'm curious why you chose to build a struct `MasterNetworkAccessProfile` with hardcoded `SSH` and `TLS` properties instead of builds an array of `NetworkAccessProfile`s, i.e: 
```go
NetworkAccessProfiles []NetworkAccessProfile `json:"networkAccessProfiles,omitEmpty"
```

Wouldn't that be more flexible if we want to add more rules to the NSG in the future ?

vosst: Definitely. I kept the scope of the change limited to a precise purpose to facilitate the discussion a little. With that, I'm obviously happy to adjust once we agree on the general approach.

jackfrancis:
@vosst I will try to make time to add some remarks this week. Please @ me if I don't!

vosst:
@jackfrancis ack and thx, I merged master and resolved the conflict. Looking forward to your feedback.

pkg/api/types.go
```diff
@@ -302,24 +302,43 @@ type DcosConfig struct {
 	RegistryPass            string `json:"registryPassword,omitempty"`
 }
 
+// NetworkAccessProfile describes inbound traffic rules for network access.
+type NetworkAccessProfile struct {
+	SourceAddressPrefix        string   `json:"sourceAddressPrefix"`
```
jackfrancis: Let's get rid of the "single" `SourceAddressPrefix` property. Reading the templates above, it looks like we have two paths: user specified one rule, or a bunch of rules, if the former, apply it, if the latter, enumerate and apply them all. I would prefer we just do the enumeration no matter what, so if we want to specify a single source address prefix (or source port range, or dest address prefix, or dest port range) we the config simply declares an array with one item.

pkg/api/types.go
jackfrancis: I agree with @sylr that we want to ship this with a general implementation, as changing the type spec later will be costly.

jackfrancis:
@vosst I added a couple notes, overally I'm pleased that you are successfully using this pattern, and I'd like to see it adopted!

pkg/api/types.go
vosst: ACK, how would we map to the individual nsg's then? Right now, when applying the template, I do an explicit selection. Maybe a map indexed with the name is more flexible and still allows us to select the right profile for a specific NSG.

vosst:
Great, thanks for the feedback. I'll take care of merging master tomorrow morning.

CecileRobertMichon:
Closing since we've moved to https://github.com/Azure/aks-engine

PR #2525, opened 2018-03-26
Fix the issue that azure file volume could not be dynamically created in v1.7

andyzhangx:
current `kubernetesmasteraddons-azure-cloud-provider-deployment.yaml` does not work for k8s 1.7, which means azure file volume could not be dynamically created in v1.7 now, this PR fix this issue. use v1beta1 for ClusterRole to support k8s 1.7

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Fix the issue that azure file volume could not be dynamically created in v1.7
```

@jackfrancis

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: rbac.authorization.k8s.io/v1
+apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


andyzhangx:
This PR should also back ported to AKS

jackfrancis:
@andyzhangx  is this change compatible with all k8s versions >= 1.7?

andyzhangx:
@jackfrancis yes, all `ClusterRole` objects are all using `v1beta1`, here is an example:
https://github.com/Azure/acs-engine/blob/master/parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml#L7

PR #2528, opened 2018-03-26
Move some AAD args out of the static list

### LINKED ISSUE #2527: Custom user defined apiserver args overriden
https://github.com/Azure/acs-engine/blob/08ac0e253c9a7835a1182cc373c69c8ebd022be9/pkg/acsengine/defaults-apiserver.go#L121-L131

Why is there a orchestratorProfile.kubernetesConfig.apiServerConfig section if we can't use it ?

I need to override the default value of --oidc-username-claim to "upn" but this code won't let me.

sylr:
**Which issue this PR fixes** : fixes #2527

### DIFF
--- a/pkg/acsengine/defaults-apiserver.go
+++ b/pkg/acsengine/defaults-apiserver.go
@@ -37,6 +37,21 @@ func setAPIServerConfig(cs *api.ContainerService) {
 		"--v":                          "4",
 	}
 
+	// Windows apiserver config overrides
+	// TODO placeholder for specific config overrides for Windows clusters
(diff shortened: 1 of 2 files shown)


jackfrancis:
Thanks for this @sylr. To explain the reasoning behind the initial implementation:

When making these configuration interfaces (kubelet, apiserver, controller-manager, etc) user-configurable and generally, our approach is conservative in terms of exposing something for user-override. Essentially, we default to "static" (i.e. non-user-overridable), and then as use-cases arise for overridable config (such as yours w/ OID user/group claim ids) we evolve the code appropriately.

PR #2541, opened 2018-03-27
remove k8s v1.10.0 pre-release versions

jackfrancis:
**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
NONE
```


jackfrancis:
This change is not backwards compatible with the current upgrade implementation!

PR #2547, opened 2018-03-28
containerd: just install tarball release

jessfraz:
make building so much faster :-)


jessfraz:
jenkins build started here: https://jenkins.azure-containers.io/view/ACS-E/job/k8s-clear-containers/56/console

PR #2557, opened 2018-03-29
Improve non-masquerade-cidr

sylr:
--non-masquerade-cidr defaults to 10.0.0.0/8 if you don't define it.

That's way too large. If you are in a VNet within 10.0.0.0/8 setting this will prevent pods to contact resources in the VNet because everything in 10.0.0.0/8 will be considered to be within the kubernetes cluster.

Instead let us define it to the value of cluster subnet.


mcwienczek:
Yes, that one! Thanks!

PR #2566, opened 2018-03-30
Configure kube-dns to expose metrics

sylr:


jackfrancis:
@lachie83 Can you kindly take a look at this and weigh in?

parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
     metadata:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ""
+        prometheus.io/scrape: "true"
```
lachie83: Should these annotations be on the service rather than the deployment?

parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -78,6 +80,9 @@ spec:
         - "--dns-port=10053"
         - "--v=2"
         - "--config-dir=/kube-dns-config"
+        env:
+        - name: PROMETHEUS_PORT
+          value: "10055"
```
lachie83: Is this port fixed? Can you please point me at the docs for this?

jackfrancis:
@lachie83 based on @sylr's responses are there any outstanding items in this PR?

parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
carlpett: I believe the recommended approach is to annotate the service (though I cannot find a black and white statement from the team about it now). The pods of the service will still get scraped if using a `role` of `endpoints` in the `kubernetes_sd_configs`.
The exposed metrics will be the same either way, but there are a bunch of extra labels available this way.

Also, do note that this annotation is not really standardized, even if it is common in examples. We do not use it at all, and rather use `ServiceMonitor`s from the prometheus-operator. But perhaps it is a reasonable default to ship anyway? Is it set on other kube* things with metrics in acs-engine?

sylr:
@jackfrancis can we merge this please ?

PR #2581, opened 2018-04-03
[feature] Prevent overwriting existing cluster resources in output directories

skinny:

This PR prevents accidentally overwriting existing cluster deployment files in any output directory. This can be overruled by passing the `-f/--force-overwrite` parameter with the deploy command

n/a

**Release note**:
```NONE
```


CecileRobertMichon:
I think the change makes sense from a user's perspective, especially since the _output directory cannot be re-obtained as acs-engine stands right now... I think it'd be good to document the new flag in /docs/deploy.md and clearly state the default behaviour. I think not overwriting by default makes sense and, frankly, is better UX so IMO changing the default behaviour is completely acceptable. Also, the default output directory is the DNS prefix so in most cases (except when a user re-deploys) an existing cluster it should be unique. Only thing we need to verify is whether this change can have an impact on upgrade and scale as they overwrite the output dir with the newest version of the cluster (off the top of my head I don't think so)....

skinny:
I only added the check to the deploy command especially for the reason of allowing upgrade and scale to work as expected. They read the existing generated apimodel.json so that should be fine. This is merely a safeguard for yourself. (have had two coworkers new to acs-engine complaining about this while they were experimenting with this)

jackfrancis:
@CecileRobertMichon added docs, ready for a review

docs/kubernetes/deploy.md
```diff
@@ -67,6 +67,8 @@ kubernetes-dashboard is running at https://contoso-apple-59769a59.westus2.clouda
 To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
 ```
 
+Administrative note: By default, the directory where acs-engine stores cluster configuration (`_output/contoso-apple` above) won't be overwritten as a result of subsequent attempts to deploy a cluster using the same `--dns-prefix`) To re-use the same resource group name repeatedly, include the `--force-overwrite` command line option with your `acs-engine deploy` command. On a related note, include an `--auto-suffix` option to append a randomly generated suffix to the dns-prefix, if your workflow requires a common prefix across multiple cluster deployments. Using the `--auto-suffix` pattern will guarantee that each deployment's configuration artifacts will be stored locally under a discrete `_output/<resource-group-name>/` directory.
```
CecileRobertMichon: I would just make it clear that auto-suffix `appends a compressed timestamp to the dnsPrefix to ensure a unique cluster name`, as opposed to only storing the output in a unique directory. Otherwise looks good.

PR #2585, opened 2018-04-03
[PoC] Add 'dump agent raw data' option for 'generate' command

yolo3301:
After a cluster is provisioned, sometimes a user might want to add extra agent nodes by themselves. All they need is the custom data and the provision script. Today the user will have to dig into the ARM template generated to find those, which is kind of hard. The user needs to know which custom data to look at and the provision script is even base64 encoded.

This change can optionally dump the raw data (aka. custom data and provision script) when generating the ARM template. Later, the user can leverage the dumped files to add new agent nodes.

**How is this different from scale command?**
1. No dependency on acs-engine. Users can directly use the dumped files to provision new agent nodes
2. No limited to existing agent pools. Scale command only supports scale an existing agent pool. With dumped files, users can freely add agent nodes in existing or new agent pools.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/Azure/acs-engine/pkg/acsengine"
 	"github.com/Azure/acs-engine/pkg/acsengine/transform"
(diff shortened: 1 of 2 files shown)


yolo3301:
@yangl900 @wenwu449

CecileRobertMichon:
Closing because we've moved to https://github.com/Azure/aks-engine

PR #2600, opened 2018-04-05
Add max pods per node support at the agent pool level.

JunSun17:
We need to add max pods per node support at the agent pool level. This is part of feature request for the AKS Custom VNET feature.

Also remove the original non-working way of passing MAX_PODS logic.


examples/kubernetes-custom-vnet/kubernetes-custom-vnet-maxpods.json
```diff
@@ -0,0 +1,60 @@
+{
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
+      "kubernetesConfig": {
+        "networkPolicy": "azure",
+        "maxPods": 30
```
CecileRobertMichon: Is this number going to be used at all or is it overridden everywhere?

pkg/acsengine/defaults-kubelet.go
```diff
@@ -35,7 +35,6 @@ func setKubeletConfig(cs *api.ContainerService) {
 		"--cluster-domain":                  "cluster.local",
 		"--network-plugin":                  "cni",
 		"--pod-infra-container-image":       cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase + KubeConfigs[o.OrchestratorVersion]["pause"],
-		"--max-pods":                        strconv.Itoa(DefaultKubernetesKubeletMaxPods),
```
CecileRobertMichon: Seems like DefaultKubernetesKubeletMaxPods isn't used anywhere in the code anymore? If that's the case we should remove it from const.go

pkg/acsengine/defaults-kubelet.go
```diff
@@ -97,6 +96,22 @@ func setKubeletConfig(cs *api.ContainerService) {
...
 
+		var defaultMasterMaxPods int
+		// set defaultMasterMaxPods to OrchestratorProfile.KubernetesConfig.MaxPods if exist.
+		if o.KubernetesConfig != nil && o.KubernetesConfig.MaxPods != 0 {
+			defaultMasterMaxPods = o.KubernetesConfig.MaxPods
+		} else {
+			if o.IsAzureCNI() {
+				defaultMasterMaxPods = DefaultKubernetesMaxPodsVNETIntegrated
+			} else {
+				defaultMasterMaxPods = DefaultKubernetesMaxPods
```
CecileRobertMichon: Is this supposed to be `DefaultKubernetesKubeletMaxPods` instead of `DefaultKubernetesMaxPods`?

examples/kubernetes-custom-vnet/kubernetes-custom-vnet-maxpods.json
JunSun17: This number will be honored if no `--max-pods` is set in KubernetesConfig.KubeletConfig. This is backward compatibility case talked in the previous comment.

pkg/acsengine/defaults-kubelet.go
JunSun17: Yes, I will remove it.

pkg/acsengine/defaults-kubelet.go
JunSun17: I think this is due to the historical changes discusses in the previous comments. Right now, we are converging to use DefaultKubernetesMaxPodsVNETIntegrated for Azure CNI and DefaultKubernetesMaxPods for Kubenet. So DefaultKubernetesKubeletMaxPods can be retired.

examples/kubernetes-custom-vnet/kubernetes-custom-vnet-maxpods.json
CecileRobertMichon: I mean in this example file, since --max-pods is set for both agent and master kubelet. Won't it be confusing for users if we include a value in an example file that is ignored?

pkg/acsengine/defaults-kubelet.go
CecileRobertMichon: Sounds good

examples/kubernetes-custom-vnet/kubernetes-custom-vnet-maxpods.json
JunSun17: Actually no master KubernetesConfig.KubeletConfig is set in this example. The above value will be used for max pods at master nodes. The other two values are used for agentpool1 and nodepool2.

CecileRobertMichon: My bad, I misread, I saw two values and assumed it was master and agent. Looks good!

JunSun17:
@jackfrancis the default values are not properly honored if no "--max-pods" is defined in kubeletConfig. It was always set to 110 no matter network policy is Azure CNI or Kubenet, and it does not honor the old cluster wide setting OrchestratorProfile.KubernetesConfig.MaxPods.

jackfrancis:
Understood. Can confirm that this is the acs-engine-provided configuration (no user override) setting for both "azure" and "none" vals of "networkPolicy":

`"--max-pods": "110"`

What is the problem with the above behavior? Do we want different settings depending on Azure CNI vs kubenet?

Also, the deprecation of `maxPods` is intentional (it is non-trivial to remove an API typed property and not break backwards compatibility so that is why it remains).

What is the high level intention here? What are we trying to do?

JunSun17:
@jackfrancis Based on our existing code and doc, it seems the default max pods per node should be 30 for Azure CNI and 110 for Kubenet. Also, this change makes backward compatibility support for OrchestratorProfile.KubernetesConfig.MaxPods.

I just happen to see this when testing max pods settings and think this PR should complement the current way of handling max pods.

jackfrancis:
O.K., good callout on the broken docs. I propose that we just delete the lines in `docs/kubernetes/features.md` that mention `maxPods`. Basically the paragraph "When using Azure integrated networking the maxPods setting will be..." here:

https://github.com/Azure/acs-engine/blob/master/docs/kubernetes/features.md

I'd prefer to maintain the practical deprecation of `OrchestratorProfile.KubernetesConfig.MaxPods` and encourage folks to use the `kubeletConfig` overrides for tuning clusters.

@sharmasushant @tamilmani1989 for any thoughts on the criticality of changing the current `--max-pods` default setting of 110 for Azure CNI clusters.

JunSun17:
@jackfrancis I think initially there are gonna be a reason why Azure CNI does not choose the default 110 for kubenet, maybe probably the 4k limit from Azure CNI?

By saying "practical deprecation of OrchestratorProfile.KubernetesConfig.MaxPods", do you mean we do not honor this value anymore? But is that breaking backward compatibility?

jackfrancis:
To your 2nd question: yes, we are breaking backwards compatibility as a trade-off for not having to maintain multiple interface inputs for a single configuration property. Not ideal, but I think the best choice for the long-term. The practical effect is:

1) a user's prior setting will be ignored and overriden by the new defaults applied to `--max-pods`
2) a user could optionally manually alter the api model before upgrade to transfer the configuration to the new pattern

I'm happy to defer to the Azure CNI team as far as providing a different value for `--max-pods` for Azure CNI schedulable nodes. It's been 110 for several versions at this juncture.

JunSun17:
@jackfrancis Let's finish this PR to reduce one task from my basket. This PR should only make things better since it adapts to the our current documents and also provide backward compatibility. At least, the current way of setting profile.IPAddressCount (from cluster level max pods setting) will not work for the custom VNET feature and it needs to be changed.

How about I check in this now to unblock the custom VNET work? we can advertise the deprecation of the maxPods parameter and make doc adjustment later in a non-urgent manner.

jackfrancis:
Thanks @JunSun17 for identifying all this inconsistent behavior around MaxPods. I'd like to formalize/rationalize the deprecation of `KubernetesConfig.MaxPods`, and also unblock your work. I've opened this PR, which I am adovcating will replace this one: #2618

PR #2604, opened 2018-04-05
attach NSGs to nics, not to the subnet

jim-minter:
This PR attaches the NSG to the nics, not to the subnet, in k8s deployments.  This is a precursor to being able to attach different NSGs to different VMs where required.

Issue #2540, opened 2018-03-27
API model for "azure context"


jim-minter:
It would be useful to have a single standard place in the acs-engine API model to contain the cluster "azure context" (e.g. tenant id, subscription id, resource group and location).

Having these values defined up front includes the following advantages:
- these data can be statically templated into cluster configuration files, e.g. azure.conf, up front on the client side rather than as a dynamic server-side script
- strengthened data model with all input data in one place rather than some pieces learned ad-hoc during ARM template execution
- improved testability

Proposal:

```golang
# pkg/api/vlabs/types.go

// Properties represents the ACS cluster definition
type Properties struct {
	// ...
	AzProfile AzProfile `json:"azProfile,omitempty"` // new field added to struct
}

// AzProfile holds the azure context for where the cluster resides
type AzProfile struct {
	TenantID       string `json:"tenantId,omitempty"`
	SubscriptionID string `json:"subscriptionId,omitempty"`
	ResourceGroup  string `json:"resourceGroup,omitempty"`
	Location       string `json:"location,omitempty"`
}
```

Today some acs-engine sub-commands allow command-line arguments such as --subscription-id, --resource-group, --location to be specified.  It's proposed that these command-line arguments would override fields in the AzProfile of an input model.  The command-line arguments could optionally be extended to add --tenant-id, and/or the tenant id could be defaulted based on the discovered tenant when acs-engine connects to azure.

After adding the AzProfile struct, existing implementations which currently discover these values dynamically in the execution of the ARM template should consider moving to using the preconfigured values.

jim-minter:
@jackfrancis @pweil- @kargakis ptal

khenidak:
I understand why we need this in the model. I'm however concerned with:
1. Asking the user to provide an information we get from somewhere else (Arm template?). I think today we drive these as env vars into a custom script extension. 
2. Handling the cases where the user provide `resource-group:Gx location:Lx` and during deploy he deploy to `resource-group:Gy location:Ly`

jackfrancis:
Thanks @jim-minter for scaffolding this design proposal. My thoughts:

- Basic interface spec proposal (novel struct under `Properties`) looks good
- roger on your defined order of configuration precedence:
  1. CLI context args
  2. api model
  3. inherited code-enforced defaults (probably not applicable for these properties in particular as they are all context-hungry)
[...]

PR #2611, opened 2018-04-06
Add first cut of the OpenShift orchestrator 

kargakis:
@jim-minter @pweil- @jackfrancis @khenidak


kargakis:
Stuff contained in this initial PR:
* reuse of the k8s templates
* master/nodes are configured to talk to the azure api
* basic htpasswd support
* installed components: router, registry, service catalog, template service broker

pkg/certgen/files.go
```diff
@@ -0,0 +1,97 @@
+package certgen
```
kargakis: @jim-minter we should probably move certgen behind an openshift package.

parts/agentparams.t
```diff
@@ -38,6 +38,27 @@
       },
       "type": "string"
     },
+    "{{.Name}}IsOpenShiftInfra": {
```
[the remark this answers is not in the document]
jim-minter: yes, it is not needed - will submit a PR to kargakis/acs-engine

parts/k8s/kubernetesagentvars.t
```diff
@@ -24,4 +24,5 @@
     "{{.Name}}osImageSKU": "[parameters('{{.Name}}osImageSKU')]",
     "{{.Name}}osImagePublisher": "[parameters('{{.Name}}osImagePublisher')]",
     "{{.Name}}osImageVersion": "[parameters('{{.Name}}osImageVersion')]",
-
+    "{{.Name}}osImageName": "[parameters('{{.Name}}osImageName')]",
```
kargakis: Opened https://github.com/Azure/acs-engine/pull/2613 for pushing these changes; we should already have them in this repo (mea culpa).

examples/openshift.json
```diff
@@ -0,0 +1,87 @@
...
+        "imageReference": {
+          "name": "YOUR IMAGE",
+          "resourceGroup": "YOUR IMAGE RESOURCE GROUP"
+        },
+        "name": "compute",
+        "storageProfile": "ManagedDisks",
+        "vmSize": "Standard_D4s_v3"
+      },
+      {
+        "availabilityProfile": "AvailabilitySet",
```
khenidak: hmm can we have another with VMSS. We are trying to slowly walk away from av-set

jim-minter: at the moment we're deploying openshift based on k8s 1.9 - therefore we were supposing that we couldn't use scalesets yet?

khenidak: Right. Should we add a validation for for using VMSS w/ OpenShift

examples/openshift.json
```diff
@@ -0,0 +1,87 @@
...
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "OpenShift",
+      "openShiftConfig": {
+        "clusterUsername": "demo",
+        "clusterPassword": "demo"
+      }
+    },
+    "azProfile": {
```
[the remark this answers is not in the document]
jackfrancis: I'm O.K. with this being a general property:

https://github.com/Azure/acs-engine/issues/2540

There is potential general-purpose utility here, so I think this is the right place.

parts/k8s/kubernetesagentvars.t
jackfrancis: #2613 has been merged :)

examples/openshift.json
jim-minter: https://github.com/kargakis/acs-engine/pull/83 , thanks

parts/k8s/kubernetesparams.t
```diff
@@ -570,6 +570,7 @@
       "metadata": {
         "description": "The docker engine version to install."
       },
+      {{if IsKubernetes}}
```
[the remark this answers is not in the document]
jim-minter: https://github.com/kargakis/acs-engine/pull/84 , thanks

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -31,6 +31,14 @@
               "subnet": {
                 "id": "[variables('{{$.Name}}VnetSubnetID')]"
               }
+{{if $.IsOpenShiftInfra }}
```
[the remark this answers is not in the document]
jim-minter: At the moment we are deploying OpenShift including k8s 1.9, but this LB on the infra nodes is required for OpenShift's [Route](https://docs.openshift.com/container-platform/3.9/architecture/networking/routes.html#architecture-core-concepts-routes) functionality

pkg/certgen/files.go
jim-minter: I agree

examples/openshift.json
```diff
@@ -0,0 +1,87 @@
...
+      "adminUsername": "cloud-user",
+      "ssh": {
+        "publicKeys": [
+          {
+            "keyData": "YOUR KEY"
+          }
+        ]
+      }
+    },
+    "servicePrincipalProfile": {
```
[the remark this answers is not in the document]
kargakis: We use both clientId and secret from here to construct azure.conf that is then untarred inside the VM. Unclear whether switching to MSI will affect that process. Is this a requirement for merging this PR or can it be addressed in a follow-up?

parts/k8s/kubernetesmasterresources.t
```diff
@@ -94,6 +206,36 @@
               "sourcePortRange": "*"
             }
           },
+{{end}}
+{{if IsOpenShift}}
+          {
```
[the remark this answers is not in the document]
kargakis: I am [suggesting](https://github.com/Azure/acs-engine/pull/2604#issuecomment-379757109) to address this in a follow-up once this PR is merged.

examples/openshift.json
kargakis: @jim-minter's changes pushed in this PR

kargakis:
> General q - Have you considered the impact on scale/upgrade commands?

I expect scale to be simple since we use node images and do node bootstraping. Unclear how we want to do upgrades with the unmanaged setup though. cc @jim-minter @pweil-

parts/k8s/kubernetesagentvars.t
kargakis: fixed

parts/agentparams.t
```diff
@@ -38,6 +38,27 @@
...
+      "defaultValue": "",
+      "metadata": {
+        "description": "Whether the current agent is an OpensShift infrastructure node or not."
+      },
+      "type": "bool"
+    },
+    "{{.Name}}osImageName": {
```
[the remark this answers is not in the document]
kargakis: fixed

parts/agentparams.t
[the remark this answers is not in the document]
kargakis: fixed

jim-minter:
> > General q - Have you considered the impact on scale/upgrade commands?

> I expect scale to be simple since we use node images and do node bootstraping. Unclear how we want to do upgrades with the unmanaged setup though. cc @jim-minter @pweil-

We have a card to track handling scaling.  Certainly we hope that scale up should be straightforward; we need to look at what acs-engine is doing today for scale down.  I don't know that we're planning any level of upgradability for the single VM-based master scenario.

kargakis:
Working on an e2e test [here](https://github.com/kargakis/acs-engine/pull/97). In the meantime, I would appreciate ACKs/nACKs for this PR. I don't want to push more stuff from my fork atm so consider this PR complete from my side (minus the e2e test which may also land on a separate PR).

examples/openshift.json
```diff
@@ -0,0 +1,67 @@
...
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "OpenShift",
+      "openShiftConfig": {
+        "clusterUsername": "demo",
+        "clusterPassword": "demo"
+      }
+    },
+    "azProfile": {
+      "tenantId": "YOUR TENANT ID",
```
jackfrancis: nit: the current idiom is to leave priviileged values (or properties for which it doesn't make sense to have default example vals) blank in example api models, e.g., `"tenantId": "",`

parts/k8s/openshiftmasterscript.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash -x
```
jackfrancis: Just a consideration: should we put these files in `parts/openshift`?

pkg/api/agentPoolOnlyApi/vlabs/types.go
```diff
@@ -129,6 +129,7 @@ type AgentPoolProfile struct {
 	AvailabilityProfile string `json:"availabilityProfile"`
 	StorageProfile      string `json:"storageProfile" validate:"eq=StorageAccount|eq=ManagedDisks|len=0"`
 	VnetSubnetID        string `json:"vnetSubnetID,omitempty"`
+	IsOpenShiftInfra    bool   `json:"isOpenShiftInfra,omitempty"`
```
jackfrancis: What are the pros/cons of adding this `isOpenShiftInfra` value vs inferring via the existing `orchestratorType` property value of `"OpenShift"`? Ideally, we would not add this new typed property. (Understand there may be considerations, just would like to clarify.)

jim-minter: OpenShift requires 2 types of agent nodes - one type "infra", which will run openshift value add workload (e.g. embedded registry, router, etc.), and another type (non-"infra") which runs end-user workload.  We need to distinguish these somehow and this is the switch that does it.

I suppose it could be done by the agentpoolprofile name but overloading that is a bit nasty.  I wondered whether instead of `IsOpenShiftInfra bool` a `IsInfra bool` or a `AgentRole string` (validated as an enum) might be more useful to others?

parts/k8s/openshiftmasterscript.sh
jim-minter: Makes sense to me - https://github.com/kargakis/acs-engine/issues/98 as a follow-up

examples/openshift.json
jim-minter: https://github.com/kargakis/acs-engine/issues/99 as a follow-up

jackfrancis: ACK

parts/k8s/openshiftmasterscript.sh
jackfrancis: ACK

CecileRobertMichon:
I don't see any update to the existing documentation (https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md) to detail the new apimodel properties or new documentation is there a plan to add some?

pkg/api/agentPoolOnlyApi/vlabs/types.go
kargakis: @jim-minter's changes pushed in this PR

kargakis:
@CecileRobertMichon thanks for bringing this up, I don't want to block this PR so how about delivering docs as part of a follow-up? I created https://github.com/kargakis/acs-engine/issues/104 to track, on my top priorities alongside the e2e test.

jim-minter:
@jackfrancis I believe this is ready to go.

PR #2621, opened 2018-04-06
Fix nil linuxprofile

zqingqing1:
**What this PR does / why we need it**: Need to check if LinuxProfile is nil, otherwise raised nil pointer panic.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


weinong:
UT?

jackfrancis:
@zqingqing1 @weinong the new testdata file suggests nil pointer execution flow is still possible (see latest test run)

zqingqing1:
yeh, working on this

pkg/api/apiloader.go
```diff
@@ -364,3 +381,26 @@ func setContainerServiceDefaultsv20170131(c *v20170131.ContainerService) {
 		}
 	}
 }
+
+const (
+	// SSHKeySize is the size (in bytes) of SSH key to create
+	SSHKeySize = 4096
+)
+
+func (a *Apiloader) createSSH(rg io.Reader) (publicKeyString string, err error) {
```
weinong: please refactor the one in `pkg/acsengine` into `pkg/helpers`

CecileRobertMichon: Is this function a duplicate of https://github.com/Azure/acs-engine/blob/a0ad2d7f37a2a0276660b5bf1a686e3fe3f47ea4/pkg/acsengine/ssh.go#L46? If so, is it possible to move it out in helpers/ and use the same function in both places to avoid duplicated code?

zqingqing1: I discussed with Weinong. I tried to move the code to Helpers folder, however, some structs in the ssh.go are in acsengine folder. If I import acsengine in this file, it will have the import cycle. moving has much work than we thought.

jackfrancis: I'm not seeing why we shouldn't do, as Cecile suggests:

1) move the `SSHCreator` type struct + methods in `acs-engine/pkg/acsengine/ssh.go` to `pkg/acgengine/helpers/helpers.go`
2) convert existing usage patterns of `github.com/Azure/acs-engine.SSHCreator` to `github.com/Azure/acs-engine/helpers.SSHCreator`
3) follow the above, new usage pattern in `pkg/api`

jackfrancis: I would move any tests of moved code to the new package namespace. I want to keep things as simple as possible, we just don't want to have duplicate code.

PR #2630, opened 2018-04-07
remove deprecated etcd-quorum-read flag from apiserver

seanknox:
The apiserver flag `--etcd-quorum-read was deprecated in Kubernetes v1.9 and removed in v1.10 because setting it to false causes bugs in controllers because of stale reads. It is true by default in from >= v1.8 upward until v1.10, where it was removed entirely.

Note: starting in v1.7 users had this option, and had to explicitly disable etcd quorum reads if desired, so this shouldn't affect older acs-engine clusters nor should it affect upgrades. This is effectively a non-issue since we use etcd3 by default which is much more performant with reads.

More info:
- https://github.com/kubernetes/kubernetes/pull/53795

**If applicable**:
- [x] documentation
- [x] unit tests

### DIFF
--- a/docs/clusterdefinition.md
+++ b/docs/clusterdefinition.md
@@ -297,7 +297,7 @@ Or perhaps you want to customize/override the set of admission-control flags pas
           "--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,AlwaysPullImages"
         }
       }
-    } 
+    }
 ```
 
 See [here](https://kubernetes.io/docs/reference/generated/kube-apiserver/) for a reference of supported apiserver options.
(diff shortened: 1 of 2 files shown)


docs/clusterdefinition.md
```diff
@@ -1,559 +1,558 @@
...
-|"--anonymous-auth"|"false|
-|"--audit-log-path"|"/var/log/apiserver/audit.log"|
-|"--insecure-port"|"8080"|
-|"--secure-port"|"443"|
-|"--service-account-lookup"|"true"|
-|"--etcd-cafile"|"/etc/kubernetes/certs/ca.crt"|
-|"--etcd-certfile"|"/etc/kubernetes/certs/etcdclient.crt"|
-|"--etcd-keyfile"|"/etc/kubernetes/certs/etcdclient.key"|
-|"--etcd-servers"|*calculated value that represents etcd servers*|
-|"--etcd-quorum-read"|"true"|
```
seanknox: My editor removed a lot of erroneous whitespace apparently, but this is the only line that actually changed (was removed.

Issue #2430, opened 2018-03-12
Heapster nanny container fails to start with Kubernetes 1.9.3


adelina-t:
**ISSUE**:

---

**v0.13.1**:

---

**Kubernetes 1.9.3**

**What happened**:

Cluster deployed with Kubernetes 1.9.3 has failed heapster pod.

```
NAMESPACE     NAME                                            READY     STATUS             RESTARTS   AGE
kube-system   heapster-7cd8dd888b-4556z                       1/2       CrashLoopBackOff   821        2d
```

**How to reproduce it** (as minimally and precisely as possible):

Deploy basic cluster from template with orchestrator Kubernetes 1.9.3

**Anything else we need to know**:

Logs from inside the heapster nanny pod:

```
kubectl logs heapster-7cd8dd888b-4556z -n kube-system heapster-nanny
ERROR: logging before flag.Parse: I0312 09:56:19.299122       1 pod_nanny.go:64] Invoked by [/pod_nanny --cpu=80m --extra-cpu=0.5m --memory=140Mi --extra-memory=4Mi --threshold=5 --deployment=heapster --container=heapster --poll-period=300000 --estimator=exponential]
ERROR: logging before flag.Parse: I0312 09:56:19.299261       1 pod_nanny.go:76] Watching namespace: kube-system, pod: heapster-7cd8dd888b-4556z, container: heapster.
ERROR: logging before flag.Parse: I0312 09:56:19.299274       1 pod_nanny.go:77] storage: MISSING, extra_storage: 0Gi
ERROR: logging before flag.Parse: I0312 09:56:19.395294       1 pod_nanny.go:162] Failed to read data from config file "MISSING/NannyConfiguration": open MISSING/NannyConfiguration: no such file or directory, using default parameters
ERROR: logging before flag.Parse: I0312 09:56:19.395356       1 pod_nanny.go:166]
ERROR: logging before flag.Parse: I0312 09:56:19.395394       1 pod_nanny.go:101] cpu: , extra_cpu: , memory: , extra_memory:
panic: cannot parse '': quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'

goroutine 1 [running]:
k8s.io/autoscaler/addon-resizer/vendor/k8s.io/apimachinery/pkg/api/resource.MustParse(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
[...]

PR #2637, opened 2018-04-09
Kubernetes Cluster Autoscaler (VMSS) addon

sozercan:

This PR adds Azure Virtual Machine Scale Set cluster autoscaler (https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/azure) as an add-on for acs-engine

It automatically pulls in and converts to base64 all required values for credentials (such as clientId, clientSecret, subscriptionId, tentantId, etc) and other required values (scale set name, resource group, etc) from the deployment.

User can override minNodes (default to 1), maxNodes (default to 5), image, limits, and requests.

Documentation: https://github.com/sozercan/acs-engine/blob/autoscaler-vmss/examples/addons/cluster-autoscaler/README.md

Merge after #2620 

**If applicable**:
- [x] documentation
- [x] unit tests
- [x] managed identity support

**Release note**:

```release-note
Added cluster-autoscaler addon for Virtual Machine Scale Sets
```


khenidak:
@kkmsft who is working on Autoscaler for AKS RP. Can we align both work streams?

feiskyer:
@ewok2030 You can add more agent pools by adding 

```
- --nodes=<kubernetesClusterAutoscalerMinNodes>:<kubernetesClusterAutoscalerMaxNodes>:<kubernetesClusterAutoscalerVMSSName>
- --nodes=<kubernetesClusterAutoscalerMinNodes>:<kubernetesClusterAutoscalerMaxNodes>:<kubernetesClusterAutoscalerVMSSName>
```

sozercan:
@ewok2030 like @feiskyer said, you can edit the yaml to include multiple agent pools, but that functionality isn't there in the add-on logic yet (it'll automatically include the primaryScaleSet which is the first agent pool). I have been thinking about adding some extra configuration options to include multiple agent pools.

sozercan:
@jackfrancis ready for review when you get a chance

pkg/acsengine/const.go
```diff
@@ -104,8 +104,8 @@ const (
 	DefaultACIConnectorAddonName = "aci-connector"
 	// DefaultDashboardAddonName is the name of the kubernetes-dashboard addon deployment
 	DefaultDashboardAddonName = "kubernetes-dashboard"
-	// DefaultACIConnectorImage defines the ACI Connector deployment version on Kubernetes Clusters
-	DefaultACIConnectorImage = "virtual-kubelet:latest"
```
jackfrancis: Did we want to delete this const?

pkg/acsengine/k8s_versions.go
```diff
@@ -36,58 +36,60 @@ var k8sComponentVersions = map[string]map[string]string{
...
-		"backoffduration": strconv.Itoa(DefaultKubernetesCloudProviderBackoffDuration),
-		"backoffexponent": strconv.FormatFloat(DefaultKubernetesCloudProviderBackoffExponent, 'f', -1, 64),
-		"ratelimitqps":    strconv.FormatFloat(DefaultKubernetesCloudProviderRateLimitQPS, 'f', -1, 64),
-		"ratelimitbucket": strconv.Itoa(DefaultKubernetesCloudProviderRateLimitBucket),
-		"gchighthreshold": strconv.Itoa(DefaultKubernetesGCHighThreshold),
-		"gclowthreshold":  strconv.Itoa(DefaultKubernetesGCLowThreshold),
+		"dockerEngine":       "1.13.*",
+		"dashboard":          "kubernetes-dashboard-amd64:v1.8.3",
+		"exechealthz":        "exechealthz-amd64:1.2",
+		"addon-resizer":      "addon-resizer:1.7",
```
jackfrancis: I assume we're downgrading addon-resizer to 1.7 because 1.8.1 is not mutually compatible with cluster-autoscaler?

Can we be confident that we won't re-introduce this bug:

https://github.com/Azure/acs-engine/issues/2430

Which was fixed here:

https://github.com/Azure/acs-engine/pull/2753

(the addon-resizer manfest definition was changed with the above PR, not just the version; perhaps the changes in how addon-resizer was expressed is the important part and not the upgrade to 1.8.1?)

jackfrancis:
This looks great, couple questions. Let's add validation unit tests (see `pkg/api/vlabs/validate_test.go`) to cover your new validations.

Thanks so much for this!!

pkg/acsengine/const.go
sozercan: yea, this one is no longer used (since it moved to `k8s_versions`), thought i would clean it up

pkg/acsengine/k8s_versions.go
sozercan: didn't mean to downgrade addon-resizer, probably didn't notice when rebasing. reverted that change.

sozercan:
@jackfrancis thanks! added validation unit tests. should be good to review when you get a chance

robsonpeixoto:
How to change the mix/max after the cluster been created?

ewok2030:
At this time, it’ll have to be done in the same way a second agent pool may be added: by SSH’ing into each master node an editing the addon YAML.

PR #2647, opened 2018-04-09
add EnableRBAC to v20180331 api model

### LINKED ISSUE #2646: Need to add EnableRBAC as a flag in v20180331 api model

---

Feature Request

---

---

Kubernetes

**What happened**:
We don't have a flag now

We need a flag

rjtsdl:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2646 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -349,6 +350,13 @@ func convertV20180331AgentPoolOnlyWindowsProfile(obj *v20180331.WindowsProfile)
 	}
 }
 
+func convertV20180331AgentPoolOnlyKubernetesConfig(enableRBAC bool) *KubernetesConfig {
```
amanohar: @rjtsdl thanks the making the change as this also has a good side effect of resolving bug 2265988 in VSTS! You can also now remove lines 126-141 from `putmanagedcluster.go` and close out bug 2265988 in VSTS when new ACS E is vendored.

pkg/api/converterfromagentpoolonlyapi.go
```diff
@@ -130,12 +130,37 @@ func convertResourcePurchasePlanToV20180331AgentPoolOnly(api *ResourcePurchasePl
...
+	// Only if both KubernetesConfig.EnableRbac and KubernetesConfig.EnableSecureKubelet true
+	// We think EnableRBAC is on
+	enableRBAC := true
+	if kc.EnableRbac == nil {
+		enableRBAC = enableRBAC && false
+	} else {
+		enableRBAC = enableRBAC && *kc.EnableRbac
+	}
+
+	if kc.EnableSecureKubelet == nil {
```
amanohar: IMO if the code is `convertertoagentpoolonlyapi.go` is written as expected then we shouldn't need to check for `EnableSecureKubelet` state. Checking `enableRBAC` flag state should suffice.

pkg/api/agentPoolOnlyApi/v20180331/types.go
```diff
@@ -40,6 +40,7 @@ type Properties struct {
 	AccessProfiles          map[string]AccessProfile `json:"accessProfiles,omitempty"`
 	AddonProfiles           map[string]AddonProfile  `json:"addonProfiles,omitempty"`
 	NodeResourceGroup       string                   `json:"nodeResourceGroup,omitempty"`
+	EnableRBAC              bool                     `json:"enableRBAC"`
```
JunSun17: does omitempty matter here?

pkg/api/converterfromagentpoolonlyapi.go
```diff
@@ -130,12 +130,37 @@ func convertResourcePurchasePlanToV20180331AgentPoolOnly(api *ResourcePurchasePl
 	v20180331.Publisher = api.Publisher
 }
 
+func convertKubernetesConfigToEnableRBACV20180331AgentPoolOnly(kc *KubernetesConfig) bool {
+	if kc == nil {
+		return false
+	}
+	// Only if both KubernetesConfig.EnableRbac and KubernetesConfig.EnableSecureKubelet true
+	// We think EnableRBAC is on
+	enableRBAC := true
```
JunSun17: the code seems pretty complicated to me. why not base on the comments:

enableRBAC := (KubernetesConfig.EnableRbac == true) && (KubernetesConfig.EnableSecureKubelet == true)

pkg/api/converterfromagentpoolonlyapi.go
```diff
@@ -130,12 +130,37 @@ func convertResourcePurchasePlanToV20180331AgentPoolOnly(api *ResourcePurchasePl
...
+
 func convertPropertiesToV20180331AgentPoolOnly(api *Properties, p *v20180331.Properties) {
 	p.ProvisioningState = v20180331.ProvisioningState(api.ProvisioningState)
+
 	if api.OrchestratorProfile != nil {
 		if api.OrchestratorProfile.OrchestratorVersion != "" {
 			p.KubernetesVersion = api.OrchestratorProfile.OrchestratorVersion
 		}
+		p.EnableRBAC = convertKubernetesConfigToEnableRBACV20180331AgentPoolOnly(api.OrchestratorProfile.KubernetesConfig)
+		p.EnableRBAC = *api.OrchestratorProfile.KubernetesConfig.EnableRbac &&
```
JunSun17: is this line redundant?

pkg/api/converterfromagentpoolonlyapi.go
rjtsdl: The way we want to make it, is to have `EnableSecureKubelet` also enabled when use specify enableRBAC. I should make it clear in the enableRBAC flag description. Check the email thread I included you.

pkg/api/converterfromagentpoolonlyapi.go
rjtsdl: EnableRbac is *bool !

pkg/api/converterfromagentpoolonlyapi.go
rjtsdl: Yeah, it is :)

rjtsdl:
Btw, I also need to make `EnableRBAC` to default to true.

pkg/api/converterfromagentpoolonlyapi.go
amanohar: Yes, agreed. We are on the same page regarding:
The way we want to make it, is to have EnableSecureKubelet also enabled when use specify enableRBAC.

I am suggesting that if rbac was enabled in API model then `EnableSecureKubelet` would have been set to true. So when converting it back to versioned API model state of  `EnableSecureKubelet` should not need to be validated. Another way to put it is rbac can still be technically enabled if even  `EnableSecureKubelet` was set to false.

weinong: why `enableRBAC && false` when it must be `false`?

pkg/api/agentPoolOnlyApi/v20180331/types.go
weinong: 1. should this be on by default?
2. can this be changed? (turning on and off?)

pkg/api/converterfromagentpoolonlyapi.go
weinong: what email thread?

> Check the email thread I included you.

pkg/api/converterfromagentpoolonlyapi.go
JunSun17: I see, could this work then:
enableRBAC := (KubernetesConfig.EnableRbac != nil) && (*(KubernetesConfig.EnableRbac) == true) && (KubernetesConfig.EnableSecureKubelet != nil) && (*(KubernetesConfig.EnableSecureKubelet) == true)

no big deal, just want to make code a bit more easy to read. :)

rjtsdl:
I have change the default behavior to be true for EnableRBAC api flag. 
TODO: I need to modify the older agentPoolOnly api model to always set those to be false, then we can remove the hack in RP codebase.

pkg/api/convertertoagentpoolonlyapi.go
rjtsdl: Made the change for v20170831 to make sure those two flags to be false ptr. Then RP don't need to deal with that.

pkg/api/agentPoolOnlyApi/v20180331/types.go
rjtsdl: Change to *bool, so we have 3 state

nil -> true
true -> true
false -> false

Note, we default behavior is to set it to true.

pkg/api/converterfromagentpoolonlyapi.go
rjtsdl: Made it easier to read.

pkg/api/converterfromagentpoolonlyapi.go
```diff
@@ -130,12 +133,27 @@ func convertResourcePurchasePlanToV20180331AgentPoolOnly(api *ResourcePurchasePl
...
 }
 
+func convertKubernetesConfigToEnableRBACV20180331AgentPoolOnly(kc *KubernetesConfig) *bool {
+	if kc == nil {
+		return common.BoolPtr(false)
+	}
+	// Only if both KubernetesConfig.EnableRbac and KubernetesConfig.EnableSecureKubelet true
+	// We think EnableRBAC is on
+	if kc != nil && kc.EnableRbac != nil && *kc.EnableRbac &&
+		kc.EnableSecureKubelet != nil && *kc.EnableSecureKubelet {
```
amanohar: I still think that `EnableSecureKubelet ` should not factor into converting back to versioned API model.  `if  *kc.EnableRbac`  then it should be enough to say that RBAC is enabled. You can also always log a warning line that `EnableSecureKubelet` was not enabled.

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -349,6 +354,20 @@ func convertV20180331AgentPoolOnlyWindowsProfile(obj *v20180331.WindowsProfile)
 	}
 }
 
+func convertV20180331AgentPoolOnlyKubernetesConfig(enableRBAC *bool) *KubernetesConfig {
+	if enableRBAC == nil || *enableRBAC == true {
```
amanohar: If `if enableRBAC == nil` shouldn't `EnableRbac` be false?

pkg/api/converterfromagentpoolonlyapi.go
rjtsdl: I don't have strong opinion on this. Updated only use `*kc.EnableRbac` to deduct the versioned api model.

pkg/api/convertertoagentpoolonlyapi.go
rjtsdl: Oh, we want the default behavior to be true.

Default behavior is when `enableRBAC == nil`

amanohar: Ok, cool. Thanks!

pkg/api/common/helper.go
```diff
@@ -51,3 +51,8 @@ func HandleValidationErrors(e validator.ValidationErrors) error {
 	}
 	return fmt.Errorf("Namespace %s is not caught, %+v", ns, e)
 }
+
+// BoolPtr convert a bool to *bool
+func BoolPtr(b bool) *bool {
```
CecileRobertMichon: there is already a function doing this at https://github.com/Azure/acs-engine/blob/7272611344e09bf7e59d565bb39457d9736511d5/pkg/helpers/helpers.go#L55 I think you could re-use it

rjtsdl: yep, will do

PR #2648, opened 2018-04-10
Custom VNET API model.

JunSun17:
Custom VNET API model.

Manual E2E test passed.


pkg/api/agentPoolOnlyApi/v20180331/errors.go
```diff
@@ -0,0 +1,45 @@
+package v20180331
```
amanohar: Why does this need a new `errors.go` file? Can you consolidate this where other errors like `ErrorResourceGroupNotMatch` are?

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -130,45 +133,117 @@ func validateUniqueProfileNames(profiles []*AgentPoolProfile) error {
...
+				// TODO: validate dockerBridgeCidr is not within VnetSubnetCidr
+
+				// TODO: validate ServiceCidr does not overlap with VnetSubnetCidr
+
+				// TODO: validate VnetSubnetCidr does n ot overlap with other VnetSubnetCidr
+
+				// TODO: validate IP address pre-allocation does not exceed VnetSubnetCidr capacity
+
+				// validate subsricption, resource group and vnet are the same among subnets
+				subnetSubscription, subnetResourceGroup, subnetVnet, _, err := GetVNETSubnetIDComponents(agentPool.VnetSubnetID)
```
amanohar: Instead of doing nil checks, do a regex match of the expected resource ID (subnet).

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -130,45 +133,117 @@ func validateUniqueProfileNames(profiles []*AgentPoolProfile) error {
...
+	} else {
+		return ErrorInvalidNetworkPlugin
 	}
 
 	return nil
 }
 
+func (n *NetworkProfile) isCustomVNET() bool {
+	// use ServiceCidr check to decide if Custom VNET is triggered
+	return n.ServiceCidr != ""
```
amanohar: If `n.ServiceCidr == ""` and `n.NetworkPlugin == string(azure)`, it won't be custom VNET? Tighten this validation so that both are true (`n.ServiceCidr != "" && n.NetworkPlugin == string(azure)`)

pkg/api/agentPoolOnlyApi/v20180331/errors.go
JunSun17: I separate the errors out and make them each a variable, so I can catch the exact one when running unit tests. Otherwise, the returned error from unit tests will be a generic error and I will need to match the wording in the error object.

Please let me know if you have a better suggestion for such error handling.

JunSun17: Talked with Jiantian and his error case is different. So for now, will keep as it is. Later we can think about error standardization/centralization.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
JunSun17: As discussed, will use agentpool subnets to implement isCustomVNET function.

PR #2653, opened 2018-04-10
add Merge function for v20180331 to guard update scenario

### LINKED ISSUE #2646: Need to add EnableRBAC as a flag in v20180331 api model

rjtsdl:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2646

following #2647, this PR is to add validation on update scenario. So we catch all the cases when enableRBAC could possibly be changed. And return as error

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```


weinong:
`DNSPrefix` will need to be optional. Otherwise, it will fail at [here](https://github.com/rjtsdl/acs-engine/blob/92229f60642f6fedee237902a2ac37e56dc79731/pkg/api/apiloader.go#L220)
Can you also merge `DNSPrefix`?

rjtsdl:
@weinong can we do in a different PR? 
I set up the framework here, then @zqingqing1 can add other fields needed to be merged, and also unittests.

PR #2654, opened 2018-04-10
Use latest patch of DefaultKubernetesRelease as default k8s version

CecileRobertMichon:
**What this PR does / why we need it**: Update k8s default version from 1.8.9 to 1.8.11

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -50,12 +50,17 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.10.0":        true,
 }
 
+// GetDefaultKubernetesVersion returns the default Kubernetes version, that is the latest patch of the default release
+func GetDefaultKubernetesVersion() string {
+	return GetLatestPatchVersion(KubernetesDefaultRelease, GetAllSupportedKubernetesVersions())
+}
+
 // GetSupportedKubernetesVersion verifies that a passed-in version string is supported, or returns a default version string if not
 func GetSupportedKubernetesVersion(version string) string {
 	if k8sVersion := version; AllKubernetesSupportedVersions[k8sVersion] {
 		return k8sVersion
 	}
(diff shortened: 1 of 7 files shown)


jackfrancis:
Should we instead prefer a release channel (i.e., 1.8) and then just make the default version the latest patch release in that channel?

CecileRobertMichon:
Yes I think that makes sense. The only reason I see that would prevent us from doing that is baking newly released k8s minor versions in acs-engine before we make them default. For example, when k8s 1.8.12 comes out, we add support for it right away but we only promote it to the default version after it has been passing our 1.8 tests for a while. What do you think?

jackfrancis:
I think we defer to k8s to bake basic stability into the most current release. But open to doing anything reasonable that is automateable.

CecileRobertMichon:
Ok agreed, I'll modify this PR to implement a default release instead of a default version

pkg/api/common/versions.go
```diff
@@ -214,9 +219,9 @@ func GetSupportedVersions(orchType string, hasWindows bool) (versions []string,
 	switch orchType {
 	case Kubernetes:
 		if hasWindows {
-			return GetAllSupportedKubernetesVersionsWindows(), string(KubernetesDefaultVersion)
+			return GetAllSupportedKubernetesVersionsWindows(), GetDefaultKubernetesVersion()
 		}
-		return GetAllSupportedKubernetesVersions(), string(KubernetesDefaultVersion)
```
CecileRobertMichon: Any reason why a string conversion was necessary here? I removed it but let me know if I'm missing something @jackfrancis

jackfrancis: not a sane one that I can think of, nope!

PR #2661, opened 2018-04-12
Update prometheus-grafana-k8s example and documentation

devcfgc:
The current example of how to install the prometheus-grafana-k8s extension used the agentPoolProfile for it (I tried several times to deploy it this way and it didn't work), but the prometheus-grafana-k8s.sh executes some commands that only will work on the master so the installation of the extension must be done in the masterProfile instead.

- [x] documentation
- [x] tested deploying k8s cluster in Azure

### DIFF
--- a/examples/extensions/prometheus-grafana-k8s.json
+++ b/examples/extensions/prometheus-grafana-k8s.json
@@ -7,19 +7,19 @@
     "masterProfile": {
       "count": 1,
       "dnsPrefix": "",
-      "vmSize": "Standard_DS2_v2"
+      "vmSize": "Standard_DS2_v2",
+      "extensions": [
+        {
+          "name": "prometheus-grafana-k8s"
+        }
+      ]
     },
     "agentPoolProfiles": [
       {
         "name": "agentpool1",
         "count": 3,
         "vmSize": "Standard_DS2_v2",
-        "availabilityProfile": "AvailabilitySet",
-        "extensions": [
-          {
-            "name": "prometheus-grafana-k8s"
-          }
-        ]
+        "availabilityProfile": "AvailabilitySet"
       }
     ],
     "linuxProfile": {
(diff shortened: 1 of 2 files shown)


jackfrancis:
@ritazh @trstringer for reviews

trstringer:
> the prometheus-grafana-k8s.sh executes some commands that only will work on the master

Can you be more specific? We have tested out this extension on both master nodes as well as agent nodes successfully. But if there is an issue, we'd love to troubleshoot it.

devcfgc:
@trstringer For example `/usr/bin/docker ps | grep apiserver` will not work in the agent, won't it? but my main concern is that I deployed at least 3 times adding the extension to the agent nodes and it never installed anything... in the master node it works perfectly.

trstringer:
That is a call from a fallback code path to determine if Kubernetes is up and running. You are correct, grepping running containers for the apiserver will only succeed on master, but the first condition there tests for the existence of kubectl, which exists on the agent nodes. So on agent nodes, instead of grepping running containers for apiserver, we run `kubectl cluster-info` to determine if Kubernetes is up and running. So this should be fine on agent nodes.

Regarding other causes why this isn't working when running on your agent nodes, can you send us the logs? They can be retrieved from the agent nodes. Here's a way to grab them:

```
$ ssh -J <adminUsername>@<master_name> \
    <adminUsername>@<agent_name> \
    -- cat /var/log/azure/prometheus-grafana-k8s.log
```

Please pull these logs from all agent nodes.

neurot1cal:
@trstringer Is the recommendation here still to enable this for agentPoolProfile instead of masterProfile?

ritazh:
@neurot1cal Both agentPoolProfile and masterProfile should work. But we recommend users to use agentPoolProfile.

neurot1cal:
@ritazh @trstringer I also noticed that in `prometheus_values.yaml` - `nodeselector` is not set. If there are windows nodes in the cluster, will the deployment script for prometheus be smart enough not to schedule on these?

devcfgc:
Regrettably I don't have time to deal with this PR I'll start working on it again as soon as I can =) thanks everyone for you comments

PR #2666, opened 2018-04-12
Standardize systemd implementation for k8s provisioning

jackfrancis:
**What this PR does / why we need it**: restart systemd services and validate everywhere

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Standardize systemd implementation for k8s provisioning
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -224,7 +224,9 @@ function installClearContainersRuntime() {
 	# Enable and start Clear Containers proxy service
 	echo "Enabling and starting Clear Containers proxy service..."
 	systemctlEnableAndCheck cc-proxy
-	retrycmd_if_failure 100 1 10 systemctl daemon-reload && systemctl restart cc-proxy
+	systemctl_restart 100 1 10 cc-proxy
+    retrycmd_if_failure 10 1 3 systemctl status cc-proxy --no-pager -l > /var/log/azure/cc-proxy-status.log
+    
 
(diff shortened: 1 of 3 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -263,8 +265,9 @@ function ensureContainerd() {
 			# Enable and start cri-containerd service
 			# Make sure this is done after networking plugins are installed
 			echo "Enabling and starting cri-containerd service..."
-			systemctl enable containerd
-			retrycmd_if_failure 100 1 10 systemctl daemon-reload && systemctl restart containerd
+			systemctlEnableAndCheck containerd
+			systemctl_restart 100 1 10 containerd
```
CecileRobertMichon: +1 on standardizing the system implementation

Small suggestion: maybe you could move `systemctl_restart 100 1 10 containerd` and `retrycmd_if_failure 10 1 3 systemctl status containerd --no-pager -l > /var/log/azure/containerd-status.log` inside the `systemctlEnableAndCheck` function like so:
```function systemctlEnableAndCheck() {
    systemctl enable $1
    systemctl is-enabled $1
    enabled=$?
    for i in {1..900}; do
        if [ $enabled -ne 0 ]; then
            systemctl enable $1
            systemctl is-enabled $1
            enabled=$?
        else
            echo "$1 took $i seconds to be enabled by systemctl"
            break
        fi
        sleep 1
    done
    if [ $enabled -ne 0 ]
    then
        echo "$1 could not be enabled by systemctl"
        exit 5
    fi
    systemctl_restart 100 1 10 $1
    retrycmd_if_failure 10 1 3 systemctl status $1 --no-pager -l > /var/log/azure/$1-status.log 
}```

PR #2675, opened 2018-04-13
Add openshift e2e test

kargakis:
Adds a minimal e2e test for OpenShift (cluster stands up; nodes are ready)

TODO:
- [ ] test that healthchecks router, registry, service catalog, template service broker
- [ ] test that deploys a sample app and tries to access it from a route\
- [ ] (maybe in this PR) use an image that includes `oc` and drop `quay.io/deis/go-dev:v1.9.1`.

@jackfrancis @jim-minter @pweil-


.circleci/config.yml
```diff
@@ -364,6 +364,38 @@ jobs:
...
+      - run: |
+          echo 'export TIMEOUT=20m' >> $BASH_ENV
+          echo 'export SKIP_TEST=true' >> $BASH_ENV
+          echo 'export LOCATION=eastus' >> $BASH_ENV
+          echo 'export ORCHESTRATOR_RELEASE=3.9' >> $BASH_ENV
+          echo 'export CLUSTER_DEFINITION=examples/openshift.json' >> $BASH_ENV
+          echo 'export CREATE_VNET=false' >> $BASH_ENV
+          echo 'export CLEANUP_ON_EXIT=false' >> $BASH_ENV
+          echo 'export RETAIN_SSH=false' >> $BASH_ENV
+          echo 'export SUBSCRIPTION_ID=${SUBSCRIPTION_ID_E2E_OPENSHIFT}' >> $BASH_ENV
```
kargakis: @jackfrancis @jim-minter @pweil- I am using our own envs to test but we will need to sync on who is going to control these IDs (I guess MS). In that case we will likely need to start using our published node images so they can be accessed by the CI subscription, does that sound right?

pweil-: yes, we should ensure we can white list the CI subscription for our images
/cc @kennywoodson

kwoodson: If you send me the subscription ids to be shared I can add them in the portal.

kargakis:
@jim-minter can be tested with:
```
docker pull registry.svc.ci.openshift.org/ci/acs-engine-tests:v3.9
docker run --rm -v ${PWD}:/go/src/github.com/Azure/acs-engine -w /go/src/github.com/Azure/acs-engine -e ORCHESTRATOR_RELEASE=3.9 -e CLUSTER_DEFINITION=examples/openshift.json -e CLIENT_ID=${CLIENT_ID} -e CLIENT_SECRET=${CLIENT_SECRET} -e TENANT_ID=${TENANT_ID} -e SUBSCRIPTION_ID=${SUBSCRIPTION_ID} -e ORCHESTRATOR=openshift -e TIMEOUT=30m -e CLEANUP_ON_EXIT=false -e REGIONS=eastus  registry.svc.ci.openshift.org/ci/acs-engine-tests:v3.9 make test-openshift
```

jim-minter:
I took a read through and this is looking good.  Dependent on #2708.

pkg/api/common/const.go
```diff
@@ -84,13 +84,13 @@ var AllDCOSSupportedVersions = []string{
 }
 
 const (
-	// OpenShiftVersion3Dot9Dot0 is the major.minor.patch string for the 3.9.0 version of OpenShift
-	OpenShiftVersion3Dot9Dot0 string = "3.9.0"
+	// OpenShiftVersion3Dot9Dot14 is the major.minor.patch string for the 3.9.14 version of OpenShift
+	OpenShiftVersion3Dot9Dot14 string = "3.9.14"
```
kargakis: @jim-minter this depends on the binaries installed in the node image. Unfortunately it seems we have different patch versions between rhel and centos so the version check inside the e2e tests fails. Any idea what should we do (eg. we could ignore the patch version in the test?) and whether tracking the correct patch version here makes sense or not?

cc @kwoodson

test/e2e/openshift/openshift_test.go
```diff
@@ -0,0 +1,131 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(running).To(Equal(true))
+	})
+
+	It("should have have the appropriate node count", func() {
+		ready := knode.WaitOnReady(eng.NodeCount(), 10*time.Second, cfg.Timeout)
+		Expect(ready).To(Equal(true))
+	})
+
+	It("should be running the expected version", func() {
```
kargakis: @jim-minter this is the test that also needs the correct patch version today.

kargakis:
Assuming the tenant id is added in circle ci and we whitelist the acs-engine ci subscription in our image, this is ready to merge.

.circleci/config.yml
```diff
@@ -367,6 +367,37 @@ jobs:
...
+          echo 'export LOCATION=eastus' >> $BASH_ENV
+          echo 'export ORCHESTRATOR_RELEASE=3.9' >> $BASH_ENV
+          echo 'export CLUSTER_DEFINITION=examples/openshift.json' >> $BASH_ENV
+          echo 'export CREATE_VNET=false' >> $BASH_ENV
+          echo 'export CLEANUP_ON_EXIT=false' >> $BASH_ENV
+          echo 'export RETAIN_SSH=false' >> $BASH_ENV
+          echo 'export SUBSCRIPTION_ID=${SUBSCRIPTION_ID_E2E_KUBERNETES}' >> $BASH_ENV
+          echo 'export CLIENT_ID=${SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES}' >> $BASH_ENV
+          echo 'export CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET_E2E_KUBERNETES}' >> $BASH_ENV
+          echo 'export TENANT_ID=${TENANT_ID_E2E_KUBERNETES}' >> $BASH_ENV
```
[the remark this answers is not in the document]
kargakis: I need to reference TENANT_ID in my test because we need to populate azProfile.

jim-minter:
@jackfrancis @CecileRobertMichon do you think it'd be possible to merge this today?  @kargakis is out at Kubecon this week and getting this in would help testing on our side too.  If it's clear to you that TENANT_ID is already available to the test via the environment without line 349, are you able to administratively add a one-line commit that removes that in order to progress this?  Thanks!

jackfrancis:
@jim-minter Take a look at the result from the openshift E2E run above. The provision script (cse in Azure ARM speak) exited non-zero with the outputted error. Does that suggest something you know how to fix?

kargakis:
We need to rollback the version of the RPMs inside our published image to a version where the images in registry.redhat.com can match

cc @kwoodson

kargakis:
We could also potentially switch the test to test centos and add a rhel test once we have a published image with RPMs that match with what's in the registry.

jim-minter:
afaics this comes down to very stale testing image.  Images built using our new process should not hit this error.

jackfrancis:
@jim-minter @kargakis should I re-run the openshift E2E test following the image publications?

jim-minter:
> should I re-run the openshift E2E test following the image publications?

Please do.  @kwoodson please can you let @jackfrancis know when the new image has arrived?

kwoodson:
@jim-minter, ack.  I suspect it will be done shortly as the origin version just completed.  

@jackfrancis, Will reply back when notified.

kwoodson:
@jackfrancis I just received the notification. Please give it a go. Thanks

Issue #1160, opened 2017-08-02
i18n.go - undefined: Asset


dcieslak19973:
**Is this a request for help?**: yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: master

---

running "go get github.com/Azure/acs-engine" yields the following error:

src/github.com/Azure/acs-engine/pkg/i18n/i18n.go:65: undefined: Asset

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**

**What happened**:
Undefined: Asset

**What you expected to happen**:

No error

**How to reproduce it** (as minimally and precisely as possible):

go get github.com/Azure/acs-engine

**Anything else we need to know**:

dcieslak19973:
This is with go version 1.8.3

dcieslak19973:
Also, after running "go get all", getting:

package github.com/go-playground/pure/examples/middleware/logging-recovery: cannot find package "github.com/go-playground/pure/examples/middleware/logging-recovery" in any of:

With various GOROOT and GOPATH directories listed.

seanknox:
This is the same problem mentioned https://github.com/Azure/acs-engine/issues/1068#issuecomment-318776434. We have text templates that need to be compiled before running, which `go get` doesn't do. It's an ugly error on cloning but harmless otherwise. If you don't need to develop on acs-engine, you're better off downloading the binary releases.

dcieslak19973:
Got it. I used the docker method and it works

uday31in:
Is there any ETA for the fix? Thanks.

masroorhasan:
Getting the same error with `go get github.com/Azure/acs-engine` on windows.

Issue #1068, opened 2017-07-21
Error on Go get and acs-engine build


mfaizanse:
**Is this a request for help?**:
YES
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
ISSUE
---

**What version of acs-engine?**:
commit 42b4fa3  
---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes

**What happened**:
acs-engine failed to get and build

**What you expected to happen**:
Build success

**How to reproduce it** (as minimally and precisely as possible):
https://github.com/Azure/acs-engine/blob/master/docs/acsengine.md#downloading-and-building-acs-engine

**Anything else we need to know**:
![image](https://user-images.githubusercontent.com/9897945/28460822-4006b958-6e2d-11e7-9682-135b03ed5f84.png)

If I deploy a kubernetes cluster using azure portal then it deploy using the kubernetes 1.6.6 version and that results in dns resolutoin failure and No connectivity inside containers. And when I was tried to deploy using the acs-engine then this issue came up.

weinong:
have you tried
`make prereqs`
`make build`

mfaizanse:
@weinong I am on windows?

mfaizanse:
@weinong Where should I run these commands? I have tried in "C:\gopath\src\github.com\Azure\acs-engine" but its failing.

![image](https://user-images.githubusercontent.com/9897945/28512333-eaf67ff8-706a-11e7-8d6f-d8f877df16d2.png)

amanohar:
@mfaizanse you need to install make (since you are running this on windows) or use the instructions at: https://github.com/Azure/acs-engine/blob/master/docs/acsengine.md to mount the repo in a container.

mfaizanse:
@amanohar I installed make and tried but failed. Please see my above comment.

amanohar:
@mfaizanse from the link I shared above can you try mounting your repo into a container by running: Windows (PowerShell): .\scripts\devenv.ps1 and then trying make prepreqs? Easiest way to start developing on ACS Engine is to use Docker.

mfaizanse:
@amanohar I have tried the container method and its working fine.

seanknox:
I can confirm this is an issue when cloning the repo via `go get`:

```
$ go get github.com/seanknox/acs-engine
# github.com/Azure/acs-engine/pkg/acsengine
[...]

PR #2678, opened 2018-04-13
start commiting bindata file for openshift

jim-minter:
among other things, committing bindata file allows other packages to vendor should they need to without having to run `go generate`


jim-minter:
@jackfrancis fyi
other bindata files not currently committed include pkg/acsengine/templates.go, pkg/i18n/translations.go.
I believe that an equivalent PR for these files would solve #1068 and #1160.

pkg/openshift/certgen/templates/templates.go
```diff
@@ -1,4 +1,4 @@
 //go:generate go get -u github.com/go-bindata/go-bindata/...
-//go:generate go-bindata -pkg templates master/... node/...
+//go:generate go-bindata -modtime 1 -pkg templates master/... node/...
```
pweil-: why are you overriding the mod time stamp here?

jim-minter: Git doesn't persist timestamps so if `-modtime 1` is not included here it results in a lot of churn in commits.

jim-minter: (origin uses -nometadata instead, it turns out)

PR #2686, opened 2018-04-14
fix clear containers cc-proxy

jessfraz:
Jenkins running here: https://jenkins.azure-containers.io/job/k8s-clear-containers/78/console

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -217,19 +217,26 @@ function installClearContainersRuntime() {
 	apt-get update && apt-get install --no-install-recommends -y \
 		cc-runtime
 
+	# Install the systemd service and socket files.
+	local repo_uri="https://raw.githubusercontent.com/clearcontainers/proxy/3.0.23"
+	curl -sSL --retry 5 --retry-delay 10 --retry-max-time 30 "${repo_uri}/cc-proxy.service.in" | sed 's#@libexecdir@#/usr/libexec#' > /etc/systemd/system/cc-proxy.service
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -217,10 +217,14 @@ function installClearContainersRuntime() {
 	apt-get update && apt-get install --no-install-recommends -y \
 		cc-runtime
 
+	# Install the systemd service and socket files.
+	local repo_uri="https://raw.githubusercontent.com/clearcontainers/proxy/3.0.23"
+	curl -sSL --retry 5 --retry-delay 10 --retry-max-time 30 "${repo_uri}/cc-proxy.service.in" | sed 's#@libexecdir@#/usr/libexec#' > /etc/systemd/system/cc-proxy.service
```
jackfrancis: I recommend doing all the curls as separate commands (curl w/ its own retry doesn't retry every possible, error, such as a corrupt/partly-finished download). We have a `retrycmd_get_tarball` convenience inside `kubernetesprovisionsource.sh` that you could use to resiliently get a "valid tarball" (see the code to understand our definition of valid tarball) to disk. Of course for the above because it's not a tarball you'll need another some other way to validate that the downloaded file is "valid".

Thanks for fixing this!

jessfraz: I updated the tarball one to use `retrycmd_get_tarball` I think the other is ok, no?

jessfraz:
jenkins running here: https://jenkins.azure-containers.io/job/k8s-clear-containers/81/console

PR #2693, opened 2018-04-14
Validation logic backward compatibility support for custom VNET

JunSun17:
Validation logic backward compatibility support for API models converted from prior v20180331 models.

This PR adds validation support for API models generated by serialization of
1) un-versioned models created prior to v20180331.
2) use template generator generated model with default values set.

Also adds various testings.

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20180331/errors.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/errors.go
@@ -2,9 +2,6 @@ package v20180331
 
 import "fmt"
 
-// ErrorNilAgentPoolProfile error
-var ErrorNilAgentPoolProfile = fmt.Errorf("Agent pool profile can not be nil")
-
 // ErrorInvalidNetworkProfile error
 var ErrorInvalidNetworkProfile = fmt.Errorf("ServiceCidr, DNSServiceIP, DockerBridgeCidr should all be empty or neither should be empty")
 
--- a/pkg/api/agentPoolOnlyApi/v20180331/errors.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/errors.go
@@ -2,14 +2,14 @@ package v20180331
 
 import "fmt"
 
-// ErrorNilNetworkProfile error
-var ErrorNilNetworkProfile = fmt.Errorf("Network profile can not be nil")
-
 // ErrorNilAgentPoolProfile error
 var ErrorNilAgentPoolProfile = fmt.Errorf("Agent pool profile can not be nil")
 
+// ErrorInvalidNetworkProfile error
+var ErrorInvalidNetworkProfile = fmt.Errorf("ServiceCidr, DNSServiceIP, DockerBridgeCidr should all be empty or neither should be empty")
+
 // ErrorInvalidNetworkPlugin error
-var ErrorInvalidNetworkPlugin = fmt.Errorf("Network plugin should be either \"azure\" or \"kubenet\"")
+var ErrorInvalidNetworkPlugin = fmt.Errorf("Network plugin should be either Azure or Kubenet")
 
 // ErrorInvalidServiceCidr error
 var ErrorInvalidServiceCidr = fmt.Errorf("ServiceCidr is not a valid CIDR")
@@ -26,11 +26,8 @@ var ErrorDNSServiceIPNotInServiceCidr = fmt.Errorf("DNSServiceIP is not within S
 // ErrorDNSServiceIPAlreadyUsed error
 var ErrorDNSServiceIPAlreadyUsed = fmt.Errorf("DNSServiceIP can not be the first IP address in ServiceCidr")
 
-// ErrorAgentPoolNoSubnet error
-var ErrorAgentPoolNoSubnet = fmt.Errorf("Agent pool does not have subnet defined")
-
-// ErrorKubenetNoCustomization error
-var ErrorKubenetNoCustomization = fmt.Errorf("Kubenet does not support ServiceCidr or DNSServiceIP or DockerBridgeCidr customization")
+// ErrorAtLeastAgentPoolNoSubnet error
+var ErrorAtLeastAgentPoolNoSubnet = fmt.Errorf("At least one agent pool does not have subnet defined")
 
 // ErrorParsingSubnetID error
 var ErrorParsingSubnetID = fmt.Errorf("Failed to parse VnetSubnetID")
--- a/pkg/api/agentPoolOnlyApi/v20180331/types.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/types.go
@@ -59,7 +59,7 @@ const (
 	// Azure represents Azure CNI network plugin
 	Azure NetworkPlugin = "azure"
 	// Kubenet represents Kubenet network plugin
-	Kubenet NetworkPlugin = "none"
+	Kubenet NetworkPlugin = "kubenet"
 )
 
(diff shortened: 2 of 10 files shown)


pkg/api/agentPoolOnlyApi/v20180331/errors.go
```diff
@@ -2,14 +2,14 @@ package v20180331
...
-
 // ErrorNilAgentPoolProfile error
 var ErrorNilAgentPoolProfile = fmt.Errorf("Agent pool profile can not be nil")
 
+// ErrorInvalidNetworkProfile error
+var ErrorInvalidNetworkProfile = fmt.Errorf("ServiceCidr, DNSServiceIP, DockerBridgeCidr should all be empty or neither should be empty")
+
 // ErrorInvalidNetworkPlugin error
-var ErrorInvalidNetworkPlugin = fmt.Errorf("Network plugin should be either \"azure\" or \"kubenet\"")
+var ErrorInvalidNetworkPlugin = fmt.Errorf("Network plugin should be either Azure or Kubenet")
```
weinong: `Kubenet` or `none`? Also, is it case insensitive?

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -141,108 +141,121 @@ func validateVNET(a *Properties) error {
 
 	n := a.NetworkProfile
 
-	if n == nil {
-		return ErrorNilNetworkProfile
-	}
+	// validate network profile settings
+	if n != nil {
+		if string(n.NetworkPlugin) == string(Azure) || string(n.NetworkPlugin) == string(Kubenet) {
```
weinong: is it easier to read using `switch` and `case`?

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -141,108 +141,121 @@ func validateVNET(a *Properties) error {
...
-		}
-		if dnsServiceIP.String() == kubernetesServiceIP.String() {
-			return ErrorDNSServiceIPAlreadyUsed
+	// validate agent pool custom VNET settings
+	if a.AgentPoolProfiles != nil {
+		if e := validateAgentPoolVNET(a.AgentPoolProfiles); e != nil {
+			return e
 		}
+	} else {
+		return ErrorNilAgentPoolProfile
```
weinong: In Update case, it may not be required.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -141,108 +141,121 @@ func validateVNET(a *Properties) error {
...
+		if agentPool.IsCustomVNET() {
+			return true
 		}
 	}
 
-	return true
+	return false
 }
 
 // GetVNETSubnetIDComponents extract subscription, resourcegroup, vnetname, subnetname from the vnetSubnetID
```
weinong: The regex is static. Can be saved at package level. Also, most literals used in Azure are case insensitive. For instance, `resourcegroups` is as good as `resourceGroups`

pkg/api/agentPoolOnlyApi/v20180331/errors.go
JunSun17: The plugin name is Kubenet, but in the model, we represent it internally as "none". Since the error message is more for customer, I put the name to Kubenet. Case does not matter here.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
JunSun17: In this case, Azure and Kubenet is handled through the same code path, so switch does not help.

NetworkPlugin can be instantiated by taking a string, I think it can not be compared directly to constant Azure or Kubenet, even though the string value are the same.

The model now requires case matching too.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
JunSun17: Yes, it is probably redundant and has been checked before, but I think it is no harm to be on the safe side to check this.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
weinong: In `pkg/acsengine/testdata/agentPoolOnly/v20180331/agentsWithFullNetworkProfile.json`, it says `azure`, though..

JunSun17: Actually tested (NetworkPlugin("azure") == Azure) is true, will change the code to make it simple.

"azure" is the serialized representation of object Azure, NetworkPlugin("Azure") will not equal to object Azure. So it is case sensitive.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
JunSun17: Yes, we will look into this in a separate task.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
weinong: if you are sure that update without `AgentPoolProfile` will be fine, then it's ok...

weinong: cant this be done at https://github.com/JunSun17/acs-engine/blob/65d35f4907534ef3db028fdd4302c03f62bbf17e/pkg/api/agentPoolOnlyApi/v20180331/validate.go#L20 ?

pkg/api/agentPoolOnlyApi/v20180331/types.go
```diff
@@ -59,7 +59,7 @@ const (
 	// Azure represents Azure CNI network plugin
 	Azure NetworkPlugin = "azure"
 	// Kubenet represents Kubenet network plugin
-	Kubenet NetworkPlugin = "kubenet"
+	Kubenet NetworkPlugin = "none"
```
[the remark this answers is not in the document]
JunSun17: sure, will do.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
JunSun17: Since we allow partial API model in the update case, remove ErrorNilAgentPoolProfile.

I still want to leave the validateAgentPoolVNET() method in VNET validation since VNET validation will cross multiple profiles (for example, we will need to validate subnetCidr from agentPoolProfile against serviceCidr from networkProfile).

pkg/api/convertertoagentpoolonlyapi_test.go
```diff
@@ -54,19 +54,19 @@ func TestConvertV20180331AgentPoolOnlyOrchestratorProfile(t *testing.T) {
 		t.Error("error in orchestrator profile kubernetesVersion conversion")
 	}
 
-	if api.KubernetesConfig.NetworkPolicy != "" {
+	if api.KubernetesConfig.NetworkPolicy != "none" {
 		t.Error("error in orchestrator profile networkPlugin conversion")
 	}
 
-	if api.KubernetesConfig.ServiceCIDR != "" {
+	if api.KubernetesConfig.ServiceCIDR != "10.0.0.0/16" {
```
weinong: can you define consts for `10.0.0.0/16` and `172.17.0.1/16`

JunSun17: Sure, will do.

PR #2699, opened 2018-04-16
enable prometheus metrics gathering with openshift-ansible playbooks

maxamillion:
Signed-off-by: Adam Miller <admiller@redhat.com>

Adds prometheus metrics collection, web console, and alert monitor to the OpenShift deployment type.

### DIFF
--- a/parts/k8s/openshiftmasterscript.sh
+++ b/parts/k8s/openshiftmasterscript.sh
@@ -4,10 +4,14 @@
 # probably shouldn't be
 
 SERVICE_TYPE=origin
-IMAGE_BASE=openshift/origin
+IMAGE_TYPE="${SERVICE_TYPE}"
+IMAGE_PREFIX="openshift"
+ANSIBLE_DEPLOY_TYPE="origin"
 if [ -f "/etc/sysconfig/atomic-openshift-node" ]; then
 	SERVICE_TYPE=atomic-openshift
-	IMAGE_BASE=registry.reg-aws.openshift.com:443/openshift3/ose
+	ANSIBLE_DEPLOY_TYPE="openshift-enterprise"
+	IMAGE_TYPE=ose
+	IMAGE_PREFIX="registry.reg-aws.openshift.com:443/openshift3"
 fi
 VERSION="$(rpm -q $SERVICE_TYPE --queryformat %{VERSION})"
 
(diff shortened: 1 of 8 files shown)


pkg/openshift/certgen/templates/master/tmp/ansible/azure-local-master-inventory.yml
```diff
@@ -52,9 +52,11 @@ localmaster:
     openshift_hosted_router_replicas: 1
     openshift_hosted_registry_replicas: 1
 
+    openshift_prometheus_state: "present"
 
     openshift_web_console_install: True
 
+    openshift_prometheus_image_prefix: "registry.reg-aws.openshift.com:443/openshift3/"
```
jim-minter: shouldn't have any references to registry.reg-aws.openshift.com in this file.  One slipped in above on line 17 at some point too.  Please factor them out.  In theory should write TEMPIMAGEBASE but that won't work directly because it ends in ose-.  I guess IMAGE_BASE needs to be refactor into separate parts.

parts/k8s/openshiftmasterscript.sh
```diff
@@ -145,6 +145,7 @@ docker run \
 	-e IMAGE_BASE="$IMAGE_BASE" \
 	-e VERSION="$VERSION" \
 	-e HOSTNAME="$(hostname)" \
+    --network="host" \
```
jim-minter: nit indent

parts/k8s/openshiftmasterscript.sh
```diff
@@ -5,9 +5,11 @@
 
 SERVICE_TYPE=origin
 IMAGE_BASE=openshift/origin
+IMAGE_PREFIX=openshift/origin
 if [ -f "/etc/sysconfig/atomic-openshift-node" ]; then
 	SERVICE_TYPE=atomic-openshift
 	IMAGE_BASE=registry.reg-aws.openshift.com:443/openshift3/ose
+	IMAGE_PREFIX=registry.reg-aws.openshift.com:443/openshift3
```
jim-minter: on the right lines adding IMAGE_PREFIX but I'd like to see IMAGE_BASE go away and redefine existing uses of it in terms of IMAGE_PREFIX
if it's necessary to add a separate env var to hold the 'ose' prefix, that's fine; it's the duplication of IMAGE_PREFIX and IMAGE_BASE that I don't like

jim-minter:
Continuing to hold this PR - at the moment, separately, we're erroneously hard-coding ose-web-console; I'd like the variable scheme proposed in this PR to allow us to fix that out

jim-minter:
lgtm, @CecileRobertMichon please could you merge?

PR #2701, opened 2018-04-16
rationalize exit codes in CSE

CecileRobertMichon:
**What this PR does / why we need it**: exit for every timeout, different exit code for every error.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -24,12 +24,20 @@ ensureRunCommandCompleted()
 {
     echo "waiting for runcmd to finish"
     wait_for_file 900 1 /opt/azure/containers/runcmd.complete
+    if [ ! -f /opt/azure/containers/runcmd.complete ]; then
+        echo "Timeout waiting for runcmd to complete"
+        exit 6
+    fi
 }
 
 ensureDockerInstallCompleted()
 {
     echo "waiting for docker install to finish"
     wait_for_file 3600 1 /opt/azure/containers/dockerinstall.complete
+    if [ ! -f /opt/azure/containers/dockerinstall.complete ]; then
+        echo "Timeout waiting for docker install to finish"
+        exit 3
+    fi
 }
 
 echo `date`,`hostname`, startscript>>/opt/m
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -24,12 +24,20 @@ ensureRunCommandCompleted()
 {
     echo "waiting for runcmd to finish"
     wait_for_file 900 1 /opt/azure/containers/runcmd.complete
+    if [ ! -f /opt/azure/containers/runcmd.complete ]; then
+        echo "Timeout waiting for runcmd to complete"
```
jackfrancis: can you add a mention of cloud-init here? maybe "Timeout waiting for cloud-init runcmd to complete"

CecileRobertMichon:
> should we add comments at the top of this file that document the exit codes and what they indicate?


done

parts/k8s/kubernetescustomscript.sh
```diff
@@ -1,5 +1,17 @@
 #!/bin/bash
 
+# This script runs on every Kubernetes VM
+# Exit codes represent the following:
+# | exit code number | meaning |
```
jackfrancis: How about if we reserve 1-9 for "general" exit failures, and then a decimal range (e.g., 10-19) per application-specific error condition. So, for example, "Etcd data dir not found" would be 10; "Timeout waiting for etcd to be accessible" would be 11; and then "Timeout waiting for docker install to finish" would be 20; etc

jackfrancis: Sorry, 1 and 2 should be reserved:

http://www.tldp.org/LDP/abs/html/exitcodes.html

jackfrancis: So, 3-9 for general errors:

- a service could not be enabled by systemctl
- a service could not be started by systemctl
- timeout waiting for cloud-init runcmd to complete
- timeout waiting for a file (ideally we'd have separate error messages per file we are waiting for, as they indicate a particular serial dependency)

CecileRobertMichon: updated them, is it what you had in mind?

PR #2706, opened 2018-04-17
WIP: Unit tests

CecileRobertMichon:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


scripts/ginkgo.coverage.sh
```diff
@@ -23,9 +23,8 @@ profile="${coverdir}/cover.out"
 hash goveralls 2>/dev/null || go get github.com/mattn/goveralls
 hash godir 2>/dev/null || go get github.com/Masterminds/godir
 
-# TODO this appears to run all unit tests, rather than just generate coverage data
 generate_cover_data() {
-  ginkgo -skipPackage test/e2e -cover -r .
+  ginkgo -skipPackage test/e2e/dcos,test/e2e/kubernetes -cover -r .
```
jackfrancis: I think we should add a backlog item to re-add `test/e2e/kubernetes` as it's an important piece of code!

CecileRobertMichon: Agreed that we should also run unit tests in the `test/e2e/kubernetes` (although right now it contains 0 unit tests). Note that I expanded the coverage (before it was skipping the whole e2e package (we were using the recursive `-r` flag), now it's only skipping two sub-packages of `test/e2e/` and running unit tests on the rest. Also this change is meant to match `make test`: https://github.com/Azure/acs-engine/blob/35722ba0b95df7aadf5fcd7821042ad5cc1b5b66/Makefile#L100

PR #2708, opened 2018-04-17
Set published openshift image if requested

kargakis:
@kwoodson ptal


pkg/acsengine/defaults.go
```diff
@@ -83,6 +83,14 @@ var (
 		ImageVersion:   "latest",
 	}
 
+	//DefaultCentOSImageConfig is the OpenShift on Centos distribution.
+	DefaultCentOSImageConfig = AzureOSImageConfig{
+		ImageOffer:     "CentOS",
+		ImageSku:       "7.3",
+		ImagePublisher: "OpenLogic",
+		ImageVersion:   "latest",
```
kargakis: We should version these images I believe.

@pweil- @jim-minter

jim-minter: What's this for? We'll be building our own CentOS images, hopefully very shortly.

kargakis: I obviously don't understand what's going on with our published images. @kwoodson pointed me to this image and I thought it was one we published. Do you know how can I use our already existing published rhel image? It will need to be added in acs-engine if we want to use it in the ARM template.

pweil-: We will use something like this once we have the correct images published but it is probably too early now.  I also think we'll probably need to cue based on orchestrator version in some fashion so that we can differentiate between image versions since that directly corresponds with the OpenShift binary version.

jim-minter: Hey @sylus, I've heard that non-OpenLogic azure CentOS images are in the pipeline, still a little way out, but it's a different team to the folks here.  You'll have to ask via CentOS channels.

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -239,7 +249,7 @@
          "count": "[variables('{{.Name}}Count')]",
          "name": "vmLoopNode"
        },
-       "apiVersion": "2015-05-01-preview",
+       "apiVersion": {{if IsKubernetes}}"2015-05-01-preview"{{else}}"[variables('apiVersionVMDefault')]"{{end}},
```
pweil-: `apiVersionVMDefault` is only set if `IsOpenShift` do we need to be safe here and check that rather than `IsKubernetes`? (same for below)

kargakis: updated

pkg/acsengine/defaults.go
```diff
@@ -83,6 +83,14 @@ var (
 		ImageVersion:   "latest",
 	}
 
+	//DefaultOpenShift39RHELImageConfig is the OpenShift on RHEL distribution.
+	DefaultOpenShift39RHELImageConfig = AzureOSImageConfig{
+		ImageOffer:     "acsengine-preview",
```
kwoodson: do we want our default to be origin?  or do we prefer rhel here?

kargakis: This is not actually a default that is set in the ARM template and the variable name is probably confusing but I just followed the naming convention used in the rest of the image config vars in this file.

kargakis:
I think this is ready for reviews @pweil- @jim-minter, ptal

Prerequisite for getting the e2e test merged.

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -117,7 +117,9 @@
     },
 {{end}}
   {
-    {{if .IsManagedDisks}}
+    {{if IsOpenShift}}
+      "apiVersion": "[variables('apiVersionVMDefault')]",
```
jim-minter: @jackfrancis @CecileRobertMichon 
so, we need a newer apiversion to handle including the "plan" fields.  We're guessing we can't just increase apiVersionDefault.  Are you OK with adding apiVersionVMDefault?  Should the name be OpenShift specific?

parts/k8s/kubernetesmasterresources.t
```diff
@@ -894,7 +904,7 @@
           {{end}}
         },
         "storageProfile": {
-          {{if not UseMasterCustomImage}}
+          {{if and (not UseMasterCustomImage) IsKubernetes}}
```
jim-minter: @kargakis can we prevent specification of both an imageref and a plan in validation instead?

parts/k8s/kubernetesagentresourcesvmas.t
jim-minter: @jackfrancis @CecileRobertMichon please ignore the above questions ;)

parts/k8s/kubernetesmasterresources.t
kargakis: Seems irrelevant to this change? Background: this is adding a disk for etcd that we don't configure in openshift-ansible so I am disabling it for us altogether.

jim-minter:
@kargakis I was able to make acs-engine deploy from acsengine-preview with much less intrusive changes to the JSON.  No API version changes.  Here's the patch that I made to my azuredeploy.json.

```diff
--- jminterose/azuredeploy.json	2018-04-19 11:45:29.232157676 -0500
+++ jminterosenew/azuredeploy.json	2018-04-19 11:37:21.453438293 -0500
@@ -2571,7 +2571,10 @@
         },
         "storageProfile": {
           "imageReference": {
-            "id": "[resourceId(variables('computeosImageResourceGroup'), 'Microsoft.Compute/images', variables('computeosImageName'))]"
+            "offer": "acsengine-preview",
+            "publisher": "redhat",
+            "sku": "rhel74",
+            "version": "latest"
           },
           "osDisk": {
             "caching": "ReadWrite",
@@ -2579,6 +2582,11 @@
           }
         }
       },
+      "plan": {
+        "name": "rhel74",
+        "publisher": "redhat",
+        "product": "acsengine-preview"
+      },
       "tags": {
         "creationSource": "[concat(variables('generatorCode'), '-', variables('computeVMNamePrefix'), copyIndex(variables('computeOffset')))]",
         "orchestrator": "[variables('orchestratorNameVersionTag')]",
@@ -2697,7 +2705,10 @@
         },
         "storageProfile": {
           "imageReference": {
-            "id": "[resourceId(variables('infraosImageResourceGroup'), 'Microsoft.Compute/images', variables('infraosImageName'))]"
+            "offer": "acsengine-preview",
+            "publisher": "redhat",
+            "sku": "rhel74",
+            "version": "latest"
           },
           "osDisk": {
             "caching": "ReadWrite",
@@ -2705,6 +2716,11 @@
           }
         }
       },
+      "plan": {
+        "name": "rhel74",
+        "publisher": "redhat",
+        "product": "acsengine-preview"
+      },
       "tags": {
         "creationSource": "[concat(variables('generatorCode'), '-', variables('infraVMNamePrefix'), copyIndex(variables('infraOffset')))]",
         "orchestrator": "[variables('orchestratorNameVersionTag')]",
@@ -3145,7 +3161,10 @@
         },
         "storageProfile": {
           "imageReference": {
-            "id": "[resourceId(variables('osImageResourceGroup'), 'Microsoft.Compute/images', variables('osImageName'))]"
+            "offer": "acsengine-preview",
+            "publisher": "redhat",
+            "sku": "rhel74",
+            "version": "latest"
           },
           "osDisk": {
             "caching": "ReadWrite",
@@ -3153,6 +3172,11 @@
           }
         }
       },
+      "plan": {
+        "name": "rhel74",
+        "publisher": "redhat",
+        "product": "acsengine-preview"
+      },
       "tags": {
         "creationSource": "[concat(variables('generatorCode'), '-', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')))]",
         "orchestrator": "[variables('orchestratorNameVersionTag')]",
```

I think this PR should be possible without the API version changes, and without the changes to parts/k8s/kubernetesparams.t.  Worst case, guarded by an "if OpenShift", pipe the publisher offer and sku if they exist through to plan.publisher, plan.product, plan.name?

kargakis:
> I think this PR should be possible without the API version changes

I prefer we start using the latest API version before others start using our work. There are breaking changes in between `2016-04-30-preview` and `2017-03-30` (see https://docs.microsoft.com/en-us/azure/virtual-machines/windows/using-managed-disks-template-deployments#managed-disks-template-formatting). Also, if you use unmanaged disks in your template, VMs are defaulting to an even older api version that has no support for `plan` or `imageReference.id`.

> and without the changes to parts/k8s/kubernetesparams.t. 

Are these (`plan` values and `imageReference` related values) always the same?

kargakis:
> Are these (plan values and imageReference related values) always the same?

Also a question for @jackfrancis @CecileRobertMichon, cs.Plan wasn't hooked to the k8s templates before. Any reason why?

CecileRobertMichon:
> Also a question for @jackfrancis @CecileRobertMichon, cs.Plan wasn't hooked to the k8s templates before. Any reason why?

Not that I'm aware of @kargakis

jim-minter:
> I prefer we start using the latest API version before others start using our work. There are breaking changes in between 2016-04-30-preview and 2017-03-30 (see https://docs.microsoft.com/en-us/azure/virtual-machines/windows/using-managed-disks-template-deployments#managed-disks-template-formatting). Also, if you use unmanaged disks in your template, VMs are defaulting to an even older api version that has no support for plan or imageReference.id.

I believe OpenShift is intending always to use managed disks - is that right?

I appreciate the wish to use the latest API but we should also balance that against drift from k8s and additional complexity.  It's not clear to me what the best choice is at this point.  Do others have a view?

kargakis:
> and without the changes to parts/k8s/kubernetesparams.t.

Will this work if you don't need a plan to deploy a published image like you can do today with stock images? is another question that needs to be answered assuming we don't configure `plan` in the VM based on acs-engine's `plan` but use the imageReference fields.

parts/k8s/kubernetesmasterresources.t
jim-minter: Wonder whether we should discuss this in a separate issue or card? Wonder whether we should be making use of that etcd disk?

pweil-: +1 worth at least discussing at the next arch call

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -151,6 +151,14 @@
         "type": "systemAssigned"
       },
       {{end}}
+      {{if HasPlan}}
```
jim-minter: does this need to be if IsOpenShift && HasPlan?

parts/k8s/kubernetesparams.t
```diff
@@ -1,3 +1,31 @@
+    "planName": {
```
jim-minter: I don't understand why these new parameters are being added and we can't use osImageOffer, osImageSKU, etc?

kargakis: I don't think `plan` should be set unconditionally in VMs. I think we need to explicit about when we use it.

kargakis: Well, for image references we would avoid setting plan...

jim-minter:
Discussed on BlueJeans - aiming to gate plan changes on OpenShift and !UseMasterCustomImage and remove additional parameters before merging

kargakis:
@jim-minter ptal in the last commit

kargakis:
I can confirm this works with both published and custom images

jim-minter:
lgtm, @CecileRobertMichon please merge

jim-minter:
many thanks @kargakis !

kargakis:
@CecileRobertMichon ptal, I need this to move forward with the openshift e2e test

parts/k8s/kubernetesmasterresources.t
CecileRobertMichon: Let's change `IsKubernetes` to `not IsOpenshift`. I assume that's the intention here. It might be confusing to have `IsKubernetes` in a kubernetes template file.

jim-minter: @CecileRobertMichon I agree, any chance this can be done as a follow-up?  @kargakis is blocked on this for his CI work and would lose most of another day if we wait.  Also I opened #2750 to handle all instances of this issue

jackfrancis: @jim-minter @CecileRobertMichon that's fine, thanks for tracking this work generally. Let's aim to get #2750 in place this week so it makes it into v0.17.0 (to avoid unnecessary incremental changes to `kubernetesmasterresources.t` across minor versions

kargakis: Opened https://github.com/Azure/acs-engine/pull/2765

Issue #2690, opened 2018-04-14
openshift: move parts/k8s/openshiftmasterscript.sh and friends to parts/openshift


kargakis:
Moved from https://github.com/kargakis/acs-engine/issues/98

@jim-minter

jim-minter:
@kwoodson - worth taking a look at?

kwoodson:
@jim-minter https://github.com/Azure/acs-engine/pull/2746

PR #2746, opened 2018-04-22
Moving openshift scripts to parts/openshift.

kwoodson:
In order to create PR gating for Openshift scripts we need to separate the openshift parts into their own directory.

fixes: #2690

I was able to perform an installation and verify these scripts landed on the nodes.


kargakis:
lgtm

pkg/acsengine/engine.go
```diff
@@ -1746,7 +1746,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			return helpers.IsTrueBoolPointer(cs.Properties.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy)
 		},
 		"OpenShiftGetMasterSh": func() string {
-			tb := MustAsset("k8s/openshiftmasterscript.sh")
+			tb := MustAsset("openshift/openshiftmasterscript.sh")
```
jim-minter: use `openshiftMasterScript` here

pkg/acsengine/engine.go
```diff
@@ -1763,7 +1763,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			return b.String()
 		},
 		"OpenShiftGetNodeSh": func(profile *api.AgentPoolProfile) string {
-			tb := MustAsset("k8s/openshiftnodescript.sh")
+			tb := MustAsset("openshift/openshiftnodescript.sh")
```
jim-minter: use `openshiftNodeScript` here

jim-minter:
minor nits - also you can place `fixes: #2690` in the PR header if you like

pkg/acsengine/engine.go
kargakis: Let's stick with lowercased letters since that's what's followed elsewhere?

jim-minter: sorry, unclear review from me - I mean, use the constant called `openshiftMasterScript` which is already defined

jim-minter:
@CecileRobertMichon we were asked to merge our work into parts/k8s to keep the openshift and k8s templates as identical as possible and be able to immediately leverage future improvements made to parts/k8s.  Clearly, the cost of that is the `{{if IsOpenShift}}`s which we're trying to minimise.  I don't think it makes sense to fully duplicate parts/k8s to parts/openshift.  Also I don't in general like the idea of a `{{if IsOpenShift}}{{template "../openshift/foo.t}}{{end}}` pattern although it could be done in a few places if essential.  On my side I think I'd rather take the hit of non-automatic PR merge to parts/k8s than duplicate the codebase.

CecileRobertMichon:
@jim-minter that seems reasonable, as long as we all agree that this means that PRs that touch the `parts/k8s` templates by the openshift team will need review from someone from acs-engine and, similarly, we will need to ensure that changes made in those files to address k8s specific scenarios don't cause regressions for openshift (which e2e tests will help us do!).

PR #2759, opened 2018-04-24
stop using registry.reg-aws.openshift.com for now

jim-minter:
also adds some missing error checking


jim-minter:
@CecileRobertMichon or @jackfrancis please merge.

@kargakis @pweil- fyi

PR #2762, opened 2018-04-24
Provisioning of custom files when deploying k8s cluster

pontusarfwedson:
This PR adds possibility to pass in custom files to the master nodes. The use case for this is when adding an admission control config file as 
```
"apiServerConfig": {
          "--admission-control-config-file":  "/etc/kubernetes/admissioncontrol.yaml"
        }
```
In that case, we need to provision the yaml file somehow. And since I did not find anyway to do this I added so it can be added to the `kubernetes.json`. It currently only works with the `vlabs` api version (which is the one we use). Since mapping is done between all the version more work might be required to support it or other api versions.

The way we now can use this in our `kubernetesConfig` is

 ```{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorVersion": "1.10.0",
      "kubernetesConfig": {
        "apiServerConfig": {
          "--admission-control-config-file":  "/etc/kubernetes/admissioncontrol.yaml"
        },
        "customFiles":[
          {"source":"<path>/admissioncontrol.yaml",
           "dest":"/etc/kubernetes/admissioncontrol.yaml"
          },
          {"source":"<path>/podnodeselector.yaml",
           "dest":"/etc/kubernetes/podnodeselector.yaml"
          }
        ]
      }

...

```
where we need to use the full absolute path for both source and dest.

This PR might as mentioned not work for all api versions so are open for feedback regarding implementing for all the api versions it should work for.

**If applicable**:

**Release note**:
NONE

### DIFF
--- a/examples/customfiles/kubernetes-customfiles-podnodeselector.json
+++ b/examples/customfiles/kubernetes-customfiles-podnodeselector.json
@@ -1,0 +1,59 @@
+{
+    "apiVersion": "vlabs",
+    "properties": {
+      "orchestratorProfile": {
+        "orchestratorType": "Kubernetes",
+        "orchestratorVersion": "1.10.2",
+        "kubernetesConfig": {
+          "enableRbac" : true,
+          "enableAggregatedAPIs": true,
+          "apiServerConfig": {
+            "--admission-control-config-file":  "/etc/kubernetes/admissioncontrol.yaml"
+          }
+        }
+      },
+      "aadProfile": {
+        "serverAppID": "",
+        "clientAppID": "",
+        "tenantID":    ""
+      },
+      "masterProfile": {
+        "count": 3,
+        "dnsPrefix": "",
+        "vmSize": "Standard_D4s_v3",
+        "customFiles": [
+            { 
+                "source":"/Users/user/files/admissioncontrol.yaml",
+                "dest":"/etc/kubernetes/admissioncontrol.yaml"
+            },
+            { 
+                "source":"/Users/user/files/podnodeselector.yaml",
+                "dest":"/etc/kubernetes/podnodeselector..yaml"
+            }
+        ]
+      },
+      "agentPoolProfiles": [
+        {
+          "name": "agentpool1",
+          "count": 4,
+          "vmSize": "Standard_D4s_v3",
+          "availabilityProfile": "VirtualMachineScaleSets",
+          "storageProfile": "ManagedDisks"
+        }
+      ],
+      "linuxProfile": {
+        "adminUsername": "admin",
+        "ssh": {
+          "publicKeys": [
+            {
+              "keyData": ""
+            }
+          ]
+        }
+      },
+      "servicePrincipalProfile": {
+        "clientId": "",
+        "secret": ""
+      }
+    }
+  }
--- a/examples/customfiles/kubernetes-customfiles-podnodeselector.json
+++ b/examples/customfiles/kubernetes-customfiles-podnodeselector.json
@@ -28,7 +28,7 @@
             },
             { 
                 "source":"/Users/user/files/podnodeselector.yaml",
-                "dest":"/etc/kubernetes/podnodeselector..yaml"
+                "dest":"/etc/kubernetes/podnodeselector.yaml"
             }
         ]
       },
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: pkg/api/types.go [hunk 1]
```diff
@@ -306,6 +306,14 @@ type KubernetesConfig struct {
 	CtrlMgrNodeMonitorGracePeriod    string            `json:"ctrlMgrNodeMonitorGracePeriod,omitempty"`
 	CtrlMgrPodEvictionTimeout        string            `json:"ctrlMgrPodEvictionTimeout,omitempty"`
 	CtrlMgrRouteReconciliationPeriod string            `json:"ctrlMgrRouteReconciliationPeriod,omitempty"`
+	CustomFiles                      *[]CustomFile     `json:"customFiles,omitempty"`
```
CecileRobertMichon: What is the reasoning behind putting this field in `KubernetesConfig`? In the example you gave it makes sense but could this feature be used to upload files to a master node for more general purposes? Does it only work for Kubernetes clusters?

pkg/api/vlabs/types.go
```diff
@@ -300,6 +300,14 @@ type KubernetesConfig struct {
 	CloudProviderRateLimit          bool              `json:"cloudProviderRateLimit,omitempty"`
 	CloudProviderRateLimitQPS       float64           `json:"cloudProviderRateLimitQPS,omitempty"`
 	CloudProviderRateLimitBucket    int               `json:"cloudProviderRateLimitBucket,omitempty"`
+	CustomFiles                     *[]CustomFile     `json:"customFiles,omitempty"`
```
CecileRobertMichon: Could you please add a short description of `CustomFiles` in the docs? https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md is a good place to start

CecileRobertMichon:
Hi @pontusarfwedson, thank you for the PR / sorry for the slow response... I've added a few comments, otherwise lgtm!

jackfrancis:
/approve
/lgtm

pending E2E

@pontusarfwedson Could you submit a follow-up PR with some real-world use-case documentation, preferably a new directory under `examples/` with an api model using this new interface, and a file in the same directory to demonstrate how to push files to a cluster?

Thanks for this!

pkg/api/types.go
pontusarfwedson: Good point. The feature itself can provision whatever files to whatever directory (where there's permission) so yeah, it's not conceptually a kubernetes feature. The current state of the PR only provisions files to master nodes. So either add as a field to `masterProfile` => `$.properties.masterProfile.customFiles or simply have it on its own, so `$.properties.masterCustomFiles` (in the kubernetes.json). What do you think?

jackfrancis: I like the former approach, let's move everything over into the `masterProfile` struct. Thanks!

examples/customfiles/kubernetes-customfiles-podnodeselector.json
```diff
@@ -0,0 +1,59 @@
...
+        "dnsPrefix": "",
+        "vmSize": "Standard_D4s_v3",
+        "customFiles": [
+            { 
+                "source":"/Users/user/files/admissioncontrol.yaml",
+                "dest":"/etc/kubernetes/admissioncontrol.yaml"
+            },
+            { 
+                "source":"/Users/user/files/podnodeselector.yaml",
+                "dest":"/etc/kubernetes/podnodeselector..yaml"
```
CecileRobertMichon: I think this is a typo `podnodeselector..yaml` --> `podnodeselector.yaml`

CecileRobertMichon:
Thanks @pontusarfwedson! I let you fix the merge conflict + the typo in the example and then we'll run e2e and merge

jackfrancis:
FYI @pontusarfwedson I rebased/force-pushed to resolve `engine.go` conflict (my change was at fault :) )

Also fixed the typo in the examples json

pontusarfwedson:
@jackfrancis alright, sounds good :)

pontusarfwedson:
Alright, one test still failing. Any leads to why? 😛

jackfrancis:
/approve
/lgtm

@pontusarfwedson flaky test :)

PR #2764, opened 2018-04-24
normalizeazureregion added in scale, upgrade, and deploy

### LINKED ISSUE #2258: acs-engine admits location format that causes failure much later

---

---

**What version of acs-engine?**: Current master at 12 Feb 2018

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**  Kubernetes

**What happened**:

I created a cluster using `acs-engine deploy --location "Australia Southeast"`.  This worked fine and the cluster was created in the Australia Southeast region.

Then I upgraded the cluster using `acs-engine upgrade --location "Australia Southeast" ...other parameters...`.  This command was accepted and the upgrade process was kicked off.  After many minutes the upgrade command attempted an ARM deployment which appeared to succeed, but then failed validation because it used the region name in the URL:

```
INFO[0279] Starting ARM Deployment (master-18-02-12T21.42.04-813459685). This will take some time...
INFO[0741] Finished ARM Deployment (master-18-02-12T21.42.04-813459685). Error: <nil>
INFO[0741] Error validating upgraded master VM: k8s-master-23389264-0
FATA[0741] Error upgrading cluster: host must be a URL or a host:port pair:
"https://itowlsonacsupgradetest.Australia Southeast.cloudapp.azure.com"
                                ^^^^^^^^^^^^^^^^^^^
```

**What you expected to happen**:  Either acs-engine should accept only the URL forms of region names (e.g. `australiasoutheast`), so that it fails quickly rather than during an upgrade, or it should map the friendly names when required.

Upgrade a cluster using a friendly region name that contains spaces (e.g. "Australia Southeast").

elizabethhalper:
**What this PR does / why we need it**: It solves the issue outlined in: https://github.com/Azure/acs-engine/issues/2258

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2258 

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -115,6 +115,8 @@ func (dc *deployCmd) validate(cmd *cobra.Command, args []string) error {
 
 	if dc.location == "" {
 		return fmt.Errorf(fmt.Sprintf("--location must be specified"))
+	} else {
+		dc.location = NormalizeAzureRegion(dc.location);
 	}
 	// skip validating the model fields for now
 	dc.containerService, dc.apiVersion, err = apiloader.LoadContainerServiceFromFile(dc.apimodelPath, false, false, nil)
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/leonelquinteros/gotext"
@@ -20,6 +19,7 @@ import (
 	"github.com/Azure/acs-engine/pkg/acsengine/transform"
 	"github.com/Azure/acs-engine/pkg/api"
 	"github.com/Azure/acs-engine/pkg/armhelpers"
+	"github.com/Azure/acs-engine/pkg/helpers"
 	"github.com/Azure/acs-engine/pkg/i18n"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/deploy.go
```diff
@@ -83,6 +84,11 @@ func newDeployCmd() *cobra.Command {
 	return deployCmd
 }
 
+// NormalizeAzureRegion returns a normalized Azure region with whilte spaces removed and converted to lower case
```
CecileRobertMichon: typo in the comment `whilte` --> `white`

CecileRobertMichon: We could add this function to https://github.com/Azure/acs-engine/blob/master/pkg/helpers/helpers.go and re-use it here and in pkg/api

pkg/helpers/helpers_test.go
```diff
@@ -9,3 +9,32 @@ func TestPointerToBool(t *testing.T) {
 		t.Fatalf("expected PointerToBool(true) to return *true, instead returned %#v", ret)
 	}
 }
+
+func TestIsRegionNormalized(t *testing.T) {
```
CecileRobertMichon: 👍

pkg/helpers/helpers_test.go
```diff
@@ -9,3 +9,34 @@ func TestPointerToBool(t *testing.T) {
...
+		{
+			input:          "",
+			expectedResult: "",
+		},
+	}
+
+	for _, c := range cases {
+		result := NormalizeAzureRegion(c.input)
+		if c.expectedResult != result {
+			t.Fatalf("NormalizeAzureRegion returned unexpected result: expected %t but got %t", c.expectedResult, result)
```
CecileRobertMichon: `%s` instead of `%t` here since we're printing string types

PR #2772, opened 2018-04-25
fix origin openshift-ansible deploys on centos

maxamillion:
Signed-off-by: Adam Miller <admiller@redhat.com>

This fixes a number of issues with the OpenShift Origin openshift-ansible deploy on CentOS images.


parts/openshift/openshiftmasterscript.sh
```diff
@@ -3,18 +3,34 @@
...
+	ANSIBLE_CONTAINER_VERSION="v${VERSION}"
+	PROMETHEUS_VERSION="v${VERSION}"
+	PROMETHEUS_EXPORTER_VERSION="v${VERSION}"
+	PROMETHEUS_ALERT_BUFFER_VERSION="v${VERSION}"
+	PROMETHEUS_ALERT_MANAGER_VERSION="v${VERSION}"
+else
+	ANSIBLE_DEPLOY_TYPE="origin"
+	IMAGE_TYPE="${SERVICE_TYPE}"
+	IMAGE_PREFIX="openshift"
+	# FIXME: These versions are set to deal with differences in how Origin and OCP
```
kargakis: Topic for discussion in the next arch call @pweil- @jim-minter ?

jim-minter: I was discussing this on IRC with maxamillion and didn't update the PR - I'm of the opinion that this can/should be removed from the PR before it merges.

jim-minter: (at least all the new PROMETHEUS_VERSION bits - the existing ANSIBLE_DEPLOY_TYPE/IMAGE_PREFIX stuff is required and will end up getting baked into the image one day)

maxamillion: What specifically should be removed?

pkg/openshift/certgen/templates/master/tmp/ansible/azure-local-master-inventory.yml
```diff
@@ -74,6 +72,10 @@ localmaster:
     template_service_broker_image_name: "IMAGE_TYPE-template-service-broker"
     template_service_broker_version: "vVERSION"
 
+    # NOTE: This is redundant, but required to handle version skew between
+    #       openshift-ansible in Origin and OCP
+    openshift_examples_content_version: "vSHORT_VER"
```
jim-minter: is SHORT_VER still defined in this PR?  Looks like it's been dropped?

parts/openshift/openshiftmasterscript.sh
```diff
@@ -62,6 +71,15 @@ set -x
 routerLBHost="{{.RouterLBHostname}}"
 routerLBIP=$(dig +short $routerLBHost)
 
+# NOTE: The version of openshift-ansible for origin defaults the ansible var
+#       openshift_prometheus_node_exporter_image_version correctly as needed by
+#       origin, but for OCP it does not.
```
jim-minter: This is fixed in https://github.com/openshift/openshift-ansible/commit/c27a0f46c2250ee8c58ce2f050070cd7e42a4e49, which is in >= 3.9.15.  Please add a note to that effect, and a TODO to remove this once we're installing >= 3.9.15.

pkg/openshift/certgen/templates/master/tmp/ansible/azure-local-master-inventory.yml
maxamillion: it's not defined explicitly because I can do a bash parameter expansion to obtain the value and the sed statement is still there with `s|SHORT_VER|${VERSION%.*}|g;`

parts/openshift/openshiftmasterscript.sh
maxamillion: +1 - will do

pkg/openshift/certgen/templates/master/tmp/ansible/azure-local-master-inventory.yml
jim-minter: thanks, was looking in the wrong place

jim-minter:
lgtm, @CecileRobertMichon please merge

maxamillion:
@CecileRobertMichon looking at the CircleCI output, I don't think the failure is related to my change. Any chance we can restart the checks?

PR #2774, opened 2018-04-25
Use cloud init to mound etcd disks instead of using bash script

### LINKED ISSUE #676: Switch to using recently added udev symlinks for mounting etcd disks

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

Use the recently added udev symlinks instead of /dev/sdc for mounting etcddisk in Kubernetes clusters.

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes

**What happened**:
Current implementation does explicit mount of  /dev/sdc1 to /var/lib/etcddisk instead of using symlinks which might be error prone and not stable as it takes explicit dependency on /dev/sdc1.

Use the recently added udev symlinks instead of /dev/sdc. The kernel device names are not guaranteed to be stable. Use /dev/disk/azure/scsi1/lun0 for the disk attached on lun 0 and /dev/disk/azure/scsi1/lun0-part1 for the first partition on the disk.

### LINKED ISSUE #1643: Use cloud init to mound etcd disks instead of using bash script
NO
---

FEATURE REQUEST
---

0.8.0
---

PR #661 added a temporary fix for mounting disks until issue https://bugs.launchpad.net/cloud-init/+bug/1692093 was resolved. The issue has now been resolved in the latest Ubuntu image. PR #1521 updated the Ubuntu image in acs-engine. PR #661 should be reverted to switch back to using cloud-init.

Here are the steps to make sure the issue is still fixed after reverting the changes introduced in PR #661:

1.	Setup a simple service in Kubernetes
2.	Reboot the master
3.	Verify you still have the service

Repeat with 1 and 3 master solutions.  If etcd does not get reformatted and you still see the service, then this is working.

aheumaier:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1643

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
fixes #676
```


parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -301,13 +319,6 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
   content: !!binary |
     {{WrapAsVariable "provisionScript"}}
 
-- path: "/opt/azure/containers/mountetcd.sh"
-  permissions: "0744"
-  encoding: gzip
-  owner: "root"
-  content: !!binary |
-    {{WrapAsVariable "mountetcdScript"}}
```
CecileRobertMichon: We can probably remove the variable and script since they aren't being used anymore.

CecileRobertMichon:
https://github.com/Azure/acs-engine/blob/1b7cef869e51512b1e218ab80855f535da484a6d/parts/k8s/kubernetescustomscript.sh#L406 for validation

CecileRobertMichon:
@khenidak if you have a minute to take a look at this PR...

khenidak:
I think we should offload the entire thing to systemd unit files with `prestart` doing:

1- creating the mount point.
2- format if not formatted.

the advantages here is we can build dependancy kubelet-unit->etcd-unit->mount-unit that will work across all the operating systems we support (CoreOs, RHEL, CentOs and Ubuntu) and will give us better error reporting when things do not work as expected

aheumaier:
@khenidak  good point. Etcd mounting moved to etcd which helps with 1) . Autoformating with systemd.mounts seems just possible via fstab - not via unit files

CecileRobertMichon:
Closing because we moved to https://github.com/Azure/aks-engine/

PR #2795, opened 2018-04-27
OpenShift AAD automatic service principal and authentication integration

jim-minter:
With this PR it's no longer necessary to have a predefined service principal, and you can log into the deployed cluster using Azure AD credentials.


jim-minter:
@pweil- @kargakis this PR contains an optional commit which removes htpasswd-based authentication.  Now would be a good time to decide what to do about that.

jim-minter:
My take on htpasswd: it's not scalable long term, embedding a password in the openshift.json is bad, and it shouldn't be necessary for automated testing (use admin.kubeconfig and create other users on the fly if required).  The sooner it's removed, the fewer people will complain about it being removed.
OTOH, at the moment Azure AD authentication requires the caller of acs-engine to have the rights to create an application object (or possibly, a correctly preconfigured object to exist).  That represents some level of additional bar to entry for evaluation/test.

jim-minter:
also cc @khenidak 

Would appreciate feedback on this PR to take it forwards.

jackfrancis:
@jim-minter thanks for asking for some sanity checks about the version revs.

Reading through the azure-sdk-for-go docs, there are not breaking changes until v15, I wonder if it makes sense to go to `v14.6.0` and get a little more current. (ditto for go-autorest)

@amanohar @weinong are there any good historical reasons for using such old versions of the above packages?

jim-minter:
OK, htpasswd gets a reprieve for now.  Squashed to 2 commits.  @pweil- @kargakis ptal

jim-minter:
(I can't remove the DO-NOT-MERGE label) - reviewers: if happy, please request merge.

cmd/deploy.go
```diff
@@ -221,7 +223,29 @@ func autofillApimodel(dc *deployCmd) {
...
+			var replyURLs *[]string
+			var requiredResourceAccess *[]graphrbac.RequiredResourceAccess
+			if dc.containerService.Properties.OrchestratorProfile.OrchestratorType == api.OpenShift {
+				appName = fmt.Sprintf("%s.%s.cloudapp.azure.com", appName, dc.containerService.Properties.AzProfile.Location)
+				appURL = fmt.Sprintf("https://%s:8443/", appName)
+				replyURLs = to.StringSlicePtr([]string{fmt.Sprintf("https://%s:8443/oauth2callback/Azure%%20AD", appName)})
+				ra := []graphrbac.ResourceAccess{
+					{
+						// TODO: where is this UUID defined?
+						ID:   to.StringPtr("311a71cc-e848-46a1-bdf8-97ff7156d8e6"),
```
kargakis: ew, what's this?

cmd/deploy.go
```diff
@@ -221,7 +223,29 @@ func autofillApimodel(dc *deployCmd) {
...
 			appName := dc.containerService.Properties.MasterProfile.DNSPrefix
 			appURL := fmt.Sprintf("https://%s/", appName)
-			applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(appName, appURL)
+			var replyURLs *[]string
+			var requiredResourceAccess *[]graphrbac.RequiredResourceAccess
+			if dc.containerService.Properties.OrchestratorProfile.OrchestratorType == api.OpenShift {
+				appName = fmt.Sprintf("%s.%s.cloudapp.azure.com", appName, dc.containerService.Properties.AzProfile.Location)
+				appURL = fmt.Sprintf("https://%s:8443/", appName)
+				replyURLs = to.StringSlicePtr([]string{fmt.Sprintf("https://%s:8443/oauth2callback/Azure%%20AD", appName)})
+				ra := []graphrbac.ResourceAccess{
```
kargakis: Are we not able to express these resources as part of the ARM template? If not, then we are tied to `acs-engine deploy` for using AAD and a flow where we `acs-engine generate` then `az deploy` is not going to work.

pkg/api/vlabs/types.go
```diff
@@ -332,10 +332,12 @@ type DcosConfig struct {
 type OpenShiftConfig struct {
 	KubernetesConfig *KubernetesConfig `json:"kubernetesConfig,omitempty"`
 
-	// ClusterUsername and ClusterPassword are temporary before AAD
-	// authentication is enabled, and will be removed subsequently.
+	// ClusterUsername and ClusterPassword are temporary, do not rely on them.
```
kargakis: How about a top-level struct dedicated to authentication?

pkg/openshift/certgen/templates/master/etc/origin/master/master-config.yaml
```diff
@@ -156,10 +156,33 @@ oauthConfig:
...
+        - unique_name
+        name:
+        - name
+        email:
+        - email
+      urls:
+        authorize: https://login.microsoftonline.com/{{ .AzureConfig.TenantID }}/oauth2/authorize
+        token: https://login.microsoftonline.com/{{ .AzureConfig.TenantID }}/oauth2/token
+{{- end}}
+  - name: Local password
```
kargakis: I thought these two options were going to be mutually exclusive. Any reason why I would want to use htpasswd assuming I can already use AAD?

pkg/api/vlabs/types.go
pweil-: Will be talking about this at the next arch call, the first thought was a slimmed down usage of our IdentityProviders in order to keep  the configuration space the same (or nearly) as a normal master config but that had some complications that @jim-minter will explain

cmd/deploy.go
jim-minter: These grant the application service principal the necessary rights to carry out oauth against AAD. I haven't found an obvious explanation to exactly what these UUIDs mean or predefined consts in a library anywhere though.

pweil-: This corresponds with the User.Read scope.  I can't find where it's documented anywhere, I expected it here: https://msdn.microsoft.com/library/azure/ad/graph/howto/azure-ad-graph-api-permission-scopes 

We should create a constant for this at least.  Found [this](https://github.com/Azure/AzureStack-Tools/blob/master/Identity/GraphAPI/GraphAPI.psm1#L113-L120) on github.  Note it is slightly different for China cloud in that code.

```
AadPermissions = [HashTable]@{
                    AccessDirectoryAsSignedInUser      = "a42657d6-7f20-40e3-b6f0-cee03008a62a"
                    EnableSignOnAndReadUserProfiles    = "311a71cc-e848-46a1-bdf8-97ff7156d8e6"
                    ReadAllGroups                      = "6234d376-f627-4f0f-90e0-dff25c5211a3"
                    ReadAllUsersBasicProfile           = "cba73afc-7f69-4d86-8450-4978e04ecd1a"
                    ReadAllUsersFullProfile            = "c582532d-9d9e-43bd-a97c-2667a28ce295"
                    ReadDirectoryData                  = "5778995a-e1bf-45b8-affa-663a9f3f4d04"
                    ManageAppsThatThisAppCreatesOrOwns = "824c81eb-e3f8-4ee6-8f6d-de7f50d565b7"
                }
```

and [this](http://www.cloudidentity.com/blog/2015/09/01/azure-ad-permissions-summary-table/) reference.

pkg/openshift/certgen/templates/master/etc/origin/master/master-config.yaml
jim-minter: Arch call discussion.

cmd/deploy.go
```diff
@@ -221,7 +223,29 @@ func autofillApimodel(dc *deployCmd) {
...
+					{
+						// TODO: where is this UUID defined?
+						ID:   to.StringPtr("311a71cc-e848-46a1-bdf8-97ff7156d8e6"),
+						Type: to.StringPtr("Scope"),
+					},
+				}
+				rra := []graphrbac.RequiredResourceAccess{
+					{
+						// TODO: where is this UUID defined?
+						ResourceAppID:  to.StringPtr("00000002-0000-0000-c000-000000000000"),
```
pweil-: ditto on the constant here.  From what I could find this is always the reserved id for AD in the tenant.  It can be found by listing the service principals:

```
$ az ad sp list -o table

AppId                                 DisplayName                                                   
------------------------------------  ------------------------------------------------------------
00000002-0000-0000-c000-000000000000  Windows Azure Active DirectoryServicePrincipal
```

(output abbreviated)

kargakis:
@jim-minter question: why is [AADProfile](https://github.com/Azure/acs-engine/blob/1d7b57593028d901cb74a91d7892e2db52f29c32/pkg/api/types.go#L508-L522) not used here?

PR #2798, opened 2018-04-27
Validation for WindowsProfile

### LINKED ISSUE #2788: Windows profile validation missing when there is at least one windows agentpool
Repro:
- `acs-engine generate --api-model myapimodel` where `myapimodel` has at least one agentpoolprofile but no `windowsProfile`
- get an ugly error about has secret

We should validate that windowsProfile != nil when there is a windows agentPool and return an error along the lines of "windowsProfile is required when the cluster definition contains Windows agent pool(s)".

JamesEarle:
fixes #2788

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -235,6 +235,12 @@ func (a *AgentPoolProfile) Validate(orchestratorType string) error {
 		return e
 	}
 
+	if a.osType != nil {
+		if e := validatePoolOsType(a.osType); e != nil {
+			return e
+		}
+	}
+
 	// for Kubernetes, we don't support AgentPoolProfile.DNSPrefix
 	if orchestratorType == Kubernetes {
 		if e := validate.Var(a.DNSPrefix, "len=0"); e != nil {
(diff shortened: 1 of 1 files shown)


pkg/api/vlabs/validate.go
```diff
@@ -818,6 +828,13 @@ func validatePoolName(poolName string) error {
 	return nil
 }
 
+func validatePoolOSType(os OSType) error {
+	if os != "linux" && os != "windows" {
```
CecileRobertMichon: here you can use `api.Linux` and `api.Windows` instead

CecileRobertMichon:
You can remove https://github.com/Azure/acs-engine/pull/2798/files#diff-f45af0b14c127459636440a5981a6934R500 since it is redundant now

JamesEarle:
Requested changes made. Let me know if there is anything more that needs to be done for this

pkg/api/vlabs/validate.go
```diff
@@ -818,6 +823,14 @@ func validatePoolName(poolName string) error {
 	return nil
 }
 
+func validatePoolOSType(os OSType) error {
+	// Should we use a.IsLinux() and a.IsWindows() instead?
+	if os != Linux && os != Windows {
```
CecileRobertMichon: @JamesEarle just realized `""` is also a default value... If the user leaves osType blank the apimodel is valid (it defaults to Linux). that's why unit tests are failing :)

CecileRobertMichon: Regarding your comment, either way is fine. Feel free to do it this way and remove the comment.

PR #2813, opened 2018-04-30
rationalize network config and k8s NetworkPolicy

### LINKED ISSUE #1930: Separate network policy and IPAM
It would be nice to separate network policy from ip address management (IPAM). Currently they are overloaded with `networkpolicy` in the cluster definition. 

|IPAM|Policy|
|---|---|
|azure-vnet-ipam|NA|
|host-local|calico|
|azure-vnet-ipam|calico|

jackfrancis:
**What this PR does / why we need it**: This PR adds a `networkConfig` property to `kubernetesConfig` which carries configuration for the kubelet runtime `--network-config` option (and enforces related cluster configuration). Prior to this change, both network config (e.g., "kubenet") and networkPolicy enforcement (e.g., what "calico" and "cilium" do) were defined under a common `networkPolicy` configuration key. Now `networkPolicy` carries only "network policy" config; and `networkConfig` carries only "network config" config.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1930

**Special notes for your reviewer**: This PR aims to deliver changes that are backwards-compatible w/ existing usage patterns (though documentation has been update to reflect only the new pattern):

Here are the before/after cases:

# Kubenet
```
"kubernetesConfig": {
    "networkPolicy": "none"
}
```
is permissible and automatically converted at generation time to:
```
"kubernetesConfig": {
    "networkPlugin": "kubenet"
}
```
# Azure CNI
```
"kubernetesConfig": {
    "networkPolicy": "azure"
}
```
is permissible and automatically converted at generation time to:
```
"kubernetesConfig": {
    "networkPlugin": "azure"
}
```

**If applicable**:
- [x] documentation
- [x] unit tests

**Release note**:

```
rationalize network config and k8s NetworkPolicy
```


jackfrancis:
@slack @JunSun17 FYI

parts/k8s/kubernetescustomscript.sh
```diff
@@ -225,10 +225,10 @@ function configKubenet() {
     chmod -R 755 $CNI_BIN_DIR
 }
 
-function configNetworkPolicy() {
-    if [[ "${NETWORK_POLICY}" = "azure" ]]; then
+function configNetworkPlugin() {
```
CecileRobertMichon: this makes so much more sense now :)

pkg/api/vlabs/validate.go
```diff
@@ -805,6 +808,33 @@ func (a *KubernetesConfig) Validate(k8sVersion string) error {
 	return nil
 }
 
+func (a *Properties) validateNetworkPlugin() error {
```
CecileRobertMichon: Are there any values of `NetworkPlugin` that are not compatible with `NetworkPolicy` values right now other than using deprecated values of `NetworkPolicy` with a different `NetworkPlugin`? Maybe we can add a check and return an error if the user tries to set policy+plugin values that aren't compatible

pkg/acsengine/defaults.go
```diff
@@ -327,6 +327,17 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 		if o.KubernetesConfig == nil {
 			o.KubernetesConfig = &api.KubernetesConfig{}
 		}
+		// For backwards compatibility with original, overloaded "NetworkPolicy" config vector
+		// we translate deprecated NetworkPolicy usage to the NetworkConfig equivalent
+		// and set a default network policy enforcement configuration
+		switch o.KubernetesConfig.NetworkPolicy {
```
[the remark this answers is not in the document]
jackfrancis: The answer your 1st question: what happens in the current PR implementation is that any "deprecated" `NetworkPolicy` usage patterns are translated to their `NetworkPlugin` equivalents, and then `NetworkPolicy` is set to the default (currently `""`, i.e., network policy not enabled).

So, if someone uses a deprecated pattern (e.g., `"networkPolicy": "none") _and_ uses the new pattern in a pathological way (e.g., `"networkPlugin": "azure"`) we overwrite the pathology by assuming the deprecated usage pattern is authoritative.

jackfrancis:
@CecileRobertMichon see my penultimate commit (not the lint one)

pkg/api/vlabs/validate.go
```diff
@@ -27,6 +27,36 @@ var (
 		"3.1.0", "3.1.1", "3.1.2", "3.1.2", "3.1.3", "3.1.4", "3.1.5", "3.1.6", "3.1.7", "3.1.8", "3.1.9", "3.1.10",
 		"3.2.0", "3.2.1", "3.2.2", "3.2.3", "3.2.4", "3.2.5", "3.2.6", "3.2.7", "3.2.8", "3.2.9", "3.2.11", "3.2.12",
 		"3.2.13", "3.2.14", "3.2.15", "3.2.16", "3.3.0", "3.3.1"}
+	networkPluginPlusPolicyAllowed = []k8sNetworkConfig{
```
CecileRobertMichon: What about the case where policy is cilium or calico and plugin is empty?

jackfrancis: thanks! fixed

CecileRobertMichon: Still missing "" with "calico" I think

jackfrancis: thanks! fixed déja vu

pkg/acsengine/const.go
```diff
@@ -38,16 +38,20 @@ const (
...
+	// NetworkPluginAzure is the string expression for Azure CNI network policy
+	NetworkPluginAzure = "azure"
 	// NetworkPluginKubenet is the string expression for kubenet network plugin
 	NetworkPluginKubenet = "kubenet"
-	// DefaultNetworkPolicy defines the network policy to use by default
-	DefaultNetworkPolicy = NetworkPolicyNone
-	// DefaultNetworkPolicyWindows defines the network policy to use by default for clusters with Windows agent pools
-	DefaultNetworkPolicyWindows = NetworkPolicyNone
+	// DefaultNetworkPlugin defines the network plugin to use by default
+	DefaultNetworkPlugin = NetworkPluginKubenet
```
CecileRobertMichon: This is inconsistent with docs

CecileRobertMichon: nevermind this is the default for non-vlabs ^

PR #2820, opened 2018-05-01
Fix deploy using client-id and client-secret

### LINKED ISSUE #2657: Deploy using client-id and client-secret command line switches throws an error
**Is this a request for help?**: Yes

---

---

**What version of acs-engine?**:  v0.15.1

---

**What happened**:
We generate service principle on the fly and are not available to put in the api-model json file. I used the following command:
 `acs-engine deploy --subscription-id (redacted)  --resource-group (redacted) --dns-prefix (redacted) --location westus2 --auto-suffix --api-model clusterDefinition.json --client-id (redacted) --client-secret (redacted) --auth-method client_secret`.
However I go the following error message:
`FATA[0000] --client-id and --client-secret must be specified when --auth-method="client_secret"`

I expected a new cluster to be created.

Replace the redacted values and execute the following command:
 `acs-engine deploy --subscription-id (redacted)  --resource-group (redacted) --dns-prefix (redacted) --location westus2 --auto-suffix --api-model clusterDefinition.json --client-id (redacted) --client-secret (redacted) --auth-method client_secret`

I tried to run the same command but escaped the client-id and client-secret using quotes but I still get the same error.

maniSbindra:
**What this PR does / why we need it**: Fix Issue #2657 

**Which issue this PR fixes** *(optional, in `fixes #2657` format, will close that issue when PR gets merged)*: fixes #2657 `#2657`

* Added condition to check if Client Id and Secret are specified in the deployment command
* Added tests to cover conditions when Client Id and Client secret are provided in the deployment command, and when id and secret are not specified in either the api model or deployment command

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -241,6 +241,11 @@ func autofillApimodel(dc *deployCmd) {
 				Secret:   secret,
 				ObjectID: servicePrincipalObjectID,
 			}
+		} else if dc.containerService.Properties.ServicePrincipalProfile == nil && dc.ClientID.String() != "" && dc.ClientSecret != "" {
+			dc.containerService.Properties.ServicePrincipalProfile = &api.ServicePrincipalProfile{
+				ClientID: dc.ClientID.String(),
+				Secret:   dc.ClientSecret,
+			}
 		}
 	}
 }
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -216,7 +216,7 @@ func autofillApimodel(dc *deployCmd) {
 
 	if !useManagedIdentity {
 		spp := dc.containerService.Properties.ServicePrincipalProfile
-		if spp != nil && spp.ClientID == "" && spp.Secret == "" && spp.KeyvaultSecretRef == nil {
+		if spp != nil && spp.ClientID == "" && spp.Secret == "" && spp.KeyvaultSecretRef == nil && (dc.ClientID.String() == "" || dc.ClientID.String() == "00000000-0000-0000-0000-000000000000") && dc.ClientSecret == "" {
 			log.Warnln("apimodel: ServicePrincipalProfile was missing or empty, creating application...")
 
 			// TODO: consider caching the creds here so they persist between subsequent runs of 'deploy'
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -241,6 +241,11 @@ func autofillApimodel(dc *deployCmd) {
 				Secret:   secret,
 				ObjectID: servicePrincipalObjectID,
 			}
+		} else if dc.containerService.Properties.ServicePrincipalProfile == nil && dc.ClientID.String() != "" && dc.ClientSecret != "" {
```
CecileRobertMichon: what about the case where `dc.containerService.Properties.ServicePrincipalProfile` is not nil but is empty? For example, ``` "servicePrincipalProfile": {
      "clientId": "",
      "secret": ""
    }```

CecileRobertMichon: We need to address this case since it applies to most example json apimodel in the `examples/` folder

cmd/deploy_test.go
```diff
@@ -40,6 +41,19 @@ const ExampleAPIModelWithDNSPrefix = `{
   }
   `
 
+const ExampleAPIModelWithoutServicePrincipalProfile = `{
```
CecileRobertMichon: Let's also add a test case for empty `ServicePrincipalProfile`

CecileRobertMichon: 1. That's correct
2. Yes, exactly. It shouldn't be very complicated to add to this PR. I believe all you need to do is to change your if statement in deploy.go to `} else if (dc.containerService.Properties.ServicePrincipalProfile == nil || (dc.containerService.Properties.ServicePrincipalProfile.ClientID == ""  && dc.containerService.Properties.ServicePrincipalProfile.ClientSecret == "") ) && dc.ClientID.String() != "" && dc.ClientSecret != "" ` and add the appropriate unit test case. We just want to make sure that the change you made applies if the sp profile is declared but contains no ID and secret values.

maniSbindra: Thanks. Makes sense. So test cases for conditions where id and secret are specified in both the api model and deployment command can be taken up later.

maniSbindra: @CecileRobertMichon i have made the requested changes, please review and let me know if any further changes are needed

PR #2825, opened 2018-05-01
Deploying DCOS clusters using bootstrap node (starting v1.11)

dmitsh:
The proper way to deploy DCOS cluster is by using a bootstrap node : https://docs.mesosphere.com/1.11/installing/oss/custom/
This will also enable DCOS upgrades in the future.

This PR adds new set of templates for DCOS 1.11.0, while keeping intact the templates of former versions.


pkg/acsengine/azureconst.go
```diff
@@ -47,6 +47,15 @@ var AzureLocations = []string{
 	"francecentral",
 }
 
+// GetDCOSBootstrapAllowedSizes returns the bootstrap allowed sizes
```
jackfrancis: This file is auto-generated, so these changes need to derive from `pkg/acsengine/Get-AzureConstants.py` (and then an execution of that script).

PR #2834, opened 2018-05-01
remove route table when using cilium CNI

### LINKED ISSUE #2829: cilium + vmss creates uanessacry route table
When creating a cilium (CNI) based cluster a route table is created and  `controller-manager` is not configured with `--configure-cloud-routes=false` This causes 

1. Route controller is enabled and creates route for each node (not needed for cilium CNI)
2. Cluster scale will be limited by 400 (limits of the route table)

**Action**:
When `cilium` is used then don't create `route-table` resource in ARM template and disable the `route controller` in `controller manager`

khenidak:
fixes #2829 -- no route table + disable route controller when using cilium

WIP note: this PR is still under test.

### DIFF
--- a/parts/k8s/kubernetesmasterresources.t
+++ b/parts/k8s/kubernetesmasterresources.t
@@ -194,10 +194,10 @@
     },
 {{end}}
 {{if not .MasterProfile.IsCustomVNET}}
-    {
+{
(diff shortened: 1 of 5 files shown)


khenidak:
@jackfrancis @CecileRobertMichon ready for review..

PR #2860, opened 2018-05-04
add exit for docker pull in cse

CecileRobertMichon:
+ check if systemd services are active after starting them


parts/k8s/kubernetescustomscript.sh
```diff
@@ -3,17 +3,44 @@
...
+# | 65 | hyperkube-extract could not be started by systemctl |
 
 set -x
 source /opt/azure/containers/provision_source.sh
+globaltimeout=7200 # TODO: make this configurable
+exitcode=-1
+
+function timeout_script() {
+   sleep "$globaltimeout"
+   kill "$1"
```
CecileRobertMichon: currently script exits with code 143

PR #2883, opened 2018-05-08
[WIP] Adding necessary bits to master.sh to bootstrap 3.10 Openshift nodes.

kwoodson:
@jim-minter @kargakis @pweil- 

When moving to 3.10 Openshift nodes we need to bootstrap the node image so that the hosted control plane will run during start.


parts/openshift/openshiftmasterscript.sh
```diff
@@ -134,8 +148,11 @@ oc create configmap node-config-infra --namespace openshift-node --from-file=nod
 
 # must start ${SERVICE_TYPE}-node after master is fully up and running
 # otherwise the implicit dns change may cause master startup to fail
-systemctl enable ${SERVICE_TYPE}-node.service
-systemctl start ${SERVICE_TYPE}-node.service &
+cp /etc/origin/node/bootstrap.kubeconfig.orig /etc/origin/node/bootstrap.kubeconfig
```
kargakis: Should be a mv probably

parts/openshift/openshiftmasterscript.sh
```diff
@@ -95,14 +97,25 @@ for i in /etc/origin/master/master-config.yaml /tmp/bootstrapconfigs/* /tmp/ansi
...
+
+systemctl start ${SERVICE_TYPE}-node
+
 # note: ${SERVICE_TYPE}-node crash loops until master is up
-for unit in etcd.service ${SERVICE_TYPE}-master-api.service ${SERVICE_TYPE}-master-controllers.service; do
-	systemctl enable $unit
-	systemctl start $unit
+for pod in /etc/origin/node/disabled/*;
+do
+    cp "$pod" /etc/origin/node/pods
```
kwoodson: This should be a `mv`.

PR #2896, opened 2018-05-09
add VMAS scale up/down support for openshift

jim-minter:
adds VMAS scale up/down support for openshift

`acs-engine scale --resource-group $RG --location $LOCATION --subscription-id $SUB --new-node-count $COUNT --deployment-dir _output/$RG --node-pool $POOL --master-FQDN https://$RG.$LOCATION.cloudapp.azure.com:8443/`

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -18,6 +19,7 @@ import (
 	"github.com/Azure/acs-engine/pkg/armhelpers/utils"
 	"github.com/Azure/acs-engine/pkg/helpers"
 	"github.com/Azure/acs-engine/pkg/i18n"
+	"github.com/Azure/acs-engine/pkg/openshift/filesystem"
 	"github.com/Azure/acs-engine/pkg/operations"
 	"github.com/leonelquinteros/gotext"
 
@@ -51,8 +53,8 @@ type scaleCmd struct {
 
 const (
 	scaleName             = "scale"
-	scaleShortDescription = "Scale an existing Kubernetes cluster"
-	scaleLongDescription  = "Scale an existing Kubernetes cluster by specifying increasing or decreasing the node count of an agentpool"
+	scaleShortDescription = "Scale an existing Kubernetes or OpenShift cluster"
+	scaleLongDescription  = "Scale an existing Kubernetes or OpenShift cluster by specifying increasing or decreasing the node count of an agentpool"
 )
 
 // NewScaleCmd run a command to upgrade a Kubernetes cluster
@@ -237,11 +239,31 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				vmsToDelete = append(vmsToDelete, indexToVM[i])
 			}
 
-			if orchestratorInfo.OrchestratorType == api.Kubernetes {
-				err = sc.drainNodes(vmsToDelete)
+			switch orchestratorInfo.OrchestratorType {
+			case api.Kubernetes:
+				kubeConfig, err := acsengine.GenerateKubeConfig(sc.containerService.Properties, sc.location)
+				if err != nil {
+					log.Errorf("failed to generate kube config: %v", err)
+					return err
+				}
+				err = sc.drainNodes(kubeConfig, vmsToDelete)
 				if err != nil {
 					log.Errorf("Got error %+v, while draining the nodes to be deleted", err)
 					return err
+				}
+			case api.OpenShift:
+				bundle := bytes.NewReader(sc.containerService.Properties.OrchestratorProfile.OpenShiftConfig.ConfigBundles["master"])
+				fs, err := filesystem.NewTGZReader(bundle)
+				if err != nil {
+					return fmt.Errorf("failed to read master bundle: %v", err)
+				}
+				kubeConfig, err := fs.ReadFile("etc/origin/master/admin.kubeconfig")
+				if err != nil {
+					return fmt.Errorf("failed to read kube config: %v", err)
+				}
+				err = sc.drainNodes(string(kubeConfig), vmsToDelete)
+				if err != nil {
+					return fmt.Errorf("Got error %v, while draining the nodes to be deleted", err)
 				}
 			}
 
@@ -320,6 +342,14 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	addValue(parametersJSON, sc.agentPool.Name+"Count", countForTemplate)
 
 	switch orchestratorInfo.OrchestratorType {
+	case api.OpenShift:
+		err = transformer.NormalizeForOpenShiftVMASScalingUp(sc.logger, sc.agentPool.Name, templateJSON)
+		if err != nil {
+			return fmt.Errorf("error tranforming the template for scaling template %s: %v", sc.apiModelPath, err)
+		}
+		if sc.agentPool.IsAvailabilitySets() {
+			addValue(parametersJSON, fmt.Sprintf("%sOffset", sc.agentPool.Name), highestUsedIndex+1)
+		}
 	case api.Kubernetes:
 		err = transformer.NormalizeForK8sVMASScalingUp(sc.logger, templateJSON)
 		if err != nil {
@@ -387,11 +417,7 @@ func addValue(m paramsMap, k string, v interface{}) {
 	}
 }
 
-func (sc *scaleCmd) drainNodes(vmsToDelete []string) error {
-	kubeConfig, err := acsengine.GenerateKubeConfig(sc.containerService.Properties, sc.location)
-	if err != nil {
-		log.Fatalf("failed to generate kube config: %v", err) // TODO: cleanup
-	}
+func (sc *scaleCmd) drainNodes(kubeConfig string, vmsToDelete []string) error {
 	masterURL := sc.masterFQDN
 	if !strings.HasPrefix(masterURL, "https://") {
 		masterURL = fmt.Sprintf("https://%s", masterURL)
@@ -401,7 +427,7 @@ func (sc *scaleCmd) drainNodes(vmsToDelete []string) error {
 	defer close(errChan)
 	for _, vmName := range vmsToDelete {
 		go func(vmName string) {
-			err = operations.SafelyDrainNode(sc.client, sc.logger,
+			err := operations.SafelyDrainNode(sc.client, sc.logger,
 				masterURL, kubeConfig, vmName, time.Duration(60)*time.Minute)
 			if err != nil {
 				log.Errorf("Failed to drain node %s, got error %v", vmName, err)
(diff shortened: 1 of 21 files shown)


jim-minter:
@kargakis @pweil- ptal

@kargakis, do you think this functionality can/should be integrated into e2e tests? any thoughts on how?

@kargakis assuming we go ahead persisting the bundles in the apimodel it should be possible to have the e2e test retrieve the kubeconfig from there -- shall I open a follow-up issue to track that?

cmd/scale.go
```diff
@@ -320,6 +345,15 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	addValue(parametersJSON, sc.agentPool.Name+"Count", countForTemplate)
 
 	switch orchestratorInfo.OrchestratorType {
+	case api.OpenShift:
+		err = transformer.NormalizeForOpenShiftVMASScalingUp(sc.logger, sc.agentPool.Name, templateJSON)
+		if err != nil {
+			log.Fatalf("error tranforming the template for scaling template %s: %s", sc.apiModelPath, err.Error())
+			os.Exit(1)
```
kargakis: Fatal already os.Exits

cmd/scale.go
```diff
@@ -237,8 +239,31 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
...
 			}
 
-			if orchestratorInfo.OrchestratorType == api.Kubernetes {
-				err = sc.drainNodes(vmsToDelete)
+			switch orchestratorInfo.OrchestratorType {
+			case api.Kubernetes:
+				kubeConfig, err := acsengine.GenerateKubeConfig(sc.containerService.Properties, sc.location)
+				if err != nil {
+					log.Errorf("failed to generate kube config: %v", err)
+					return err
```
kargakis: Logging an error and returning it seems redundant. What is the caller doing with it? Maybe just 
```go
return fmt.Errorf("failed to generate kube config: %v", err)
```

cmd/scale.go
```diff
@@ -237,8 +239,31 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
...
+				}
+			case api.OpenShift:
+				bundle := bytes.NewReader(sc.containerService.Properties.OrchestratorProfile.OpenShiftConfig.ConfigBundles["master"])
+				fs, err := filesystem.NewTGZReader(bundle)
+				if err != nil {
+					log.Errorf("failed to read master bundle: %v", err)
+					return err
+				}
+				kubeConfig, err := fs.ReadFile("etc/origin/master/admin.kubeconfig")
```
kargakis: FWIW, In Kubernetes it seems they are generating a new kubeconfig so they are not reusing the one that was generated during cluster startup.

pkg/acsengine/transform/transform.go
```diff
@@ -97,6 +99,54 @@ func (t *Transformer) NormalizeForVMSSScaling(logger *logrus.Entry, templateMap
 	return nil
 }
 
+// NormalizeForOpenShiftVMASScalingUp takes a template and removes elements that are unwanted in a K8s VMAS scale up/down case
```
kargakis: s/K8s/OpenShift/

pkg/acsengine/transform/transform.go
```diff
@@ -97,6 +99,54 @@ func (t *Transformer) NormalizeForVMSSScaling(logger *logrus.Entry, templateMap
...
+func (t *Transformer) NormalizeForOpenShiftVMASScalingUp(logger *logrus.Entry, agentPoolName string, templateMap map[string]interface{}) (err error) {
```
kargakis: Please add a unit test for this

kargakis:
> @kargakis assuming we go ahead persisting the bundles in the apimodel it should be possible to have the e2e test retrieve the kubeconfig from there -- shall I open a follow-up issue to track that?

How is this going to look like when we allow users bring their own certificates? Maybe we can populate the certificate profile based on the config bundle?

kargakis:
> Maybe we can populate the certificate profile based on the config bundle?

In case we generate.

pweil-:
> do you think this functionality can/should be integrated into e2e tests?

Yes to the should question.  If not here then I want it in our OpenShift testing.

pkg/api/vlabs/validate.go
```diff
@@ -280,7 +280,10 @@ func validateImageNameAndGroup(name, resourceGroup string) error {
 }
 
 // Validate implements APIObject
-func (m *MasterProfile) Validate() error {
+func (m *MasterProfile) Validate(o *OrchestratorProfile) error {
+	if o.OrchestratorType == OpenShift && m.Count != 1 {
```
pweil-: unit test (reading commit by commit, if it comes later than ignore :smile:)

cmd/scale.go
```diff
@@ -237,8 +239,31 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
...
+				if err != nil {
+					log.Errorf("failed to read kube config: %v", err)
+					return err
+				}
+				err = sc.drainNodes(string(kubeConfig), vmsToDelete)
```
pweil-: the drain node calls can be removed from the case statement, the only difference to account for is getting the kube config to pass in.

parts/openshift/openshiftnodescript.sh
```diff
@@ -23,3 +23,8 @@ update-ca-trust
 # note: ${SERVICE_TYPE}-node crash loops until master is up
 systemctl enable ${SERVICE_TYPE}-node.service
 systemctl start ${SERVICE_TYPE}-node.service &
+
+while [[ $(KUBECONFIG=/etc/origin/node/node.kubeconfig oc get node $(hostname) -o template \
```
pweil-: It's ok for a node to be unready as the system comes up, why wait?

pkg/acsengine/defaults.go
```diff
@@ -740,6 +740,11 @@ func setStorageDefaults(a *api.Properties) {
 }
 
 func openShiftSetDefaultCerts(a *api.Properties) (bool, error) {
+	if len(a.OrchestratorProfile.OpenShiftConfig.ConfigBundles["master"]) > 0 &&
```
pweil-: wish there was a clearer indicator that this is a non-generate command

CecileRobertMichon:
> I can't find any scale tests in circleci e2e tests. Unclear whether there are any that run out of bound in Jenkins @jackfrancis @CecileRobertMichon

Yes, our scale and upgrade tests run on Jenkins. Let us know which apimodel definition you would like to test and we can set up a new job.

cmd/scale.go
jim-minter: Yes, but in the k8s case the actual authentication certificate within the kubeconfig is still the originally generated one.  While not perfect I think our approach is an improvement.

cmd/scale.go
jim-minter: This is a code move for k8s and I'm reluctant to refactor.  I'll ensure our newly added code is cleaner in this regard.  https://github.com/Azure/acs-engine/issues/2913 for the other code.

cmd/scale.go
jim-minter: the whole file needs a cleanup - I'm a bit torn between A) matching existing code, B) adding cleaner code, C) trying to improve the whole thing.  I'll move my stance from A) to B), how about that?

cmd/scale.go
jim-minter: however the drain node calls still must only run when the orchestrator is k8s or ocp - that's why they're in the switch still

parts/openshift/openshiftnodescript.sh
jim-minter: it has the advantage that the scale up call doesn't return success until all the requested nodes have successfully joined the cluster, and that if any new node fails to register, the error is reported as a deployment failure

pkg/acsengine/defaults.go
jim-minter: I think the key here is "Default"

jim-minter:
reviews addressed, including adding unit tests (thanks! :grimacing:)
need to decide what to do about e2e testing.

jim-minter:
> Yes, our scale and upgrade tests run on Jenkins. Let us know which apimodel definition you would like to test and we can set up a new job.

Wondering if we can/should do this in the e2e so we can make additional checks (e.g. that the right node labels are applied after scale up)?

CecileRobertMichon:
> Wondering if we can/should do this in the e2e so we can make additional checks (e.g. that the right node labels are applied after scale up)?

@jim-minter our Jenkins jobs run e2e tests before and after scaling the cluster so we do make those checks. Rationale is to avoid slowing down circleci with scale and upgrade since they take a long time to test (we run multiple scale operations on the same cluster and check it's healthy each time with e2e)

jim-minter:
Have added to the e2e test to ensure that created nodes are labeled correctly, plus 1 more commit with non-controversial fixes to the e2e tests

cmd/scale.go
jackfrancis: Thanks for filing the backlog item ❤️

pkg/api/vlabs/validate_test.go
```diff
@@ -829,6 +829,41 @@ func TestValidateImageNameAndGroup(t *testing.T) {
...
+		expectedErr      bool
+	}{
+		{
+			orchestratorType: OpenShift,
+			count:            1,
+		},
+		{
+			orchestratorType: OpenShift,
+			count:            3,
+			expectedErr:      true,
```
pweil-: tests need to test against a known, expected error in failure cases to really know you're hitting the code you think you are.  I'd also make the test provide the `MasterProfile` so this really can be used as a  `TestMasterProfileValidate` in the future and not just a count test.

test/e2e/openshift/openshift_test.go
```diff
@@ -56,6 +56,32 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		Expect(ready).To(Equal(true))
 	})
 
+	It("should label nodes correctly", func() {
```
pweil-: :heart:

pkg/api/vlabs/validate_test.go
kargakis: agreed

parts/k8s/kubernetesparams.t
```diff
@@ -494,7 +494,7 @@
       "type": "string"
     },
     "kubernetesClusterAutoscalerEnabled": {
-      {{PopulateClassicModeDefaultValue "kubernetesClusterAutoscalerEnabled"}}
+      "defaultValue": false,
```
kargakis: Fix appropriately or should we fix this in a separate PR?

kargakis: Seems to be fixed in https://github.com/Azure/acs-engine/pull/2915, ignore

jim-minter:
@pweil- updated - ptal

current thinking after discussion with @jackfrancis is that for now we'll instigate a periodic test on our side which proves that the scale up use case works to our satisfaction - trello card opened for this

jim-minter:
gofmt error fixed, thanks

@jackfrancis or @CecileRobertMichon please merge

parts/openshift/openshiftnodescript.sh
kargakis: Is there any timeout used during scale up?

jim-minter: same as on deploy - arm api will timeout, not sure exactly after how long - something like 45 minutes?

PR #2908, opened 2018-05-10
Version OpenShift code paths

kargakis:
Moves our scripts and certgen codepath into versioned directories.


kargakis:
ready for reviews

kargakis:
Still need to version certgen otherwise this is not complete.

kargakis:
Seeing the following error during scale up, trying to figure out why
```
INFO[0004] Finished ARM Deployment (mkargaki-cluster-1716432361). Error: resources.DeploymentsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidRequestContent" Message="The request content was invalid and could not be deserialized: 'Cannot deserialize the current JSON array (e.g. [1,2,3]) into type 'Microsoft.WindowsAzure.ResourceStack.Common.Collections.InsensitiveDictionary`1[Microsoft.WindowsAzure.ResourceStack.Frontdoor.Templates.Schema.TemplateOutputParameter]' because the type requires a JSON object (e.g. {\"name\":\"value\"}) to deserialize correctly.\r\nTo fix this error either change the JSON to a JSON object (e.g. {\"name\":\"value\"}) or change the deserialized type to an array or a type that implements a collection interface (e.g. ICollection, IList) like List<T> that can be deserialized from a JSON array. JsonArrayAttribute can also be added to the type to force it to deserialize from a JSON array.\r\nPath 'properties.template.outputs', line 1, position 157.'." 
```

pkg/api/converterfromapi.go
```diff
@@ -633,8 +634,12 @@ func convertOrchestratorProfileToVLabs(api *OrchestratorProfile, o *vlabs.Orches
 
 	if api.OrchestratorVersion != "" {
 		o.OrchestratorVersion = api.OrchestratorVersion
-		sv, _ := semver.NewVersion(o.OrchestratorVersion)
-		o.OrchestratorRelease = fmt.Sprintf("%d.%d", sv.Major(), sv.Minor())
+		// Enable using "unstable" as a valid version in the openshift orchestrator.
+		// Required for progressing on an unreleased version.
+		if api.OpenShiftConfig == nil || api.OrchestratorVersion != common.OpenShiftVersionUnstable {
```
pweil-: why not check OrchestratorType?

pkg/acsengine/engine.go
```diff
@@ -123,7 +125,31 @@ var dcos2TemplateFiles = []string{dcos2BaseFile, dcosAgentResourcesVMAS, dcosAge
...
+	openshift39NodeScript,
+	openshift39MasterScript,
+)
+
+func getOpenshiftMasterShAsset(version string) string {
+	switch version {
+	case common.OpenShiftVersion3Dot9Dot0:
+		return openshift39MasterScript
+	default:
+		return openshiftMasterScript
```
pweil-: shouldn't the default correspond to `OpenShiftDefaultVersion`?

pweil-: Actually, thinking on this a bit more, if the default version gets defaulted into the api model someone would have to set `unstable` in order to use it which makes me think that this should just check specific versions.  The default case should never be hit unless there is programmatic error?  Maybe I'm thinking too hard on it :smile:

kargakis: I don't feel strong about doing any sort of error handling here since version errors are already be managed during validation.

kargakis: @jim-minter thoughts?

pkg/api/types.go
```diff
@@ -345,7 +345,7 @@ type OpenShiftConfig struct {
 	// EnableAADAuthentication is temporary, do not rely on it.
 	EnableAADAuthentication bool `json:"enableAADAuthentication,omitempty"`
 
-	ConfigBundles map[string][]byte `json:"configBundles,omitempty"`
+	ConfigBundles map[string]map[string][]byte `json:"configBundles,omitempty"`
```
jim-minter: why does this change?

pkg/acsengine/engine.go
jim-minter: I'm not worried about it.

scripts/validate-generated.sh
```diff
@@ -20,7 +20,10 @@ cp -a "${DIR}/.." "${T}/"
 (cd "${T}/" && go generate ./...)
 
 
-GENERATED_FILES=("pkg/openshift/certgen/templates/bindata.go")
+GENERATED_FILES=(
```
jim-minter: need update to scripts/validate-go.sh as well?

pkg/acsengine/engine.go
pweil-: > I'm not worried about it.

Famous last words :smile: 

I'll just leave this existing example here (also noting that only specific versions are returned here, I'd argue they need another default handler and could/should make this something that can be passed back :smile: ).

```go
switch orchestratorType {
	case api.DCOS:
		switch orchestratorVersion {
		case api.DCOSVersion1Dot8Dot8:
			return dcosCustomData188
		case api.DCOSVersion1Dot9Dot0:
			return dcosCustomData190
		case api.DCOSVersion1Dot9Dot8:
			return dcosCustomData198
		case api.DCOSVersion1Dot10Dot0:
			return dcosCustomData110
		case api.DCOSVersion1Dot11Dot0:
			return dcos2CustomData111
		}
	default:
		// it is a bug to get here
		panic(fmt.Sprintf("BUG: invalid orchestrator %s", orchestratorType))
	}
	return ""
```

scripts/validate-generated.sh
kargakis: updated

pkg/acsengine/engine.go
kargakis: OK, I don't mind switching to panic.

kargakis: updated

kargakis:
> Looks like I am seeing the same error on master @jim-minter ^

Opened https://github.com/Azure/acs-engine/issues/2938

jim-minter:
@pweil- when upstream merges they squash as part of the process

pweil-:
> @pweil- when upstream merges they squash as part of the process

this won't be the last time you remind me of this, I guarantee it.

PR #2935, opened 2018-05-14
Enable AAD authentication for agent pool only clusters

amanohar:
This PR enables AAD support in API version 2018-03-31

Exert of changes to API are:
```
 "properties": {
    "enableRBAC": true,
    "aadProfile": {
        "clientAppID": "<NATIVE_APP_ID>",
        "serverAppID": "<WEB_APP_ID>",
        "serverAppSecret": "<WEB_APP_Secret>",
        "tenantID": "<AAD_TENANT_ID>"
    }
}
```
API version 2018-03-31 will enable AAD integration using Webhook authentication method (and is not enabled in ACS Engine generated templates yet). ACS Engine generated templates support AAD integration via OIDC authentication and this remains as is.


amanohar:
@mboersma  @jackfrancis  @slack PTAL. Thanks!

kargakis:
@jim-minter fyi - we should probably be using AADProfile, too.

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -510,3 +516,15 @@ func convertV20180331AgentPoolOnlyAddonProfiles(obj map[string]v20180331.AddonPr
 	}
 	return api
 }
+
+func convertV20180331AgentPoolOnlyAADProfile(obj *v20180331.AADProfile) *AADProfile {
+	aadProfile := &AADProfile{
```
mboersma: We can just return this struct directly:
```go
return &AADProfile {
...
```

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -353,10 +353,16 @@ func convertV20180331AgentPoolOnlyProperties(obj *v20180331.Properties) *Propert
 	if obj.ServicePrincipalProfile != nil {
 		properties.ServicePrincipalProfile = convertV20180331AgentPoolOnlyServicePrincipalProfile(obj.ServicePrincipalProfile)
 	}
+
 	if obj.AddonProfiles != nil {
 		properties.AddonProfiles = convertV20180331AgentPoolOnlyAddonProfiles(obj.AddonProfiles)
 	}
 
+	// TODO: validate that API is v20180331
```
mboersma: I left his comment here so we wouldn't forget, but IDK where such validation should actually go.

amanohar: `convertV20180331AgentPoolOnlyProperties` is invoked for API version 2018-03-31 only so I removed the comment. I don't think we need separate validation for this.

pkg/api/convertertoagentpoolonlyapi.go
amanohar: Updated. Pushed new changes PTAL.

pkg/api/types.go
```diff
@@ -503,12 +503,21 @@ type HostedMasterProfile struct {
 	Subnet string `json:"subnet"`
 }
 
+type AuthenticatorType string
+
+const (
+	OIDC    AuthenticatorType = "OIDC"
```
jackfrancis: To fix lint, you could just convert these to `oidc` and `webHook` because they don't seem to be consumed as exported consts currently

amanohar: @jackfrancis fixed. Pushed new commit.

pkg/api/types.go
```diff
@@ -503,12 +503,23 @@ type HostedMasterProfile struct {
 	Subnet string `json:"subnet"`
 }
 
+type AuthenticatorType string
```
jackfrancis: same lint comment here :)

mboersma: This line needs a comment to satisfy `make test-style`:

```
$ make test-style

==> Running static validations <==

==> Running linters <==
pkg/api/types.go:506:6:warning: exported type AuthenticatorType should have comment or be unexported (golint)
make: *** [test-style] Error 1
```

PR #2954, opened 2018-05-15
feat(*): Bumps client-go to v7.0.0

thomastaylor312:
This updates client-go to v7.0.0 and adds the needed dependencies on `k8s.io/api`. Also fixes a small issue with conflicting `uuid` library versions

**Release note**:
```release-note
Updates client-go to v7.0.0 and adds the needed dependencies on `k8s.io/api`.
```


jackfrancis:
@jim-minter FYI

This is a rather complex, highly valuable change that updates k8s.io/client-go to v7.0.0 (in order to permit vendoring in of a whole host of native k8s libs that depend on a more recent version of k8s.io/client-go).

I know openshift E2E is flaky at this point, but I'd like to get a passing run as a sanity check. If you could scan the changes in glide.yaml for anything that looks suspicious from your perspective, that would be great!

kargakis:
@jackfrancis this change does not affect openshift e2e tests since we use `oc` for requests. Also, none of our flakes today seem related to the client (we have an etcd flake, and then the rest are related to azure/image publishing).

PR #2961, opened 2018-05-15
Enhance e2e debuggability

kargakis:
Every time a test fails or an interrupt is received (^C), tests dump info about pods, nodes, and a run of diagnostics.

@jim-minter @pweil-

### DIFF
--- a/test/e2e/openshift/openshift_test.go
+++ b/test/e2e/openshift/openshift_test.go
@@ -1,6 +1,7 @@
 package openshift
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


kargakis:
Alternative: Instead of running diagnostics which seems like it has some extra noise we may not care about we can potentially start small and dump logs from selected components (or every pod that failed or crashloops).

pweil-:
> Alternative: Instead of running diagnostics which seems like it has some extra noise we may not care about we can potentially start small and dump logs from selected components (or every pod that failed or crashloops).

I'm not opposed to having diagnostics output but I suspect we'll want the logs anyway for useful debugging.  I'd at least dump the infra components.

kargakis:
This should be ready do go but the only thing that worries me today is that it's hard to access CI artifacts and I don't think it will make our lifes easier to just use stdout so we need to figure this out.

cc @jackfrancis @CecileRobertMichon @jim-minter

jim-minter:
lgtm - @kargakis are you ready for this to be merged?

kargakis:
Yes, this is ready to go in

kargakis:
Opened https://github.com/Azure/acs-engine/issues/2982 for the openshift test failure.

jim-minter:
@CecileRobertMichon please merge

jackfrancis:
Merging because I'm seeing this openshift E2E flakiness elsewhere, does not appear causal from this PR.

kargakis:
@jackfrancis thanks

PR #2980, opened 2018-05-16
Support for Low-priority VMs on Virtual Machine Scale Sets (VMSS) for Kubernetes

asw101:
- Adds support for Low-priority VMs on Virtual Machine Scale Sets (VMSS) for Kubernetes v1.10+

**If applicable**:
- [x] documentation
- [ ] tested backward compatibility (n/a)
- [x] unit tests

**Release note**:
```release-note
Add Support for Low-priority VMs on Virtual Machine Scale Sets (VMSS) for Kubernetes v1.10+ agent availability profile
```

cc: @CecileRobertMichon @jackfrancis @sozercan (extends @sozercan's https://github.com/Azure/acs-engine/pull/2620 ).

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmss.t
+++ b/parts/k8s/kubernetesagentresourcesvmss.t
@@ -43,6 +43,10 @@
         "mode": "Manual"
       },
       "virtualMachineProfile": {
+        {{if .IsLowPriorityScaleSet}}
+        "priority": "[variables('{{.Name}}ScaleSetPriority')]",
+        "evictionPolicy": "[variables('{{.Name}}ScaleSetEvictionPolicy')]",
+        {{end}}
         "networkProfile": {
           "networkInterfaceConfigurations": [
             {
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -764,6 +764,9 @@ func setStorageDefaults(a *api.Properties) {
 		if len(profile.AvailabilityProfile) == 0 {
 			profile.AvailabilityProfile = api.VirtualMachineScaleSets
 		}
+		if len(profile.ScaleSetEvictionPolicy) == 0 && profile.ScaleSetPriority == api.ScaleSetPriorityLow {
+			profile.ScaleSetEvictionPolicy = api.ScaleSetEvictionPolicyDelete
+		}
 	}
 }
 
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -402,21 +402,23 @@ type Extension struct {
 
 // AgentPoolProfile represents an agent pool definition
 type AgentPoolProfile struct {
-	Name                string               `json:"name"`
-	Count               int                  `json:"count"`
-	VMSize              string               `json:"vmSize"`
-	OSDiskSizeGB        int                  `json:"osDiskSizeGB,omitempty"`
-	DNSPrefix           string               `json:"dnsPrefix,omitempty"`
-	OSType              OSType               `json:"osType,omitempty"`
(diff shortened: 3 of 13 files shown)


pkg/api/types.go
```diff
@@ -708,6 +710,11 @@ func (a *AgentPoolProfile) IsVirtualMachineScaleSets() bool {
 	return a.AvailabilityProfile == VirtualMachineScaleSets
 }
 
+// IsLowPriorityScaleSet returns true if the VMSS is Low Priority
+func (a *AgentPoolProfile) IsLowPriorityScaleSet() bool {
+	return a.ScaleSetPriority == ScaleSetPriorityLow
```
CecileRobertMichon: Should we also check that `availabilityProfile == VirtualMachineScaleSets ` here? Another option would be to add validation to ensure that the priority is empty when  `availabilityProfile != VirtualMachineScaleSets `. Depends on whether we want the user to be able to specify a priority even if they aren't using VMSS (and ignore that priority) or if we prefer strict checking that priority is only specified when VMSS is used (which means we can always use it if it's specified). What do you think?

pkg/acsengine/defaults.go
```diff
@@ -764,6 +764,9 @@ func setStorageDefaults(a *api.Properties) {
 		if len(profile.AvailabilityProfile) == 0 {
 			profile.AvailabilityProfile = api.VirtualMachineScaleSets
 		}
+		if len(profile.ScaleSetEvictionPolicy) == 0 && profile.ScaleSetPriority == api.ScaleSetPriorityLow {
```
CecileRobertMichon: Extra credit: add a unit test in default_test.go to make sure EvictionPolicy / Priority are set coherently 🙂

pkg/api/types.go
asw101: Thank you for catching this! I think the former check works well here and I've just added it.

pkg/acsengine/defaults.go
asw101: Great point. I've just added TestAgentPoolProfile which check default values remain empty, and that the ScaleSetEvictionPolicy is set correctly when ScaleSetPriority is Low.

asw101:
@CecileRobertMichon thank you for your feedback. I've just committed those changes, and also confirming I had tested manual deployment with the example apimodel is working fine.

parts/k8s/kubernetesagentresourcesvmss.t
```diff
@@ -43,6 +43,10 @@
         "mode": "Manual"
       },
       "virtualMachineProfile": {
+        {{if .IsLowPriorityScaleSet}}
```
jackfrancis: Becuse we're simply assigning the `"priority"` value that is present in the `{{.Name}}ScaleSetPriority` variable, would it make more sense to remove this `{{if .IsLowPriorityScaleSet}}` clause? In other words, let's just make sure that a sane default value is always present for template variable expansion (either "Regular" or "Low").

The advantage of doing this is to remove an unnecessary `if` clause from the template (which already has a lot of conditional blocks and is rather hard to read!)

asw101: Thanks Jack, I was in two minds on this and leant towards making the output template less verbose (whilst also reducing the number of parameters, and variables), despite setting the default values, as people who don't use or care about Low-priority VMSS aren't likely to encounter this in their templates (particularly because "Regular" is implicit). And also because it is currently preview. But happy to simplify if you prefer. Any thoughts?

jackfrancis:
@asw101 This is fine for now, especially as it unblocks folks. We'll do another pass over this sometime in the future. Thanks so much!

PR #2994, opened 2018-05-17
removing the vnet from scale templates

JackQuincy:
This allows a cluster to scale even if the user has added a subnet to the vnet

This removes the vnet from update deployments for agents. This allows scaling operations to run on a cluster the user has added a subnet to the vnet.
https://github.com/Azure/AKS/issues/349 
**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
NONE
```


JackQuincy:
back compat shouldn't be needed. This is called by the scale command so no change in user behavior so I wasn't planning on writing docs, but I could be persuaded.

CecileRobertMichon:
How does this PR relate to https://github.com/Azure/acs-engine/pull/2095?

JackQuincy:
Should I be concerned by the dcos and openshift failures I might of affected open shift expect the error doesn't make sense to me if that was the case. And I didn't affect dcos. I actually improved code coverage marginally.

CecileRobertMichon:
@JackQuincy probably not, both openshift and dcos e2e have been unstable this week. I restarted the tests to make sure.

JackQuincy:
This is ready for review when someone has the time

CecileRobertMichon:
@JackQuincy is this also going to change upgrade behavior (ie. upgrade without vnet)?

JackQuincy:
@CecileRobertMichon yes for the agent parts of it, no for the master parts. Honestly we should do a small refactor here to make removing the unneeded nrp resources it's own method that gets called by the transform for scaling, that way it could be referenced by master upgrade and avoid the issues I had here previously, like load balancer services have a short outage around the template deployment.

CecileRobertMichon:
I'm running an upgrade test at https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/96/console and it's been stuck on the first master deployment for a while... let's keep watching it to make sure it succeeds

PR #3014, opened 2018-05-18
Update cluster-autoscaler addon to v1.2.2

sozercan:
Updates `cluster-autoscaler` addon to v1.2.2 for k8s v1.10


feiskyer:
👍

Issue #2926, opened 2018-05-11
Extensions not working


kichristensen:
**Is this a request for help?**:
Yes
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
ISSUE
---

**What version of acs-engine?**:
v0.16.2
---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes

**What happened**:
Running the extension examples provided doesn't work

**What you expected to happen**:
It worked

**How to reproduce it** (as minimally and precisely as possible):
Add the prometheus-grafana-k8s extension to deployment

**Anything else we need to know**:
After some fiddling I got the extension to be accepted in the ARM template, but then I instead got the error below during deployment:

```
{
  "error": {
    "code": "BadRequest",
    "message": "Multiple VMExtensions per handler not supported for OS type 'Linux'. VMExtension 'cse0' with handler 'Microsoft.Azure.Extensions.CustomScript' already added or specified in input."
  }
}
```

ritazh:
@kichristensen can you please share how you are adding the extension to your cluster definition? thanks!

kichristensen:
This is my config with extensions:

```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes", 
      "orchestratorRelease": "1.9",
      "kubernetesConfig": {
        "apiServerConfig": {
          "--admission-control":  "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DenyEscalatingExec,Initializers,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota",
          "--runtime-config": "admissionregistration.k8s.io/v1alpha1"
        },
        "useInstanceMetadata": true,
        "networkPolicy": "azure",
        "enableRbac": true
      }
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "$ACS_DNS_PREFIX",
      "vmSize": "Standard_D2_v2"
    },
    "agentPoolProfiles": [
      {
        "name": "linuxpool",
        "count": 2,
        "vmSize": "Standard_DS11_v2",
        "availabilityProfile": "AvailabilitySet",
        "osType": "Linux",
        "extensions": [
          {
            "name": "prometheus-grafana-k8s"
          }
        ]},
      {
        "name": "windowspool",
        "count": 1,
        "vmSize": "Standard_DS11_v2",
[...]

PR #3018, opened 2018-05-19
Fix prometheus extension due to renaming cse

ritazh:
Update prometheus extension with new cse naming convention
Tested this extension on both master and agent

Fixes https://github.com/Azure/acs-engine/issues/2926

FYI @trstringer

### DIFF
--- a/extensions/prometheus-grafana-k8s/v1/prometheus-grafana-k8s.sh
+++ b/extensions/prometheus-grafana-k8s/v1/prometheus-grafana-k8s.sh
@@ -341,6 +341,9 @@ DS_NAME=prometheus1
 
 PROM_URL=http://monitoring-prometheus-server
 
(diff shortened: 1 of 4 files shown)


trstringer:
Looks good to me! Thanks, @ritazh! :+1: :smile:

neurot1cal:
@CecileRobertMichon @ritazh Do we think this should resolve the issues with extensions in VMSS as well? Or was there anything specific to VMSS still that will prevent extensions from working?

CecileRobertMichon:
@neurot1cal I think VMSS might be a separate problem. I'm working on a PR to fix it.

CecileRobertMichon:
Yes @ritazh is correct there are two separate issues, the timeout one won't be addressed by my PR, only the naming incompatibility. PR opened at #3031.

Issue #2918, opened 2018-05-10
openshift: tracker for etcd timeout issues


jim-minter:

jim-minter:
@kargakis @pweil-

jim-minter:
https://circleci.com/gh/Azure/acs-engine/25995

```
TASK [openshift_hosted : Create OpenShift router] ******************************
[...]

PR #3020, opened 2018-05-19
add default distro for agents.

mjudeikis:
If you generate template for Openshift now without distro specified (as per docs section in https://github.com/Azure/acs-engine/blob/master/docs/openshift/deploy.md), you will get agent distro as Ubuntu and masters as RHEL.

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -605,9 +605,6 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 			}
 		}
 	case api.OpenShift:
-		if a.MasterProfile.Distro == "" {
-			a.MasterProfile.Distro = api.RHEL
-		}
 		kc := a.OrchestratorProfile.OpenShiftConfig.KubernetesConfig
 		if kc == nil {
 			kc = &api.KubernetesConfig{}
@@ -645,10 +642,18 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
 	if a.MasterProfile == nil {
 		return
 	}
-
-	// Set default Distro to Ubuntu
-	if a.MasterProfile.Distro == "" {
-		a.MasterProfile.Distro = api.Ubuntu
+	o := a.OrchestratorProfile
+	// set default Distro based on orchestrator
+	switch o.OrchestratorType {
+	case api.OpenShift:
+		if a.MasterProfile.Distro == "" {
+			a.MasterProfile.Distro = api.OpenShiftLatestRHEL
+		}
+	default:
+		// Set default Distro to Ubuntu
+		if a.MasterProfile.Distro == "" {
+			a.MasterProfile.Distro = api.Ubuntu
+		}
 	}
 
 	if !a.MasterProfile.IsCustomVNET() {
@@ -734,14 +739,24 @@ func setAgentNetworkDefaults(a *api.Properties) {
 		}
 	}
 
+	o := a.OrchestratorProfile
 	for _, profile := range a.AgentPoolProfiles {
 		// set default OSType to Linux
 		if profile.OSType == "" {
 			profile.OSType = api.Linux
 		}
-		// set default Distro to Ubuntu
-		if profile.Distro == "" {
-			profile.Distro = api.Ubuntu
+
+		// set default Distro based on orchestrator
+		switch o.OrchestratorType {
+		case api.OpenShift:
+			if profile.Distro == "" {
+				profile.Distro = api.OpenShiftLatestRHEL
+			}
+		default:
+			// Set default Distro to Ubuntu
+			if profile.Distro == "" {
+				profile.Distro = api.Ubuntu
+			}
 		}
 
 		// Set the default number of IP addresses allocated for agents.
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -642,14 +642,8 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
 	if a.MasterProfile == nil {
 		return
 	}
-	o := a.OrchestratorProfile
-	// set default Distro based on orchestrator
-	switch o.OrchestratorType {
-	case api.OpenShift:
-		if a.MasterProfile.Distro == "" {
-			a.MasterProfile.Distro = api.OpenShiftLatestRHEL
-		}
-	default:
+	// don't default Distro for OpenShift
+	if !a.OrchestratorProfile.IsOpenShift() {
 		// Set default Distro to Ubuntu
 		if a.MasterProfile.Distro == "" {
 			a.MasterProfile.Distro = api.Ubuntu
@@ -739,20 +733,14 @@ func setAgentNetworkDefaults(a *api.Properties) {
 		}
 	}
 
-	o := a.OrchestratorProfile
 	for _, profile := range a.AgentPoolProfiles {
 		// set default OSType to Linux
 		if profile.OSType == "" {
 			profile.OSType = api.Linux
 		}
 
-		// set default Distro based on orchestrator
-		switch o.OrchestratorType {
-		case api.OpenShift:
-			if profile.Distro == "" {
-				profile.Distro = api.OpenShiftLatestRHEL
-			}
-		default:
+		// don't default Distro for OpenShift
+		if !a.OrchestratorProfile.IsOpenShift() {
 			// Set default Distro to Ubuntu
 			if profile.Distro == "" {
 				profile.Distro = api.Ubuntu
--- a/pkg/acsengine/defaults_test.go
+++ b/pkg/acsengine/defaults_test.go
@@ -519,6 +519,68 @@ func TestAgentPoolProfile(t *testing.T) {
 	}
 }
 
+// TestSetComponentsNetworkDefaults covers tests for setMasterNetworkDefaults and setAgentNetworkDefaults
+// TODO: Currently this test covers only api.Distro setting. Extend test cases to cover network configuration too.
+func TestSetComponentsNetworkDefaults(t *testing.T) {
+
+	var tests = []struct {
+		name                string                  // test case name
+		orchestratorProfile api.OrchestratorProfile // orchestrator to be tested
+		expectedDistro      api.Distro              // expected result default disto to be used
+		result              bool                    // test result expectation for false positives (false - fail, true - success)
+	}{
+		{
+			"rhel_openshift",
+			api.OrchestratorProfile{
+				OrchestratorType: api.OpenShift,
+			},
+			api.OpenShiftLatestRHEL,
+			true,
+		},
+		{
+			"ubuntu_kubernetes",
+			api.OrchestratorProfile{
+				OrchestratorType: api.Kubernetes,
+			},
+			api.Ubuntu,
+			true,
+		},
+		{
+			"ubuntu_kubernetes_fail",
+			api.OrchestratorProfile{
+				OrchestratorType: api.OpenShift,
+			},
+			api.Ubuntu,
+			false,
+		},
+		{
+			"ubuntu_openshift_fail",
+			api.OrchestratorProfile{
+				OrchestratorType: api.Kubernetes,
+			},
+			api.OpenShiftLatestRHEL,
+			false,
+		},
+	}
+
+	for _, test := range tests {
+		mockAPI := getMockAPIProperties("1.0.0")
+		mockAPI.OrchestratorProfile = &test.orchestratorProfile
+		setMasterNetworkDefaults(&mockAPI, false)
+		setAgentNetworkDefaults(&mockAPI)
+
+		if test.result == !reflect.DeepEqual(mockAPI.MasterProfile.Distro, test.expectedDistro) {
+			t.Fatalf("setMasterNetworkDefaults() test case %v did not return right Distro configurations %v != %v", test.name, mockAPI.MasterProfile.Distro, test.expectedDistro)
+		}
+		for _, agent := range mockAPI.AgentPoolProfiles {
+			if test.result == !reflect.DeepEqual(agent.Distro, test.expectedDistro) {
+				t.Fatalf("setAgentNetworkDefaults() test case %v did not return right Distro configurations %v != %v", test.name, agent.Distro, test.expectedDistro)
+			}
+		}
+	}
+
+}
+
 func getMockAddon(name string) api.KubernetesAddon {
 	return api.KubernetesAddon{
 		Name:    name,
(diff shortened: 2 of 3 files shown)


kargakis:
lgtm

Even though distro is not used when we get defaulted (which happens when we use an image reference).

mjudeikis:
Got the same impression but just to keep consistent. As now if failed you get error for ubuntu and rhel. Confusing... took me a while to understand why ubuntu even in the picture :)

CecileRobertMichon:
@jim-minter PTAL

jim-minter:
lgtm, please merge

jim-minter:
@mjudeikis thanks and congrats on the first PR :)  for now please cc @jim-minter @kargakis @pweil- on acs-engine PRs so we get to them!

pkg/acsengine/defaults.go
```diff
@@ -587,6 +587,11 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 		if a.MasterProfile.Distro == "" {
 			a.MasterProfile.Distro = api.RHEL
 		}
+		for key, agent := range a.AgentPoolProfiles {
```
pweil-: strangely, it looks like the default to ubuntu happens in `setMasterNetworkDefaults` and `setAgentNetworkDefaults`.  That seems incorrect to me.  I like it here.  Just wanted to point out that if we commit here this logic is now separated for better or worse.

In any case....I'm playing unit test police for everything.  Where them tests at? (looks like it will be the first one for this case, you only have to cover what you added)  :smile: :police_car: :oncoming_police_car:

jim-minter:
@CecileRobertMichon apols, please hold off from merge at least until unit test added (my bad for not thinking of this)

pkg/acsengine/defaults.go
mjudeikis: @pweil- are you saying that `set[components]NetworkDefaults` should not be tempering with `Distro` in the first place? And we should just move all `Distro` logic to `setOrchestratorDefaults` or all orchestrators? (assuming I will write the testcase first and it will backward compatible :blush:

jim-minter: it shouldn't, but until there's a better interface around the orchestrators I'd personally be tempted to condense all the `.Distro = api.RHEL` code next to the ubuntu equivalents (@pweil- don't look at me like that) and write a unit test.

pweil-: > don't look at me like that

+1 for consistency,  -0.5 for location.  Overall we're still ahead, go for it.

mjudeikis:
Single test run:

Old code:
```
=== RUN   TestSetComponentsNetworkDefaults
--- FAIL: TestSetComponentsNetworkDefaults (0.00s)
	/home/mjudeiki/go/src/github.com/Azure/acs-engine/pkg/acsengine/defaults_test.go:538: setMasterNetworkDefaults() test case rhel_openshift did not return right Distro configurations ubuntu != rhel
FAIL
exit status 1
FAIL	github.com/Azure/acs-engine/pkg/acsengine	0.005s
Error: Tests failed.
```


new code:
```
Running tool: /usr/bin/go test -v -timeout 30s github.com/Azure/acs-engine/pkg/acsengine -run ^TestSetComponentsNetworkDefaults$

=== RUN   TestSetComponentsNetworkDefaults
--- PASS: TestSetComponentsNetworkDefaults (0.00s)
PASS
ok  	github.com/Azure/acs-engine/pkg/acsengine	0.006s
Success: Tests passed.
```

@pweil- unit test just for you :)
@jim-minter Moved together with ubuntu logic :)

pkg/acsengine/defaults.go
```diff
@@ -624,10 +621,18 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
 	if a.MasterProfile == nil {
 		return
 	}
-
-	// Set default Distro to Ubuntu
```
kargakis: Unclear why defaulting of the distro is part of `setMasterNetworkDefaults` but this lgtm

mjudeikis: I think this need to be moved to separate code block. But one change at the time.

pkg/acsengine/defaults.go
```diff
@@ -69,10 +69,11 @@ var (
...
 	//DefaultRHELOSImageConfig is the RHEL Linux distribution.
+	//Being used if no image or distro specified. Default Distro -> RHEL - > this
 	DefaultRHELOSImageConfig = AzureOSImageConfig{
-		ImageOffer:     "RHEL",
-		ImageSku:       "7.3",
-		ImagePublisher: "RedHat",
+		ImageOffer:     "acsengine-preview",
+		ImageSku:       "rhel74",
+		ImagePublisher: "redhat",
 		ImageVersion:   "latest",
```
mjudeikis: This should change when code goes live and we will know which images will be available for generic public. Now RHEL does not work but acsengine-preview does.

pkg/acsengine/defaults.go
```diff
@@ -69,10 +69,11 @@ var (
 	}
 
 	//DefaultRHELOSImageConfig is the RHEL Linux distribution.
+	//Being used if no image or distro specified. Default Distro -> RHEL - > this
 	DefaultRHELOSImageConfig = AzureOSImageConfig{
```
mjudeikis: I would like to keep this as `default`, as its aligns with other code base In this case no input -> RHEL -> DefaultRHELOSImageConfig

kargakis: I don't think we should change this default. Other non-openshift deployments may use it.

pkg/acsengine/defaults.go
```diff
@@ -624,10 +622,18 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
...
-
-	// Set default Distro to Ubuntu
-	if a.MasterProfile.Distro == "" {
-		a.MasterProfile.Distro = api.Ubuntu
+	o := a.OrchestratorProfile
+	// set default Distro based on orchestrator
+	switch o.OrchestratorType {
+	case api.OpenShift:
+		if a.MasterProfile.Distro == "" {
+			a.MasterProfile.Distro = api.RHEL
```
kargakis: Instead of `api.RHEL` default to `api.OpenShift39RHEL`.

pkg/acsengine/defaults.go
```diff
@@ -713,14 +719,24 @@ func setAgentNetworkDefaults(a *api.Properties) {
...
 		}
-		// set default Distro to Ubuntu
-		if profile.Distro == "" {
-			profile.Distro = api.Ubuntu
+
+		// set default Distro based on orchestrator
+		switch o.OrchestratorType {
+		case api.OpenShift:
+			if profile.Distro == "" {
+				profile.Distro = api.RHEL
```
kargakis: Instead of `api.RHEL` default to `api.OpenShift39RHEL`.

pkg/acsengine/defaults_test.go
```diff
@@ -503,6 +503,71 @@ func TestAgentPoolProfile(t *testing.T) {
 	}
 }
 
+// TestSetComponentsNetworkDefaults covers tests for setMasterNetworkDefaults and setAgentNetworkDefaults
+// TODO: Currently this test covers only api.Distro setting. Extend test cases to cover network configuration too.
+func TestSetComponentsNetworkDefaults(t *testing.T) {
+
+	var tests = []struct {
+		name string
+		//orchestrator to be tested
```
jim-minter: pls put each comment at the end of the preceding line

pkg/acsengine/defaults_test.go
```diff
@@ -503,6 +503,71 @@ func TestAgentPoolProfile(t *testing.T) {
...
+		api.OrchestratorProfile
+		//expected result default disto
+		api.Distro
+		//test result expectation for false positives (false - fail, true - succ)
+		bool
```
jim-minter: give the field a name please

pkg/acsengine/defaults_test.go
jim-minter: `//orchestrator to be tested` really this is just a friendly test name, right?

pkg/acsengine/defaults_test.go
jim-minter: call api.Distro `expectedDistro` then test against it and you shouldn't need the bool member at all?

pkg/acsengine/defaults_test.go
```diff
@@ -520,20 +585,34 @@ func getMockAddon(name string) api.KubernetesAddon {
...
+		OrchestratorProfile: &api.OrchestratorProfile{
+			OrchestratorVersion: orchestratorVersion,
+			KubernetesConfig:    &api.KubernetesConfig{},
+			OpenShiftConfig:     &api.OpenShiftConfig{},
+			DcosConfig:          &api.DcosConfig{},
+		},
+		MasterProfile: &api.MasterProfile{
+			Count:        1,
+			DNSPrefix:    "apps.example.com",
+			VMSize:       "Standart_D4s_v3",
```
jim-minter: Standard

pkg/api/const.go
```diff
@@ -28,8 +28,9 @@ const (
 	RHEL   Distro = "rhel"
 	CoreOS Distro = "coreos"
 	// Supported distros by OpenShift
-	OpenShift39RHEL Distro = "openshift39_rhel"
-	OpenShiftCentOS Distro = "openshift39_centos"
+	OpenShift39RHEL     Distro = "openshift39_rhel"
+	OpenShiftCentOS     Distro = "openshift39_centos"
+	OpenShiftLatestRHEL Distro = OpenShift39RHEL
```
jim-minter: why?

pkg/acsengine/defaults_test.go
mjudeikis: yes

pkg/acsengine/defaults_test.go
mjudeikis: bool member is to detect false/positive. If code changes to the point where it gives results where tests should fail, we dont test failures. This allows to test failed code usecases

pkg/acsengine/defaults_test.go
mjudeikis: did copy paste :) will update

pkg/api/const.go
mjudeikis: this follows unix alternatives pattern. if no distro is defined in the `openshift.json` it fill default to `OpenShiftLatestRHEL` which we need to update each time we release stable version. So if somebody will be using "manual" deployment using acs-engine cli (not web ui) we could enforce defaults stable.

pkg/acsengine/defaults.go
```diff
@@ -624,10 +621,18 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
...
-
-	// Set default Distro to Ubuntu
-	if a.MasterProfile.Distro == "" {
-		a.MasterProfile.Distro = api.Ubuntu
+	o := a.OrchestratorProfile
+	// set default Distro based on orchestrator
+	switch o.OrchestratorType {
+	case api.OpenShift:
+		if a.MasterProfile.Distro == "" {
+			a.MasterProfile.Distro = api.OpenShiftLatestRHEL
```
mjudeikis: @jim-minter this is there latest default is being used. setting default if not specified. instead of copy pase we have "soft link" style pointer

jim-minter:
@pweil- @kargakis this one is turning into a bit of a saga.  Do you think we should default Distro at all?  Why should we default to RHEL over CentOS for example?  Wouldn't we be best to keep it explicit?

mjudeikis:
Where is 3 scenarios here:
1. leave it as it is and k8s defaulting kicks in (ubuntu errors)
2. fail early and give error that "distro (rhel or centos image) is required to be provided.
3. default to something consistently (lets say rhel).

in case 3, outcome is consistent error message. I suspect this to fail for anybody who uses acs-engine from cli, just failure is controller and consistent. In the end majority of users will not see this part as image will be provided by azure UI.

kargakis:
I am not married to defaulting either just need to ensure the generated api
model is valid.

kargakis:
testing

/lgtm

jim-minter:
@mjudeikis please simplify this PR so that we default distro to ubuntu only if the orchestrator is not openshift, otherwise we leave it blank as a default.
also please make sure the testing changes are minimal, and remove `result bool` from the testing struct, it's not necessary.

pkg/acsengine/defaults_test.go
```diff
@@ -536,20 +578,34 @@ func getMockAddon(name string) api.KubernetesAddon {
 }
 
 func getMockBaseContainerService(orchestratorVersion string) api.ContainerService {
+	mockAPIProperties := getMockAPIProperties(orchestratorVersion)
```
jim-minter: I don't believe any of the changes from here downwards are needed in this PR - is there any reason to keep them?

pkg/acsengine/defaults.go
```diff
@@ -645,10 +642,12 @@ func setMasterNetworkDefaults(a *api.Properties, isUpgrade bool) {
 	if a.MasterProfile == nil {
 		return
 	}
-
-	// Set default Distro to Ubuntu
-	if a.MasterProfile.Distro == "" {
-		a.MasterProfile.Distro = api.Ubuntu
+	// set default Distro for all what is not OpenShift
```
jim-minter: // don't default Distro for OpenShift
(also change comment further down please)

pkg/acsengine/defaults_test.go
mjudeikis: I needed just `mockAPI` and we had just `MockBaseContainerService`, So to not to duplicate code I did this split. They are needed for other tests so this is for backwards compatibility for existing tests

jim-minter: sorry, my bad - there's no reason to change the contents of the api.Properties at this stage, though?

mjudeikis:
Content is just for visibility. I tend to learn about how unknown code
expects to behaves by reading tests. When reading `openshift.json` example
and test you can map data structs fast just by skipping code fast. Just one
of the patterns I picked...

pkg/acsengine/defaults_test.go
jim-minter: in general I don't think it's expected for more than one of orchestratorprofile.{kubernetes,dcos,openshift}config to be non-nil.  Also the masterprofile pieces may or may not be relevant to other users.  If it's not necessary for the test, please leave it out.  Please fix this and the comments and we'll finally merge.

pkg/acsengine/defaults.go
CecileRobertMichon: Agreed it should be moved out in a separate PR

pweil-:
Looks like the failure may be the timeout issue.  I see:

```
/usr/bin/oc create -f /tmp/Secret_PBL0Z -n default\", \"results\": {}, \"returncode\": 1, \"stderr\": \"Error from server: error when creating \\\"/tmp/Secret_PBL0Z\\\": timeout\\n\", \"stdout\": \"\"}
```
https://github.com/Azure/acs-engine/issues/2918

A subsequent command seems to have completed fine

CecileRobertMichon:
Would you like me to override and merge? Or do you want to keep retrying until you get a successful pass?

pweil-:
Please merge. We are still figuring out the fix for the flake. Thanks!

PR #3030, opened 2018-05-21
DCOS: implement upgrade operation for clusters with bootstrap node

dmitsh:
implement upgrade operation for clusters with bootstrap node


jackfrancis:
Running upgrade sanity check, will merge after

PR #3031, opened 2018-05-21
Fix extensions not working

CecileRobertMichon:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2926 

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
```

PR #3059, opened 2018-05-23
Revert vmss extension name

### LINKED ISSUE #3064: VMSS deployment is broken in 0.17.1
PR that caused the regression: #3031 

PR to fix it: #3059 
 
Planning to release 0.17.2 with the fix shortly.

CecileRobertMichon:
**What this PR does / why we need it**: PR #3031 changed the CSE name for vmss agents. This broke vmss deployments because vmss agentpools do not have a variable "offset".

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3064 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,7 @@ jobs:
       - checkout
       - run: |
           echo 'export TIMEOUT=10m' >> $BASH_ENV
-          echo 'export CLUSTER_DEFINITION=examples/kubernetes.json' >> $BASH_ENV
+          echo 'export CLUSTER_DEFINITION=examples/kubernetes-vmss/kubernetes.json' >> $BASH_ENV
           echo 'export CREATE_VNET=false' >> $BASH_ENV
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       - checkout
       - run: |
           echo 'export TIMEOUT=10m' >> $BASH_ENV
-          echo 'export CLUSTER_DEFINITION=examples/kubernetes.json' >> $BASH_ENV
+          echo 'export CLUSTER_DEFINITION=examples/kubernetes-vmss/kubernetes.json' >> $BASH_ENV
```
CecileRobertMichon: `VMSS` is the default so updating the "k8s Linux default" to use vmss since all the other e2e tests use `AvailabilitySet`

jackfrancis: Instead of this, let's just update the `examples/kubernetes.json` api model to no longer specify an `"availabilityProfile"` (it should use vmss by default)

PR #3065, opened 2018-05-23
Rationalize AvailabilityProfile default

CecileRobertMichon:
**What this PR does / why we need it**: We were setting AvailabilityProfile to VMSS by default. However, VMSS is not supported for K8s < 1.10. This PR ensures that we set AvailabilityProfile to AvailabilitySet if using K8s < 1.10.

Backwards compatibility should not be affected because we were validating before to make sure that VMSS was not set with k8s < 1.10. This change just makes it easier for the user by setting a default that works with validation instead of requiring an override.

**If applicable**:
- [x] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -587,7 +587,7 @@ func (a *Properties) Validate(isUpdate bool) error {
 			}
 		}
 
-		// validation for VMSS for Kubernetes
+		// validation for VMSS with Kubernetes
 		if a.OrchestratorProfile.OrchestratorType == Kubernetes && agentPoolProfile.AvailabilityProfile == VirtualMachineScaleSets {
 			version := common.RationalizeReleaseAndVersion(
 				a.OrchestratorProfile.OrchestratorType,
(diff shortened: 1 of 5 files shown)


CecileRobertMichon:
@sozercan could you please take a look at this PR?

pkg/api/vlabs/validate.go
```diff
@@ -611,25 +611,9 @@ func (a *Properties) Validate(isUpdate bool) error {
 				return fmt.Errorf("VirtualMachineScaleSets are only available in Kubernetes version %s or greater; unable to validate for Kubernetes version %s",
 					minVersion, version)
 			}
-		}
-
-		// validation for instanceMetadata using VMSS on Kubernetes
-		if a.OrchestratorProfile.OrchestratorType == Kubernetes && (agentPoolProfile.AvailabilityProfile == VirtualMachineScaleSets || len(agentPoolProfile.AvailabilityProfile) == 0) {
```
CecileRobertMichon: This validation was duplicated

Issue #3072, opened 2018-05-24
go-bindata dependency update


pweil-:
In the `Makefile` we are pulling github.com/jteeuwen/go-bindata which is an unmaintained repository.  

@jackfrancis any reason not to use github.com/go-bindata/go-bindata?  I had that on my path so I wasn't using the makefile's binary and the only diff I noted was a `//DO NOT EDIT` comment.

/cc @jim-minter

pweil-:
> only diff I noted was a //DO NOT EDIT comment.

actually, my local might be old too.  I renamed the binary and didn't see any diffs.  Will submit a PR for you to review.

pweil-:
https://github.com/Azure/acs-engine/pull/3074

PR #3073, opened 2018-05-24
remove go get for bindata and rely on Makefile bootstrap target used …

### LINKED ISSUE #3032: openshift: should not to go get in generate - use make bootstrap
The openshift code contains `//go:generate ... go get ...` which creates a dependency on network access to build.  These should move to the bootstrap Makefile target.

pweil-:
…in generate

Fixes #3032 

The `bootstrap` target of the Makefile already contains a get for `go-bindata`, we do not need to add anything there.  However, it is fetching `github.com/jteeuwen/go-bindata` which appears to be unmaintained.  Issue #3072

```
ifndef HAS_GOBINDATA
	go get github.com/jteeuwen/go-bindata/...
endif
```

@jim-minter


pweil-:
> need to change pkg/openshift/certgen/release39/templates/templates.go as well

dang it, had that in my search window and forgot to do it.  Thanks for catching.  Updated

PR #3079, opened 2018-05-24
GetSupportedVersion returns the right win default

CecileRobertMichon:
**What this PR does / why we need it**: Hotfix for ACS RP, wrong default version was returned for windows (1.8.13, which isn't supported).

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -68,11 +68,20 @@ func GetDefaultKubernetesVersionWindows() string {
 }
 
 // GetSupportedKubernetesVersion verifies that a passed-in version string is supported, or returns a default version string if not
-func GetSupportedKubernetesVersion(version string) string {
-	if k8sVersion := version; AllKubernetesSupportedVersions[k8sVersion] {
-		return k8sVersion
+func GetSupportedKubernetesVersion(version string, hasWindows bool) string {
+	var k8sVersion string
+	if hasWindows {
+		k8sVersion = GetDefaultKubernetesVersionWindows()
+		if AllKubernetesWindowsSupportedVersions[version] {
+			k8sVersion = version
+		}
+	} else {
+		k8sVersion = GetDefaultKubernetesVersion()
+		if AllKubernetesSupportedVersions[version] {
+			k8sVersion = version
+		}
 	}
-	return GetDefaultKubernetesVersion()
+	return k8sVersion
 }
 
 // GetAllSupportedKubernetesVersions returns a slice of all supported Kubernetes versions
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -257,13 +257,13 @@ func GetSupportedVersions(orchType string, hasWindows bool) (versions []string, 
 }
 
 //GetValidPatchVersion gets the current valid patch version for the minor version of the passed in version
-func GetValidPatchVersion(orchType, orchVer string) string {
+func GetValidPatchVersion(orchType, orchVer string, hasWindows bool) string {
 	if orchVer == "" {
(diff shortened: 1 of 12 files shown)


pkg/api/common/versions.go
```diff
@@ -68,11 +68,20 @@ func GetDefaultKubernetesVersionWindows() string {
 }
 
 // GetSupportedKubernetesVersion verifies that a passed-in version string is supported, or returns a default version string if not
-func GetSupportedKubernetesVersion(version string) string {
-	if k8sVersion := version; AllKubernetesSupportedVersions[k8sVersion] {
-		return k8sVersion
+func GetSupportedKubernetesVersion(version string, hasWindows bool) string {
```
JackQuincy: we shouldn't silently override. If a value is passed in we should throw an error if invalid.

CecileRobertMichon: I would tend to agree with you but this PR is intended to fix the bug, not change the behavior of default k8s version definition... The silent override was already in place for getting the k8s version for API version `v20170701` so I don't think we want to change the behavior in this PR.

JackQuincy: does validation happen before this so we don't have to worry about it?

CecileRobertMichon: We do validate beforehand. So if I pass in something random as `orchestratorVersion` such as `0.5` I get `FATA[0000] error loading API model in generateCmd: error parsing the api model: OrchestratorProfile has unknown orchestrator version: 0.5`
The only case where this function overrides the version is if it is empty or if the cluster is a Windows cluster and the passed in version is valid but not supported for Windows, in which case the user gets default. I looked into fixing validation to also take into account "is this a Windows cluster? If so, is the passed in version valid for Windows?" but it's not straight forward since it requires having the agent pool profiles. I can make that an additional change in this PR if we see value in it.
TLDR: Is it okay to override the version if the value passed in for the user is not supported with windows but is a valid version or should we fail instead?

pkg/api/vlabs/validate.go
```diff
@@ -649,9 +649,6 @@ func (a *Properties) Validate(isUpdate bool) error {
 		}
 
 		if a.OrchestratorProfile.OrchestratorType == Kubernetes {
-			if i == 0 {
-				continue
```
CecileRobertMichon: This was causing a bug where we would not check anything for the first agentpool

pkg/api/vlabs/validate.go
```diff
@@ -663,25 +660,13 @@ func (a *Properties) Validate(isUpdate bool) error {
 			case Swarm:
 			case SwarmMode:
 			case Kubernetes:
-				var version string
-				if a.HasWindows() {
```
CecileRobertMichon: No need to check if the cluster has a Windows pool since this is run on Windows agents

pkg/api/vlabs/validate.go
```diff
@@ -663,25 +660,13 @@ func (a *Properties) Validate(isUpdate bool) error {
...
-						a.OrchestratorProfile.OrchestratorVersion,
-						true)
-				} else {
-					version = common.RationalizeReleaseAndVersion(
-						a.OrchestratorProfile.OrchestratorType,
-						a.OrchestratorProfile.OrchestratorRelease,
-						a.OrchestratorProfile.OrchestratorVersion,
-						false)
-				}
-				if version == "" {
```
jackfrancis: Let's put this empty string check back, and remove the `common.AllKubernetesWindowsSupportedVersions[version]` check

PR #3082, opened 2018-05-25
run restorecon on /var/lib/docker

jim-minter:
this PR is a precursor for using the azure ephemeral disk for docker https://github.com/openshift/openshift-ansible/pull/8488


jim-minter:
@kargakis @kwoodson @pweil- ptal - wdyt?

I tried the alternative approach of using docker-storage-setup, but the issue is that /dev/sdb arrives out of the factory pre-partitioned and with an NTFS (!) volume, and that confuses docker-storage-setup, necessitating a fixup before docker-storage-setup runs.  With this approach, waagent does the heavy lifting, but unfortunately it doesn't set the SELinux attributes correctly on /var/lib/docker.  IMO it's easier to fix that up afterwards, at least for now.

jim-minter:
Part 2 is https://github.com/openshift/openshift-ansible/pull/8488 -- this PR would need to go in first.  It should be a no-op for older images.

jim-minter:
/hold

this doesn't appear to be working right, will debug

pweil-:
Is adding the context= option to the mount options usable here?  Might be necessary for a reboot/remount anyway.

jim-minter:
Thanks for the link, didn't know about that mount option.  Will investigate.

kargakis:
cool, works

/lgtm cancel

jim-minter:
> Is adding the context= option to the mount options usable here? Might be necessary for a reboot/remount anyway.

@pweil- so nearly.  Turns out the waagent config parser is b0rked in the version we have and truncates after '=' characters in configuration parameters.

PR #3095, opened 2018-05-27
Update go-autorest to v9.10.0

### LINKED ISSUE #3094: Authentication for US Government uses Commercial AAD endpoint

NO

---

ISSUE

---

Latest, 0.17

---

Any

**What happened**:

When authentication, e.g. when using `acs-engine deploy`, the commercial endpoint, `login.microsoftonline.com` and `https://microsoft.com/devicelogin` are used instead of `login.microsoftonline.us` and `https://microsoft.com/deviceloginus`. 

This is will likely cause acs-engine to fail for US Government soon: https://blogs.msdn.microsoft.com/azuregov/2018/04/03/azure-government-aad-authority-endpoint-update/

The problem is caused by the use of an older (0.8.4) version of https://github.com/Azure/go-autorest

When deploying to Azure Government, ACS Engine should use `login.microsoftonline.us` endpoint. 

Do any deployment to Azure Government, notice the devicelogin URL.

gsacavdm:
Update go-autorest to v9.7.0 to address issue #3094 , also good to keep any dependencies as up-to-date as possible.

**Which issue this PR fixes**: fixes #3094 

According the [go-autorest's changelog notes on v9.0.0](https://github.com/Azure/go-autorest/blob/master/CHANGELOG.md#v900):

> IMPORTANT: This release was initially labeled incorrectly as v8.4.0. From the time it was released, it should have been marked v9.0.0 because it contains breaking changes to the MSI packages. We apologize for any inconvenience this causes.

Therefore upgrading from v8.4.0/v9.0.0 to v9.7.0 isn't upgrading across major releases. Also, no breaking changes are indicated in the releases that occurred between these versions.

I manually tested acs-engine generate, deploy, scale and upgrade  to ensure they still run fine.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Upgrade go-autorest to v9.7.0
```


gsacavdm:
FYI @jackfrancis

CecileRobertMichon:
> I see that k8s-linux-default-e2e failed though it seems to have timed out. What's the recommended way to troubleshoot this? Or can I just retrigger the test to see if ti passes?


@gsacavdm this is circle ci being flaky, I just re-triggered the test

gsacavdm:
k8s-linux-default-e2e passed :)
Not sure what the deal is with the others.

Thoughts @CecileRobertMichon ?

CecileRobertMichon:
@tariq1890 did you test backwards compatibility? ie. deploy from master, upgrade with your branch

jackfrancis:
/hold

This PR doesn't actually update `vendor/`

jackfrancis:
We'll have to update azure-sdk-for-go as well, the version we're currently pinned to (`~v11.0.0-beta`) requires go-autorest `~8.4.0`. See:

https://github.com/Azure/azure-sdk-for-go/blob/2592daf71ab6b95dcfc7f7437ecc1afb9ddb7360/glide.yaml

gsacavdm:
@jackfrancis , after making the change locally, I rebuilt ACS engine and confirmed my change did update `/vendor` and that the AAD endpoint for Azure Government was the newer one (`microsoft.com\deviceloginus`).

gsacavdm:
@jackfrancis, as discussed, I updated the azure-sdk-for-go to v12.3.0 which has go-autorest v9.6.0.

I tested `acs-engine deploy` and confirmed that it uses the right Azure Government AAD endpoint. Deploy failed because the AKS.Linux extension hasn't been published to Azure Government, but outside of that, everything else succeeded and I have a working Kubernetes cluster in Azure Gov.

I'll file a separate issue for the AKS.Linux extension to make that optional.

Can you trigger your tests to see if anything breaks?

jackfrancis:
@gsacavdm Did you run into any issues w/ the changes described here (from v11 to v12):

https://github.com/Azure/azure-sdk-for-go/releases/tag/v12.0.0-beta

@amanohar @weinong @nicolerenee FYI, this update to go-autorest and azure-sdk-for-go is to protect Fairfax (US Gov) acs-engine use cases against the broken pattern described here:

https://github.com/Azure/acs-engine/issues/3094

jackfrancis:
Also @gsacavdm FYI I've been rebase/force-pushing to fix some test flakiness, so your local copy will need to be manually re-pulled!

gsacavdm:
@jackfrancis , I rebased as per your suggestion.
Also, I did look at the release notes, but without looking thoroughly at how acs-engine uses the azure-sdk-for-go, I really can't speculate. My guess is only the first three items have the potential to break things:

>* Long Running Operations now return a Future type. Go routines and return via channel features have been removed.
>* Added support for services that make use of polymorphic types.
>* Operations now make use of the Context Pattern.

It did compile and run without issues (outside of the missing extension issue which is unrelated) I would have expected those three to cause it not to compile at all of they were affecting anything.

jackfrancis:
@gsacavdm updating go-autorest to v9.10.0 and azure-sdk-for-go to ~v12.4.0-beta to match AKS

gsacavdm:
Sounds good. What do we need to move this forward. All my tests looked good (I'll retest with your changes but I don't expect issues).

PR #3100, opened 2018-05-29
Replacing the static pod definition for etcd ip.

kwoodson:
With recent changes the etcd health check url IP was incorrect.  This provides us a method to update it.

Also includes a fix for the latest 3.10 changes to subsequent roles.
@jim-minter @kargakis @pweil-


kwoodson:
Still seeing errors with the template_service_broker.  It requires this variable be set `l_os_registry_url`.  I attempted to include the vars from openshift_facts but this variable was not updated as denoted by the debug statement:
```
TASK [debug] *******************************************************************
ok: [localhost] => {
    "l_os_registry_url": "VARIABLE IS NOT DEFINED!"
}
```

parts/openshift/unstable/openshiftmasterscript.sh
```diff
@@ -9,6 +9,7 @@ else
 	SERVICE_TYPE=origin
 fi
 VERSION="$(rpm -q $SERVICE_TYPE --queryformat %{VERSION})"
+IP_ADDRESS="$(ip route get 1 | sed -n 's/.* src \([^ ]*\) .*/\1/p')"
```
pweil-: would a `dig +short`  on the hostname be easier here?  Maybe with all the dns funkiness we've seen it isn't a good idea :smile: 

If this is the best way are the fields variable requiring the sed+regex or would a `ip route get 1 | awk '{print $(NF-2);exit}'` work?  I'm a regex hater, they're hard to read, but ack that it may be the best thing so that's just a nit.

kwoodson: @pweil- I'm debated the different methods of finding the IP address.  The issue I want to eliminate is hosts with multiple IP addresses.  I want this to be a reliable way of returning it.  I asked around and the `ip route get 1` was recommended because it is very generic and worked with hosts that have multiple interfaces.  This method chooses the IP address that is routing to the internet which is most likely the one we want to use.

In this case on my master this was what was needed with your suggestion.  Does that change with multiple addresses?
```
 ip route get 1 | awk '{print $(NF); exit}'
10.0.0.11
```

I attempted this a few different ways that seemed to produce the answer we wanted but I'm not sure how these behave in a multi-ip environment:
```
hostname -i
ifconfig eth0 | awk '/inet/{ print $2 ; exit}'
ip a s eth0 | awk '/inet/{ split($2, a, "/"); print a[1]; exit}'
python -c 'import socket; print socket.gethostbyname(socket.gethostname())'
```

@pweil- @jim-minter @kargakis @mjudeikis, thoughts here?

pweil-: Pretty certain that hostname -i doesn’t guarantee order so that won’t work. Dig or what you have now I guess

kwoodson: The dig command didn't return any results.
```
[root@ocp-master-18884060-0 ~]# dig +short $(hostname)
[root@ocp-master-18884060-0 ~]# 
```

mjudeikis: Ideally Azure should give API to get this where we could curl and get the value. But I agree with `ip route get 1` and awk over regexp as is the one I used in most of my scripts. If we can have boxes with multiple interfaces `ToDo` write a test for this case.

kwoodson: @mjudeikis, the curl does work but then we need to make assumptions or have a tool that parses it.

```
curl -qs -H Metadata:true "http://169.254.169.254/metadata/instance/network/interface/0/ipv4/ipAddress/0/privateIpAddress?api-version=2017-08-01&format=text"
```

The assumption here is that we are taking interface[0] and ipAddress[0].  I think this could work.

Any thoughts?

mjudeikis: my vote for now is `ip get route 1` . But most of this script must go... :dagger: 

As a long term I would like to consider writing our own small `golang` binary which would do all the `curl`,`wget`,`ip` etc. Which would not relay on the underlying infrastructure/image binaries. it would do simple things like `cli get ip`, `cli get version` etc etc? This could enable us to have more control over how we construct metadata.  Overkill?

pweil-: derp...why was I saying dig...we were using `host $(hostname)`. Anyway, the default route or the metadata service is fine.  If it is any comfort, the assumption of interface[0] and ipaddress[0] is documented.  I'm fine with the `ip get route`.

kwoodson: On my local machine the awk command returns different output so the version using NF wouldn't work.

If we are ok with the sed version I'll remove `[WIP]`.  If you'd prefer the curl instead of sed then I'll swap it out.

mjudeikis: I might be overprotective, but if we both and compare? And fail if they dont match?

jim-minter:
cf https://github.com/openshift/openshift-ansible/pull/8560

kwoodson:
@jim-minter, I have templated the master script as requested.

pkg/openshift/certgen/unstable/templates/master/tmp/ansible/azure-ocp-deploy.yml
```diff
@@ -2,7 +2,11 @@
   hosts: localmaster
   gather_facts: false
 
-  tasks:
+  # we need variables from openshift_facts to be in scope
```
jim-minter: @kwoodson why does this file change?

kwoodson: @jim-minter, inside of openshift-ansible the template service broker has the 'l_os_registry_url` variable defined.
```
https://github.com/openshift/openshift-ansible/blob/master/roles/template_service_broker/defaults/main.yml#L11
```

This variable is being set in openshift_facts and the expectation is that we include these facts to run these other roles:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/defaults/main.yml#L15

I included the facts which brings those variables into scope.

Hope this helps clarify.

jim-minter: thanks!

jim-minter:
/assign CecileRobertMichon 

@CecileRobertMichon please merge

kargakis:
openshift test failure is the etcd timeout: https://github.com/Azure/acs-engine/issues/2918

kargakis:
flake was https://github.com/Azure/acs-engine/issues/3114

kargakis:
@CecileRobertMichon should these codecov contexts show up here?

CecileRobertMichon:
@kargakis it should. Was the PR rebased since https://github.com/Azure/acs-engine/commit/cb48589541903dd5df18f8ff820091446e73600c? I'll merge this one anyway

PR #3101, opened 2018-05-29
Revert "use vars for nsg, ip, and lb.  Add prefix and suffix."

kargakis:
/hold

For debugging


kargakis:
Figured out the issue, commentor name needs to get lowercased :)

/close

Issue #3098, opened 2018-05-29
Kubelet failed to start on master with custom hyperkube from kubernetes master


adelina-t:
**ISSUE**

---

**0.17.1 or later**:

---

**Kubernetes 1.11 with custom hyperkube built from kubernetes master**

Deploying cluster results in 

`"VM has reported a failure when processing extension 'cse-master-0'`

ssh into the master machine and looking over logs I can see that kubelet failed to start:

```
journalctl -u kubelet 

May 29 09:48:38 k8s-master-17277121-0 systemd[1]: Stopped Kubelet.
May 29 09:48:38 k8s-master-17277121-0 systemd[1]: Starting Kubelet...
May 29 09:48:39 k8s-master-17277121-0 sysctl[8407]: net.ipv4.tcp_retries2 = 8
May 29 09:48:39 k8s-master-17277121-0 ebtables[8410]: Bridge table: nat
May 29 09:48:39 k8s-master-17277121-0 ebtables[8410]: Bridge chain: PREROUTING, entries: 0, policy: ACCEPT
May 29 09:48:39 k8s-master-17277121-0 ebtables[8410]: Bridge chain: OUTPUT, entries: 0, policy: ACCEPT
May 29 09:48:39 k8s-master-17277121-0 ebtables[8410]: Bridge chain: POSTROUTING, entries: 0, policy: ACCEPT
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: Chain PREROUTING (policy ACCEPT)
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: target     prot opt source               destination
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: DOCKER     all  --  anywhere             anywhere             ADDRTYPE match dst-type LOCAL
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: Chain INPUT (policy ACCEPT)
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: target     prot opt source               destination
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: Chain OUTPUT (policy ACCEPT)
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: target     prot opt source               destination
May 29 09:48:39 k8s-master-17277121-0 iptables[8413]: DOCKER     all  --  anywhere            !127.0.0.0/8          ADDRTYPE match dst-type LOCAL
[...]

PR #2838, opened 2018-05-02
add support for k8s 1.11.0-alpha.2

jackfrancis:
**What this PR does / why we need it**: See:

https://github.com/kubernetes/kubernetes/releases/tag/v1.11.0-alpha.2

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
add support for k8s 1.11.0-alpha.2
```

PR #3117, opened 2018-05-31
Fix k8s 1.11.0-alpha.2+ cluster bootstrap issue

karataliu:
Fix systemd service quote issue.

New validation was introduced in kubelet that does not allow empty argument, but using empty variable ${} in systemd will interpret as empty argument. This is to address the issue.

Reference:
https://github.com/kubernetes/kubernetes/pull/61833
https://unix.stackexchange.com/questions/175461/how-to-use-quote-and-dollarsign-with-systemd

Fixes https://github.com/Azure/acs-engine/issues/3098
Related https://github.com/Azure/acs-engine/pull/2838

cc @jackfrancis @CecileRobertMichon


adelina-t:
Can confirm, I cherry-picked this, solves my issue.

jackfrancis:
Thank you @karataliu!

PR #3126, opened 2018-05-31
Add agent pool only api for openshift

kargakis:
@pweil- @jim-minter @khenidak @ritazh @sozercan

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -13,8 +14,6 @@ import (
 	"github.com/leonelquinteros/gotext"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-
-	"encoding/json"
 
 	"github.com/Azure/acs-engine/pkg/acsengine"
 	"github.com/Azure/acs-engine/pkg/acsengine/transform"
@@ -80,10 +79,10 @@ func newDeployCmd() *cobra.Command {
 				log.Fatalf(fmt.Sprintf("error merging API model in deployCmd: %s", err.Error()))
 			}
 			if err := dc.loadAPIModel(cmd, args); err != nil {
-				log.Fatalln("failed to load apimodel: %s", err.Error())
+				log.Fatalf("failed to load apimodel: %s", err.Error())
 			}
 			if _, _, err := dc.validateApimodel(); err != nil {
-				log.Fatalln("Failed to validate the apimodel after populating values: %s", err.Error())
+				log.Fatalf("Failed to validate the apimodel after populating values: %s", err.Error())
 			}
 			return dc.run()
 		},
@@ -152,7 +151,7 @@ func (dc *deployCmd) mergeAPIModel() error {
 			return fmt.Errorf(fmt.Sprintf("error merging --set values with the api model: %s", err.Error()))
 		}
 
-		log.Infoln(fmt.Sprintf("new api model file has been generated during merge: %s", dc.apimodelPath))
+		log.Infof(fmt.Sprintf("new api model file has been generated during merge: %s", dc.apimodelPath))
 	}
 
 	return nil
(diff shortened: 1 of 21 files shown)


pkg/api/agentPoolOnlyApi/vlabs/types.go
```diff
@@ -30,15 +30,37 @@ type ManagedCluster struct {
...
 	DNSPrefix               string                   `json:"dnsPrefix" validate:"required"`
 	FQDN                    string                   `json:"fqdn,omitempty"`
 	AgentPoolProfiles       []*AgentPoolProfile      `json:"agentPoolProfiles,omitempty" validate:"dive,required"`
 	LinuxProfile            *LinuxProfile            `json:"linuxProfile,omitempty" validate:"required"`
 	WindowsProfile          *WindowsProfile          `json:"windowsProfile,omitempty"`
 	ServicePrincipalProfile *ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
 	AccessProfiles          map[string]AccessProfile `json:"accessProfiles,omitempty"`
-	CertificateProfile      *CertificateProfile      `json:"certificateProfile,omitempty" validate:"required"`
+	CertificateProfile      *CertificateProfile      `json:"certificateProfile,omitempty"`
+	AzProfile               *AzProfile               `json:"azProfile,omitempty"`
```
kargakis: @jim-minter @pweil- fyi, we are not getting rid of azProfile with MSI - MSI is one more field in azure.conf

jim-minter:
looking good - waiting on discussion with MSFT around how RP passes in parameters (certs, AzProfile contents), also need to avoid calling the master templating code when in "agent pool only" mode - that way can avoid making the changes like removing the master IP address from the TLS certs, etc.

kargakis:
Tests added.

pkg/api/agentPoolOnlyApi/vlabs/const.go
```diff
@@ -31,3 +31,18 @@ const (
 	// ManagedDisks means that the nodes use managed disks for their os and attached volumes
 	ManagedDisks = "ManagedDisks"
 )
+
+// distro values
```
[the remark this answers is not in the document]
jim-minter: Open an issue.

pkg/api/agentPoolOnlyApi/vlabs/types.go
jim-minter: If we're having second thoughts about AzProfile is it worth leaving it out for now?

pkg/api/agentPoolOnlyApi/vlabs/validate.go
```diff
@@ -118,6 +143,58 @@ func validateUniqueProfileNames(profiles []*AgentPoolProfile) error {
...
+		}
+		var found bool
+		for _, validRole := range validRoles {
+			if agentPoolProfile.Role == validRole {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("role %q is only supported for the OpenShift orchestrator", agentPoolProfile.Role)
```
jim-minter: is not supported by orchestrator %q

jim-minter:
@pweil- and @jackfrancis please can you look this PR over?

pkg/api/agentPoolOnlyApi/vlabs/types.go
kargakis: Then we need to figure out a different way for delivering azure.conf inside the VM. FWIW, I have opened https://github.com/kubernetes/kubernetes/issues/64912.

pkg/api/agentPoolOnlyApi/vlabs/validate.go
kargakis: done

pkg/api/agentPoolOnlyApi/vlabs/types.go
jim-minter: Agreed, but given this api is shared with aks, I'm reluctant to add something we're not 100% sure about.  If we become sure we need it, we can add it subsequently.

kargakis: I can move it inside openShiftConfig. What's the alternative?

pkg/api/agentPoolOnlyApi/vlabs/validate_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/Azure/acs-engine/pkg/api/common"
+)
+
+func TestValidateAgentPoolProfile(t *testing.T) {
+	tests := []struct {
+		name string
+
```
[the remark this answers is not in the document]
kargakis: Yeah, it's:
```
test name

inputs

outputs
```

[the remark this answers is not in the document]
pweil-: that was more joking at how many new lines were in the definitions :smile:

pkg/openshift/certgen/unstable/defaults.go
```diff
@@ -17,15 +17,25 @@ func OpenShiftSetDefaultCerts(a *api.Properties, orchestratorName, clusterID str
 		return true, nil
 	}
 
+	var ips []net.IP
+	var dnsPrefix string
+	if a.MasterProfile != nil {
+		ips = []net.IP{net.ParseIP(a.MasterProfile.FirstConsecutiveStaticIP)}
+		dnsPrefix = a.MasterProfile.DNSPrefix
+	} else if a.HostedMasterProfile == nil {
+		return false, fmt.Errorf("no masterProfile or hostedMasterProfile found")
```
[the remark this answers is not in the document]
kargakis: Right, now this code needs to get called for generating the node boostrap certificate (and I think I need to make a change and expect the master CA as an input instead of generating it here) but all of this changeset depends on what happens in the RP, ie. how are we going to generate certs for HCP.

cc @sozercan @khenidak

kargakis:
@jim-minter @pweil- I have updated this PR to accept the master CA certs as an input to the api model and generate the node bootstrap certificates and kubeconfig. I also opened https://github.com/Azure/acs-engine/pull/3310 as a separate PR that is a prerequisite for this PR; the infra resources need to be moved out of the masterresources file since that template is not executed when running the agent pool api. PTAL

jim-minter:
> accept the master CA certs as an input to the api model

I don't think that conforms with what we were discussing yesterday, but it's not an acs-engine API change I can see that it's probably expedient for now.

kargakis:
Managed to get the agents up, still don't see the CSRs in the HCP; will keep digging

Reviews in the existing commits are welcome

kargakis:
> Managed to get the agents up, still don't see the CSRs in the HCP; will keep digging

Figured this out, bad bootstrap kubeconfig

I can deploy the nodes now just fine :) I think this PR is ready to go in, will open follow-up issues.

pkg/api/agentPoolOnlyApi/vlabs/types.go
pweil-: I feel like in our conversations we are on the fence about using this structure to hold anything for OpenShift.  There has also been talk about a different top level resource (other than `ManageCluster`).  

Should we hold for now?

cmd/deploy.go
```diff
@@ -238,34 +237,49 @@ func autofillApimodel(dc *deployCmd) error {
...
 	}
 
-	if dc.dnsPrefix != "" && dc.containerService.Properties.MasterProfile.DNSPrefix != "" {
-		return fmt.Errorf("invalid configuration: the apimodel masterProfile.dnsPrefix and --dns-prefix were both specified")
-	}
-	if dc.containerService.Properties.MasterProfile.DNSPrefix == "" {
-		if dc.dnsPrefix == "" {
-			return fmt.Errorf("apimodel: missing masterProfile.dnsPrefix and --dns-prefix was not specified")
+	switch {
+	case dc.containerService.Properties.MasterProfile != nil:
```
pweil-: I feel like we're changing behavior here and I'm not sure of the knock-on effects.  Presumably, since there was never a check for MasterProfile != nil above this logic was always being run and these fields were always being set.  Now they will not be. Looks like you found a couple of cases just in this method.

kargakis:
Follow-up issues: https://github.com/Azure/acs-engine/issues/3338 https://github.com/Azure/acs-engine/issues/3339

kargakis:
Closing in favor of https://github.com/Azure/acs-engine/pull/3456

Issue #3113, opened 2018-05-30
Heapster service not found


shayansarkar:
**Is this a request for help?**:
Yes
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
Issue
---

**What version of acs-engine?**:
Version: canary
GitCommit: f8753adf
---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes 1.10.3

**What happened**:
I deployed a new cluster, and in the new cluster I ran the following command

`kubectl top pods --all-namespaces`

and I received the following error

`Error from server (NotFound): the server could not find the requested resource (get services http:heapster:)`

**What you expected to happen**:

I expected to see a list of pods running and the CPU and Memory usage

**How to reproduce it** (as minimally and precisely as possible):

This is the json file i used for the creation of the cluster and then after it is deployed, I ran the kubectl top pods command.
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorVersion": "1.10.3",
      "kubernetesConfig": {
        "enableRbac": true,
        "dockerEngineVersion": "17.05.*",
        "addons": [
          {
            "name": "cluster-autoscaler",
            "enabled": true,
            "config": {
              "minNodes": "2",
              "maxNodes": "40"
            }
          }
        ]
      }
    },
    "masterProfile": {
      "count": 3,
      "dnsPrefix": "",
      "vmSize": "Standard_D2s_v3",
      "vnetSubnetId": "",
      "firstConsecutiveStaticIP": "",
      "vnetCidr": ""
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 2,
        "vmSize": "Standard_E2s_v3",
        "availabilityProfile": "VirtualMachineScaleSets",
        "vnetSubnetId": "",
        "storageProfile": "ManagedDisks"
      }
    ],
    "linuxProfile": {
      "adminUsername": "",
      "ssh": {
        "publicKeys": [
          {
            "keyData":""
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "",
      "secret": ""
    }
  }
}
```

**Anything else we need to know**:

shayansarkar:
For some additional information and questions, it seems as though if I ssh into the master node and run 

`kubectl top pods --all-namespaces` 

it works as expected.  However, if I run the same command from my machine, I still get the following message:
[...]

PR #2753, opened 2018-04-23
WIP use addon-resizer 1.8.1 for k8s 1.9 and 1.10

jackfrancis:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
use addon-resizer 1.8.1 for k8s 1.9 and 1.10
```

PR #3127, opened 2018-05-31
Heapster Service missing from addon and is breaking dashboard charts

### LINKED ISSUE #3124: Heapster Service missing from addon is breaking dashboard charts

---

---

**What version of acs-engine?**: v0.17.1

---

---

**What happened**: When using the dashboard, no charts are appearing for any of the nodes/pods.

---

**What you expected to happen**: Charts to appear in dashboard

---

**How to reproduce it** (as minimally and precisely as possible): Deploy heapster and dashboard addons. Navigate to dashboards and no charts appear.

---

The dashboard addon is configured to point at http://heapster.kube-system:80 for metrics, but the heapster addon does not define a Service, so the DNS lookup fails. The dashboard cant find heapster and so it hides the charts. Manually adding the Service back to the heapster addon fixes the problem.

This may be related to #3113

ewok2030:
**What this PR does / why we need it**: Dashboard charts not working without it

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3124 and maybe #3113

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-heapster-deployment.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-heapster-deployment.yaml
@@ -68,6 +68,21 @@ data:
     apiVersion: nannyconfig/v1alpha1
     kind: NannyConfiguration
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    kubernetes.io/cluster-service: "true"
+    kubernetes.io/name: Heapster
+  name: heapster
+  namespace: kube-system
+spec:
+  ports:
+  - port: 80
+    targetPort: 8082
+  selector:
+    k8s-app: heapster
+---
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:


jackfrancis:
Thanks @ewok2030! Was this only recently noticed <just because>? Or is this actually a regression?

ewok2030:
It’s included in the previous k8s version addons, but it looks like it got accidentally removed during some recent maintenance. 

It wouldn’t have raised any errors during testing because the dashboard just assumes healster is not available and hides the charts.

ewok2030:
@jackfrancis Yes, this is only an issue in 1.9+. The Service object is present in the 1.6/1.7/1.8 addons - it just looks like it got dropped by mistake in #2753.

jackfrancis:
Thanks for confirming @ewok2030!

/approve
/lgtm

PR #3131, opened 2018-05-31
OMS agent addon

jackfrancis:
**What this PR does / why we need it**: Adds a Kubernetes addon for OMS agent

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
OMS agent addon
```


pkg/acsengine/defaults.go
```diff
@@ -313,6 +313,24 @@ var (
...
 	}
+
+	// DefaultContainerMonitoringAddonsConfig is the default container monitoring Kubernetes addon Config
+	DefaultContainerMonitoringAddonsConfig = api.KubernetesAddon{
+		Name:    ContainerMonitoringAddonName,
+		Enabled: helpers.PointerToBool(api.DefaultContainerMonitoringAddonEnabled),
+		Config: map[string]string{
+			"omsAgentVersion":       "1.6.0-42",
+			"dockerProviderVersion": "2.0.0-2",
+			"workspaceGuid":         "c5905df1-b3b9-42b9-acf0-14a4c4ef028c",
```
jackfrancis: @rashmichandrashekar What is the value of having a default workspaceGuid and workspaceKey? Don't these always have to be user-provided?

jackfrancis:
I'd like to advocate that we merge this in in an "unknown working" state to unblock the other parts of this effort. @rashmichandrashekar can you commit to adding documentation and advising on E2E test coverage in future PRs?

pkg/acsengine/defaults.go
rashmichandrashekar: They do.. But for testing purposes we want to use this workspace, since the functionality we are testing doesn't involve validating data in the workspace.

rashmichandrashekar:
@jackfrancis : Yes I will add documentation and add tests/help with adding tests for E2E testing.

pkg/acsengine/defaults.go
```diff
@@ -313,6 +313,24 @@ var (
...
+			"workspaceKey":          "b4xMPEns/5Oo61hQbRLJQoPhbmAHdBn2eDjkCbRzgEISiKi9m1CR+093sczO9E8iA5w1EyGVCVkwsnuWt1MD9w==",
```
CecileRobertMichon: we should probably not have a hardcoded key in the code

pkg/acsengine/defaults.go
```diff
@@ -313,6 +313,24 @@ var (
...
+		},
+		Containers: []api.KubernetesContainerSpec{
+			{
+				Name:  "omsagent",
+				Image: "dockerio.azureedge.net/microsoft/oms:ciprod05082018",
```
CecileRobertMichon: Are CPU/Memory Limits and Requests not set intentionally?

pkg/acsengine/defaults.go
jackfrancis: @rashmichandrashekar this question is for you :)

rashmichandrashekar: Sorry missed this :)
That is a good point. Is there a way for the acs-engine to talk to a secret store or key vault so that we can put these values there?

pkg/acsengine/defaults.go
rashmichandrashekar: Yes. This is the yaml file we will be using with AKS cluster creation. We intend to use the same yaml so that we are close to testing out the real scenario.

pkg/acsengine/defaults.go
rashmichandrashekar: On a second thought - since we will be testing this out as an addon, is there a way the cluster can take these as input parameters and pass them in during testing?

jackfrancis: These params are user-configurable, e.g., in your cluster api model:

```
"addons": [
          {
            "name": "container-monitoring",
            "enabled": true,
            "config": {
                "omsAgentVersion": "1.6.0-42",
                "dockerProviderVersion": "2.0.0-2",
		"workspaceGuid": "c5905df1-b3b9-42b9-acf0-14a4c4ef028c",
		"workspaceKey": "b4xMPEns/5Oo61hQbRLJQoPhbmAHdBn2eDjkCbRzgEISiKi9m1CR+093sczO9E8iA5w1EyGVCVkwsnuWt1MD9w=="
            }
            "containers": [
              {
                "name": "omsagent",
                "cpuRequests": "50m",
                "memoryRequests": "150Mi",
                "cpuLimits": "50m",
                "memoryLimits": "150Mi"
              }
            ]
          }
```

rashmichandrashekar: okay. In that case we can update this to be passed in from the cluster api model.

jackfrancis: ACK. Should I remove the hardcoded "workspaceGuid" and "workspaceKey" default values, then?

rashmichandrashekar: Yes please. Thank you.

Issue #2832, opened 2018-05-01
Windows doesn't support pre-release k8s versions


CecileRobertMichon:
Deploying a pre-release (ex. 1.11.0-alpha1) k8s version throws the following error:

```

Mode                LastWriteTime         Length Name                          
----                -------------         ------ ----                          
d-----         5/1/2018   9:33 PM                volumeplugins                 
C:\AzureData\CustomDataSetupScript.ps1 : Cannot convert value "1.11.0-alpha.1" 
to type "System.Version". Error: "Input string was not in a correct format."
At line:1 char:1
+ C:\AzureData\CustomDataSetupScript.ps1 -MasterIP 10.255.255.5 -KubeDn ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorExcep 
   tion
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorExceptio 
   n,CustomDataSetupScript.ps1
 ```

For now all pre-release versions are disabled on Windows, we should enable pre-release version parsing and enable future pre-releases for Windows.

PatrickLang:
is this the issue? https://github.com/Azure/acs-engine/blob/b7364feda03b86967b6f401d5518c8edad73be68/parts/k8s/kuberneteswindowssetup.ps1#L318

adelina-t:
@PatrickLang yep, looked over a deployment, that's the reason the script errors out. Made pull request  to fix this, tested it on 1.11.0-alpha.2, setup works just fine.

PR #3135, opened 2018-06-01
Strip k8s version to Major.Minor.Build format

adelina-t:
System.Version will fail if versions are of the form x.y.z-alpha . Strip version before checks involving System.Version.

Fixes: #2832


jackfrancis:
@PatrickLang for sanity check

/approve
/lgtm

pending Windows E2E

Thanks @adelina-t!

PR #3138, opened 2018-06-01
Feat/custom dns

aminebizid:
When using custom VNET, I had to update manually /etc/hosts files on whole nodes to add nodes IP and hostname entries. This is makes autoscaling unusable.
Thanks to this PR, I can choose a custom DNS server I created (Ubuntu/Bind9) and I'm able by adding an extension to update the DNS entry of my nodes on their creation.

### DIFF
--- a/parts/k8s/kubernetesmasterresources.t
+++ b/parts/k8s/kubernetesmasterresources.t
@@ -467,11 +467,13 @@
         ,
         "enableIPForwarding": true
 {{end}}
+{{if HasCustomNodesDNS}}
  ,"dnsSettings": {
           "dnsServers": [
               "[variables('dnsServer')]"
           ]
       }
+{{end}}
 {{if or .MasterProfile.IsCustomVNET IsOpenShift}}
         ,"networkSecurityGroup": {
           "id": "[variables('nsgID')]"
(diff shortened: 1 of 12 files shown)


pkg/api/vlabs/types.go
```diff
@@ -604,6 +610,16 @@ func (l *LinuxProfile) HasSearchDomain() bool {
 	return false
 }
 
+// HasSearchDomain returns true if the customer specified secrets to install
```
jackfrancis: copy/paste typo, comment should be `// HasCustomNodesDNS` to pass lint

parts/k8s/kubernetesmasterresources.t
```diff
@@ -467,6 +467,11 @@
         ,
         "enableIPForwarding": true
 {{end}}
+ ,"dnsSettings": {
```
jackfrancis: Doesn't this need to be guarded inside a `HasCustomNodesDNS` block?

parts/k8s/kubernetesmasterresources.t
```diff
@@ -543,6 +548,11 @@
           ,
           "enableIPForwarding": true
   {{end}}
+   ,"dnsSettings": {
```
jackfrancis: ditto

jackfrancis:
@aminebizid FYI, I rebased off of master because of some nasty changes to `engine.go` that you didn't anticipate :)

I'll review and run E2E, thanks so much for this!

jackfrancis:
@aminebizid Should we file a follow-up item to support > 1 DNS server?

PR #3141, opened 2018-06-01
openshift master scripts refactor

mjudeikis:
This pr does few things:

* make script idempotent, so it can be rerun when developing
* adds ip addr validation (each time when we call `dig`, `hostname`, we should always validate the output!)
*  splits code in functions so they could be deletes or commented out one by one if needed when developing. 

Next steps would be to remove some of the code. But when code is idempotent we can move agent install to base image and dont worry it will break downstream code (less code orchestration)  

@kwoodson @jim-minter @pweil- @kargakis


mjudeikis:
I know we want to get rid of this script, but this will enable easier development for now :)

parts/openshift/unstable/openshiftmasterscript.sh
```diff
@@ -53,13 +111,10 @@ oc adm create-bootstrap-policy-file --filename=/etc/origin/master/policy.json
...
-# additional functionality in this script that requires it.  One route to remove
-# this code is to bake this script into the base image, then pass in parameters
-# such as the registry storage account name and key direct from ARM.
-rpm -i https://packages.microsoft.com/yumrepos/azure-cli/azure-cli-2.0.31-1.el7.x86_64.rpm
-
+# configure azure cloud provider
+function configure_cloud_provider() {
 set +x
 . <(sed -e 's/: */=/' /etc/origin/cloudprovider/azure.conf)
 az login --service-principal -u "$aadClientId" -p "$aadClientSecret" --tenant "$aadTenantId" &>/dev/null
```
mjudeikis: this one need better error handling as if variables are passed wrong it errors to /dev/null

kargakis:
I would try to keep this in sync with the script in release39 if we want to move forward with this.

mjudeikis:
Agreed. Let first get consensus if people fine with structure and Will change 3.9 accordingly too.

jim-minter:
I don't want to do this.  The master script goes away with HCP.

PR #3147, opened 2018-06-01
Merge AADProfile during update

amanohar:
This PR merges old and new AADProfile API models for the update case as all properties are not required during update and ACS Engine merge API models for the caller.


rjtsdl:
LGTM

amanohar:
Thanks @jackfrancis for the quick review!

Issue #2952, opened 2018-05-14
openshift: "503 Service Unavailable" flake


jim-minter:
```
$ oc get route/nginx-example -n default -o jsonpath={.spec.host}
2018/05/14 17:44:00 got status "503 Service Unavailable" while trying to access nginx-example-default.40.114.111.175.nip.io
```

jim-minter:
https://circleci.com/gh/Azure/acs-engine/26335

jim-minter:
@kargakis can you take a look?

jim-minter:
Seems improbable that it was DNS if it got as far as receiving a 503 response?

kargakis:
Need to check whether the pod uses a readiness check and also whether the deployment config waits for it to turn ready.

kargakis:
> Need to check whether the pod uses a readiness check and also whether the deployment config waits for it to turn ready.

Both readiness probe is used and the DC has sane rolling parameter defaults.

kargakis:
ugh, retries were using microseconds for intervals instead of the intended milliseconds - opened https://github.com/Azure/acs-engine/pull/3202

PR #3154, opened 2018-06-04
Move artifacts gathering around

kargakis:
* moves logging of node and pod info into artifact gathering in `teardown`
* removes running diagnostics at least for now
* avoid version check panic

Depends on https://github.com/Azure/acs-engine/pull/3152

### DIFF
--- a/test/e2e/openshift/util/util.go
+++ b/test/e2e/openshift/util/util.go
@@ -166,13 +166,14 @@ func FetchClusterInfo(logPath string) error {
 
 // FetchOpenShiftLogs returns logs for all OpenShift components
 // (control plane and infra).
-func FetchOpenShiftLogs(distro, version, sshKeyPath, adminName, name, location, logPath string) {
+func FetchOpenShiftLogs(distro, version, sshKeyPath, adminName, name, location, logPath string) error {
 	if err := fetchControlPlaneLogs(distro, version, sshKeyPath, adminName, name, location, logPath); err != nil {
-		log.Printf("Cannot fetch logs for control plane components: %v", err)
+		return fmt.Errorf("cannot fetch logs for control plane components: %v", err)
 	}
 	if err := fetchInfraLogs(logPath); err != nil {
-		log.Printf("Cannot fetch logs for infra components: %v", err)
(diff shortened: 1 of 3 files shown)


kargakis:
Rebased

kargakis:
Flaked on https://github.com/Azure/acs-engine/issues/2952

test/e2e/openshift/util/util.go
```diff
@@ -153,6 +139,27 @@ func FetchLogs(kind, namespace, name string) string {
 	return string(out)
 }
 
+// FetchClusterInfo returns node and pod information about the cluster.
+func FetchClusterInfo(logPath string) {
+	needsLog := map[string]string{
```
jim-minter: I'm happy for this to merge, but you don't think `map[string]func() error` would be a better foundation here?

kargakis: I don't feel strong either way, updated to your suggestion (which needs to be `map[string]func() string`).

jim-minter: apols, I meant `map[string]func() (string, error)` - in which case the changes to the other functions aren't needed and they're more reusable

kargakis: ack, updated

test/e2e/openshift/util/util.go
```diff
@@ -174,7 +185,8 @@ func fetchControlPlaneLogs(distro, version, sshKeyPath, adminName, name, locatio
 	case common.OpenShiftVersionUnstable:
 		return fetchUnstableControlPlaneLogs(distro, sshKeyPath, sshAddress, name, logPath)
 	default:
-		panic(fmt.Sprintf("BUG: invalid OpenShift version %s", version))
+		log.Printf("Invalid OpenShift version %q - won't gather logs from the control plane", version)
+		return nil
```
jim-minter: as discussed - return fmt.Errorf() here

kargakis: done

kargakis:
@jackfrancis @CecileRobertMichon please merge

PR #3157, opened 2018-06-04
Replace tabs with spaces

sylr:


jackfrancis:
Hi @sylr, thanks for spending the time to do this! I know it's annoying, but we actually care about the data real estate in these files, as byte counts matter in terms of overall stability.

I've filed a work item to rationalize those files so it's clear what our intent is:

https://github.com/Azure/acs-engine/issues/3161

PR #3159, opened 2018-06-04
Fix test-style broken since bcf7d3b

sylr:
After bcf7d3b err var is directly re-assigned without being
used and this breaks the static validation.


jackfrancis:
Thanks @sylr!

PR #3160, opened 2018-06-04
Drop redundant prowjob namespace

kargakis:
/assign CecileRobertMichon 

The namespace is already defaulted to "default" but we don't use prowjobs so we don't really care about this config option.


kargakis:
@cecilerobertmichon works great! I think we are done with prow (at least
for now)

PR #3162, opened 2018-06-04
Adding a validation when using existing vnet.

kwoodson:
Adding validation to the `masterProfile` when specifying an existing vnetsubnetid and/or firstconsecutivestaticip.

@kargakis @jim-minter

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -304,12 +304,19 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
 
 func (a *Properties) validateMasterProfile() error {
 	m := a.MasterProfile
-	if a.OrchestratorProfile.OrchestratorType == OpenShift && m.Count != 1 {
-		return errors.New("openshift can only deployed with one master")
-	}
-
-	if a.OrchestratorProfile.OrchestratorType == OpenShift && m.StorageProfile != ManagedDisks {
-		return errors.New("OpenShift orchestrator supports only ManagedDisks")
+	if a.OrchestratorProfile.OrchestratorType == OpenShift {
+		if m.Count != 1 {
+			return errors.New("openshift can only deployed with one master")
+		}
+		if m.VnetSubnetID != "" && m.FirstConsecutiveStaticIP == "" {
+			return errors.New("when specifying a vnetsubnetid the firstconsecutivestaticip is required")
+		}
+		if m.FirstConsecutiveStaticIP != "" && m.VnetSubnetID == "" {
+			return errors.New("when specifying the firstconsecutivestaticip the vnetsubnetid is required")
+		}
+		if m.StorageProfile != ManagedDisks {
+			return errors.New("OpenShift orchestrator supports only ManagedDisks")
+		}
 	}
 
 	if m.ImageRef != nil {
--- a/pkg/api/vlabs/validate_test.go
+++ b/pkg/api/vlabs/validate_test.go
@@ -1000,6 +1000,31 @@ func TestMasterProfileValidate(t *testing.T) {
 			},
 			expectedErr: "openshift can only deployed with one master",
 		},
+		{ // test existing vnet: run with only specifying vnetsubnetid
+			orchestratorType: OpenShift,
+			masterProfile: MasterProfile{
+				VnetSubnetID: "testvnetstring",
+				Count:        1,
+			},
+			expectedErr: "when specifying a vnetsubnetid the firstconsecutivestaticip is required",
+		},
+		{ // test existing vnet: run with only specifying firstconsecutivestaticip
+			orchestratorType: OpenShift,
+			masterProfile: MasterProfile{
+				FirstConsecutiveStaticIP: "10.0.0.1",
+				Count: 1,
+			},
+			expectedErr: "when specifying the firstconsecutivestaticip the vnetsubnetid is required",
+		},
+		{ // test existing vnet: run with specifying both vnetsubnetid and firstconsecutivestaticip
+			orchestratorType: OpenShift,
+			masterProfile: MasterProfile{
+				DNSPrefix:                "dummy",
+				VnetSubnetID:             "testvnetstring",
+				FirstConsecutiveStaticIP: "10.0.0.1",
+				Count: 1,
+			},
+		},
 	}
 
 	for i, test := range tests {


pkg/api/vlabs/validate.go
```diff
@@ -295,8 +295,13 @@ func (o *OpenShiftConfig) Validate() error {
 
 // Validate implements APIObject
 func (m *MasterProfile) Validate(o *OrchestratorProfile) error {
-	if o.OrchestratorType == OpenShift && m.Count != 1 {
-		return errors.New("openshift can only deployed with one master")
+	if o.OrchestratorType == OpenShift {
+		if m.Count != 1 {
+			return errors.New("openshift can only deployed with one master")
+		}
+		if (m.VnetSubnetID != "" && m.FirstConsecutiveStaticIP == "") || (m.FirstConsecutiveStaticIP != "" && m.VnetSubnetID == "") {
```
jim-minter: 1) I think it should be allowed to specify/override FirstConsecutiveStaticIP in the non-existing vnet case.
2) We default FirstConsecutiveStaticIP somewhere - bearing that in mind, does this work correctly?  Perhaps we should not default FirstConsecutiveStaticIP when VnetSubnetID is set?
3) Needs updated unit test.

kargakis: > We default FirstConsecutiveStaticIP somewhere - bearing that in mind, does this work correctly? Perhaps we should not default FirstConsecutiveStaticIP when VnetSubnetID is set?

I briefly looked at this today, it seems that validation is run before defaulting

validation: [loadApiModel](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/cmd/generate.go#L57) calls [LoadContainerServiceFromFile](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/cmd/generate.go#L135) which calls [DeserializeContainerService](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/pkg/api/apiloader.go#L34) which calls [LoadContainerService](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/pkg/api/apiloader.go#L45) which finally does [validation](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/pkg/api/apiloader.go#L171).
defaulting: [run](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/cmd/generate.go#L61) calls [GenerateTemplate](https://github.com/Azure/acs-engine/blob/bed3f27da4b42f6a446683889f5c54c4c49b87e5/cmd/generate.go#L185) which is setting the defaults [here](https://github.com/Azure/acs-engine/blob/0dececb663cc4e792b786f274690c69ce057f1c0/pkg/acsengine/template_generator.go#L57)

I would be surprised if this worked differently for `acs-engine deploy`.

kwoodson: After chatting with @jim-minter, I'll implement the validation unit test and go from there.

pkg/api/vlabs/validate.go
```diff
@@ -295,8 +295,13 @@ func (o *OpenShiftConfig) Validate() error {
...
+			return errors.New("when deploying openshift using an existing vnet the vnetSubnetId and firstConsecutiveStaticIP are required")
```
kargakis: if m.VnetSubnetID == "" then we are not deploying with an an existing vnet, right?

kwoodson:
@jim-minter, added unit tests.
1. Fails when specifying only vnetsubnetid
2. Fails when specifying only firstconsecutiveip
3. Succeeds when passing both parameters
4. Succeeds when not passing both parameters.

pkg/api/vlabs/validate.go
kwoodson: @kargakis,  @jim-minter and I chatted about this.  We would expect the vnetsubnetid and the firstconsecutivestaticip to be defined if deploying into an existing vnet.

@jim-minter, does this sound correct?

pkg/api/vlabs/validate.go
```diff
@@ -302,14 +302,20 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
 	return nil
 }
 
+// Validate implements APIObject
```
jim-minter: I don't understand this comment?

pkg/api/vlabs/validate.go
jim-minter: @kwoodson the error string needs to be reworded or maybe become 2 error strings - i.e. "if you specify an existing vnet you must specify the ip" and "you can't specify the ip unless you specify an existing vnet"

jim-minter:
@kwoodson just nits - please fix up and ping me then will lgtm

pkg/api/vlabs/validate.go
kwoodson: Looks like some sort of copy pasta from here: https://github.com/Azure/acs-engine/blob/master/pkg/api/vlabs/validate.go#L97

Will remove.

pkg/api/vlabs/validate.go
kwoodson: Ok, will break into 2 separate statements.

kwoodson:
@jim-minter, addressed your concerns.

kwoodson:
@CecileRobertMichon, Mind merging this one if its to your linking?

kwoodson:
Here was the error output from e2e:

```
$ ./bin/acs-engine generate /go/src/github.com/Azure/acs-engine/_output/openshift-eastus-42464.json --output-directory /go/src/github.com/Azure/acs-engine/_output/openshift-eastus-42464
INFO[0017] Error returned by LoadContainerService: when specifying the firstconsecutivestaticip the vnetsubnetid is required 
2018/06/13 21:22:40 Cannot execute remote command "sudo cat /var/lib/waagent/custom-script/download/0/stderr": exit status 255
2018/06/13 21:22:40 Cannot execute remote command "sudo cat /var/lib/waagent/custom-script/download/0/stdout": exit status 255
```

@jim-minter, @kargakis,

I'm guessing that the defaulted firstconsecutivestaticip is coming into play here.  I need to address the defaulted variable when checking for whether the vnetsubnetid is passed.

PR #3166, opened 2018-06-05
Replace semver package

CecileRobertMichon:
**What this PR does / why we need it**: The Masterminds semver package has a bug when comparing pre-release versions, use github.com/blang/semver instead.

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```


pkg/acsengine/template_generator.go
```diff
@@ -174,17 +173,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			return cs.Properties.OrchestratorProfile.IsKubernetes() && common.IsKubernetesVersionGe(cs.Properties.OrchestratorProfile.OrchestratorVersion, version)
 		},
 		"IsKubernetesVersionLt": func(version string) bool {
-			orchestratorVersion, _ := semver.NewVersion(cs.Properties.OrchestratorProfile.OrchestratorVersion)
-			constraint, _ := semver.NewConstraint("<" + version)
-			return cs.Properties.OrchestratorProfile.IsKubernetes() && constraint.Check(orchestratorVersion)
-		},
-		"IsKubernetesVersionTilde": func(version string) bool {
```
CecileRobertMichon: removed this function as it's not used anywhere in the code. Any objections?

pkg/api/common/versions.go
```diff
@@ -108,19 +108,13 @@ func GetAllSupportedKubernetesVersions() []string {
...
 	var ret []string
-	var cons *semver.Constraints
-	minVersion, _ := semver.NewVersion(version)
+	minVersion, _ := semver.Make(version)
 	for _, v := range versions {
-		sv, _ := semver.NewVersion(v)
-		if preReleases && minVersion.Prerelease() == "" {
-			sv, _ = semver.NewVersion(fmt.Sprintf("%d.%d.%d", sv.Major(), sv.Minor(), sv.Patch()))
+		sv, _ := semver.Make(v)
+		if !preReleases && len(sv.Pre) != 0 {
```
CecileRobertMichon: if `preReleases` is false we don't want to include pre-releases so we skip any version with a pre-release part

pkg/api/common/versions.go
```diff
@@ -133,18 +127,13 @@ func GetVersionsGt(versions []string, version string, inclusive, preReleases boo
...
-		sv, _ := semver.NewVersion(v)
-		if preReleases {
-			sv, _ = semver.NewVersion(fmt.Sprintf("%d.%d.%d", sv.Major(), sv.Minor(), sv.Patch()))
-		}
-		if inclusive {
-			cons, _ = semver.NewConstraint("<=" + version)
-		} else {
-			cons, _ = semver.NewConstraint("<" + version)
+		sv, _ := semver.Make(v)
+		if !preReleases && len(sv.Pre) != 0 {
```
CecileRobertMichon: if `preReleases` is false we don't want to include pre-releases so we skip any version with a pre-release part

pkg/api/common/versions.go
```diff
@@ -342,29 +331,29 @@ func RationalizeReleaseAndVersion(orchType, orchRel, orchVer string, hasWindows
 	return version
 }
 
-// IsKubernetesVersionGe returns if a semver string is >= to a compare-against semver string (suppport "-" suffixes)
+// IsKubernetesVersionGe returns true if actualVersion is greater than or equal to version
 func IsKubernetesVersionGe(actualVersion, version string) bool {
-	orchestratorVersion, _ := semver.NewVersion(strings.Split(actualVersion, "-")[0]) // to account for -alpha and -beta suffixes
```
CecileRobertMichon: we don't need to do this anymore because the new semver package handles pre-release versions

pkg/api/common/versions_test.go
```diff
@@ -55,7 +55,7 @@ func TestGetVersionsGt(t *testing.T) {
 		"1.2.0":      true,
 		"1.2.1":      true,
 	}
-	v := GetVersionsGt(versions, "1.1.0-alpha.1", false, false)
+	v := GetVersionsGt(versions, "1.1.0-alpha.1", false, true)
```
CecileRobertMichon: this was a bug in the unit test since the test expects pre-release versions

pkg/api/common/versions_test.go
```diff
@@ -115,6 +115,21 @@ func TestIsKubernetesVersionGe(t *testing.T) {
 			actualVersion:  "1.8.7",
 			expectedResult: true,
 		},
+		{
```
CecileRobertMichon: New UT cases to check pre-releases are handled correctly

pkg/api/common/versions_test.go
```diff
@@ -130,7 +145,7 @@ func TestIsKubernetesVersionGe(t *testing.T) {
 
 func TestGetVersionsLt(t *testing.T) {
 	versions := []string{"1.1.0", "1.2.0-rc.1", "1.2.0", "1.2.1"}
-	expected := []string{"1.2.0", "1.2.1"}
+	expected := []string{"1.1.0", "1.2.0"}
```
CecileRobertMichon: the expected versions here were wrong

pkg/api/common/versions_test.go
```diff
@@ -275,6 +290,34 @@ func TestGetVersionsBetween(t *testing.T) {
 			t.Errorf(errStr)
 		}
 	}
+
+	versions = []string{"1.11.0-alpha.1", "1.11.0-alpha.2", "1.11.0-beta.1"}
```
CecileRobertMichon: New UT cases to check handling pre-releases

pkg/api/orchestrators.go
```diff
@@ -181,12 +182,12 @@ func kubernetesInfo(csOrch *OrchestratorProfile) ([]*OrchestratorVersionProfile,
...
 	ret := []*OrchestratorProfile{}
 
-	currentVer, err := semver.NewVersion(csOrch.OrchestratorVersion)
+	currentVer, err := semver.Make(csOrch.OrchestratorVersion)
 	if err != nil {
 		return nil, err
 	}
-	nextNextMinorVersion := currentVer.IncMinor().IncMinor()
-	upgradeableVersions := common.GetVersionsBetween(common.GetAllSupportedKubernetesVersions(), csOrch.OrchestratorVersion, nextNextMinorVersion.String(), false, true)
+	nextNextMinorString := strconv.FormatUint(currentVer.Major, 10) + "." + strconv.FormatUint(currentVer.Minor+2, 10) + ".0"
```
CecileRobertMichon: this is a workaround I had to apply because the `IncMinor()` function equivalent doesn't exist in this package. All it does is keep the same `Major`, increment `Minor` by 2 and set `patch` to 0. Validated that this works in unit tests.

pkg/api/types.go
```diff
@@ -859,8 +859,8 @@ func (o *OrchestratorProfile) GetAPIServerEtcdAPIVersion() string {
 	ret := "etcd3"
 	if o.KubernetesConfig != nil {
 		// if we are here, version has already been validated..
-		etcdversion, _ := semver.NewVersion(o.KubernetesConfig.EtcdVersion)
-		if etcdversion != nil && 2 == etcdversion.Major() {
```
CecileRobertMichon: //TODO: re-add nil check

CecileRobertMichon: done

CecileRobertMichon: I remembered the reason I removed it is because `Make` will never return nil, it returns an empty Version if anything

CecileRobertMichon:
The unit tests are expected to fail on this PR because the blang/semver package is missing from the `vendor` dir. For the result of tests with the new package vendoring see #3184

CecileRobertMichon:
Closing in favor of #3184

Issue #3026, opened 2018-05-21
Increase log level in the openshift control plane


kargakis:
@pweil- @jim-minter opening for discussion since I am torn.

We are setting up and starting the control plane ourselves (no ansible) as part of the master script. Once we switch to static pods there won't be any big difference. I see two possible actions for increasing the loglevel in the e2e tests:

1. update log level in the config (sysconfig or pod manifests) after a deployment is done but before tests start + restart services (ugly plus we miss debug logs during post-install tasks)
2. expose loglevel in openshift.json - as I was writting this I realized we cannot configure this today at all so I think it makes sense to do it here and now.

I think we should do 2. but wanted to discuss before I move forward with opening a PR, maybe there is something else? Thoughts? Objections?

/label orchestrator/openshift
/kind feature

jim-minter:
I'm tempted to do nothing for now.  Would it help to express more clearly /why/ we think it would be necessary to do this?

kargakis:
The control plane currently runs on loglevel=2 which is too low for debugging. In Origin, e2e tests run with 4.

pweil-:
we really need the ability to change loglevel at runtime.  

2 seems like the best option but it seems less useful if it cannot be used on an existing cluster.  That would be the bar for me.   Not sure how easy that is.

jim-minter:
@kargakis can we table this for the architecture call?  I feel there's way more to talk about here than meets the eye.  I've added a topic to the arch topics doc.

PR #3169, opened 2018-06-05
openshift: increase loglevel for control plane components 

kargakis:
Needed for debugging existing flakes. As per yesterday's arch call discussion, we agreed to bump the loglevel for now and can revisit later.

/assign jim-minter pweil- 
/label orchestrator/openshift
/approve

Fixes https://github.com/Azure/acs-engine/issues/3026


pkg/openshift/certgen/release39/templates/master/etc/etcd/etcd.conf
```diff
@@ -48,6 +48,7 @@ ETCD_PEER_KEY_FILE=/etc/etcd/peer.key
 
 #[logging]
 ETCD_DEBUG="False"
```
kargakis: `ETCD_DEBUG="True"` is too chatty because of the gossip protocol so I didn't enable it here - we may want to use it eventually if bumping the etcdserver loglevel is not enough.

parts/openshift/unstable/openshiftmasterscript.sh
```diff
@@ -106,14 +106,17 @@ done
...
 
+export KUBECONFIG=/etc/origin/master/admin.kubeconfig
+
 # Move each static pod into place so the kubelet will run it.
 # Pods: [apiserver, controller, etcd]
-mv /etc/origin/node/disabled/* /etc/origin/node/pods
+oc set env --local -f /etc/origin/node/disabled/apiserver.yaml DEBUG_LOGLEVEL=4 -o yaml --dry-run > /etc/origin/node/pods/apiserver.yaml
+oc set env --local -f /etc/origin/node/disabled/controller.yaml DEBUG_LOGLEVEL=4 -o yaml --dry-run > /etc/origin/node/pods/controller.yaml
+mv /etc/origin/node/disabled/etcd.yaml /etc/origin/node/pods/etcd.yaml
+rm /etc/origin/node/disabled/*
```
jim-minter: `rm /etc/origin/node/disabled` better?

jim-minter:
nit/question otherwise looks fine

parts/openshift/unstable/openshiftmasterscript.sh
kargakis: updated

Issue #3114, opened 2018-05-30
openshift: infra node is not created


kargakis:
```
FAILED - RETRYING: command (7 retries left).
FAILED - RETRYING: command (6 retries left).
FAILED - RETRYING: command (5 retries left).
FAILED - RETRYING: command (4 retries left).
FAILED - RETRYING: command (3 retries left).
FAILED - RETRYING: command (2 retries left).
FAILED - RETRYING: command (1 retries left).
fatal: [localhost]: FAILED! => {"attempts": 30, "changed": true, "cmd": "/usr/bin/oc get nodes -l \"region=infra\" | wc -l", "delta": "0:00:01.313789", "end": "2018-05-30 18:21:41.381096", "failed": true, "rc": 0, "start": "2018-05-30 18:21:40.067307", "stderr": "No resources found.", "stderr_lines": ["No resources found."], "stdout": "0", "stdout_lines": ["0"]}

PLAY RECAP *********************************************************************
localhost                  : ok=52   changed=25   unreachable=0    failed=1  
```
/label orchestrator/openshift
https://circleci.com/gh/Azure/acs-engine/29050#artifacts/containers/0
/kind flake
@jim-minter @pweil- @mjudeikis @kwoodson I have seen this a couple of times when doing my own deployments

mjudeikis:
bumpt. Got this one too. Any idea how to debug?
api model has infra defined:
```
    {
        "name": "infra",
        "count": 1,
        "vmSize": "Standard_D4s_v3",
        "osType": "Linux",
        "availabilityProfile": "AvailabilitySet",
        "storageProfile": "ManagedDisks",
        "distro": "ubuntu",
        "imageReference": {
          "name": "centos7-3.10-xxxx",
          "resourceGroup": "images"
        },
        "role": "infra",
        "fqdn": "",
        "preProvisionExtension": null,
        "extensions": []
      }
    ],
```
infra cound:
```
  "infraCount": {
      "value": 1
    },
```

mjudeikis:
Just after this got this one:
```
 "status": "Failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "AllocationFailed",
        "message": "Allocation failed. We do not have sufficient capacity for the requested VM size in this region. Read more about improving likelihood of allocation success at http://aka.ms/allocation-guidance"
      }
    ]
  }
}
```
[...]

PR #3173, opened 2018-06-05
openshift: infra nic depends on router lb

kargakis:
Likely fixes https://github.com/Azure/acs-engine/issues/3114


jim-minter:
good find, kargakis!

kargakis:
Failed on the etcd timeout :)

PR #3184, opened 2018-06-06
Make vendor semver package

CecileRobertMichon:
**What this PR does / why we need it**: The Masterminds semver package has a bug when comparing pre-release versions, use github.com/blang/semver instead.

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```


pkg/api/types.go
```diff
@@ -859,8 +859,8 @@ func (o *OrchestratorProfile) GetAPIServerEtcdAPIVersion() string {
 	ret := "etcd3"
```
jackfrancis: Let's change this entire function to:

```
etcdversion, _ := semver.Make(o.KubernetesConfig.EtcdVersion)
return "etcd" + etcdversion.Major
```

pkg/api/vlabs/validate.go
```diff
@@ -149,17 +149,15 @@ func (o *OrchestratorProfile) Validate(isUpdate, HasWindows bool) error {
 					return err
 				}
 				minVersion := "1.7.0"
+				cons, _ := semver.Make(minVersion)
```
jackfrancis: Change these two lines to:

```
minVersion, _ := semver.Make("1.7.0")
```

jackfrancis: And then use `minVersion` below in place of `cons`

jackfrancis: Also return errors when it's practical

PR #3189, opened 2018-06-06
Prow updates

kargakis:
First commit adds two options in the prow config:
`implicit_self_approve: true` means every PR opened is implicitly self-approved by the author. So if I open a PR against a directory I am an owner of, I don't need to explicitly `/approve` it.
`lgtm_acts_as_approve: true` means every PR I am `/lgtm`ing is also approved by me. So if I `/lgtm` a PR which I can also `/approve`, I don't need to do `/approve` manually.

Second commit syncs the manifests in acs-engine with what's deployed in-cluster.

/assign CecileRobertMichon

### DIFF
--- a/.prowci/tide.yaml
+++ b/.prowci/tide.yaml
@@ -38,7 +38,7 @@ items:
         serviceAccountName: tide
         containers:
         - name: tide
-          image: registry.svc.ci.openshift.org/ci/tide:latest
+          image: quay.io/kargakis/tide:workaround
           imagePullPolicy: IfNotPresent
           args:
           - --dry-run=false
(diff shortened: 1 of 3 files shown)


.prowci/tide.yaml
```diff
@@ -38,7 +38,7 @@ items:
         serviceAccountName: tide
         containers:
         - name: tide
-          image: registry.svc.ci.openshift.org/ci/tide:latest
+          image: quay.io/kargakis/tide:workaround
```
kargakis: For future reference, we are working around https://github.com/kubernetes/kubernetes/issues/64594

PR #3181, opened 2018-06-05
add GPU ExtendedResourceToleration admission controller support

lachie83:
Use-case - Users only want workloads that require GPU resources to be scheduled to nodes that have GPUs. 

Adds the following support:
* Adds ExtendedResourceToleration to the admission controller list - The allows for workloads with resource requires/limits of `nvidia.com/gpu` to be auto-tolerated by the admission controller to enable scheduling to the GPU node pool(s) - https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#extendedresourcetoleration
* Adds the toleration to the nvidia-deviceplugin daemonset so that will only be scheduled to nodes with the supported GPU hardware

PR #3192, opened 2018-06-06
fix #3190 only use nvidia-container-runtime on gpu nodes

### LINKED ISSUE #3190: Hybrid clusters with GPU & CPU do not work
NO

---

ISSUE
---

0.18.1
---

Kubernetes

**What happened**:
Tried to deploy a cluster with two agent pools; one with reguler VMs (CPU only) and one with GPU's.
Similar to this blog post: https://www.microsoft.com/developerblog/2017/11/21/autoscaling-deep-learning-training-kubernetes/

The ARM deployment failed with a VMExtensionProvisioningError. 

The master nodes start correctly. Also, the nodes with GPU start correctly. The error is in the CPU nodes.

Logging into a the node with CPU, we see that the hyperkube-extract service won't start:

```
systemd[1]: hyperkube-extract.service: Service hold-off time over, scheduling restart.
systemd[1]: Stopped kubectl and kubelet extraction.
systemd[1]: Starting kubectl and kubelet extraction...
docker[38944]: v1.10.3: Pulling from hyperkube-amd64
docker[38944]: Digest: sha256:00d814b1f7763f4ab5be80c58e98140dfc69df107f253d7fdd714b30a714260a
docker[38944]: Status: Image is up to date for k8s-gcrio.azureedge.net/hyperkube-amd64:v1.10.3
docker[38959]: /usr/bin/docker: Error response from daemon: shim error: fork/exec /usr/bin/nvidia-container-runtime: no such file or directory.
systemd[1]: hyperkube-extract.service: Control process exited, code=exited status=127
systemd[1]: Failed to start kubectl and kubelet extraction.
systemd[1]: hyperkube-extract.service: Unit entered failed state.
systemd[1]: hyperkube-extract.service: Failed with result 'exit-code'.
```

We observe that the CPU-only nodes get the same docker-deamon.json as the GPU ones:

```
$ cat /etc/docker/daemon.json
{
  "live-restore": true,
  "log-driver": "json-file",
  "log-opts":  {
     "max-size": "50m",
     "max-file": "5"
  }
  ,"default-runtime": "nvidia",
  "runtimes": {
     "nvidia": {
         "path": "/usr/bin/nvidia-container-runtime",
         "runtimeArgs": []
    }
  }
}
```
This fails because the the regular VM with CPU only, do not have the nvidia container runtime.

A working cluster. Different docker-daemon depending on the VM type.

Follow instructions in the blog post above, deploying a cluster with 2 agent pools with both GPU and CPU.

We tried with an older version of the acs-engine, 0.9.4. We used this version by correlating the date of the blog post (nov 2017) with releases of the acs-engine. This older acs-engine, does work correctly.

The previous version of the acs-engine (0.17.1) also works correctly. So this must be a regression in the most recent release of acs-engine. 

We think the refactoring to NVIDIA device-plugins may be related (#2545)

cc @johnhofman

johnhofman:
Allows deployment of a cluster with cpu and gpu agent pools with the nvidia-device-plugin enabled. Without this the docker daemon on the cpu nodes fails to start.

**Which issue this PR fixes** *
This fixes #3190 by only making the changes to `/etc/docker/daemon.json` on gpu enabled nodes using the same template switch which enables/disables the installation of the GPU drivers.

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -34,7 +34,7 @@ write_files:
       "log-opts":  {
          "max-size": "50m",
          "max-file": "5"
-      }{{if IsNVIDIADevicePluginEnabled}}
+      }{{if and (IsNSeriesSKU .) (IsNVIDIADevicePluginEnabled)}}
       ,"default-runtime": "nvidia",
       "runtimes": {
          "nvidia": {


sozercan:
Thanks @johnhofman!

#3181 also addresses this issue and adds toleration for device plugin so it doesn't get scheduled on CPU nodes. I'll close this in favor of #3181 unless you see anything else missing?

PR #3196, opened 2018-06-06
When UseMasterCustomImage or UseAgentCustomImage is set, ensure that …

jim-minter:
…the OS

disk is Premium_LRS (SSD-backed).

It is much easier to ensure this in the acs-engine template rather than in the
image itself, because the `az` command-line tooling does not allow the image
disk SLA to be specified (even though it can be done in the API).

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmss.t
+++ b/parts/k8s/kubernetesagentresourcesvmss.t
@@ -125,6 +125,11 @@
           "osDisk": {
             "createOption": "FromImage",
             "caching": "ReadWrite"
+            {{if UseAgentCustomImage .}}
+            ,"managedDisk": {
+              "storageAccountType": "Premium_LRS"
+            }
+            {{end}}
           {{if ne .OSDiskSizeGB 0}}
             ,"diskSizeGB": {{.OSDiskSizeGB}}
           {{end}}
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -232,8 +232,13 @@
             {{end}}
           },
           "osDisk": {
-            "createOption": "FromImage"
-            ,"caching": "ReadWrite"
+            "createOption": "FromImage",
+            "caching": "ReadWrite"
+            {{if UseAgentCustomImage .}}
+            ,"managedDisk": {
+              "storageAccountType": "Premium_LRS"
+            }
+            {{end}}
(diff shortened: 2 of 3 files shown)


jim-minter:
@kargakis @pweil- fyi

jim-minter:
https://github.com/openshift/openshift-ansible/pull/8658 is a similar part of the same work

jim-minter:
n.b. it appears to be the case that when provisioning from a marketplace image, premium os disks are used automatically.

parts/k8s/kubernetesagentresourcesvmss.t
```diff
@@ -125,6 +125,11 @@
           "osDisk": {
             "createOption": "FromImage",
             "caching": "ReadWrite"
+            {{if UseAgentCustomImage .}}
```
kargakis: Unfortunately this won't help cut down test flakes since tests use the marketplace image.

jim-minter: see above - afaics, when provisioning from a marketplace image, premium os disks are used automatically.

kargakis: Interesting - you are right but then on the machine when I try to verify whether this is an SSD, I am not seeing what I expect.
```
[root@ocp-master-38049041-0 ~]# cat /sys/block/sda/queue/rotational 
1
[root@ocp-master-38049041-0 ~]# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
fd0      2:0    1    4K  0 disk 
sda      8:0    0   32G  0 disk 
├─sda1   8:1    0   10G  0 part /
└─sda2   8:2    0   22G  0 part /var
sdb      8:16   0   32G  0 disk 
└─sdb1   8:17   0   32G  0 part /mnt/resource
sr0     11:0    1  628K  0 rom  
[root@ocp-master-38049041-0 ~]# uname -r
3.10.0-862.el7.x86_64
```
https://unix.stackexchange.com/a/65602

Is there a different way to figure this out? It's very unfortunate if we already run it :(

kargakis: ssd
:'(

jim-minter:
@pweil- I see your point.  I think it depends on @jackfrancis - Jack: would you accept this without the additional `{{if}}`s?

jackfrancis:
Sorry I missed this, is this PR still current?

PR #3197, opened 2018-06-06
adding nil check for managed cluster and its properties object

shrutir25:
Checks for managed cluster object and properties object to make sure it is not nil

Fixes nil panics seen here https://jarvis-west.dc.ad.msft.net/C7CCDED7

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


pkg/api/apiloader.go
```diff
@@ -200,6 +200,10 @@ func (a *Apiloader) LoadContainerServiceForAgentPoolOnlyCluster(
 		if e := json.Unmarshal(contents, &managedCluster); e != nil {
 			return nil, IsSSHAutoGenerated, e
 		}
+		// verify managedCluster
+		if managedCluster == nil || managedCluster.Properties == nil {
+			return nil, false, a.Translator.Errorf("properties object in managed cluster should not be nil")
```
amanohar: @shrutir25 actually the properties can be nil and is a supported ARM scenario. RPs need to be able to handle this.

amanohar: Here is the documentation: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/resource-api-reference.md#put-resource -


```
properties | Optional. Format not defined by ARM. Settings used to provision or configure the resource. All Resource Provider specific properties must be placed within this. The order of parameters in the request is unspecified. RPs should not rely on any particular ordering.
-- | --
```

amanohar: Here is the documentation: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/resource-api-reference.md#put-resource -

```
properties | Optional. Format not defined by ARM. Settings used to provision or configure the resource. All Resource Provider specific properties must be placed within this. The order of parameters in the request is unspecified. RPs should not rely on any particular ordering.
-- | -
```

shrutir25: @amanohar - so then you are suggesting that we make this check in RP before calling this function in acs-engine right ?

amanohar: @shrutir25 I meant RP has to support a payload with nil `properties` and PUT should not fail.

weinong: I talked to Gaurav to get clarification. So long as the fields inside `Properties` are required, we can return 400 if it's not supplied.

amanohar: In case of update correct behavior is to accept `properties`=nil as user is not required to supply `required` properties during update.

amanohar:
@shrutir25 Couple of suggestions:
1)  I recommend using `isUpdate` param to make a determination of what kind of error to return. If it's an existing MC and nil then create `managedCluster.Properties` and then merge should happen.
2) Fix this for both (all) MC versions and not just `2018-03-31`.

pkg/api/apiloader.go
```diff
@@ -200,6 +200,10 @@ func (a *Apiloader) LoadContainerServiceForAgentPoolOnlyCluster(
 		if e := json.Unmarshal(contents, &managedCluster); e != nil {
 			return nil, IsSSHAutoGenerated, e
 		}
+		// verify managedCluster.Properties is not nil for creating case
+		if (managedCluster == nil || managedCluster.Properties == nil) && !hasExistingCS {
```
amanohar: `Properties` is a pointer. In update case if customer didn't set `Properties` would it be nil? Will that pointer need initialization in update case?

amanohar: Also add a test for this in: `apiloader_test.go`

weinong: in update case, wouldn't nil `managedCluster.Properties` cause panic in line 208?

PR #3202, opened 2018-06-06
openshift: fix retry interval in rollout e2e test

kargakis:
Fixes https://github.com/Azure/acs-engine/issues/2952 (hopefully)

@jackfrancis @CecileRobertMichon please merge; potential flake fix


kargakis:
@mjudeikis ptal

mjudeikis:
/lgtm 
I wonder is it worth parameterise  it? Or its one time value?

jim-minter:
@kargakis you don't think 200ms is still pretty short?  1s?

kargakis:
The interval is doubled after each retry and we retry 10 times so I think
starting off with 200ms is reasonable.

jim-minter:
ah, cool, didn't realise it doubled.

PR #3198, opened 2018-06-06
Azure CNI networkmonitor addon

jackfrancis:
**What this PR does / why we need it**: Installs an Azure CNI "networkmonitor" addon whenever Azure CNI is present

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
Azure CNI networkmonitor addon
```

PR #3205, opened 2018-06-06
Add Azure-npm to provide native k8s network policy support

saiyan86:

This PR provides native k8s policy support. 
Azure-npm is a Microsoft Azure's fault-tolerant and scalable implementation of [Kubernetes Network Policy plugin](https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/).
Azure-npm supports all [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) specified by current version of Kubernetes, and will continue to support future versions.

Azure-npm only supports [Kubernetes v1.8](https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/) or later, as it requires NetworkPolicy API v1beta1.


pkg/acsengine/const.go
```diff
@@ -50,6 +50,8 @@ const (
 	NetworkPolicyCalico = "calico"
 	// NetworkPolicyCilium is the string expression for cilium network policy config option
 	NetworkPolicyCilium = "cilium"
+	// NetworkPolicyAzure is the string expression for cilium network policy config option
```
CecileRobertMichon: copy paste typo in the comment, should say Azure CNI, not cilium

pkg/acsengine/const.go
```diff
@@ -50,6 +50,8 @@ const (
...
+	NetworkPolicyAzure = "azure"
 	// NetworkPluginAzure is the string expression for Azure CNI network policy
```
CecileRobertMichon: This should probably be `network plugin`  instead of `network policy` since you are adding one for network policy. Or maybe we could reuse the same constant for both? Your call.

pkg/acsengine/defaults.go
```diff
@@ -384,7 +384,7 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 		switch o.KubernetesConfig.NetworkPolicy {
 		case NetworkPluginAzure:
 			o.KubernetesConfig.NetworkPlugin = NetworkPluginAzure
-			o.KubernetesConfig.NetworkPolicy = DefaultNetworkPolicy
+			o.KubernetesConfig.NetworkPolicy = NetworkPolicyAzure
```
CecileRobertMichon: What is the reason for changing the default here? This might break backwards compatibility so we'll have to do some extensive testing if this is something we want. cc: @jackfrancis

pkg/api/convertertoapi.go
```diff
@@ -718,10 +718,7 @@ func setVlabsKubernetesDefaults(vp *vlabs.Properties, api *OrchestratorProfile)
 
 	if vp.OrchestratorProfile.KubernetesConfig != nil {
 		// Included here for backwards compatibility with deprecated NetworkPolicy usage patterns
-		if vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy == NetworkPolicyAzure {
-			api.KubernetesConfig.NetworkPlugin = vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy
-			api.KubernetesConfig.NetworkPolicy = "" // no-op but included for emphasis
-		} else if vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy == NetworkPolicyNone {
+		if vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy == NetworkPolicyNone {
```
CecileRobertMichon: Could you explain why this is needed?

pkg/acsengine/const.go
saiyan86: @jackfrancis separated NetworkPolicy & NetworkPlugin. So basically NetworkPlugin means CNI and NetworkPolicy means network policy plugin. These are two different things.

pkg/api/convertertoapi.go
saiyan86: For setting default NetworkPolicy to azure

pkg/acsengine/defaults.go
saiyan86: Previously we don't have network policy support in Azure CNI, now we just released it. Therefore we want to set default NetworkPolicy to azure.

pkg/acsengine/const.go
CecileRobertMichon: Yes I meant the comment should reflect that, right it's `// NetworkPluginAzure is the string expression for Azure CNI network policy` which is confusing

pkg/acsengine/defaults_test.go
```diff
@@ -430,10 +430,6 @@ func TestNetworkPolicyDefaults(t *testing.T) {
 		t.Fatalf("NetworkPlugin did not have the expected value, got %s, expected %s",
 			properties.OrchestratorProfile.KubernetesConfig.NetworkPlugin, "azure")
 	}
-	if properties.OrchestratorProfile.KubernetesConfig.NetworkPolicy != "" {
```
CecileRobertMichon: Instead of removing a unit test I think you should update it to the actual expected value that way we don't reduce coverage.

pkg/api/convertertoapi.go
CecileRobertMichon: I don't understand how this is setting the default NetworkPolicy to azure? What I'm reading is "if NetworkPolicy is none. set NetworkPolicy to "" and NetworkPlugin to kubenet". What am I missing?

CecileRobertMichon:
`Azure-npm only supports Kubernetes v1.8 or later, as it requires NetworkPolicy API v1beta1.` does this mean we should add some validation on the k8s version? Right now it looks like it is being enabled by default, even if I specify k8s 1.7.x

saiyan86:
@CecileRobertMichon You're right. We could overwrite the NetworkPolicy to DefaultNetworkPolicy for k8s < 1.8.

pkg/api/convertertoapi.go
```diff
@@ -718,12 +718,15 @@ func setVlabsKubernetesDefaults(vp *vlabs.Properties, api *OrchestratorProfile)
...
 			api.KubernetesConfig.NetworkPlugin = NetworkPluginKubenet
 			api.KubernetesConfig.NetworkPolicy = "" // no-op but included for emphasis
+		} else if vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy == NetworkPolicyAzure {
```
CecileRobertMichon: I think this should happen in defaults.go since it is modifying the default behavior

CecileRobertMichon:
How does this PR relate to #3198 ?

saiyan86:
@CecileRobertMichon PR #3198 is Azure Container Network Monitering Service, which is used for container networking telemetry purposes. This PR adds native k8s network policy support.

pkg/acsengine/defaults.go
```diff
@@ -394,6 +394,12 @@ func setOrchestratorDefaults(cs *api.ContainerService) {
 			o.KubernetesConfig.NetworkPlugin = NetworkPolicyCilium
 		}
 
+		if o.KubernetesConfig.NetworkPolicy == NetworkPolicyAzure {
+			if !common.IsKubernetesVersionGe(k8sVersion, "1.8.0") {
```
jackfrancis: Let's move this < 1.8.0 logic up into the above switch/case statement, as it is executing against an identical condition (NetworkPolicy  == "azure")

pkg/api/convertertoapi.go
jackfrancis: These changes need to be removed from this PR. The entire reason for this code (as described in the comments) is to maintain backwards-compatibility with prior usage patterns. This code is not meant to accommodate ongoing usage pattern changes/additions.

CecileRobertMichon:
What is the expected default behavior for network policy after this PR if a user does not pass in any netork plugin/policy value? ie. with the default kubernetes example apimodel

We need to update the relevant docs (especially since the default behavior isn't straightforward by looking at the code because of all the back compat cruft)

jackfrancis:
This change should have no impact on the current expected default configuration outcome. From a high level, this PR:

- introduces a new allowd `networkPlugin/networkPolicy` combination of "azure"/"azure"
- The above, new, configuration installs the Azure NetworkPolicy daemonset via the addons vector onto their cluster

That should be it. I believe our unit tests should ensure no regressions in our default configuration cases.

marrobi:
@saiyan86 Great to see this progressing. Where do we report issues with azure-npm? Is there a repo available? Certain policies do not seem to function correctly, for example try https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/04-deny-traffic-from-other-namespaces.md . This does not seem to work with azure-npm, but does with kube-router. Policies using pod selectors do seem to work.

@jackfrancis @CecileRobertMichon are you aware of a project that tests various network policies? 

Thanks.

sharmasushant:
@marrobi Thank you for trying it out. Please file the issue @ https://github.com/Azure/azure-container-networking/issues

jackfrancis:
@marrobi I'm not! We have a very basic NetworkPolicy validation in our E2E tests, would love to introduce something more thorough if you discover a comprehensive lib/project.

saiyan86:
@jackfrancis I will add some tests soon that should work for both Calico and us.

PR #2521, opened 2018-03-24
Update Calico to 3.1

dtzar:
**What this PR does / why we need it**:
Updates Calico from 2.6 to the latest 3.0 which enables [additional functionality](https://www.projectcalico.org/releasethecalico/). 

**Which issue this PR fixes** 
This lays the foundation for the requested IPAM support via these issues 
https://github.com/Azure/acs-engine/issues/2227
https://github.com/Azure/acs-engine/issues/1930

Potentially fixes #1621
fixes #2202

**If applicable**:
- [x] documentation
- [ ] network policy e2e test(s)
- [x] tested backward compatibility (ie. deploy with the previous version, upgrade with this branch)

```release-note
Updates Calico from 2.6 to the latest 3.0
```

PR #3208, opened 2018-06-07
Document required manual calico 2.6.3 -> calico 3.1.1 upgrade when upgrading from < 0.17.0-provisioned clusters

### LINKED ISSUE #3191: Calico (2.6.3 -> 3.1.1) upgrade fails between acs-engine 0.16.0 and 0.18.1
No
---

ISSUE
---

0.16.0 and 0.18.1
---

kubernetes 1.9.7 and 1.10.3
acs-engine 0.16.0 and 0.18.1

**What happened**:
* I deployed a cluster with 3 masters and 3 nodes on kubernetes 1.9.7 via acs-engine 0.16.0 to simulate an upgrade scenario on a production cluster we run.

* I then used acs-engine 0.18.1 to perform an upgrade operation on it, targeting kubernetes 1.10.3. During this process I also expected an upgrade of calico from 2.6.3 to 3.1.1 due to #2521 (which is one major reason for performing this upgrade)

* After an hour of waiting for VMs to delete/create, the cluster was successfully up, but still running calico 2.6.3.

* The old daemonset created by 0.16.0 deployment has an update strategy of OnDelete (new one has rollingupdate), so pods needed to be killed, but the daemonset actually running in the cluster hadn't been updated anyway, so I deleted the daemonset itself and watched the new one come up.

* These new calico 3.1.1 nodes came up with the following issue:

```
2018-06-06 11:49:31.577 [INFO][9] startup.go 99: Datastore is ready
2018-06-06 11:49:31.586 [INFO][9] customresource.go 217: Error getting resource Key=ClusterInformation(default) Name="default" Resource="ClusterInformations" Revision="" error=clusterinformations.crd.projectcalico.org "default" not found
2018-06-06 11:49:31.592 [INFO][9] migrate.go 811: cannot migrate from version v2.6.3: migration to v3 requires a tagged release of Calico v2.6.5+
2018-06-06 11:49:31.592 [ERROR][9] migrate.go 720: Unable to migrate data from version 'v2.6.3': migration to v3 requires a tagged release of Calico v2.6.5+
2018-06-06 11:49:31.592 [ERROR][9] startup.go 106: Unable to ensure datastore is migrated. error=unable to migrate data from version 'v2.6.3': migration to v3 requires a tagged release of Calico v2.6.5+
2018-06-06 11:49:31.593 [WARNING][9] startup.go 1058: Terminating
Calico node failed to start
```

I then reapplied the /etc/kubernetes/addons/calico-node.yaml manifest after changing node tag from v3.1.1 to v2.6.10, and cni from v3.1.1 to v2.0.6, and calico came up nicely. I then applied again with node tag v3.1.1, cni tag v3.1.1, watched it successfully migrate, and now have a functional 1.10.3 cluster with calico 3.1.1.

Calico version 3.1.1 up and running upon cluster upgrade completion.

* Deploy cluster with acs-engine 0.16.0 and calico enabled
* Upgrade cluster with acs-engine 0.18.1
* Observe calico-node version, observe complaints from calico-node when actually running 3.1.1 due to attempt to migrate from < 2.6.5 to 3.1.1.

I was a bit unsure about submitting this issue - is this whole process something we expect acs-engine to handle, or are the manual steps I had to go through expected? Could acs-engine somehow handle a multi-step process of first migrating to the most recent 2.6.x before moving to 3.1.x?

oivindoh:
Upgrading from clusters deployed by acs-engine < 0.17.0 and calico enabled had calico 2.6.3. When upgrading such a cluster with 0.17.0 and later, calico addon manifest is 3.1.x, and a migration is not supported for releases prior to 2.6.5, so we need to perform some manual steps to get up and running on 3.1.x. See Issue #3191 

fixes #3191 

I'm not entirely certain where this belongs (whether in examples/networkpolicy or examples/k8s-upgrade.

**If applicable**:
- [x] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/examples/networkpolicy/README.md
+++ b/examples/networkpolicy/README.md
@@ -31,3 +31,22 @@ To understand how to deploy this template, please read the baseline [Kubernetes]
 Once the template has been successfully deployed, following the [simple policy tutorial](https://docs.projectcalico.org/v3.0/getting-started/kubernetes/tutorials/simple-policy) or the [advanced policy tutorial](https://docs.projectcalico.org/v3.0/getting-started/kubernetes/tutorials/advanced-policy) will help to understand calico networking.
 
 > Note: `ping` (ICMP) traffic is blocked on the cluster by default.  Wherever `ping` is used in any tutorial substitute testing access with something like `wget -q --timeout=5 google.com -O -` instead.
+
+### Update guidance for clusters deployed by acs-engine releases prior to 0.17.0
+Clusters deployed with calico networkPolicy enabled prior to `0.17.0` had calico `2.6.3` deployed, and a daemonset with an `updateStrategy` of `Ondelete`.
+
+acs-engine releases starting with 0.17.0 now produce an addon manifest for calico in `/etc/kubernetes/addons/calico-daemonset.yaml` contaning calico 3.1.x, and an `updateStrategy` of `RollingUpdate`.
+
+To get up and running with the new version of calico, follow these steps:
+
+1) Edit `/etc/kubernetes/addons/calico-daemonset.yaml` on a master node, replacing `calico/node:v3.1.1` with `calico/node:v2.6.10` and `calico/cni:v3.1.1` with `calico/cni:v2.0.6`. Run `kubectl apply -f /etc/kubernetes/addons/calico-daemonset.yaml`.
+
+   We're performing this step because calico does not support migrating its datastore from v2 to v3 when coming from releases prior to `2.6.5`. `2.6.10` is chosen as a migration stepping stone because it is the most recent 2.6.x patch available.
+
+2) Confirm that all the pods in the daemonset get rotated and come up healthy, reporting ready status in logs.
+
+3) Edit `/etc/kubernetes/addons/calico-daemonset.yaml` on the master node again, replacing `calico/node:v2.6.10` with `calico/node:v3.1.1` and `calico/cni:v2.0.6` with `calico/cni:v3.1.1`. Run `kubectl apply -f /etc/kubernetes/addons/calico-daemonset.yaml`.
+
+    Propagate this updated manifest to any other master nodes in the cluster.
+
+4) Confirm that all the pods in the daemonset get rotated and come up healthy after finishing tasks logged by migrate.go


examples/networkpolicy/README.md
```diff
@@ -31,3 +31,32 @@ To understand how to deploy this template, please read the baseline [Kubernetes]
 Once the template has been successfully deployed, following the [simple policy tutorial](https://docs.projectcalico.org/v3.0/getting-started/kubernetes/tutorials/simple-policy) or the [advanced policy tutorial](https://docs.projectcalico.org/v3.0/getting-started/kubernetes/tutorials/advanced-policy) will help to understand calico networking.
 
 > Note: `ping` (ICMP) traffic is blocked on the cluster by default.  Wherever `ping` is used in any tutorial substitute testing access with something like `wget -q --timeout=5 google.com -O -` instead.
+
+### Update guidance for clusters deployed by acs-engine releases prior to 0.17.0
+Clusters deployed with calico networkPolicy enabled prior to `0.17.0` had calico `2.6.3` deployed, and a daemonset with an `updateStrategy` of `Ondelete`.
+
+acs-engine releases starting with 0.17.0 now produce an addon manifest for calico in `/etc/kubernetes/addons/calico-daemonset.yaml` contaning calico 3.1.x, and an `updateStrategy` of `RollingUpdate`.
+
+To get up and running with the new version of calico after upgrading a cluster with acs-engine `0.17.0` and up, follow these steps:
```
CecileRobertMichon: Thank you @oivindoh! Looks great. Could you please add something like "As per the Calico v0.3.1 release notes, Calico v0.3.x includes breaking changes:
> Some highlights include:
> You must upgrade to Calico v2.6.5 before you can upgrade to v3.0.1 (see https://docs.projectcalico.org/v3.0/getting-started/kubernetes/upgrade/) 
> Calico deployments that access the etcd datastore directly must complete
a one-time migration.
> You must convert any customized Calico manifests via calicoctl convert
before you can use them with v3.0.1.

Here some instructions to get up and running with the new version of Calico […]" here?

CecileRobertMichon: So people can have context on the breaking changes (and know it's a calico breaking change, not acs-engine)

oivindoh: Good point! I added some text to be more explicit about this being due to calico breaking changes and not acs-engine 👍

oivindoh:
@dtzar could you sanity-check this?

dtzar:
@oivindoh Thanks for the docs.  Although I'm sure the steps you list *work*, this does not follow the [upgrade guidance provided by Calico](https://docs.projectcalico.org/v3.0/getting-started/kubernetes/upgrade/upgrade#upgrading-a-self-hosted-installation-that-uses-the-kubernetes-api-datastore).  I know there are significant changes to the RBAC config as well as the manifest - see the changes in [my PR which did the upgrade ](https://github.com/Azure/acs-engine/pull/2521/files) to get an idea.  

Soo... I wouldn't recommend using this guidance unless it maps directly to what is on Calico's webpage.

oivindoh:
@dtzar I guess I'm not entirely clear on where I diverge from the linked upgrade guidance - applying the 3.x manifest with node/cni changed to 2.6.10 and 2.0.6 effectively performs step 1, and handles upgrade to 2.6.5+ in the process. Applying that manifest again with node/cni 3.1.1 performs the rest of the steps required, keeping cluster-cidr as appropriate.

What I wasn't clear on after reading [Calico docs](https://docs.projectcalico.org/v3.1/getting-started/kubernetes/upgrade/upgrade#upgrading-an-installation-that-uses-the-kubernetes-api-datastore) (and what I wanted to document here) was the *how* of upgrading to 2.6.5+, given that I now had a cluster with a 3.1.1 manifest to be managed as an addon and calico 2.6.3 actually running in the cluster.

examples/networkpolicy/README.md
```diff
@@ -35,29 +35,25 @@ Once the template has been successfully deployed, following the [simple policy t
...
+acs-engine releases starting with 0.17.0 now produce an addon manifest for calico in `/etc/kubernetes/addons/calico-daemonset.yaml` contaning calico 3.1.x, and an `updateStrategy` of `RollingUpdate`. Due to breaking changes introduced by calico 3, one must first migrate through calico `2.6.5` or a later 2.6.x release in order to migrate to calico 3.1.x. as described in the [calico kubernetes upgrade documentation](https://docs.projectcalico.org/v3.1/getting-started/kubernetes/upgrade/). The acs-engine manifest for calico uses the [kubernetes API datastore, policy-only setup](https://docs.projectcalico.org/v3.1/getting-started/kubernetes/upgrade/upgrade#upgrading-an-installation-that-uses-the-kubernetes-api-datastore).
 
-To get up and running with the new version of calico after upgrading a cluster with acs-engine `0.17.0` and up, follow these steps:
+1. To update to `2.6.5+` in preparation of an upgrade to 3.1.x as specified, edit `/etc/kubernetes/addons/calico-daemonset.yaml` on a master node, replacing `calico/node:v3.1.1` with `calico/node:v2.6.10` and `calico/cni:v3.1.1` with `calico/cni:v2.0.6`. Run `kubectl apply -f /etc/kubernetes/addons/calico-daemonset.yaml`.
```
dtzar: The place where you diverge from the upgrade guidance is here because you are only replacing the image versions, not the entire template/manifest.  The `calico-daemonset.yaml` is a merge of the Calico RBAC + manifest files and needs more changes than just bumping the image versions.

examples/networkpolicy/README.md
```diff
@@ -35,29 +35,25 @@ Once the template has been successfully deployed, following the [simple policy t
...
-
-   We're performing this step because calico does not support migrating its datastore from v1 to v3 when coming from releases prior to `2.6.5`. `2.6.10` is chosen as a migration stepping stone because it is the most recent 2.6.x patch available.
-
-2) Wait until all the pods in the daemonset get rotated and come up up-to-date, healthy and ready:
+    Wait until all the pods in the daemonset get rotated and come up up-to-date, healthy and ready:
 
     `YYYY-MM-DD HH:MM:SS.FFF [INFO][n] health.go 150: Overall health summary=&health.HealthReport{Live:true, Ready:true}`
 
-3) Edit `/etc/kubernetes/addons/calico-daemonset.yaml` on the master node again, replacing `calico/node:v2.6.10` with `calico/node:v3.1.1` and `calico/cni:v2.0.6` with `calico/cni:v3.1.1`. Run `kubectl apply -f /etc/kubernetes/addons/calico-daemonset.yaml`.
+2. To complete the upgrade to 3.1.x, edit `/etc/kubernetes/addons/calico-daemonset.yaml` on the master node again, replacing `calico/node:v2.6.10` with `calico/node:v3.1.1` and `calico/cni:v2.0.6` with `calico/cni:v3.1.1`. Run `kubectl apply -f /etc/kubernetes/addons/calico-daemonset.yaml`.
```
dtzar: Same thing here.

examples/networkpolicy/README.md
dtzar: It would be easiest if you reference the updated calico-daemonset file already created (per the link above) and then you'd only really need to replace the values which get auto-generated during ACS-Engine runtime for the CIDR.. look for the `<...>` in the template.

dtzar:
@oivindoh I read through the document again end-to-end and I realize what you're doing now.  The missing explanation is that I think you're saying to push through with the actual acs-engine upgrade (which will break Calico) and then go into each master node and do the following steps, correct?  

If so, then yes the only thing I'm unsure of is if there are any potential problems with the temporary upgrade step you have of 2.6.10 using a 3.x RBAC/Policy template.

oivindoh:
@dtzar Exactly - however upgrading the cluster shouldn’t break calico because it will happily continue to run using its old 2.6.3 definitions until the moment you apply the add on manifest manually, since it’s not set to anything but ensureexists, effectively keeping addon manager off the table.

I should update the document and remove the propagate to each master comment, since the masters will already have the updated manifest anyway (we have opted for 3.1.3 instead of the supplied 3.1.1 for richer network policy support, so needed to propagate).

I didn’t do any form of extended validation after applying 2.6.10 other the quick manual tests to see or services were still available and no alerts were triggering before updating to 3.x, so I can’t really vouch for zero negative effects during that step, but I could not detect anything FWIW.

3.1.1 and 3.1.1 are definitely chugging along happily after the upgrade process.

dtzar:
@oivindoh - I wouldn't recommend using the new yaml manifest with 2.6.10 images since it does have significant changes.  Would you be willing to test out this flow and update the document?
1. Update the daemonset yaml (pre-acs-engine-upgrade) to 2.6.10 with the steps as you described.
2. Upgrade / deploy to newer 3.x calico using acs-engine (it should just work at this point...)

Also - I just issued https://github.com/Azure/acs-engine/pull/3257 which upgrades to the latest 3.1.3 so you can update your document accordingly and not have to do another step.

PR #3209, opened 2018-06-07
add default reg dns

mjudeikis:
If we redeploy registry for any reasons and recreate service it will cause all crazy things.
Good practice is to have DNS instead of IP. 

ptal @kargakis @jim-minter


jim-minter:
great catch, thanks @mjudeikis !

kargakis:
@mjudeikis can you also add this in the release-3.9 branch so it gets tested by the e2e test?

jim-minter:
@mjudeikis is out, I've pushed an additional commit onto his branch.
/lgtm

kargakis:
/lgtm

too

PR #2859, opened 2018-05-04
Scaling Windows agent pools for Availability Sets and VMSS

danigian:
**What this PR does / why we need it**:
This code implements the scaling for Windows agent pools (both for Availability Sets and Virtual Machine Scale Sets)

**Which issue this PR fixes**: 
fixes #1972 

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:
```release-note
```

PR #3210, opened 2018-06-07
fix scale command on OpenShift

jim-minter:


jim-minter:
https://github.com/Azure/acs-engine/pull/2859 broke this - we should ideally get some scale testing included in the acs-engine tests. https://github.com/Azure/acs-engine/issues/3211.

jim-minter:
@pweil- fyi :embarrassed:

cmd/scale.go
```diff
@@ -229,8 +229,8 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				continue
 			}
 
-			osPublisher := *vm.StorageProfile.ImageReference.Publisher
-			if strings.EqualFold(osPublisher, "MicrosoftWindowsServer") {
+			osPublisher := vm.StorageProfile.ImageReference.Publisher
+			if osPublisher != nil && vm.StorageProfile.ImageReference != nil && strings.EqualFold(*osPublisher, "MicrosoftWindowsServer") {
```
CecileRobertMichon: Shouldn't the check for `vm.StorageProfile.ImageReference != nil` be done before assigning `osPublisher := vm.StorageProfile.ImageReference.Publisher ` here?

jim-minter: thanks - typo - fixed

cmd/scale.go
```diff
@@ -229,8 +229,8 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				continue
 			}
 
-			osPublisher := *vm.StorageProfile.ImageReference.Publisher
-			if strings.EqualFold(osPublisher, "MicrosoftWindowsServer") {
+			osPublisher := vm.StorageProfile.ImageReference.Publisher
```
CecileRobertMichon: Won't this be a nil pointer dereference if StorageProfile and/or ImageReference are nil?

jim-minter: It would, although OTOH these have just been returned by an Azure API query for a running VM therefore I was supposing this scenario would be improbable.

CecileRobertMichon: it would be possible if the VM returned doesn't have those fields, I thought that was the intent of this PR? Do OpenShift VMs have and ImageReference but no osPublisher?

jim-minter: Correct, VMs that are not provisioned by the marketplace have vm.StorageProfile.ImageReference.ID set, but vm.StorageProfile.ImageReference.Publisher is nil.

CecileRobertMichon: ok great, thanks for clarifying

Issue #2895, opened 2018-05-09
openshift: infra pods on compute nodes


jim-minter:
looks like registry console and asb pods x2 are being created on compute nodes - shouldn't they be on infra?

jim-minter:
@maxamillion ptal

maxamillion:
@jim-minter I'm honestly not sure which is considered "best practice", I'd defer to @kwoodson

kwoodson:
@jim-minter @maxamillion, we want our infra nodes running the cluster infrastructure (routing, registry, console, etc).  This is for a few reasons

- security (isolation)
- resource protection (ram, cpu guarantees)
- fronted by an elb for flexibility, scalability, reliability

maxamillion:
Alright, so the current situation is this:

* Registry Console is deployed via openshift-ansible's `roles/cockpit-ui` role, that uses the `registry-console` template as deployed via the openshift-ansible `roles/openshift_hosted_templates` role and that template does not have a `nodeSelector` field in it as of the openshift-ansible 3.9.29 release. I'm unsure of how to resolve this in the best way because I'm unfamiliar with the implications of adding that field to the generic template that is destined for the Service Catalog.

* The Ansible Service Broker itself is more interesting because it *should* have a `nodeSelector` as it is defined in the `oc_obj` inside the openshift-ansible playbook which is set to [ansible_service_broker_node_selector](https://github.com/openshift/openshift-ansible/blob/openshift-ansible-3.9.29-1/roles/ansible_service_broker/tasks/install.yml#L246) and that evaluates by default to [region=infra](https://github.com/openshift/openshift-ansible/blob/openshift-ansible-3.9.29-1/roles/ansible_service_broker/defaults/main.yml#L3) which should work. However, after deployment the `dc/asb` inside the OpenShift Cluster does not have a `nodeSelector` field at all as per `oc get dc/asb -n openshift-ansible-service-broker -o yaml`. I'm still trying to find a root cause for that one.

jim-minter:
Yes; what is the node selector in the default project in a normal openshift deployment?

maxamillion:
[...]

PR #3229, opened 2018-06-08
Adding test for compute node pod whitelist.

kwoodson:
Adding a test for pods on compute nodes.  The basic gist is that we are verifying that infra node pods do not get deployed onto compute nodes.  We are creating a whitelist of pods allowed on the compute nodes.

fixes: https://github.com/Azure/acs-engine/issues/2895

@jim-minter @kargakis

### DIFF
--- a/test/e2e/kubernetes/pod/pod.go
+++ b/test/e2e/kubernetes/pod/pod.go
@@ -33,15 +33,23 @@ type Pod struct {
 
 // Metadata holds information like name, createdat, labels, and namespace
 type Metadata struct {
-	CreatedAt time.Time         `json:"creationTimestamp"`
-	Labels    map[string]string `json:"labels"`
-	Name      string            `json:"name"`
-	Namespace string            `json:"namespace"`
+	CreatedAt       time.Time         `json:"creationTimestamp"`
+	Labels          map[string]string `json:"labels"`
+	Name            string            `json:"name"`
+	Namespace       string            `json:"namespace"`
+	OwnerReferences []OwnerReferences `json:"ownerReferences"`
+}
+
+// OwnerReferences describes whether pods belong to a daemonset or other deploymenttype
+type OwnerReferences struct {
+	Kind string `json:"kind"`
+	Name string `json:"name"`
 }
 
 // Spec holds information like containers
 type Spec struct {
 	Containers []Container `json:"containers"`
+	Nodename   string      `json:"nodeName"`
 }
 
 // Container holds information like image and ports
@@ -50,6 +58,12 @@ type Container struct {
 	Ports     []Port    `json:"ports"`
 	Env       []EnvVar  `json:"env"`
 	Resources Resources `json:"resources"`
+}
+
+// Condition holds pod condition
+type Condition struct {
+	Status string `json:"status"`
+	Type   string `json:"type"`
 }
 
 // EnvVar holds environment variables
@@ -84,10 +98,11 @@ type Limits struct {
 
 // Status holds information like hostIP and phase
 type Status struct {
-	HostIP    string    `json:"hostIP"`
-	Phase     string    `json:"phase"`
-	PodIP     string    `json:"podIP"`
-	StartTime time.Time `json:"startTime"`
+	HostIP     string      `json:"hostIP"`
+	Phase      string      `json:"phase"`
+	PodIP      string      `json:"podIP"`
+	StartTime  time.Time   `json:"startTime"`
+	Conditions []Condition `json:"conditions"`
 }
 
 // CreatePodFromFile will create a Pod from file with a name
@@ -107,9 +122,17 @@ func CreatePodFromFile(filename, name, namespace string) (*Pod, error) {
 	return pod, nil
 }
 
-// GetAll will return all pods in a given namespace
+// GetAll will return all pods in a given namespace. If namespace is "" will return all pods
 func GetAll(namespace string) (*List, error) {
-	cmd := exec.Command("kubectl", "get", "pods", "-n", namespace, "-o", "json")
+	args := []string{"get", "pods", "-o", "json"}
+
+	if namespace == "" {
+		args = append(args, "--all-namespaces")
+	} else {
+		args = append(args, "-n", namespace)
+	}
+
+	cmd := exec.Command("kubectl", args...)
 	util.PrintCommand(cmd)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
--- a/test/e2e/openshift/openshift_test.go
+++ b/test/e2e/openshift/openshift_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/acs-engine/pkg/api/common"
 	"github.com/Azure/acs-engine/test/e2e/config"
 	"github.com/Azure/acs-engine/test/e2e/engine"
+	"github.com/Azure/acs-engine/test/e2e/kubernetes/namespace"
 	knode "github.com/Azure/acs-engine/test/e2e/kubernetes/node"
 	"github.com/Azure/acs-engine/test/e2e/kubernetes/pod"
 	"github.com/Azure/acs-engine/test/e2e/openshift/node"
@@ -134,10 +135,13 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 	})
 
 	It("should deploy a sample app and access it via a route", func() {
-		err := util.ApplyFromTemplate("nginx-example", "openshift", "default")
+		ns := "e2e-test"
+		_, err := namespace.Create(ns)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(util.WaitForDeploymentConfig("nginx-example", "default")).NotTo(HaveOccurred())
-		host, err := util.GetHost("nginx-example", "default")
+		err = util.ApplyFromTemplate("nginx-example", "openshift", ns)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(util.WaitForDeploymentConfig("nginx-example", ns)).NotTo(HaveOccurred())
+		host, err := util.GetHost("nginx-example", ns)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(util.TestHost(host, 10, 200*time.Millisecond)).NotTo(HaveOccurred())
 	})
@@ -170,5 +174,40 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		running, err := pod.WaitOnReady("asb", "openshift-ansible-service-broker", 3, 30*time.Second, cfg.Timeout)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(running).To(Equal(true))
+	})
+
+	It("should not deploy infrastructure components on compute nodes", func() {
+		namespaces := []string{
+			"openshift-",
+			"kube-",
+			"default",
+		}
+
+		list, err := pod.GetAll("")
+		Expect(err).NotTo(HaveOccurred())
+
+	nextPod:
+		for _, pod := range list.Pods {
+			// skip if we find a pod that is a daemonset
+			for _, ref := range pod.Metadata.OwnerReferences {
+				if ref.Kind == "DaemonSet" {
+					continue nextPod
+				}
+			}
+			// registry-console was landing on compute nodes. It will now land on masters.
+			// Needs to be configurable.  Skip for now.
+			// https://github.com/Azure/acs-engine/issues/2895
+			if strings.HasPrefix(pod.Metadata.Name, "registry-console") {
+				continue
+			}
+
+			if strings.Contains(pod.Spec.Nodename, "compute") {
+				for _, pattern := range namespaces {
+					if strings.HasPrefix(pod.Metadata.Namespace, pattern) {
+						Expect(strings.HasPrefix(pod.Metadata.Namespace, pattern)).To(BeFalse())
+					}
+				}
+			}
+		}
 	})
 })


kargakis:
+1 for blacklisting

kwoodson:
@kargakis, 

@jim-minter and I had a chat about a strategy here.  We wanted to whitelist namespaces as well as pods. 

I've also added a test to check all deployed pods are not in an error state.

Please review.

kargakis:
Sorry got this backwards

+1 for a whitelist

test/e2e/kubernetes/pod/pod.go
```diff
@@ -107,9 +115,17 @@ func CreatePodFromFile(filename, name, namespace string) (*Pod, error) {
 	return pod, nil
 }
 
-// GetAll will return all pods in a given namespace
+// GetAll will return all pods in a given namespace. If namespace is "" will return all pods
 func GetAll(namespace string) (*List, error) {
-	cmd := exec.Command("kubectl", "get", "pods", "-n", namespace, "-o", "json")
+	command := []string{"get", "pods", "-o", "json"}
```
jim-minter: s/command/args/

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+		}
+		list, err := pod.GetAll("")
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, pod := range list.Pods {
+			skip := false
+			// we have a whitelist of namespaces that we will skip
+			for _, pattern := range namespaces {
+				if strings.Contains(pod.Metadata.Namespace, pattern) {
+					skip = true
```
jim-minter: `skip` is ugly - either using Go's `break LABEL` syntax (https://golang.org/ref/spec#Break_statements) or a separate function would be better.

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		}
 	})
 
+	It("should whitelist pods on compute nodes", func() {
```
jim-minter: something like "should deploy all pods onto appropriate nodes" would be clearer?

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+		namespaces := []string{
+			"openshift-infra",
+			"openshift-node",
+			"openshift-metrics",
+			"kube-",
+			"default",
+		}
+
+		whitelist := []string{
```
jim-minter: daemonsetPrefixes

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+					skip = true
+					break
+				}
+			}
+			if skip {
+				skip = false
+				continue
+			}
+
+			found := false
```
jim-minter: move down a line

test/e2e/openshift/openshift_test.go
jim-minter: why are we excluding certain namespaces?

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+	})
+
+	// Verify all pods have initialized, ready, podscheduled are true
+	It("should have all pods in ready state", func() {
+		list, err := pod.GetAll("")
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, pod := range list.Pods {
+			for _, cond := range pod.Status.Conditions {
+				Expect(cond.Status).To(BeTrue())
```
jim-minter: too broad - just assert that Ready is true - leave the others

jim-minter: it might be easier to check .status.phase is Running or Completed than iterate the conditions

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+						break
+					}
+				}
+				Expect(found).To(BeTrue())
+			}
+		}
+	})
+
+	// Verify all pods have initialized, ready, podscheduled are true
+	It("should have all pods in ready state", func() {
```
jim-minter: have you tested this passes against a cluster today? does it catch the asb problem that we've currently got? if so, to get this in, you might need to exclude the asb pods from the test and add a `// TODO: remove this when issue #xyz is closed`

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+				}
+			}
+			if skip {
+				skip = false
+				continue
+			}
+
+			found := false
+			if strings.Contains(pod.Spec.Nodename, "compute") {
+				for _, pattern := range whitelist {
```
jim-minter: think about ensuring that the intent of your test is clear to a 3rd party.  What you're testing is: "no pods should be on a compute node unless they start with a daemonsetPrefix".  Try to use patterns and variable names that make this clear, e.g. perhaps something like "isDaemonsetPod := matchPrefix()... ; assert(isDaemonsetPod)"

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +85,62 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+					if strings.Contains(pod.Metadata.Name, pattern) {
```
jim-minter: strings.HasPrefix

test/e2e/openshift/openshift_test.go
kwoodson: Currently we are only checking `compute` node types.  Should we add master and infra types as well?

test/e2e/kubernetes/pod/pod.go
kwoodson: Updated.

test/e2e/openshift/openshift_test.go
kwoodson: The idea was that any pod that is in the openshift-metrics namespace we want to skip regardless of host.  I see that this might bypass too large a space with the list that I came up with.  I will only skip `openshift-metrics` currently.

test/e2e/openshift/openshift_test.go
kwoodson: This could also include `statefulsets`.  Do you still prefer `daemonsetPrefixes`?

test/e2e/openshift/openshift_test.go
kwoodson: Updated.

test/e2e/openshift/openshift_test.go
kwoodson: This makes sense.  Will update.

test/e2e/openshift/openshift_test.go
kwoodson: done.

test/e2e/openshift/openshift_test.go
kwoodson: There are 3 condition types:
1.  Initialized
2.  Ready
3.  PodScheduled

All _should_ be `true`.  Would you prefer if we only check for `Ready`?

kargakis:
There is a PodReady helper we use elsewhere in tests. Checking for pod
readiness covers initialization and scheduling.

test/e2e/openshift/openshift_test.go
kwoodson: I checked on a cluster with an application in a CrashLoopBackOff  state:
```
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2018-06-05T02:41:43Z
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: 2018-06-12T18:34:06Z
    message: 'containers with unready status: [glass]'
    reason: ContainersNotReady
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: 2018-06-05T02:41:43Z
    status: "True"
    type: PodScheduled
```

The `phase` was set to `Running`.
```
  phase: Running
```

I believe that phase wouldn't be sufficient.

test/e2e/openshift/openshift_test.go
kwoodson: Will run and confirm that we catch this case.

kargakis:
Neither 'ready = true' is sufficient when the pod is crashlooping, that's
why the helper used in the tests checks for readiness 3-4 times in a row.

kwoodson:
@kargakis, does the PodReady helper find the asb pod that is stuck waiting for the image to deploy?  If it does then we can remove those tests.

kargakis:
We already have readiness tests for all infra pods and yes the asb test
gets stuck in 3.10. We should either have your test check only that pods
got scheduled on the correct nodes or collapse with the existing tests if
you also want to check readiness.

kwoodson:
As per @kargakis' advice, I have removed the ready test as those are implemented elsewhere.

I also ended up moving the nginx test to `e2e-test` so that the whitelist testing works.
```
Azure Container Cluster using the OpenShift Orchestrator 
  should deploy all pods onto appropriate nodes
  /go/src/github.com/Azure/acs-engine/test/e2e/openshift/openshift_test.go:89

$ kubectl get pods -o json --all-namespaces
•
------------------------------
```
The nginx test worked as well:
```
Azure Container Cluster using the OpenShift Orchestrator 
  should deploy a sample app and access it via a route
  /go/src/github.com/Azure/acs-engine/test/e2e/openshift/openshift_test.go:175

$ kubectl create namespace e2e-test

$ kubectl get namespace e2e-test -o json

$ oc process nginx-example -n openshift

$ oc apply -n e2e-test -f nginx-example

$ oc rollout status dc/nginx-example -n e2e-test

$ oc get route/nginx-example -n e2e-test -o jsonpath={.spec.host}
•
------------------------------
```

The namespace and pods:
```
[root@ocp-master-94252885-0 ~]# oc get pods -n e2e-test
NAME                    READY     STATUS      RESTARTS   AGE
nginx-example-1-build   0/1       Completed   0          21m
nginx-example-1-wbc98   1/1       Running     0          20m
```

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +86,44 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		}
 	})
 
+	It("should deploy all pods onto appropriate nodes", func() {
+		namespaces := []string{
+			"openshift-metrics",
```
jim-minter: I think you can add prometheus-node-exporter- to daemonsetPrefixes and get rid of openshift-metrics here?
The test should then say:
1) if the pod is from a daemonset, don't care about its node.
2) if the pod's name starts registry-console-, don't care.
3) the pod must be on compute if and only if the pod's namespace does not start with openshift-, kube- and the namespace is not "default"

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +86,44 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+			"e2e-test",
+		}
+
+		daemonsetPrefixes := []string{
+			"sync-",
+			"ovs-",
+			"sdn-",
+			"registry-console", // Remove this line once registry-console adds the proper nodeSelector
```
jim-minter: reference issue

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +86,44 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		}
 	})
 
+	It("should deploy all pods onto appropriate nodes", func() {
+		namespaces := []string{
+			"openshift-metrics",
+			"e2e-test",
+		}
+
+		daemonsetPrefixes := []string{
```
jim-minter: ideally you would get rid of daemonsetPrefixes entirely; use "pod has an OwnerReference of kind DaemonSet" to detect a pod owned by a daemonset.

kwoodson:
@jim-minter, implemented your latest strategy

> 1 if the pod is from a daemonset, don't care about its node.
> 2 if the pod's name starts registry-console-, don't care.
> 3 the pod must be on compute if and only if the pod's namespace does not start with openshift-, kube- > and the namespace is not "default"

```
Azure Container Cluster using the OpenShift Orchestrator 
  should deploy all pods onto appropriate nodes
  /go/src/github.com/Azure/acs-engine/test/e2e/openshift/openshift_test.go:89

$ kubectl get pods -o json --all-namespaces
•
------------------------------
```

```
[root@ocp-master-18458924-0 ~]# oc adm manage-node ocp-compute-18458924-0  --list-pods

Listing matched pods on node: ocp-compute-18458924-0

NAMESPACE           NAME                             READY     STATUS    RESTARTS   AGE
openshift-metrics   prometheus-node-exporter-l6nmj   1/1       Running   0          21m
openshift-node      sync-9m845                       1/1       Running   0          22m
openshift-sdn       ovs-wmss5                        1/1       Running   0          22m
openshift-sdn       sdn-xkxq7                        1/1       Running   0          22m
```

kwoodson:
@jim-minter, please review.

test/e2e/openshift/openshift_test.go
```diff
@@ -85,6 +86,38 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
 		}
 	})
 
+	It("should deploy all pods onto appropriate nodes", func() {
```
kargakis: Move your test at the end of the suite to ensure the pod list is not racing with the deployer pods and you are getting all pods we expect to run.

kargakis: Also I would reword the title of the test to "should not deploy infrastructure components in compute nodes" since you check just that (eg. you don't discriminate between pods that land in master vs infra).

kwoodson: Ok, I will move this to the end of the tests.  Will also rename.

kwoodson:
@kargakis, updated.

kargakis:
/lgtm

@CecileRobertMichon please approve

CecileRobertMichon:
@kwoodson please fix the lint error in the unit tests

kwoodson:
@kargakis, @CecileRobertMichon, Added doc string to clean up linting.

kwoodson:
Rebasing onto master.

kargakis:
@CecileRobertMichon please approve

kwoodson:
@kargakis, @jim-minter,

Have you seen this on our e2e?
```
Error from deployment for openshift-eastus-80970 in resource group openshift-eastus-80970:exit status 1
2018/06/21 20:52:16 Command Output: ERROR: The template deployment 'openshift-eastus-80970' is not valid according to the validation procedure. The tracking id is 'd5713f39-0adb-4318-93be-ded82aab76fe'. See inner errors for details. Please see https://aka.ms/arm-deploy for usage details.
```

kargakis:
Nothing in the activity log .......... Unclear whether we would get more info out of the returned tracking ID if we can't get any activity log from the whole resource group.

test/e2e/openshift/openshift_test.go
```diff
@@ -171,4 +175,36 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+		namespaces := []string{
+			"openshift-",
+			"kube-",
+			"default",
+		}
+
+		list, err := pod.GetAll("")
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, pod := range list.Pods {
```
jim-minter:
```
nextPod:
		for _, pod := range list.Pods {
			// skip if we find a pod that is a daemonset
			for _, ref := range pod.Metadata.OwnerReferences {
				if ref.Kind == "DaemonSet" {
					continue nextPod
				}
			}
```

^ is better because (a) it would handle multiple ownerreferences if they ever happened for some reason, and (b) testing for a non-nil slice does not imply that the slice has an element (nb you can use len > 0 for that since len(nil slice) is defined to be 0)

test/e2e/openshift/openshift_test.go
```diff
@@ -171,4 +175,36 @@ var _ = Describe("Azure Container Cluster using the OpenShift Orchestrator", fun
...
+			// Needs to be configurable.  Skip for now.
+			// https://github.com/Azure/acs-engine/issues/2895
+			if strings.HasPrefix(pod.Metadata.Name, "registry-console") {
+				continue
+			}
+
+			if strings.Contains(pod.Spec.Nodename, "compute") {
+				for _, pattern := range namespaces {
+					if strings.HasPrefix(pod.Metadata.Namespace, pattern) {
+						Expect(false).To(BeTrue())
```
jim-minter: `Expect(strings.HasPrefix(pod.Metadata.Namespace, pattern)).To(BeFalse())`

jim-minter:
@kwoodson 

> Have you seen this on our e2e?

Open a flake issue with link to the job if it recurs.

kargakis:
@kwoodson reopen and rebase if you are planning to wrap this up. Not sure we need it anymore.

/close

PR #3232, opened 2018-06-08
DCOS: added tags to VMs in cluster

dmitsh:
DCOS: added tags to VMs in cluster

### DIFF
--- a/parts/dcos/dcosagentresourcesvmss.t
+++ b/parts/dcos/dcosagentresourcesvmss.t
@@ -102,10 +102,13 @@
       ],
       "tags":
       {
-        "creationSource" : "[concat('acsengine-', variables('{{.Name}}VMNamePrefix'), '-vmss')]"
+        "creationSource" : "[concat('acsengine-', variables('{{.Name}}VMNamePrefix'), 'vmss')]",
+        "orchestratorName": "dcos",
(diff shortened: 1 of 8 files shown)


parts/dcos/dcosagentresourcesvmss.t
```diff
@@ -102,10 +102,13 @@
       ],
       "tags":
       {
-        "creationSource" : "[concat('acsengine-', variables('{{.Name}}VMNamePrefix'), '-vmss')]"
+        "creationSource" : "[concat('acsengine-', variables('{{.Name}}VMNamePrefix'), 'vmss')]",
```
radhikaj: Can we add some tag to say this is a Windows agent. Thanks.

parts/dcos/dcosmastervars.t
```diff
@@ -26,6 +26,7 @@
         {{end}}
       ],
 {{end}}
+    "orchestratorVersion": "{{.OrchestratorProfile.OrchestratorVersion}}",
```
radhikaj: Here in agentWindowsVersion can we tag with whether it is a Windows or a Linux agent?

dmitsh: OS type has its own field

PR #3243, opened 2018-06-09
Add cluster-autoscaler addon documentation

diwakar-s-maurya:
It mentions cluster-autoscaler add-on in cluster definition documentation. The feature is present in acs-engine but it is not documented and because of it users are not able to use it.

**If applicable**:
- [x] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/docs/clusterdefinition.md
+++ b/docs/clusterdefinition.md
@@ -103,7 +103,7 @@ As you can see above, `addons` is an array child property of `kubernetesConfig`.
 }
 ```
 
-More usefully, let's add some custom configuration to both of the above addons:
+More usefully, let's add some custom configuration to the above addons:
 
 ```
 "kubernetesConfig": {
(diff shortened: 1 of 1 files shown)


jackfrancis:
@sozercan Can you confirm that we will want to mention VMSS-only support for this addon?

sozercan:
@jackfrancis yes, it is VMSS only right now. It has a validation against availability sets if you try to generate/deploy.

feiskyer:
The change lgtm.

cluster-autoscaler also works for vmas. @sozercan Would you like to also add vmas support? (in another PR)

docs/clusterdefinition.md
```diff
@@ -131,6 +132,22 @@ More usefully, let's add some custom configuration to both of the above addons:
                   "memoryLimits": "512Mi"
                 }
               ]
+        },
+        {
+            "name": "cluster-autoscaler",
```
CecileRobertMichon: nit: line 106 above says `More usefully, let's add some custom configuration to both of the above addons:` , could you please change that to ``More usefully, let's add some custom configuration to the above addons:` since there are now 3 examples?

CecileRobertMichon:
@jackfrancis we should expand the addons docs to include the missing addons (nvidia-device-plugin, oms, aci-connector, etc.) - in a separate PR

docs/clusterdefinition.md
diwakar-s-maurya: Done.

Issue #3247, opened 2018-06-11
quote base64 strings in the yaml files; use k8s compatible encoding for base64 values


PaulCharlton:
**Is this a request for help?**:

Not a request for help.  Have already prepared PR for review

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

issue + bug fix

---

**What version of acs-engine?**:

all supporting encryption at rest for K8S

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**

K8S v 1.8.8

**What happened**:

KubeAPI refused to start -- was able to ssh into master node and pull the logs.  the logs revealed that the encryption key field was not considered proper base64 encoding

cut and pasted the encryption key to bash prompt and was able to properly decode the string

noticed that the yaml for the encryption key did not include quotes around the encryption key string, and that the encryption key string contained the character "-" which is interpreted by yaml.

added double quotes into the yaml file around the encryption key -- issue resolved.

incorrect yaml file -> `/etc/kubernetes/encryption-config.yaml` on master node

```
kind: EncryptionConfig
apiVersion: v1
resources:
  - resources:
      - secrets
    providers:
      - aescbc:
          keys:
            - name: key1
              secret: yXTlx7BDn0D6s41br2vp1-oNbSAo6mM8WFucwqednGI=
      - identity: {}
```

corrected yaml file
```
kind: EncryptionConfig
apiVersion: v1
resources:
  - resources:
      - secrets
    providers:
      - aescbc:
          keys:
            - name: key1
              secret: "yXTlx7BDn0D6s41br2vp1-oNbSAo6mM8WFucwqednGI="
      - identity: {}
```
I imagine that the "-" character came from some encoding of a date-specific field in the random key since this has been tested daily by our team for the past couple of months -- today, all of our new cluster deployments broke with the same issue -- with different encryption keys, yet all included a "-" character in the encoded string

**What you expected to happen**:

the encryption key field shall be interpreted as proper base64 encoding by the kubeapi module

**How to reproduce it** (as minimally and precisely as possible):

introduce any base64 encoding of an encryption key that has a "-" character into the `azuredeployparameters.json` file

**Anything else we need to know**:
PR prepared and ready for review
similar issue noted for TLS certificate quoting for the `cilium` Networking module -- PR covers that as well

the following snippets are in `azuredeploy.json` illustrating the issue:

note that the substitution string does not contain double quotes (pre-fix)
```
[...]

PR #3248, opened 2018-06-11
quote base64 strings in the yaml files; use k8s compatible encoding for base64 values

PaulCharlton:

put double quotes around the base64 strings in the yaml files to protect against base64 strings which contain '-'  (note: not all dialects of base64 have '-' in index 62, some have '+', '_', '~'

this prevents errors in parsing the yaml 

**Which issue this PR fixes**

fixes https://github.com/Azure/acs-engine/issues/3247

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:
```release-note
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -287,7 +287,7 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
 {{end}}
 
 {{if EnableDataEncryptionAtRest }}
-    sed -i "s|<etcdEncryptionSecret>|{{WrapAsVariable "etcdEncryptionKey"}}|g" "/etc/kubernetes/encryption-config.yaml"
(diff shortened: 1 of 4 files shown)


PaulCharlton:
looks like this issue was introduced in https://github.com/ElementAnalytics/acs-engine/commit/541a0febbb3be54a2e0050766a1dd95e29638dff

the code:
```
		"WrapAsVariable": func(s string) string {
			return fmt.Sprintf("',variables('%s'),'", s)
		},
```
in `pkg/acsengine/template_generator.go` is not placing a `\` in front of the `'` characters and it looks like it intended those `'` characters to pass through to the generated yaml

CecileRobertMichon:
> looks like this issue was introduced in ElementAnalytics/acs-engine@541a0fe
the code:
		"WrapAsVariable": func(s string) string {
			return fmt.Sprintf("',variables('%s'),'", s)
		},
in pkg/acsengine/template_generator.go is not placing a \ in front of the ' characters adjacent to the double quotes and it looks like it intended those ' characters to pass through to the generated yaml

The `WrapAsVariable` function wasn't added or modified in the commit you are referring to above, it was only moved to a different function. 

Running upgrade/scale tests on this branch.

PR #3264, opened 2018-06-13
Set extension for custom DNS

aminebizid:

Add an extension script called register-dns to update custom DNS with master and agent nodes.


jackfrancis:
/lgtm

@aminebizid If you end up using this widely with success, please consider adding some documentation!

PR #3227, opened 2018-06-08
Add helper function CloneAndConform

rjtsdl:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:
#3111 
And it also requires a `make build-vendor`

**If applicable**:
- [ ] documentation
- [x] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
```

PR #3272, opened 2018-06-14
Clone and conform helper function

zqingqing1:
**What this PR does / why we need it**: Add helper function CloneAndConform #3227

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/Azure/acs-engine/pkg/i18n"
+	"github.com/rjtsdl/conform"
 	"golang.org/x/crypto/ssh"
 )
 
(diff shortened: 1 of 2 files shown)


zqingqing1:
This PR is mainly created based on https://github.com/Azure/acs-engine/pull/3227

pkg/helpers/helpers.go
```diff
@@ -49,6 +50,19 @@ func JSONMarshal(content interface{}, escape bool) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// CloneAndConform doesn't change the content, it makes copy of content, and conform fields
+func CloneAndConform(src, dst interface{}) error {
```
CecileRobertMichon: where is this new function being used?

zqingqing1: It will mark secret value as 'redacted'. This function will used by RP.

CecileRobertMichon: Why not put the function in the RP codebase then? Since it won't be called anywhere in acs-engine code

weinong: @zqingqing1 will update the PR with this new annotation to aks api model. We need this function here to be able to unit test it.

CecileRobertMichon: Sounds good. Let's add a comment to explain that this function is needed for RP so that someone doesn't go in and remove it when they see it's not called.

zqingqing1:
Code will change at RP side. just add annotation for acs-engine codebase.

PR #3274, opened 2018-06-14
Adding OpenShift e2e test for existing vnet.

kwoodson:
Adding a test for deploying an openshift cluster with an existing vnet.


jim-minter:
looks good so far, I guess you need to get that vnet created and values substituted into the json file before acs-engine reads it?

kwoodson:
@CecileRobertMichon, I believe this is ready to merge.  Please review.   Thanks!

PR #3277, opened 2018-06-14
Updating the omsagent yaml file to include new changes for node zero dependency

rashmichandrashekar:
This PR is to update the omsagent yaml file to include changes for node zero dependency changes.
Also updates the agent version to the latest one in production
Updates the cpu and memory requests and limits to use the same limits in production

- This doesnt not include any new variable additions in the yaml file so there are no changes to the rest of the changes that Jack made to get Container Monitoring add on to work


rashmichandrashekar:
@jackfrancis : Thanks Jack!

PR #3278, opened 2018-06-15
don't exit CSE on apt.systemd.daily error

jackfrancis:
**What this PR does / why we need it**: Run apt.systemd.daily but don't fail CSE on a transient error

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
don't exit CSE on apt.systemd.daily error
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -177,7 +177,7 @@ function installDocker() {
 }
 
 function runAptDaily() {
-    retrycmd_if_failure 20 30 60 /usr/lib/apt/apt.systemd.daily || exit $ERR_APT_DAILY_TIMEOUT
+    /usr/lib/apt/apt.systemd.daily
 }
 
 function generateAggregatedAPICerts() {


parts/k8s/kubernetescustomscript.sh
```diff
@@ -177,7 +177,7 @@ function installDocker() {
 }
 
 function runAptDaily() {
-    retrycmd_if_failure 20 30 60 /usr/lib/apt/apt.systemd.daily || exit $ERR_APT_DAILY_TIMEOUT
+    /usr/lib/apt/apt.systemd.daily
```
CecileRobertMichon: should we keep the retry there? So that it catches transient behaviors? Or maybe keep it but reduce the number of retries

jackfrancis: I think we should fire and forget once. If there is a reservation on an apt lock while this is requested, for example, we should defer to letting apt.system.daily do its thing according to its regular schedule. In other words, the idea is that CSE runs apt.systemd.daily as a convenience, but not as a requirement.

CecileRobertMichon: 👍

PR #3295, opened 2018-06-18
Run dist-upgrade in CSE

sylr:

- Run `apt-get dist-upgrade` in CSE to upgrade system with latest kernel and security patch.
- Factorize `apt-get update` calls to decrease overall script time

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -28,6 +28,7 @@ ERR_OUTBOUND_CONN_FAIL=50 # Unable to establish outbound connection
 ERR_CUSTOM_SEARCH_DOMAINS_FAIL=80 # Unable to configure custom search domains
 ERR_APT_DAILY_TIMEOUT=98 # Timeout waiting for apt daily updates
 ERR_APT_UPDATE_TIMEOUT=99 # Timeout waiting for apt-get update to complete
+ERR_APT_DIST_UPGRADE_TIMEOUT=100 # Timeout waiting for apt-get update to complete
 
 OS=$(cat /etc/*-release | grep ^ID= | tr -d 'ID="' | awk '{print toupper($0)}')
 UBUNTU_OS_NAME="UBUNTU"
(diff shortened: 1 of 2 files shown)


sylr:
@jackfrancis  @CecileRobertMichon could you review that ?

CecileRobertMichon:
@sylr we are running tests on the pr to gather data on cse success rates

sylr:
Brilliant, thanks!

sylr:
@CecileRobertMichon any updates on this ?

sylr:
@CecileRobertMichon @jackfrancis Any update about this ?

jackfrancis:
From the man page:

```
       dist-upgrade
           dist-upgrade in addition to performing the function of upgrade, also intelligently handles changing dependencies with new versions of packages;
           apt-get has a "smart" conflict resolution system, and it will attempt to upgrade the most important packages at the expense of less important ones if
           necessary. The dist-upgrade command may therefore remove some packages. The /etc/apt/sources.list file contains a list of locations from which to
           retrieve desired package files. See also apt_preferences(5) for a mechanism for overriding the general settings for individual packages.```

It doesn't appear from the description that `dist-upgrade` as compared to `upgrade` is more likely to include kernel/security patches. Could you provide some concrete examples as to the differences between the two?

Also, we should probably just choose one or the other and standardize, rather than have separate calls.

sylr:
I've always used `dist-upgrade` cause tt will upgrade everything no matter what:

https://askubuntu.com/questions/194651/why-use-apt-get-upgrade-instead-of-apt-get-dist-upgrade

I did not find any `upgrade` calls, are there ?

CecileRobertMichon:
I think https://github.com/Azure/acs-engine/blob/5e9e37701396faffca7e4a84d6f3ec5a5345abeb/parts/k8s/kubernetescustomscript.sh#L180 does the `upgrade system with latest kernel and security patch` part @sylr

sylr:
@CecileRobertMichon I'm not sure it works very well, I'm pretty sure that I saw packages that needed to be updated after a new deployment.

jackfrancis:
Because `dist-upgrade` is less deterministic (though arguably it is less brittle and less likely to fail) we are going to close this. As mentioned in another PR, we are working on pre-baking OS images plus application layer base configuration. dist-upgrade may be a better option for our future CI workflow that does this baking.

PR #3296, opened 2018-06-18
remove docker dependency from hyperkube extract

jessfraz:
doing this in parts til I find the dns issue in https://github.com/Azure/acs-engine/pull/2973

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -90,40 +90,6 @@ write_files:
       name: localclustercontext
     current-context: localclustercontext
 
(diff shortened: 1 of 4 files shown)


jessfraz:
https://jenkins.azure-containers.io/view/acs-engine/job/k8s-containerd/36/console

jessfraz:
thanks @CecileRobertMichon

karataliu:
Seems 'img pull' has some dependency on it, shall we turn to 'docker pull' if runtime is set to docker?

jessfraz:
it has a dep on runc which docker does as well so its relative, there is no point in having a different command that is only tested under certain scenarios it would give more likely hood for failure

PR #3298, opened 2018-06-18
Adding tests to ensure plugins for container monitoring are running fine.

rashmichandrashekar:
Adding tests to make sure
1. omsagent daemonset is running
2. omsagent replicaset is running
3. kubepodinventory plugin is processing data successfully
4. kubenodeinventory plugin is processing data successfully
```

### CODE REVIEW

**Comment by CecileRobertMichon** [comment]
Hi @rashmichandrashekar, the unit tests are failing at the lint step (you can run `make test-style` locally to repro)

File: parts/k8s/addons/omsagent-daemonset.yaml [hunk 1]
```diff
@@ -222,6 +222,8 @@ metadata:
                value: my_aks_cluster
              - name: DISABLE_KUBE_SYSTEM_LOG_COLLECTION
                value: "true"
+             - name: ISTEST
+               value: true
```
CecileRobertMichon: what is the reason for adding this?

rashmichandrashekar: @CecileRobertMichon : This is used in our plugin to write logs only in case of test environment.

PR #3299, opened 2018-06-19
Enable uploading cse logs to storage account

CecileRobertMichon:
**What this PR does / why we need it**: Allows specifying parameters in the deployment to upload cse logs to a storage account at the end of cse script. 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


parts/k8s/kubernetescustomscript.sh
```diff
@@ -496,6 +496,25 @@ configAddons() {
...
+installAzCli(){
+    AZ_REPO=$(lsb_release -cs)
+    echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | \
+    sudo tee /etc/apt/sources.list.d/azure-cli.list
+    retrycmd_if_failure 3 1 30 curl -L https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
+    apt_get_update && apt_get_install 2 20 120 azure-cli
+}
+
+exitScript() {
+    if [[ "$SAS_TOKEN" != "" ]]; then
```
weinong: can we only upload when all the required env vars are set?

CecileRobertMichon: fixed

parts/k8s/kubernetescustomscript.sh
```diff
@@ -496,6 +496,25 @@ configAddons() {
     fi
 }
 
+installAzCli(){
+    AZ_REPO=$(lsb_release -cs)
+    echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | \
+    sudo tee /etc/apt/sources.list.d/azure-cli.list
+    retrycmd_if_failure 3 1 30 curl -L https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
```
jackfrancis: The two-step approach elsewhere (get artifact / add to key) is more resilient to network flakiness

parts/k8s/kubernetescustomscript.sh
```diff
@@ -496,6 +496,25 @@ configAddons() {
...
+    apt_get_update && apt_get_install 2 20 120 azure-cli
```
jackfrancis: Any reason why only 2 retries?

parts/k8s/kubernetescustomscript.sh
```diff
@@ -496,6 +496,25 @@ configAddons() {
...
+}
+
+exitScript() {
+    if [[ "$SAS_TOKEN" != "" ]] && [[ "$CONTAINER_NAME" != "" ]] && [[ "$STORAGE_ACCOUNT_NAME" != "" ]] && [[ "$BLOB_PREFIX" != "" ]]; then
+        installAzCli || echo "Failed to install az cli"
+        set +x
+        az storage blob upload -f /var/log/azure/cluster-provision.log -c $CONTAINER_NAME -n $BLOB_PREFIX-`hostname` --account-name $STORAGE_ACCOUNT_NAME --sas-token $SAS_TOKEN || echo "Failed to upload to Storage Account"
```
jackfrancis: I guess if the installAzCli fails above we'll just hit the or condition here on file not found?

parts/k8s/kubernetescustomscript.sh
CecileRobertMichon: The thought here was: we don't want to delay/block deployment for logs so being resilient in this step doesn't matter a lot

parts/k8s/kubernetescustomscript.sh
jackfrancis: Let's reduce the timeout to 5 seconds and retry, say, 10 times

parts/k8s/kubernetescustomscript.sh
jackfrancis: Let's reduce timeout to 60 and retry 3 times

PR #3320, opened 2018-06-20
Update Cluster Autoscaler to v1.3

### LINKED ISSUE #3309: Autoscaler plugin wrong image
In autoscaler deployment the container image is set to 
k8s-gcrio.azureedge.net
instead of 
k8s-gcrio.azureedge.net/cluster-autoscaler:v1.2.2

I had to update myself the deployment after cluster install.

sozercan:
* Updates cluster-autoscaler to v1.3 for k8s 1.11
* Adds a link to `cluster-autoscaler/README.md` from `cluster-definition.md`

@feiskyer can you confirm if CA v1.3 supports v1.11? should I update k8s v1.10 to CA 1.3 also? CA versioning seems confusing.

Fixes #3309


CecileRobertMichon:
Looks like the unit test needs to be updated as well... `	k8s_versions_test.go:48: getK8sVersionComponents() returned an unexpected map[string]string value for k8s 1.11.0-alpha.1: cluster-autoscaler = cluster-autoscaler:v1.3`

feiskyer:
> @feiskyer can you confirm if CA v1.3 supports v1.11? should I update k8s v1.10 to CA 1.3 also? CA versioning seems confusing.

CA v1.3.x is for Kubernetes v1.11.x, while CA v1.2.x is for Kubernetes v1.10.x.

So we should use  CA 1.3.0 only for Kubernetes v1.11.x and keep using CA 1.2.2 for Kubernetes v1.10.x.

pkg/acsengine/k8s_versions.go
```diff
@@ -24,6 +24,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"aci-connector":                    "virtual-kubelet:latest",
 		ContainerMonitoringAddonName:       "oms:ciprod05082018",
 		AzureCNINetworkMonitoringAddonName: "networkmonitor:v0.0.4",
+		"cluster-autoscaler":               "cluster-autoscaler:v1.3",
```
feiskyer: Is v1.3 the image tag? If so, I think v1.3.0 should be used here.

The original image released is `gcr.io/google_containers/cluster-autoscaler:v1.3.0`.

sozercan: thanks! updated tag

PR #3335, opened 2018-06-21
Mount audit log file to disk

kichristensen:
Audit logging in enabled by default in Kubernetes 1.8 and above, but you cannot get to the log file easy, because it is only accessible within the kube-apiserver container. This PR puts the file on the host instead.

The path of the audit log is changed, because I don't think it will be a good idea to mount the hosts ```/var/log``` to the container.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:


kichristensen:
@CecileRobertMichon what do you think about this?

PR #3340, opened 2018-06-22
undo cilium changes to base64 encoding

jackfrancis:
**What this PR does / why we need it**: Reverts part of #3248 

Our cilium E2E tests have been broken since that change.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
undo cilium changes to base64 encoding
```

### DIFF
--- a/examples/networkpolicy/kubernetes-cilium.json
+++ b/examples/networkpolicy/kubernetes-cilium.json
@@ -3,7 +3,7 @@
   "properties": {
     "orchestratorProfile": {
       "orchestratorType": "Kubernetes",
-      "orchestratorRelease": "1.9",
+      "orchestratorRelease": "1.10",
       "kubernetesConfig": {
         "networkPolicy": "cilium",
         "addons": [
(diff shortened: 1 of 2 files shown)


jackfrancis:
@PaulCharlton I've reverted the cilium changeset from #3248 because we've been unable to build v1.9._n_ clusters since then. We use the following example api model in our tests:

https://github.com/Azure/acs-engine/blob/master/examples/networkpolicy/kubernetes-cilium.json

Is there a particular subset of k8s APIs that your change was targeted toward? You should be able to repro by building a cilium-enabled v1.9.8 cluster using the above api model against master. What we're seeing is that we have no name resolution. This PR fixes that.

PaulCharlton:
@jackfrancis thanks -- I had imagined that the CI testing would have picked up such an inconsistency.  I did not personally test against 1.9.   My change set was specifically needed for the etcd encryption key.

jackfrancis:
@PaulCharlton Currently our CI testing doesn't include cilium inline. This PR doesn't revert the changes to `base64.StdEncoding` (from `base64.URLEncoding`). Any concerns with merging this reversion to the cilium-specific cert paving?

PaulCharlton:
LGTM

jackfrancis:
Thanks for weighing in @PaulCharlton :)

PR #3356, opened 2018-06-25
fix(addons): Adds missing cluster subnet substitution for flannel

thomastaylor312:
The flannel configmap was using a hardcoded value for the cluster subnet, so when you used a custom one, it would not route traffic properly


jackfrancis:
@jessfraz FYI

This lgtm, thanks @thomastaylor312!

jessfraz:
nice thanks :)

PR #3363, opened 2018-06-25
openshift: add SecurityGroupName and PrimaryAvailabilitySetName to azure.conf

jim-minter:
@pweil- @kargakis we need to set primaryAvailabilitySetName in azure.conf, otherwise when you create a LoadBalancer service the cloud provider code tries to glom its load balancer on to all the availability sets it can find.  It turns out you can have maximum one public LB per availability set (or vm scaleset), and we're using `infra`'s one for the router, hence without doing this LoadBalancer services are currently broken for us.  (Also possibly we should see if we can convert the router to instantiate the load balancer it uses via a LoadBalancer service).

We need to set securityGroupName in azure.conf to work around https://github.com/kubernetes/kubernetes/issues/64716.

I feel like all the above implies that at least for now we should validate precisely two agent pools - compute and infra - because I suspect that underlying support for Azure features with more is possibly sketchy, and *we* should take the decision to relax the validation and enable these rather than people trying and getting breakage.


jim-minter:
@kargakis please can you pull these validation changes through into your agentPoolOnlyApi PR?

kargakis:
> We need to set securityGroupName in azure.conf to work around kubernetes/kubernetes#64716.

Why are you using the master nsg? Is https://github.com/Azure/acs-engine/issues/3258 what we need to track for this?

kargakis:
We will probably get away with hardcoding these in the agent pool api

/lgtm

kargakis:
Manual approval since the changes are openshift-only

jim-minter:
Yes, master-nsg because that's what's currently being used for the compute nodes.  The cloud provider code automatically adds/removes security rules for LoadBalancers on the NSG (in case it wasn't slow enough already).

pkg/api/vlabs/validate_test.go
```diff
@@ -1316,3 +1324,85 @@ func validOpenShiftConifg() *OpenShiftConfig {
...
+					{
+						Name:                "infra",
+						Role:                "infra",
+						StorageProfile:      ManagedDisks,
+						AvailabilityProfile: AvailabilitySet,
+					},
+				},
+			},
+			expectedErr: errors.New("OpenShift requires exactly two agent pool profiles: compute and infra"),
+		},
```
pweil-: Need test for more than 2 and less than 2

pweil-: actually, nah, since you're doing the deep equal I guess it's fine.

kargakis:
@CecileRobertMichon please merge, this is stuck on the codecov context

PR #3375, opened 2018-06-26
Ensure the use of latest Ubuntu image

sylr:
This would save time if we decide to run apt dist-upgrade in the CSE and would ensure we run with the latest security patches for the kernel and libs if we don't.

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -65,9 +65,9 @@ var (
 	//DefaultUbuntuImageConfig is the default Linux distribution.
 	DefaultUbuntuImageConfig = AzureOSImageConfig{
 		ImageOffer:     "UbuntuServer",
-		ImageSku:       "16.04-LTS",
+		ImageSku:       "16.04-DAILY-LTS",
 		ImagePublisher: "Canonical",
-		ImageVersion:   "16.04.201806220",
+		ImageVersion:   "latest",
 	}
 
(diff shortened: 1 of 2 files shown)


sylr:
@jackfrancis Any updates about this ?

jackfrancis:
@sylr I've marked this as work-in-progress to clearly identify that this may or may not merge. We need to do testing and understand the risk of consuming non-tested (from and AKS perspective) images.

sylr:
@jackfrancis Any updates on the testing ?

jackfrancis:
@sylr Thanks for following up. We managed to dig up some historical incidents that derived from a daily image being flaky, leading to temporary service degradation.

For that reason, let's consider making the daily image an option as opposed to replacing the static image references.

The value of static images is that they will always matriculate through baseline functional and soak testing before being released. We think this value makes it an appropriate default path (though we're aware that this has the trade-off of the OS image being a bit stale compared to "daily".

jackfrancis:
@sylr We are going in a slightly different direction. Folks are working on a "pre-baked" image approach, which may incorporate a "latest" image ref as a baseline. See this WIP PR:

https://github.com/Azure/acs-engine/pull/3536

Feel free to get involved there and voice your opinion. The idea is: by pre-baking and testing our OS images, we can be more confident about always basing them on a "latest" tag (any instability in the OS should be revealed in the baking process), and thus not be at risk of falling behind by pinning to an explicit build in time.

Issue #2576, opened 2018-04-02
Add unit tests


CecileRobertMichon:
Currently coverage is 38%. We should add unit tests as much as possible and get closer to 100%. Good places to start: 
- https://github.com/Azure/acs-engine/tree/master/pkg/acsengine, for example https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/engine_test.go 
- https://github.com/Azure/acs-engine/tree/master/pkg/operations/
- https://github.com/Azure/acs-engine/tree/master/cmd

CecileRobertMichon:
Update: coverage is now 47% (https://codecov.io/gh/Azure/acs-engine). Codecov integration was added in #2719 to help us track coverage in PRs.

PR #3382, opened 2018-06-28
Bring the coverage of vlabs/validate to 90%

tariq1890:
**What this PR does / why we need it**: adding more coverage to validation logic

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: working towards #2576 

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -788,16 +788,13 @@ func (a *AgentPoolProfile) validateOrchestratorSpecificProperties(orchestratorTy
 			return fmt.Errorf("VirtualMachineScaleSets does not support storage account attached disks.  Instead specify 'StorageAccount': '%s' or specify AvailabilityProfile '%s'", ManagedDisks, AvailabilitySet)
 		}
 	}
-	if len(a.Ports) == 0 && len(a.DNSPrefix) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/api/vlabs/validate.go
```diff
@@ -797,7 +797,7 @@ func (a *AgentPoolProfile) validateOrchestratorSpecificProperties(orchestratorTy
 func validateKeyVaultSecrets(secrets []KeyVaultSecrets, requireCertificateStore bool) error {
 	for _, s := range secrets {
 		if len(s.VaultCertificates) == 0 {
-			return fmt.Errorf("Invalid KeyVaultSecrets must have no empty VaultCertificates")
+			return fmt.Errorf("Valid KeyVaultSecrets must have no empty VaultCertificates")
```
tariq1890: Also, Line 791 in validate.go is not really accessible via unit test. It is because of what happens in line 772. I think we can remove lines 791-793. If that's okay, I can update this PR

CecileRobertMichon: Yes I think that makes sense, I'm all for removing unreachable code 👍 Your call whether you want to update this PR or leave it as a separate task

PR #3384, opened 2018-06-28
[WIP] Add new flag to use built-in API model

### LINKED ISSUE #3378: acs-engine quick-create ...
We have talked multiple times with regards to optimized creation process (cc: @jessfraz @jackfrancis @cpuguy83, @CecileRobertMichon ) - The current process is:
1. **Lengthy** : copy api-model,  modify values, do `acs-engine generate` then deploy. Each step is prone to errors (spelling issues, copy/paste etc).
2. The current process is **not optimized the best practices** (turn on MSI, use premium disks etc). 

#### Proposal: 
1. Establish a well know golden api-model (as part of the binary). That golden model encompasses all the best practices that we learned and learning.  The api-model will also be updated as part of the release as needed.
2. Create a new CLI command `acs-engine quick-create --resource-group --location --master-count {n} --agentpool-count {n} --node-per-pool {n} --overrides [{key:val}] --output-directory {path}`  calling this command will:

1. Generate the needed api-model + ARM template
2. Create resource-group
3. Perform deployment

> It is assumed that some of the advanced feature (such as BYO VNET) will not supported by `quick-create`

@cpuguy83 is willing to carry this forward. 

comments/questions?

cpuguy83:

This allows users to use a pre-configured API model which follows best
practices.
For now this is mutually exclusive with `--api-model`, but in the future
this may support merging so users can provide their own overrides but
still use the defined best practices.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3378

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
Add `--use-builtin-model` which uses a preconfigured Kubernetes API model following best practices.
```


cmd/deploy.go
```diff
@@ -100,6 +104,8 @@ func newDeployCmd() *cobra.Command {
 	f.StringVarP(&dc.location, "location", "l", "", "location to deploy to (required)")
 	f.BoolVarP(&dc.forceOverwrite, "force-overwrite", "f", false, "automatically overwrite existing files in the output directory")
 	f.StringArrayVar(&dc.set, "set", []string{}, "set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
+	f.BoolVar(&dc.useBuiltinAPIModel, "with-builtin-model", false, "use builtin api model, see --builtin-model-name to specify a specific builtin model")
+	f.StringVar(&dc.builtinAPIModelName, "builtin-model-name", defaultBuiltinAPIModelName, "set the builtin api model with the specified name")
```
CecileRobertMichon: Why not use use one flag for the apimodel name and determine "use built-in model" by checking if that string value is empty?

cpuguy83: Thought about that. Mainly so the user doesn't have to supply the model name if they don't want to.... but not tied to this as is.

Also considered having `--api-model` look for a `builtin` value, with the tradeoff that the user can't really customize the model then (not that they can with what I have here, but there's the potential to).

CecileRobertMichon: In that case, I have a different question: why let the user specify an apimodel if the point of the feature is to allow a quick / no customization deploy? If the user wants to specify the apimodel they can always do `acs-engine deploy --api-model pkg/api/models/v0.json`. I liked @khenidak's suggestion to have a "quick-create" option, instead of making it a command it could be `acs-engine deploy --quick-create […]`. Thoughts?

cpuguy83: Thinking about this a bit more, we could just use `--api-model` and then have a separate command to dump a built-in model that the user can then manually tweak.

cpuguy83: >  why let the user specify an apimodel

I think there's really two goals, a quick way to get started with blessed defaults, and a good way to make sure users are using the current best practices even if they don't necessarily know about some new setting.

khenidak: I think this needs to be deliberate by user. Just a thought

cpuguy83: Deliberate in what way?

cpuguy83: Discussed with @khenidak, updated to have just one flag.

jackfrancis:
I like this feature at a high level (hey, just build me a cluster, I trust you to provide a best practice default config), but I think the UX for something like that should reflect the no-nonsense orientation of this user flow. I.e.:

`acs-engine generate`

Once we provide this convenience (we should do it in the `deploy` flow as well) we should just make the `--api-model` an optional flag. Can folks convince me of the value proposition of:

`acs-engine deploy --with-builtin-model=v0`

vs:

`acs-engine deploy --api-model=v0`

(we can discuss making "v0" more semantic later)

We can do the above by doing something like:

- did the user pass in a filepath as the value?
  - if so, interpret it as a filepath
- did the user pass in a "known golden config" string?
  - if so, use that golden config

cpuguy83:
Not a huge fan of guessing if it's a file path/not a file path, but a keyword like `builtin:<name>` would help.

The primary reason I chose the new flag was to allow for being able to let the user specify a builtin model as a baseline and merge in their customizations from `--api-model`.
Dumping the model for the user to manually merge could work, but also requires the user to do this for each new version they want to use.... definitely not opposed to this as it is more deliberate, just possibly more error prone and less likely to update.

cpuguy83:
ping @jackfrancis do you have any other input? Which way do you want to go?

CecileRobertMichon:
@cpuguy83 Jack is on vacation this week. Have you seen the [--set flag](https://github.com/jcorioland/acs-engine/blob/master/docs/kubernetes/deploy.md#step-4-generate-the-templates)?

cpuguy83:
@CecileRobertMichon I believe we still need to allow users to set the name/version of the builtin model or risk breaking users upon upgrade of acs-engine.
`--set` is certainly helpful (definitely need it to set the SSH key at least), though.

Are you saying with `--set` we may not need to support merging a builtin model with a user defined model?

cpuguy83:
p.s. I'm fine with whatever, just want to make sure we don't make something that's hard to work with a couple of versions from now :)

Specifying the builtin model on `--api-model` now also doesn't preclude us from potentially extending this later to support merging later, so this seems mostly fine outside of my concern for guessing what the user wanted (file path/not a file path).

CecileRobertMichon:
@cpuguy83 could you clarify what you mean by 

> I believe we still need to allow users to set the name/version of the builtin model or risk breaking users upon upgrade of acs-engine.

? Are you referring to k8s upgrade or acs-engine new versions? How would choosing the latest apimodel version break it?

cpuguy83:
> How would choosing the latest apimodel version break it?

If we update the model to include some new feature or "better" default and the change is unexpected from a user perspective.

CecileRobertMichon:
Also, we shouldn't need `--set` for the ssh key (unless the user wants to provide their own). Theoretically, it is already possible to auto-generate an ssh key and deploy a "vanilla" k8s cluster by running: ```acs-engine deploy --subscription-id xxxxxx \
    --dns-prefix xxxx --location westus2 \
    --api-model examples/kubernetes.json
```

https://github.com/Azure/acs-engine/blob/master/docs/kubernetes/deploy.md#deploy

**Comment by cpuguy83** [comment]
I was under the impression that we wanted versioned models but reading over the proposal perhaps this was just in passing conversation.

Perhaps it will be best to just introduce a new subcommand as defined originally.

**Comment by jackfrancis** [comment]
Thanks for your patience on this @cpuguy83. My skepticism thus far is due to a hesitancy to introduce anything new, as it seems like we *should* be able to re-use the existing interface(s) to do this.

How about if we introduced these patterns:

`./acs-engine generate --api-model https://raw.githubusercontent.com/Azure/acs-engine/77bd48385ba12314f1d7df36b4e0b4d53ceac0a2/examples/kubernetes.json --set <overrides> ...`

`./acs-engine deploy --api-model https://raw.githubusercontent.com/Azure/acs-engine/77bd48385ba12314f1d7df36b4e0b4d53ceac0a2/examples/kubernetes.json ...`

Basically, if we pass a URI value to `--api-model`, then we are implementing "immutable/versioned api models" as an HTTP/HTTPS endpoint. The existing `--set` option would allow for custom configurations on top of the blessed api models approach.

Thoughts? @CecileRobertMichon @khenidak as well

**Comment by khenidak** [comment]
@jackfrancis  _we are introducing something new_  The value of it outweighs the cost "something new". 

Please help us close this and so we can do other great things on `acs-engine` :-)

**Comment by jackfrancis** [comment]
Spoke with @khenidak on this. Interested for your thoughts @CecileRobertMichon.

Kal and folks want an option to easily install a "blessed" cluster configuration. Given those requirements, it makes the most sense (to me) to enable omitting the `--api-model` argument; basically enabling the following workflow:

`acs-engine generate`

or

`acs-engine generate --set <overrides>`

What are your thoughts around that? In terms of implementation, we could either:

- maintain an authoritative, verbose api model in the repo that this new "default" behavior uses when `--api-model` is not passed in, or we could simply use the vlabs defaults. The point is that we want to continuously update this "blessed" workflow with every acs-engine release.

**Comment by CecileRobertMichon** [comment]
@jackfrancis @khenidak @cpuguy83 I agree with Jack that it makes sense to reuse what is already there since to me the new ability to deploy a cluster with a blessed configuration is not different than what is already there except that we would maintain a versioned example apimodel and that the user would not need to know about it. 

In that sense I think it is reasonable to allow the user to run `acs-engine deploy` or `acs-engine deploy --set <overrides>`. I think `deploy` makes sense over `generate` (or we could allow both) because deploy is the least manual of the two since generate is a two step process.

Regarding 

> maintain an authoritative, verbose api model in the repo that this new "default" behavior uses when --api-model is not passed in, or we could simply use the vlabs defaults. The point is that we want to continuously update this "blessed" workflow with every acs-engine release.

I think both make sense. I have a slight preference for a defined apimodel in the repo so that users can know what they are getting without having to dig through the code.

**Comment by cpuguy83** [comment]
So to summarize, it sounds like for this we want to have an api-model that is maintained in repo and compiled in with the binary which gets used when `--api-model` is not set.

My one thought on this is we are now taking an error case (no API model set) to building out a cluster which may not match user intentions.

**Comment by CecileRobertMichon** [comment]
> My one thought on this is we are now taking an error case (no API model set) to building out a cluster which may not match user intentions.

I have two suggestions that might solve this:
1) When a user doesn't specify an apimodel, print an info log "No apimodel specified, deploying a built-in cluster configuration, or
2) Only allow the `--api-model` flag to not be set if the user also specifies an explicit "build me a blessed apimodel cluster" for example `--quick-create` (so something like `acs-engine deploy --quick-create`)

**Comment by jackfrancis** [comment]
I'm not super concerned with converting a current error case (no `--api-model`) and converting that into a functional workflow (deploy the default, blessed cluster config). And @CecileRobertMichon's suggestion to print a short message of what this new workflow does ("Deploying a cluster using the default Azure Kubernetes configuration...") is a good to inform users who may be surprised at this new behavior.

Thinking about this, however, I wonder if we should indeed add a new flag, like `acs-engine --show-default-config` which prints the full default api model to stdout. This way we can provide users an easy way to build upon/customize the default, as well as easily know what we're going to build on their behalf if they choose this route. It also avoids the ugly problem of having to "store an api model as a file in the repo, and somehow compile that in, as you stated @cpuguy83.

Thoughts? I do think we're almost there!

**Comment by cpuguy83** [comment]
Ok, pushed these changes.
Would be great too if someone can validate the settings on the default model I've defined.

File: pkg/api/apiloader.go [hunk 2]
```diff
@@ -35,6 +35,44 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				StorageProfile:               "ManagedDisks",
+				AvailabilityProfile:          "AvailabilitySet",
+				AcceleratedNetworkingEnabled: boolPtr(true),
+			},
+		},
+		LinuxProfile: &vlabs.LinuxProfile{AdminUsername: "azureuser"},
+	}
+}
+
+func boolPtr(b bool) *bool {
```
CecileRobertMichon: You can re-use helpers.PointerToBool() here instead of adding a new function

pkg/api/apiloader.go
```diff
@@ -35,6 +35,44 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
 	return a.DeserializeContainerService(contents, validate, isUpdate, existingContainerService)
 }
 
+// LoadDefaultContainerServiceProperties loads the default API model
+func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
+	return TypeMeta{APIVersion: vlabs.APIVersion}, &vlabs.Properties{
+		OrchestratorProfile: &vlabs.OrchestratorProfile{
+			OrchestratorType:    "kubernetes",
+			OrchestratorVersion: "1.10.3",
```
CecileRobertMichon: Why 1.10.3? I would vote for OrchestratorRelease 1.10 that way it picks the latest patch of 1.10

cpuguy83:
Fixed

pkg/api/apiloader.go
```diff
@@ -35,6 +35,40 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				NetworkPlugin:      "azure",
+				UseManagedIdentity: true,
+				Addons: []vlabs.KubernetesAddon{
+					{Name: "kubernetes-dashboard", Enabled: helpers.PointerToBool(true)},
+				},
+			},
+		},
+		MasterProfile: &vlabs.MasterProfile{
+			Count:        1,
+			VMSize:       "Standard_D2_v2",
```
khenidak: Standard_DS2_v3

pkg/api/apiloader.go
```diff
@@ -35,6 +35,40 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+			OSDiskSizeGB: 200,
+		},
+		AgentPoolProfiles: []*vlabs.AgentPoolProfile{
+			{
+				Name:                         "agent",
+				Count:                        3,
+				VMSize:                       "Standard_D2_v2",
+				OSDiskSizeGB:                 200,
+				StorageProfile:               "ManagedDisks",
+				AvailabilityProfile:          "AvailabilitySet",
```
khenidak: VirtualMachineScaleSet

pkg/api/apiloader.go
```diff
@@ -35,6 +35,40 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+		MasterProfile: &vlabs.MasterProfile{
+			Count:        1,
+			VMSize:       "Standard_D2_v2",
+			OSDiskSizeGB: 200,
+		},
+		AgentPoolProfiles: []*vlabs.AgentPoolProfile{
+			{
+				Name:                         "agent",
+				Count:                        3,
+				VMSize:                       "Standard_D2_v2",
```
khenidak: Standard_DS2_v3

pkg/api/apiloader.go
```diff
@@ -35,6 +35,40 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
 }
 
+// LoadDefaultContainerServiceProperties loads the default API model
+func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
+	return TypeMeta{APIVersion: vlabs.APIVersion}, &vlabs.Properties{
+		OrchestratorProfile: &vlabs.OrchestratorProfile{
+			OrchestratorType:    "kubernetes",
+			OrchestratorVersion: "1.10",
+			KubernetesConfig: &vlabs.KubernetesConfig{
+				NetworkPlugin:      "azure",
```
khenidak: I think we should network policy by default. @jackfrancis can we use Calico with AzureCNI?

pkg/api/apiloader.go
```diff
@@ -35,6 +35,40 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
 	return a.DeserializeContainerService(contents, validate, isUpdate, existingContainerService)
 }
 
+// LoadDefaultContainerServiceProperties loads the default API model
+func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
+	return TypeMeta{APIVersion: vlabs.APIVersion}, &vlabs.Properties{
+		OrchestratorProfile: &vlabs.OrchestratorProfile{
+			OrchestratorType:    "kubernetes",
+			OrchestratorVersion: "1.10",
```
CecileRobertMichon: This should be `OrchestratorRelease: "1.10", `

pkg/api/apiloader.go
jackfrancis: should be "VirtualMachineScaleSets"

pkg/api/apiloader.go
jackfrancis: No, though `"NetworkPolicy": "azure"` (i.e., Azure NetworkPolicy implementation) is now possible. @saiyan86 can speak to how experimental that feature is at the moment.

cpuguy83:
Copying this from @khenidak's comment above which is hidden due to a code push.
```
khenidak 21 hours ago  Member
@jackfrancis @CecileRobertMichon @cpuguy83
We need to ensure that model is golden - Here are few comments:
1- UseInstanceMetadata should be turned on
2- Kubernetes Version should be latest (or allow the user to use --kubernetes-version)
3- Use Ds* VMs (with premium disks)
4- Due to the fact that we don't have scale for master tooling, let us default to 3 masters.
5- Apply ARM limits when this is done #3319
6- Turn on KMS when multi masters is fixed on https://github.com/Azure/kubernetes-kms/issues (cc: @ritazh )
7- Can we do a quick round of asking ourselves which addons should we have on the golden model?

More importantly should we override all the defaults? I don't want this to break when the acs-engine defaults change.
```

cpuguy83:
Updated:
- Using DS2_v2 (which v3 doesn't seem to be available everywhere).
- 3 masters
- `UseInstanceMetadata` enabled.
- s/OrchestratorVersion/OrchestratorRelease/
- Using "1.10" as K8s version, this can be overwritten with `--set`
- Fixed `VirtualMachineScaleSets`

Is `azure` network policy ok to use?

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
 	return a.DeserializeContainerService(contents, validate, isUpdate, existingContainerService)
 }
 
+// LoadDefaultContainerServiceProperties loads the default API model
+func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
+	return TypeMeta{APIVersion: vlabs.APIVersion}, &vlabs.Properties{
+		OrchestratorProfile: &vlabs.OrchestratorProfile{
+			OrchestratorType:    "kubernetes",
+			OrchestratorRelease: "1.10",
```
jackfrancis: let's get rid of this and rely upon the in-code defaults enforcement for now

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+			KubernetesConfig: &vlabs.KubernetesConfig{
+				NetworkPlugin:      "azure",
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				UseManagedIdentity: true,
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				Addons: []vlabs.KubernetesAddon{
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+					{Name: "kubernetes-dashboard", Enabled: helpers.PointerToBool(true)},
+				},
+				UseInstanceMetadata: helpers.PointerToBool(true),
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+			},
+		},
+		MasterProfile: &vlabs.MasterProfile{
+			Count:          3,
+			VMSize:         "Standard_DS2_v2",
+			OSDiskSizeGB:   200,
+			StorageProfile: "ManagedDisks",
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+		},
+		AgentPoolProfiles: []*vlabs.AgentPoolProfile{
+			{
+				Name:                         "agent",
+				Count:                        3,
+				VMSize:                       "Standard_DS2_v2",
+				OSDiskSizeGB:                 200,
+				StorageProfile:               "ManagedDisks",
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				AvailabilityProfile:          "VirtualMachineScaleSets",
```
jackfrancis: ditto

pkg/api/apiloader.go
```diff
@@ -35,6 +35,42 @@ func (a *Apiloader) LoadContainerServiceFromFile(jsonFile string, validate, isUp
...
+				AcceleratedNetworkingEnabled: helpers.PointerToBool(true),
```
jackfrancis: ditto

khenidak:
wrt .. `Azure network policy`.. Yes, ready to use.

cpuguy83:
Note that without setting useManagedIdentity=true, the model validate unless the user provides service credentials.

PR #3386, opened 2018-06-29
Cleanup error usage

cpuguy83:

Use github.com/pkg/errors for defining errors.
Instead of masking errors with `fmt.Errorf()`, uses `errors.Wrap` where appropriate.
Incidentally this also fixes some formatting issues in certain error messages.

### DIFF
--- a/cmd/dcos-upgrade.go
+++ b/cmd/dcos-upgrade.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
@@ -15,6 +14,7 @@ import (
 	"github.com/Azure/acs-engine/pkg/i18n"
 	"github.com/Azure/acs-engine/pkg/operations/dcosupgrade"
 	"github.com/leonelquinteros/gotext"
+	"github.com/pkg/errors"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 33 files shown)


jackfrancis:
@cpuguy83 thanks for this! In general we like to pave vendor/ changes discretely, and in advance to avoid merge conflicts. But because you authored the PR that induced these conflicts we're probably 👍 here :)

cpuguy83:
Thanks! I did at least put the vendor change in a separate commit here (though I guess I did not on the last change).

This is rebased.

cmd/dcos-upgrade.go
```diff
@@ -76,40 +76,40 @@ func (uc *dcosUpgradeCmd) validate(cmd *cobra.Command) error {
...
 	uc.locale, err = i18n.LoadTranslations()
 	if err != nil {
-		return fmt.Errorf("error loading translation files: %s", err.Error())
+		return errors.Wrap(err, "error loading translation files")
 	}
 
 	if len(uc.resourceGroupName) == 0 {
 		cmd.Usage()
-		return fmt.Errorf("--resource-group must be specified")
+		return errors.New("--resource-group must be specified")
```
jackfrancis: Just curious: for error messages that aren't synthesizing downstream `err` objects with add'l string message context, what is the value of using `github.com/pkg/errors`? I.e., what the benefit of `errors.New()` over `fmt.Errorf()`?

cpuguy83: `errors.New()` stores the stack information which can aid in debugging... e.g. you can `fmt.Printf("+v")` and see the full stack trace of this error.

PR #3423, opened 2018-07-05
Remove DenyEscalatingExec admission controller from default list

lachie83:
We are removing this plugin by default for the following reasons:

- It only impacts access via the Kubernetes API and without other measures enabled this can be easily circumvented.
- It breaks upstream conformance tests.

If it's needed it can be enabled as part of the cluster definition.


jessfraz:
lgtm :)

jackfrancis:
For folks paying attention: the practical implication after this PR lands is for clusters that need `DenyEscalatingExec` is to explicitly state it in the cluster config (api model) from now on. E.g.:

```
<...>
    "kubernetesConfig": {
        "apiServerConfig": {
            "--admission-control": "<other stuff you need>,DenyEscalatingExec,<other stuff you need>"
    }
<...>
```

PR #3434, opened 2018-07-06
OSA external API

jim-minter:
External API + dumb converter and validation for OSA

### DIFF
--- a/pkg/api/osa/vlabs/convert.go
+++ b/pkg/api/osa/vlabs/convert.go
@@ -1,0 +1,55 @@
+package vlabs
+
+import (
+	"github.com/Azure/acs-engine/pkg/api"
+)
+
+// AsContainerService returns an OpenShiftCluster converted to an
+// api.ContainerService.
+func (oc *OpenShiftCluster) AsContainerService() *api.ContainerService {
+	cs := &api.ContainerService{
+		ID:       oc.ID,
+		Location: oc.Location,
+		Name:     oc.Name,
+		Tags:     oc.Tags,
+		Type:     oc.Type,
+		Properties: &api.Properties{
+			ProvisioningState: api.ProvisioningState(oc.Properties.ProvisioningState),
+			OrchestratorProfile: &api.OrchestratorProfile{
+				OpenShiftConfig: &api.OpenShiftConfig{
+					OpenShiftVersion:       oc.Properties.OpenShiftVersion,
+					PublicHostname:         oc.Properties.PublicHostname,
+					RoutingConfigSubdomain: oc.Properties.RoutingConfigSubdomain,
+				},
+			},
+			ServicePrincipalProfile: &api.ServicePrincipalProfile{
+				ClientID: oc.Properties.ServicePrincipalProfile.ClientID,
+				Secret:   oc.Properties.ServicePrincipalProfile.Secret,
+			},
+		},
+	}
+
+	if oc.Plan != nil {
+		cs.Plan = &api.ResourcePurchasePlan{
+			Name:          oc.Plan.Name,
+			Product:       oc.Plan.Product,
+			PromotionCode: oc.Plan.PromotionCode,
+			Publisher:     oc.Plan.Publisher,
+		}
+	}
+
+	cs.Properties.AgentPoolProfiles = make([]*api.AgentPoolProfile, 0, len(oc.Properties.AgentPoolProfiles))
+	for _, app := range oc.Properties.AgentPoolProfiles {
+		cs.Properties.AgentPoolProfiles = append(cs.Properties.AgentPoolProfiles,
+			&api.AgentPoolProfile{
+				Name:         app.Name,
+				Role:         api.AgentPoolProfileRole(app.Role),
+				Count:        app.Count,
+				VMSize:       app.VMSize,
+				VnetSubnetID: app.VnetSubnetID,
+			},
+		)
+	}
+
+	return cs
+}
--- a/pkg/api/osa/vlabs/types.go
+++ b/pkg/api/osa/vlabs/types.go
@@ -1,0 +1,99 @@
+package vlabs
+
+// OpenShiftCluster complies with the ARM model of resource definition in a JSON
+// template.
+type OpenShiftCluster struct {
+	ID         string                `json:"id,omitempty"`
+	Location   string                `json:"location,omitempty"`
+	Name       string                `json:"name,omitempty"`
+	Plan       *ResourcePurchasePlan `json:"plan,omitempty"`
+	Tags       map[string]string     `json:"tags,omitempty"`
+	Type       string                `json:"type,omitempty"`
+	Properties Properties            `json:"properties,omitempty"`
+}
+
+// ResourcePurchasePlan defines the resource plan as required by ARM for billing
+// purposes.
+type ResourcePurchasePlan struct {
+	Name          string `json:"name,omitempty"`
+	Product       string `json:"product,omitempty"`
+	PromotionCode string `json:"promotionCode,omitempty"`
+	Publisher     string `json:"publisher,omitempty"`
+}
+
+// Properties represents the cluster definition.
+type Properties struct {
+	ProvisioningState ProvisioningState `json:"provisioningState,omitempty"`
+	OpenShiftVersion  string            `json:"openShiftVersion,omitempty"`
+	// TODO: need to clarify external API for PublicHostname and decide what to
+	// implement when.  Allow users to specify a PublicHostname then have them
+	// create a CNAME to a FQDN we return?  Allow users to not specify and we
+	// return a FQDN?  In which case, how will the plugin know the FQDN?
+	PublicHostname string `json:"publicHostname,omitempty"`
+	// FQDN string `json:"fqdn,omitempty"` // TODO: do we need to add this?
+	// TODO: need to clarify external API for RoutingConfigSubdomain.  Do we
+	// create one and return it if it's not provided?  Will this be transparent
+	// to the plugin?
+	RoutingConfigSubdomain string `json:"routingConfigSubdomain,omitempty"`
+	// TODO: need to clarify the external API for AgentPoolProfiles.  Will we
+	// allow users to specify an `infra` pool?
+	AgentPoolProfiles       AgentPoolProfiles       `json:"agentPoolProfiles,omitempty"`
+	ServicePrincipalProfile ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
+}
+
+// ProvisioningState represents the current state of container service resource.
+type ProvisioningState string
+
+const (
+	// Creating means ContainerService resource is being created.
+	Creating ProvisioningState = "Creating"
+	// Updating means an existing ContainerService resource is being updated.
+	Updating ProvisioningState = "Updating"
+	// Failed means resource is in failed state.
+	Failed ProvisioningState = "Failed"
+	// Succeeded means last create/update succeeded.
+	Succeeded ProvisioningState = "Succeeded"
+	// Deleting means resource is in the process of being deleted.
+	Deleting ProvisioningState = "Deleting"
+	// Migrating means resource is being migrated from one subscription or
+	// resource group to another.
+	Migrating ProvisioningState = "Migrating"
+	// Upgrading means an existing resource is being upgraded.
+	Upgrading ProvisioningState = "Upgrading"
+)
+
+// AgentPoolProfiles represents all the AgentPoolProfiles.
+type AgentPoolProfiles []AgentPoolProfile
+
+// AgentPoolProfile represents configuration of VMs running agent daemons that
+// register with the master and offer resources to host applications in
+// containers.
+type AgentPoolProfile struct {
+	Name         string               `json:"name,omitempty"`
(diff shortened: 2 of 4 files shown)


jim-minter:
@pweil- @kargakis @mjudeikis @amanohar @qike-ms @nicolerenee ptal
xref https://github.com/jim-minter/azure-helm/pull/60

pkg/api/osa/vlabs/convert.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	cs.Properties.AgentPoolProfiles = make([]*api.AgentPoolProfile, 0, len(oc.Properties.AgentPoolProfiles))
+	for _, app := range oc.Properties.AgentPoolProfiles {
+		cs.Properties.AgentPoolProfiles = append(cs.Properties.AgentPoolProfiles,
+			&api.AgentPoolProfile{
+				Name:         app.Name,
+				Count:        app.Count,
+				VMSize:       app.VMSize,
+				VnetSubnetID: app.VnetSubnetID,
```
mjudeikis: Is it worth adding subnet name for vnet? 
Case: Customer has a big vnet and wants to split everything into subnets and dedicate one sunbet for OSA?
https://github.com/Azure/acs-engine/blob/master/pkg/api/types.go#L441

test-e2e-win:
@codecov[bot]: you can't request testing unless you are a [Azure](https://github.com/orgs/Azure/people) member.

<details>

In response to [this](https://github.com/Azure/acs-engine/pull/3434#issuecomment-403071950):

># [Codecov](https://codecov.io/gh/Azure/acs-engine/pull/3434?src=pr&el=h1) Report
>> Merging [#3434](https://codecov.io/gh/Azure/acs-engine/pull/3434?src=pr&el=desc) into [master](https://codecov.io/gh/Azure/acs-engine/commit/d4567131d1e71917e3867cedd8f56c5a55ff7687?src=pr&el=desc) will **not change** coverage.
>> The diff coverage is `n/a`.
>
>
>```diff
>@@          Coverage Diff           @@
>##           master   #3434   +/-   ##
>======================================
>  Coverage    55.9%   55.9%           
>======================================
>  Files         105     105           
>  Lines       15855   15855           
>======================================
>  Hits         8864    8864           
>  Misses       6249    6249           
>  Partials      742     742
>```
>


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

pkg/api/osa/vlabs/convert.go
jim-minter: If we follow the AKS API, the RG, vnet and subnet are all included within VnetSubnetID.

jim-minter:
@pweil- @kargakis @mjudeikis @amanohar @qike-ms @nicolerenee there are a few open questions around the design of this API and they may have knock-on effects on data flow wrt the plugin.  These are marked TODO in the PR.  Please comment or we should discuss verbally.

test-e2e-win:
@jim-minter: you can't request testing unless you are a [Azure](https://github.com/orgs/Azure/people) member.

<details>

In response to [this](https://github.com/Azure/acs-engine/pull/3434#issuecomment-403080059):

>@pweil- @kargakis @mjudeikis @amanohar @qike-ms @nicolerenee there are a few open questions around the design of this API and they may have knock-on effects on data flow wrt the plugin.  These are marked TODO in the PR.  Please comment or we should discuss verbally.


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Properties represents the cluster definition.
+type Properties struct {
+	ProvisioningState ProvisioningState `json:"provisioningState,omitempty"`
+	OpenShiftVersion  string            `json:"openShiftVersion,omitempty"`
+	// TODO: need to clarify external API for PublicHostname and decide what to
+	// implement when.  Allow users to specify a PublicHostname then have them
+	// create a CNAME to a FQDN we return?  Allow users to not specify and we
+	// return a FQDN?  In which case, how will the plugin know the FQDN?
```
amanohar: I might not have full picture of the use case but what is the plugin being referred to here?

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+// OpenShiftCluster complies with the ARM model of resource definition in a JSON
+// template.
+type OpenShiftCluster struct {
+	ID         string                `json:"id,omitempty"`
+	Location   string                `json:"location,omitempty"`
+	Name       string                `json:"name,omitempty"`
+	Plan       *ResourcePurchasePlan `json:"plan,omitempty"`
+	Tags       map[string]string     `json:"tags,omitempty"`
+	Type       string                `json:"type,omitempty"`
+	Properties Properties            `json:"properties,omitempty"`
```
amanohar: This can be a pointer.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+	// create a CNAME to a FQDN we return?  Allow users to not specify and we
+	// return a FQDN?  In which case, how will the plugin know the FQDN?
+	PublicHostname string `json:"publicHostname,omitempty"`
+	// FQDN string `json:"fqdn,omitempty"` // TODO: do we need to add this?
+	// TODO: need to clarify external API for RoutingConfigSubdomain.  Do we
+	// create one and return it if it's not provided?  Will this be transparent
+	// to the plugin?
+	RoutingConfigSubdomain string `json:"routingConfigSubdomain,omitempty"`
+	// TODO: need to clarify the external API for AgentPoolProfiles.  Will we
+	// allow users to specify an `infra` pool?
```
amanohar: My understanding is infra pool/VMs is definitely need for OS. In that case customer shouldn't have choice and should just get one (plus it can be named appropriately to additional context).

amanohar: Should infra pool be showed to the customer as one of the pools in agentPoolProfiles or should it just show as a separate entity in the public model like `infraPoolProfile` just for additional clarity (even though backend implementation is the same i.e. it's just a VMSS hosting specific components).

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+	AgentPoolProfiles       AgentPoolProfiles       `json:"agentPoolProfiles,omitempty"`
```
amanohar: Recommend making these pointers.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+
+// AgentPoolProfile represents configuration of VMs running agent daemons that
+// register with the master and offer resources to host applications in
+// containers.
+type AgentPoolProfile struct {
+	Name         string               `json:"name,omitempty"`
+	Role         AgentPoolProfileRole `json:"role,omitempty"` // TODO: should we expose this?
+	Count        int                  `json:"count,omitempty"`
+	VMSize       string               `json:"vmSize,omitempty"`
+	VnetSubnetID string               `json:"vnetSubnetID,omitempty"`
```
amanohar: Is custom VNET going to be supported in private preview?

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+	// OSDiskSizeGB int `json:"osDiskSizeGB,omitempty"` // TODO: do we need to add this?
+	// AvailabilityProfile string `json:"availabilityProfile,omitempty"` // TODO: do we need to add this?
+	// StorageProfile string `json:"storageProfile,omitempty"` // TODO: do we need to add this?
```
amanohar: If only VMSS is supported from the start then I don't there is an need to support or even expose this property.

amanohar: This for managed disk vs unmanaged disk support and isn't relevant as have been migrating towards the manageddisk path. E.g. current AKS + VM nodes only supports manageddisk option. So this field shouldn't need to be there at all for OS.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+	// OSType OSType `json:"osType,omitempty"` // TODO: do we need to add this?
```
amanohar: Do you anticipate a scenario where customer could want an OS controlplane but Windows nodes (assuming the infra pieces are on RH/OS nodes)?

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// AgentPoolProfileRole representes the role of the AgentPoolProfile.
```
amanohar: I don't know if there is any value in exposing the role name here. Are these RBAC roles that are going to be created OOB? Perhaps a specific use case will help shed more light into whether this should stay here or not.

pkg/api/types.go
```diff
@@ -368,6 +368,10 @@ type OpenShiftConfig struct {
 	EnableAADAuthentication bool `json:"enableAADAuthentication,omitempty"`
 
 	ConfigBundles map[string][]byte `json:"configBundles,omitempty"`
+
+	OpenShiftVersion       string
```
amanohar: Thinking out loud: Does a separate `OpenShiftVersion` field need to be added here? Can it use `OrchestratorType` and `OrchestratorVersion` fields? Trying to avoid redundant fields or does `OpenShiftVersion` as a concept doesn't fit with the current semantics of `OrchestratorVersion`?

pkg/api/osa/vlabs/convert.go
```diff
@@ -0,0 +1,55 @@
+package vlabs
```
amanohar: Also, need the reverse convertor from internal to public model. A TODO for the future.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,99 @@
...
+	PromotionCode string `json:"promotionCode,omitempty"`
+	Publisher     string `json:"publisher,omitempty"`
+}
+
+// Properties represents the cluster definition.
+type Properties struct {
+	ProvisioningState ProvisioningState `json:"provisioningState,omitempty"`
+	OpenShiftVersion  string            `json:"openShiftVersion,omitempty"`
+	// TODO: need to clarify external API for PublicHostname and decide what to
+	// implement when.  Allow users to specify a PublicHostname then have them
```
amanohar: Asking users to precreate CNAME will also need validation before cluster create is declared a  success and is also error prone as it needs manual steps+there is always potential for a user to misread documentation. OC should create an FQDN for the customer and return it to them.

pkg/api/osa/vlabs/types.go
pweil-: I think at a minimum we'd allow users to select some options of the infra pool.  Specifically for sizing.  Maybe they don't need all of the setting exposed there.

pweil-: This would be fine.  I will take a look at this set up while Jim is out.

pkg/api/osa/vlabs/types.go
pweil-: Yes, @mjudeikis has been working on making sure it is supported by the plugin

pkg/api/osa/vlabs/types.go
pweil-: There has been discussion about support for Windows nodes.  We may not need this right away but  I expect it to come up in the future.

pkg/api/osa/vlabs/types.go
pweil-: These were indicative of the pool's purpose when using a single agent pool profile slice.  It allows us to have special logic in the template for the role.  I think we have two choices here:

1.  use a compute and infra agent pool on the external api and convert to single agent pool slice on the internal with a role like we have today
2.  keep the internal and external apis using a role

Same result, not sure if one is easier to reason about for a user.  Separate external fields would allow us to control differences between compute and infra pools rather than having separate validation logic based on role.

pkg/api/osa/vlabs/types.go
pweil-: the OpenShift plugin in azure-helm.  It uses the publicHostName in a couple places.  Notably, generating certificates, kube config, and various api object configs

CecileRobertMichon:
closing in favor of #3456

pkg/api/osa/vlabs/types.go
jim-minter: Please could you say why?

jim-minter: @amanohar ^

pkg/api/types.go
jim-minter: I think this is possible, will remove.

pkg/api/osa/vlabs/types.go
kargakis: I guess for `omitempty` to work as expected.

PR #3436, opened 2018-07-06
Add CertificateProfile to V20180331 AgentOnly

brichins:
I'd like to add CertificateProfiles to the V20180331 AgentOnly profiles.  It's already in the vlabs API.  The code is copied/adapted from there.

This PR adds the ability to specify the certificates or keyvault references in the acs-engines model file.  The vlabs API already contains these features.  It's very useful and I'd like to see the same functionality in the v20180331 API.  Without it, I have to modify the resulting ARM parameters file and inject the certificate references.  If I don't the ARM apply fails since the certs are referenced in the ARM file but not the ARM parameters file.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20180331/types.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/types.go
@@ -43,6 +43,7 @@ type Properties struct {
 	EnableRBAC              *bool                    `json:"enableRBAC,omitempty"`
 	NetworkProfile          *NetworkProfile          `json:"networkProfile,omitempty"`
 	AADProfile              *AADProfile              `json:"aadProfile,omitempty"`
+	CertificateProfile      *CertificateProfile      `json:"certificateProfile,omitempty"`
 }
 
 // NetworkProfile represents network related definitions
@@ -103,6 +104,26 @@ type ManagedClusterAccessProfile struct {
 type ServicePrincipalProfile struct {
 	ClientID string `json:"clientId,omitempty" validate:"required"`
 	Secret   string `json:"secret,omitempty" conform:"redact"`
+}
+
+// CertificateProfile contains cert material for the Kubernetes cluster
+type CertificateProfile struct {
+	// CaCertificate is the certificate authority certificate.
+	CaCertificate string `json:"caCertificate,omitempty"`
+	// CaPrivateKey is the certificate authority key.
+	CaPrivateKey string `json:"caPrivateKey,omitempty"`
+	// ApiServerCertificate is the rest api server certificate, and signed by the CA
+	APIServerCertificate string `json:"apiServerCertificate,omitempty"`
+	// ApiServerPrivateKey is the rest api server private key, and signed by the CA
+	APIServerPrivateKey string `json:"apiServerPrivateKey,omitempty"`
+	// ClientCertificate is the certificate used by the client kubelet services and signed by the CA
+	ClientCertificate string `json:"clientCertificate,omitempty"`
+	// ClientPrivateKey is the private key used by the client kubelet services and signed by the CA
+	ClientPrivateKey string `json:"clientPrivateKey,omitempty"`
+	// KubeConfigCertificate is the client certificate used for kubectl cli and signed by the CA
+	KubeConfigCertificate string `json:"kubeConfigCertificate,omitempty"`
+	// KubeConfigPrivateKey is the client private key used for kubectl cli and signed by the CA
+	KubeConfigPrivateKey string `json:"kubeConfigPrivateKey,omitempty"`
 }
 
 // LinuxProfile represents the Linux configuration passed to the cluster
--- a/pkg/api/convertertoagentpoolonlyapi.go
+++ b/pkg/api/convertertoagentpoolonlyapi.go
@@ -296,6 +296,17 @@ func convertVLabsAgentPoolOnlyCertificateProfile(vlabs *vlabs.CertificateProfile
 	api.KubeConfigPrivateKey = vlabs.KubeConfigPrivateKey
 }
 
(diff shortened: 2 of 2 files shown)


jackfrancis:
@amanohar @weinong to explain why this may not be a good idea :)

brichins:
I really like the ability in the V20180331 API to reference secrets from the KeyVault.  If I can accomplish this another way, I'd be open to that too.

CecileRobertMichon:
@brichins not sure what you mean by "the ability in the V20180331 API to reference secrets from the KeyVault", vlabs should be able to do the same thing. If not, we should work towards fixing that instead of adding features to V20180331.

brichins:
I had to use the V20180331 because when I tried to use the vlabs API, it ignored the network profile and chose the kubenet network plugin.  If we can update vlabs to use the right network plugin, that would be great.

brichins:
One more questions.   Is there a public Slack channel or IRC where folks can ask questions of the acs-engine project team?  If not, what is the recommended method to ask questions? create issues?

CecileRobertMichon:
The way to define the network plugin to be kubenet with vlabs is `"networkPlugin": "kubenet"` in the `kubernetesConfig` api model configuration object. See https://github.com/Azure/acs-engine/blob/master/docs/clusterdefinition.md#kubernetesconfig for more details.

I know a lot of people use the #sig-azure slack channel to ask questions about acs-engine. Always feel free to open a Github issue though if that is more convenient for you.

brichins:
Thanks for the quick response.  In this case I've specified the "azure" network plugin but when it generates with agentOnly it ignores that and inserts the kubenet one.

Here's a snippet:
`{
  "apiVersion": "vlabs",
  "location": "eastus",
  "name": "foo-qa",
  "properties": {
    "kubernetesVersion": "1.10.5",
    "dnsPrefix": "foo-qa",
    "fqdn": "foo-qa.eastus.cloudapp.azure.com",
    "enableRBAC": true,
    "networkProfile": {
      "networkPlugin": "azure"
    },...`

The generated apimodel.json shows this:
`{
  "apiVersion": "vlabs",
  "location": "eastus",
  "name": "foo-qa",
  "properties": {
    "orchestratorProfile": {
      "orchestratorRelease": "1.10",
      "orchestratorVersion": "1.10.5",
      "kubernetesConfig": {
        "kubernetesImageBase": "k8s-gcrio.azureedge.net/",
        "clusterSubnet": "10.244.0.0/16",
        "dnsServiceIP": "10.0.0.10",
        "serviceCidr": "10.0.0.0/16",
        "networkPlugin": "kubenet",...
`

You can see it's ignoring the specified plugin.

CecileRobertMichon:
@brichins there is no networkProfile value in vlabs, see https://github.com/Azure/acs-engine/tree/master/examples/networkplugin for an example.

brichins:
I changed to the orchestratorProfile, but it still shows up with kubenet in the generated apimodel.json

```
{
  "apiVersion": "vlabs",
  "location": "eastus",
  "name": "foo-qa",
  "properties": {
    "kubernetesVersion": "1.10.5",
    "dnsPrefix": "foo-qa",
    "fqdn": "foo-qa.eastus.cloudapp.azure.com",
    "enableRBAC": true,
    "orchestratorProfile": {
      "kubernetesConfig": {
        "networkPlugin": "azure"
      }
    },
    "agentPoolProfiles": [
      {
        "name": "worker",
        "count": 2,
        "vmSize": "Standard_D2s_v3",
        "vnetSubnetId": "/subscriptions/c9b653b0-1f71-4650-9aaa-d29b9db06fd3/resourceGroups/foo-qa/providers/Microsoft.Network/virtualNetworks/k8s-vnet-17918083/subnets/k8s-subnet",
        "storageProfile": "ManagedDisks"
      }
    ],
    "linuxProfile": {
      "adminUsername": "admin",
      "ssh": {
        "publicKeys": [
          {
            "keyData": "ssh-rsa somekey"
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "8ed3ff25-ea24-42a4-955c-55716d1f4173",
      "secret": "supersecret"
    },
    "certificateProfile": {

    }
  }
}
```

CecileRobertMichon:
@brichins this is happening because your apimodel is being interpreted as agentpool only, any reason why you are not defining a master profile?

brichins:
I wanted to use the agent pool only because I already have a master and some agents deployed.  I'm attempting to deploy a new set of agents that can be dedicated for use by a different team.

CecileRobertMichon:
Ah in that case you should be able to use `acs-engine scale`

brichins:
In this case I want to keep my original ARM templates though (from the initial cluster).  If I attempt to modify my original apimodel and redeploy, with a newer version of ACS, it generates different template values that aren't compatible with cloud-init and I get ARM deployment errors that resources could not be modified.

I was hoping I could create sets of agents outside my original config so I could change things like the networkPlugin, kube version, etc.

Seems like if acs-engine has the feature to support agent pool only, it should allow specifying the correct network plugin.  It works correctly with the V20180331 API.  I believe it should also work with the vlabs API.  Thoughts?

CecileRobertMichon:
agent pool only is not a feature of acs-engine, it is used for some AKS flows (AKS = Azure Kubernetes Service, our Managed Kubernetes offering).

brichins:
ok.  That makes more sense.  I believe you should probably remove the agents-only.json example if it's not part of the project in general.  https://github.com/Azure/acs-engine/blob/master/examples/agents-only.json

I believe is misleading to have an example demonstrating things that aren't features of the product.  I'm happy to withdraw my PR at this point.

Thanks for the help.

PR #3449, opened 2018-07-09
Activate Accelerated Networking per default

### LINKED ISSUE #3236: enable accelerated networking by default for supported VMs
See for list of VMs that support accelerated networking:

https://azure.microsoft.com/en-us/blog/maximize-your-vm-s-performance-with-accelerated-networking-now-generally-available-for-both-windows-and-linux/

julienstroheker:

Activate the Accelerated Networking feature per default is the SKU is supported

**Which issue this PR fixes** *

 fixes #3236 

- [X] documentation (already there)
- [x] unit tests

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -1,5 +1,5 @@
     {
-{{if IsEnabledAcc .}}
+{{if .AcceleratedNetworkingEnabled}}
       "apiVersion": "[variables('apiVersionAcceleratedNetworking')]",
 {{else}}
       "apiVersion": "[variables('apiVersionDefault')]",
@@ -32,11 +32,7 @@
       "location": "[variables('location')]",
       "name": "[concat(variables('{{.Name}}VMNamePrefix'), 'nic-', copyIndex(variables('{{.Name}}Offset')))]",
       "properties": {
-{{if IsEnabledAcc .}}
-        "enableAcceleratedNetworking" : true,
-{{else}}
-        "enableAcceleratedNetworking" : false,
-{{end}}
+        "enableAcceleratedNetworking" : "{{.AcceleratedNetworkingEnabled}}",
 {{if not IsOpenShift}}
 {{if .IsCustomVNET}}
         "networkSecurityGroup": {
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -1,5 +1,5 @@
     {
-{{if .AcceleratedNetworkingEnabled}}
+{{if IsEnabledAcc .}}
       "apiVersion": "[variables('apiVersionAcceleratedNetworking')]",
 {{else}}
       "apiVersion": "[variables('apiVersionDefault')]",
@@ -32,7 +32,11 @@
       "location": "[variables('location')]",
       "name": "[concat(variables('{{.Name}}VMNamePrefix'), 'nic-', copyIndex(variables('{{.Name}}Offset')))]",
       "properties": {
-        "enableAcceleratedNetworking" : "{{.AcceleratedNetworkingEnabled}}",
+{{if IsEnabledAcc .}}
+        "enableAcceleratedNetworking" : true,
+{{else}}
+        "enableAcceleratedNetworking" : false,
+{{end}}
 {{if not IsOpenShift}}
 {{if .IsCustomVNET}}
         "networkSecurityGroup": {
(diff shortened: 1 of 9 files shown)


julienstroheker:
@lachie83 would love to have a first review on this to see if I am on the right track.

I tested using VMSS and VMAS and it works with simple and multiple agentpool profile.

Thanks !

lachie83:
cc @sozercan can you please review

pkg/api/types.go
```diff
@@ -451,6 +451,45 @@ type AgentPoolProfile struct {
...
+		}
+		if strings.EqualFold(a.VMSize, "Standard_DS1") || strings.Contains(a.VMSize, "Standard_DS1_") {
+			return false
+		}
+		if strings.EqualFold(a.VMSize, "Standard_F1") || strings.EqualFold(a.VMSize, "Standard_F1s") {
+			return false
+		}
+		return true
+	}
+	return *a.AcceleratedNetworkingEnabled
```
CecileRobertMichon: If the user manually sets a.AcceleratedNetworkingEnabled to true, I think we should also check if it is supported with the VM SKU, either by checking here and overriding to false if it's not, or by adding validation for the apimodel (https://github.com/Azure/acs-engine/blob/master/pkg/api/vlabs/validate.go) to make sure that if  a.AcceleratedNetworkingEnabled is true, then the VM SKU supports accelerated networking and returning an error if it doesn't. I personally think the latter is better UX so we don't silently override.

pkg/api/types.go
```diff
@@ -451,6 +451,45 @@ type AgentPoolProfile struct {
 	ImageRef                     *ImageReference      `json:"imageReference,omitempty"`
 }
 
+// IsEnabledAcc returns if the AcceleratedNetworkingEnabled is explicitly enabled, or the user-provided default if non explicitly enabled
+func (a *AgentPoolProfile) IsEnabledAcc() bool {
```
CecileRobertMichon: Not sure if we need an extra field/function `IsEnabledAcc()` : instead, we could set a default for `AcceleratedNetworkingEnabled` in https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults.go#L791 (using the same logic as below)

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -32,7 +32,11 @@
       "location": "[variables('location')]",
       "name": "[concat(variables('{{.Name}}VMNamePrefix'), 'nic-', copyIndex(variables('{{.Name}}Offset')))]",
       "properties": {
-        "enableAcceleratedNetworking" : "{{.AcceleratedNetworkingEnabled}}",
```
CecileRobertMichon: I think this code could stay unchanged, see my comment below

parts/k8s/kubernetesagentresourcesvmss.t
```diff
@@ -53,7 +53,11 @@
               "name": "[variables('{{.Name}}VMNamePrefix')]",
               "properties": {
                 "primary": true,
-                "enableAcceleratedNetworking" : "{{.AcceleratedNetworkingEnabled}}",
+                {{if IsEnabledAcc .}}
```
CecileRobertMichon: same here

pkg/acsengine/template_generator.go
```diff
@@ -645,6 +645,12 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			}
 			return false
 		},
+		"IsEnabledAcc": func(profile *api.AgentPoolProfile) bool {
```
CecileRobertMichon: this would also be removed

pkg/api/types.go
julienstroheker: This is a very good point, I'll work on it

pkg/api/vlabs/validate.go
```diff
@@ -347,6 +347,11 @@ func (a *Properties) validateAgentPoolProfiles() error {
 		if e := validatePoolOSType(agentPoolProfile.OSType); e != nil {
 			return e
 		}
+		if agentPoolProfile.AcceleratedNetworkingEnabled != nil {
```
CecileRobertMichon: Here you could use `helpers.IsTrueBoolPointer(agentPoolProfile.AcceleratedNetworkingEnabled)` instead of checking if it's nil and check helpers.AcceleratedNetworkingSupported(VMSize)` in validatePoolAcceleratedNetworking, no need to check if it is true or if AcceleratedNetworkingSupported == AcceleratedNetworking.

pkg/helpers/helpers_test.go
```diff
@@ -164,3 +164,58 @@ EPDesL0rH+3s1CKpgkhYdbJ675GFoGoq+X21QaqsdvoXmmuJF9qq9Tq+JaWloUNq
 		t.Fatalf("Public Key did not match expected format/value")
 	}
 }
+
+func TestAcceleratedNetworkingSupported(t *testing.T) {
```
CecileRobertMichon: 👍

PR #3456, opened 2018-07-10
OSA external API

pweil-:
Continuation of https://github.com/Azure/acs-engine/pull/3434 while @jim-minter is out

@pweil- @kargakis @mjudeikis @amanohar @qike-ms @nicolerenee ptal

I will hold on vendoring anything back in to azure-helm until we're more certain about this shape.

Initial feedback includes: separation of compute and infra pools, pointers, updated validation

### DIFF
--- a/pkg/api/osa/vlabs/convert.go
+++ b/pkg/api/osa/vlabs/convert.go
@@ -1,0 +1,55 @@
+package vlabs
+
+import (
+	"github.com/Azure/acs-engine/pkg/api"
+)
+
+// AsContainerService returns an OpenShiftCluster converted to an
+// api.ContainerService.
+func (oc *OpenShiftCluster) AsContainerService() *api.ContainerService {
+	cs := &api.ContainerService{
+		ID:       oc.ID,
+		Location: oc.Location,
+		Name:     oc.Name,
+		Tags:     oc.Tags,
+		Type:     oc.Type,
+		Properties: &api.Properties{
+			ProvisioningState: api.ProvisioningState(oc.Properties.ProvisioningState),
+			OrchestratorProfile: &api.OrchestratorProfile{
+				OpenShiftConfig: &api.OpenShiftConfig{
+					OpenShiftVersion:       oc.Properties.OpenShiftVersion,
+					PublicHostname:         oc.Properties.PublicHostname,
+					RoutingConfigSubdomain: oc.Properties.RoutingConfigSubdomain,
+				},
+			},
+			ServicePrincipalProfile: &api.ServicePrincipalProfile{
+				ClientID: oc.Properties.ServicePrincipalProfile.ClientID,
+				Secret:   oc.Properties.ServicePrincipalProfile.Secret,
+			},
+		},
+	}
+
+	if oc.Plan != nil {
+		cs.Plan = &api.ResourcePurchasePlan{
+			Name:          oc.Plan.Name,
+			Product:       oc.Plan.Product,
+			PromotionCode: oc.Plan.PromotionCode,
+			Publisher:     oc.Plan.Publisher,
+		}
+	}
+
+	cs.Properties.AgentPoolProfiles = make([]*api.AgentPoolProfile, 0, len(oc.Properties.AgentPoolProfiles))
+	for _, app := range oc.Properties.AgentPoolProfiles {
+		cs.Properties.AgentPoolProfiles = append(cs.Properties.AgentPoolProfiles,
+			&api.AgentPoolProfile{
+				Name:         app.Name,
+				Role:         api.AgentPoolProfileRole(app.Role),
+				Count:        app.Count,
+				VMSize:       app.VMSize,
+				VnetSubnetID: app.VnetSubnetID,
+			},
+		)
+	}
+
+	return cs
+}
--- a/pkg/api/osa/vlabs/convert.go
+++ b/pkg/api/osa/vlabs/convert.go
@@ -38,18 +38,28 @@ func (oc *OpenShiftCluster) AsContainerService() *api.ContainerService {
 		}
 	}
 
-	cs.Properties.AgentPoolProfiles = make([]*api.AgentPoolProfile, 0, len(oc.Properties.AgentPoolProfiles))
-	for _, app := range oc.Properties.AgentPoolProfiles {
+	cs.Properties.AgentPoolProfiles = make([]*api.AgentPoolProfile, 0, len(oc.Properties.ComputePools)+1)
+	for _, app := range oc.Properties.ComputePools {
 		cs.Properties.AgentPoolProfiles = append(cs.Properties.AgentPoolProfiles,
 			&api.AgentPoolProfile{
 				Name:         app.Name,
-				Role:         api.AgentPoolProfileRole(app.Role),
+				Role:         api.AgentPoolProfileRole(AgentPoolProfileRoleEmpty),
 				Count:        app.Count,
 				VMSize:       app.VMSize,
 				VnetSubnetID: app.VnetSubnetID,
 			},
 		)
 	}
+
+	cs.Properties.AgentPoolProfiles = append(cs.Properties.AgentPoolProfiles,
+		&api.AgentPoolProfile{
+			Name:         oc.Properties.InfraPool.Name,
+			Role:         api.AgentPoolProfileRole(AgentPoolProfileRoleInfra),
+			Count:        oc.Properties.InfraPool.Count,
+			VMSize:       oc.Properties.InfraPool.VMSize,
+			VnetSubnetID: oc.Properties.InfraPool.VnetSubnetID,
+		},
+	)
 
 	return cs
 }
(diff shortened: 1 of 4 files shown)


pkg/api/osa/vlabs/convert.go
```diff
@@ -0,0 +1,65 @@
...
+// api.ContainerService.
+func (oc *OpenShiftCluster) AsContainerService() *api.ContainerService {
+	cs := &api.ContainerService{
+		ID:       oc.ID,
+		Location: oc.Location,
+		Name:     oc.Name,
+		Tags:     oc.Tags,
+		Type:     oc.Type,
+		Properties: &api.Properties{
+			ProvisioningState: api.ProvisioningState(oc.Properties.ProvisioningState),
```
kargakis: Remind me why we need this? A comment would be nice.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,110 @@
...
+// AgentPoolProfile represents configuration of VMs running agent daemons that
+// register with the master and offer resources to host applications in
+// containers.
+type AgentPoolProfile struct {
+	Name         string `json:"name,omitempty"`
+	Count        int    `json:"count,omitempty"`
+	VMSize       string `json:"vmSize,omitempty"`
+	VnetSubnetID string `json:"vnetSubnetID,omitempty"`
+	// OSDiskSizeGB int `json:"osDiskSizeGB,omitempty"` // TODO: do we need to add this?
+	// AvailabilityProfile string `json:"availabilityProfile,omitempty"` // TODO: do we need to add this?
```
kargakis: I expect no

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,110 @@
...
+	// StorageProfile string `json:"storageProfile,omitempty"` // TODO: do we need to add this?
```
kargakis: No, only managed disks

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,110 @@
...
+	// OSType OSType `json:"osType,omitempty"` // TODO: do we need to add this?
```
kargakis: Hrm, eventually it seems we will need this (unclear whether it's a blocker for GA).

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// AgentPoolProfile represents configuration of VMs running agent daemons that
```
kargakis: s/AgentPoolProfile/InfraPoolProfile/

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,154 @@
...
+	default:
+		return fmt.Errorf("invalid provisioningState %q", p.ProvisioningState)
+	}
+
+	if p.OpenShiftVersion == "" {
+		return fmt.Errorf("openShiftVersion must not be empty")
+	}
+
+	if p.PublicHostname == "" {
+		return fmt.Errorf("publicHostname must not be empty")
```
kargakis: Move in `isValidHostname`

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,154 @@
...
+		if err := app.Validate(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if err := infra.Validate(); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
```
kargakis: return errors.NewAggregate(errs)

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,154 @@
...
+		return errors.NewAggregate(errs)
+	}
+
+	return nil
+}
+
+// Validate validates an InfraPoolProfile.
+func (infra *InfraPoolProfile) Validate() error {
+	errs := []error{}
```
kargakis: var errs []error

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if !regexAgentPoolName.MatchString(infra.Name) {
+		return fmt.Errorf("invalid name %q", infra.Name)
+	}
+
+	if infra.Count < 3 {
+		errs = append(errs, fmt.Errorf("must have at least 3 infra nodes"))
```
kargakis: Unclear why we would force HA infra profile.

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,110 @@
...
+// register with the master and offer resources to host infra applications in
+// containers.
+//
+// NOTE: right now this is just a copy of AgentPoolProfile.  It allows us to
+// control the settings in a way specific to infra nodes if we need to in the future.
+type InfraPoolProfile struct {
+	Name         string `json:"name,omitempty"`
```
kargakis: Embed AgentPoolProfile

kargakis:
Can you forsee any additional knobs we will want to enable for infra nodes?

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return p.ServicePrincipalProfile.Validate()
+}
+
+func ValidatePools(compute AgentPoolProfiles, infra *InfraPoolProfile) error {
+	if infra == nil || len(compute) == 0 {
+		return fmt.Errorf("both infra and compute pools are required")
+	}
+
+	errs := []error{}
```
kargakis: var errs []error

pkg/api/osa/vlabs/types.go
pweil-: non-blocker.

pkg/api/osa/vlabs/validate.go
pweil-: we should not leave it up to the user

pkg/api/osa/vlabs/types.go
pweil-: That is what we're trying *not* to do by separating the profiles (comments on https://github.com/Azure/acs-engine/pull/3434).  The idea/suggestion was that it is possible we want far less options in the infra pool profile than the agent pool profile.  cc @amanohar

pkg/api/osa/vlabs/validate.go
kargakis: Agreed. I was thinking of profiles: non-HA (1), HA (3+)

pkg/api/osa/vlabs/types.go
kargakis: sgtm

pweil-:
> Can you forsee any additional knobs we will want to enable for infra nodes?

Not more, but maybe less.  For instance, do we allow the user to choose the name?  Not much point in doing so.

pweil-:
@kargakis feedback in round2 commit

pkg/api/osa/vlabs/convert.go
pweil-: I would rather see this as a status struct (that also has condition fields) but this is the way other apis are reporting status.  I assume that there is reporting somewhere based on it.  Maybe @amanohar @qike-ms @nicolerenee  can tell us more

amanohar: @kargakis @pweil- here is more context about ARM's `provisioningState` field: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#provisioningstate-property

Detailed error is returned by RP to ARM via operation resource: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#provisioningstate-property

pkg/api/osa/vlabs/types.go
jim-minter: intend to add this in as it's likely enough + neater to add from the beginning than retrofit

pkg/api/osa/vlabs/validate.go
jim-minter: not doing this because of the change in semantic of the field

pkg/api/osa/vlabs/validate.go
jim-minter: I've been putting this kind of thing in plugin validation, not generic validation

pkg/api/osa/vlabs/types.go
jim-minter: I'd like to roll this back - if we're going to have compute, infra and master and who knows what set of user-allowed configurables I think they're best in one AgentPoolProfile struct not multiple

pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,147 @@
+package vlabs
+
+import (
+	"fmt"
+	"regexp"
+
+	"k8s.io/apimachinery/pkg/util/errors"
```
jim-minter: IMO doing this is very bad style in an api package.

jim-minter:
Closing for https://github.com/Azure/acs-engine/pull/3556

PR #3457, opened 2018-07-10
BUGFIX: Add TLS Certificates to Docker Image

apeschel:

The Docker image is missing TLS root certificates, and as a result the
acs-engine commands fail on TLS when connecting to Azure services via
HTTPS. This commit adds the root certificates to the image.


CecileRobertMichon:
Thanks @apeschel. Just pushed the 0.19.3 image with the updated Dockerfile.

Issue #3460, opened 2018-07-11
No remoting available from master to Windows agents


PatrickLang:
**Is this an ISSUE or FEATURE REQUEST?** (choose one): Feature request

---

**What version of acs-engine?**: 0.19.2

---

If I SSH into a Linux master node, I can SSH over to Linux agents as well which is helpful for troubleshooting. Today - Windows can use SSH port forwarding for remote desktop, but that doesn't get a command line. Running PowerShell remoting or SSH on Windows would be helpful to check service status, grep logs, and so on for Windows agents.

Here's one example of what I'd like to be able to do:

```
docker run -it mcr.microsoft.com/powershell pwsh -command "enter-pssession 20143k8s9000 -Credential (get-credential) -Authentication Basic -UseSSL
"

PowerShell credential request
Enter your credentials.
User: azureuser
Password for user azureuser: ************

PSH> get-content c:\k\kubelet.log -Last 100
```

PR #3461, opened 2018-07-11
Add manual steps for powershell remoting.

PatrickLang:
I'd like to be able to connect to Windows nodes from the command line (#3460). This is technically feasible, so I'm documenting the manual steps first. To fully automate this, I need to investigate what the best approach is to secure the connections. My hope is that we can get SSL certs from Azure, rather than self-signed certs that will expire quickly.

Another possibility would be to put this in an acs-engine extension script instead

### DIFF
--- a/docs/kubernetes/troubleshooting.md
+++ b/docs/kubernetes/troubleshooting.md
@@ -64,7 +64,17 @@ There are two symptoms where you may need to debug Custom Script Extension error
 - VMExtensionProvisioningError or VMExtensionProvisioningTimeout
 - `kubectl node` doesn't list the Windows node(s)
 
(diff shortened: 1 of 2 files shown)


PatrickLang:
@jsturtevant does this make sense to you?

PatrickLang:
`Get-Content -tail 60 .\kubelet.err.log -Wait`

PatrickLang:
`$s = New-PSSession -ComputerName 12050k8s9000 -Credential $cred -UseSSL -Authentication basic`
`invoke-command -Session $s -Command { get-content c:\k\kubelet.err.log } | Out-File kubelet.err.log`

PR #3462, opened 2018-07-11
AKS Autoscale api model update

JackQuincy:
Updates to agentpool only api model and the unversioned to support autoscaling.

None this is needed for work in AKS.
I didn't add parameter validation as for agentpools only it makes more sense to just happen in AKS for this. If we move this into standard acs-engine clusters when autoscalePool is set true we should make sure max and min counts are set and that Count is between them. And we should ensure Max and min aren't set when autoscalePool isn't set.
**If applicable**:
- [x] unit tests

**Release note**:

```release-note
NONE
```


JackQuincy:
@CecileRobertMichon do I have to manually do a backwards compatibility test or do we have an automated way to do it?

CecileRobertMichon:
@JackQuincy you can run https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-scale/build?delay=0sec and https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/build?delay=0sec by setting the upgrade/scale branch to your branch to test back compat

pkg/helpers/helpers.go
```diff
@@ -63,6 +63,12 @@ func PointerToBool(b bool) *bool {
 	return &p
 }
 
+// PointerToInt returns a pointer to a int
+func PointerToInt(i int) *int {
```
CecileRobertMichon: This would be an easy unit test to add in https://github.com/Azure/acs-engine/blob/master/pkg/helpers/helpers_test.go

pkg/api/agentPoolOnlyApi/v20180331/types.go
```diff
@@ -183,6 +183,9 @@ type AgentPoolProfile struct {
 	StorageProfile string `json:"storageProfile" validate:"eq=ManagedDisks|len=0"`
 	VnetSubnetID   string `json:"vnetSubnetID,omitempty"`
 	MaxPods        *int   `json:"maxPods,omitempty"`
+	MaxCount       *int   `json:"maxCount,omitempty"`
```
JackQuincy: @feiskyer thoughts on this? I'm doing it different than you did for acs-engine because I wanted to support a world with multiple agent pools.

pkg/helpers/helpers.go
JackQuincy: added

pkg/api/agentPoolOnlyApi/v20180331/types.go
JackQuincy: Also in AKS I can't deploy it via the addon manager. @weinong also curious on your thoughts

JackQuincy: Though it is offically an addon so possibly I should move it back to addons. Thoughts? @feiskyer @weinong @amanohar

feiskyer: Looks ok to me. But notice that the customers may want to change maxCount/minCount after deployment.

weinong:
We should bump the api version...

pkg/api/agentPoolOnlyApi/v20180331/types.go
JackQuincy: Definitely. I was just going to handle that in the service as it already has merge logic that is more complete than the half hearted merge here. That is also why I didn't try to add validation.

JackQuincy:
why bump the api version?

JackQuincy:
@jackfrancis Why was this closed?

PR #3474, opened 2018-07-13
DOCUMENTATION: Fix Documented Default Value for clusterSubnet

apeschel:
This commit corrects the default value listed in the documentation for
clusterSubnet.


apeschel:
Here's the source of confusion:

https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/defaults.go#L513-L520

When Azure VNET integration is enabled, which is the default, clusterSubnet defaults to `10.240.0.0/12`. When it is disabled, it defaults to `10.244.0.0/16`.

I'll update the documentation accordingly.

PR #3497, opened 2018-07-18
NVIDIA device plugin limits and requests

### LINKED ISSUE #3481: nvidia-device-plugin-daemonset pods w/ high memory + CPU
We're seeing high utilization of `nvidia-device-plugin-daemonset` pods on clusters. We should consider resource constraining the daemonset spec.

Data TBD.

sozercan:
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3481


jackfrancis:
Thank you @sozercan!

pkg/acsengine/template_generator.go
```diff
@@ -975,6 +975,30 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 					} else {
 						val = cloudSpecConfig.KubernetesSpecConfig.NVIDIAImageBase + KubeConfigs[k8sVersion][NVIDIADevicePluginAddonName]
 					}
+				case "kubernetesNVIDIADevicePluginCPURequests":
+					if aC > -1 {
```
ritazh: I think this should be nC, instead of aC.

nC := getAddonContainersIndexByName(nvidiaDevicePluginAddon.Containers, NVIDIADevicePluginAddonName)

jackfrancis: indeed, will file a quick fix, good find!

parts/k8s/addons/kubernetesmasteraddons-nvidia-device-plugin-daemonset.yaml
```diff
@@ -40,6 +40,13 @@ spec:
       containers:
       - image: <kubernetesNVIDIADevicePluginSpec>
         name: nvidia-device-plugin-ctr
```
ritazh: This name needs to match `NVIDIADevicePluginAddonName`, which should be `nvidia-device-plugin`

jackfrancis: @sozercan @ritazh what's the action item here? Do we change the const string value, or is this daemonset manifest broken (i.e., change the name here)?

sozercan: i think we can just change the name here to `nvidia-device-plugin`

PR #3504, opened 2018-07-18
Kubernetes E2E: check for nil node InternalIP

jackfrancis:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Kubernetes E2E: check for nil node InternalIP
```

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -278,7 +278,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 					for _, node := range nodeList.Nodes {
 						success := false
 						for i := 0; i < 60; i++ {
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -278,7 +278,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 					for _, node := range nodeList.Nodes {
 						success := false
 						for i := 0; i < 60; i++ {
-							dashboardURL := fmt.Sprintf("http://%s:%v", node.Status.GetAddressByType("InternalIP").Address, port)
+							address := node.Status.GetAddressByType("InternalIP")
```
CecileRobertMichon: do we know that node.Status won't be nil here? maybe we should add a nil check

jackfrancis: I think the "status" child property (derived from `kubectl get <resource> -o json`) can be trusted to be there. At least until it isn't :)

We're doing the same thing for svc/pod/etc

PR #3519, opened 2018-07-19
Get sub from azure CLI when not provided

cpuguy83:

When a user doesn't provide a subscription explicitly, try to use the selected subscription from azure CLI.

Not sure if it's overkill to first check what the selected cloud is vs just always looking for `AzureCloud`.

**If applicable**:
- [x] unit tests

**Release note**:

```release-note
Read subscription ID from azure CLI config when not explicitly provided 
```

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -6,6 +6,8 @@ import (
 	"crypto/rsa"
 	"encoding/json"
 	"io"
+	"os"
+	"runtime"
 	"strings"
 
 	"github.com/Azure/acs-engine/pkg/i18n"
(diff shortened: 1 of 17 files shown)


pkg/helpers/helpers.go
```diff
@@ -118,3 +120,15 @@ func AcceleratedNetworkingSupported(sku string) bool {
 	}
 	return true
 }
+
+// GetHomeDir attempts to get the home dir from env
+func GetHomeDir() string {
```
CecileRobertMichon: A unit test for this new helper function would be great

cpuguy83: Sure, I'm just not sure how to adequately test it without basically putting the same exact logic in the test itself.

PR #3477, opened 2018-07-16
Commiting the latest code gen files so that circle-ci builds pass.

tariq1890:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
```

PR #3520, opened 2018-07-19
Commit minor changes to generated code

cpuguy83:
It seems the generated code is slightly different than whats committed,
so this updates what's committed in the repo to what is actually being
generated.

No functional changes.


tariq1890:
I don't think this should have been committed. From what I've seen, the go-bindata repo hasn't changed since this commit #3477.

jackfrancis:
@tariq1890 you are correct :)

cpuguy83:
Yeah, apparently my go-bindata was outdated even though I just installed it very recently ;) Will make a revert PR.

cpuguy83:
Oh looks like @jackfrancis already took care of it!
Sorry!

tariq1890:
I guess we really need a way of locking on to a particular go-bindata version :).

PR #3522, opened 2018-07-20
Replace Glide with Golang Dep

### LINKED ISSUE #3454: Migrate to dep for dependency management
The developers of glide have recommended that we use dep instead and that they will cease active feature development from now on. Also, most of the other Azure Go repos seem to be using dep as their dependency management tool.

juan-lee:
Glide is deprecated and suggesting its users move to Golang dep.

fixes #3454


cpuguy83:
Are we not going to keep deps in `vendor/`?

jackfrancis:
@cpuguy83 Yes, no change to that strategy, just a change of tool.

cpuguy83:
I'm not sure I understand.
Right now it looks like everything in vendor is being wiped.

jackfrancis:
Yeah, I thought that too, it's that github is truncating the diff and sorting by deletes.

dep does seem to be conspicuously culling a ton of (seemingly non-used) `vendor/` artifacts compared to glide. We're currently testing to validate that nothing functional is effected.

PR #3526, opened 2018-07-21
Add docker health prope

khenidak:
Adds a docker health prope to clusters. The health prope runs by `systemd` and propes docker health restart as needed. The health prope is designed to test the entire docker stack not just `docker ps` which have proven is prone to `healthy` false positive.

@jackfrancis 

@juan-lee + @weinong -- This adds more comfort for aks nodes, as it self heals docker when it misbehavs


khenidak:
@jackfrancis @CecileRobertMichon I appreciate your help solving for this error. failing ci
```
+ echo 'go generate produced changes that were not already present'
go generate produced changes that were not already present
+ exit 1
```

jackfrancis:
redirecting here:

https://github.com/Azure/acs-engine/pull/3529

thanks @khenidak!

PR #3527, opened 2018-07-23
Installing vendored go-bindata binaries

tariq1890:
**What this PR does / why we need it**: We need this as a guard against unexpected changes in the upstream repos of command line tools that are being used in the CI build processes. This provides the means of holding onto a specific version/tag of a binary. 

- This is the approach recommended by the dep developers. More info here : https://golang.github.io/dep/docs/Gopkg.toml.html#required
 
- If this PR is accepted, we would need to remove the go-bindata from the deis' go docker image. (https://github.com/deis/docker-go-dev/blob/master/rootfs/Dockerfile#L75)

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


jackfrancis:
@jim-minter @kargakis FYI

Thanks @tariq1890 for working on moving us out of go-bindata purgatory! :)

jackfrancis:
Thanks for the callout that acs-engine build workflows using the deis/go-dev image won't reliably pick up these changes. I've filed an issue there to see if folks want to proceed:

https://github.com/deis/docker-go-dev/issues/121

Alternatively, we could fork that dev image and make an acs-engine-specific one (or just start over with a leaner image that isn't as general purpose as deis/go-dev).

jackfrancis:
@tariq1890 It's hard to track down definitive info online, but various sources suggest that `vendor/` is resolved before `GOPATH`, which should mean that these changes should work for acs-engine even if they are in an operational context that provides a different go-bindata (e.g., let's say the deis/go-dev container image). We'd just need to ensure that we our acs-engine runtime context always has a `$PWD` of the repo root.

To test that out, would you like to add these changes to this branch and validate:

https://github.com/Azure/acs-engine/pull/3510

tariq1890:
@jackfrancis You are right :). I tested it with the v1.13.0 image and it does works.

So I ran the validate-generated before bootstrap (the old go-bindata is called as a result)and it failed.

After running bootstrap, validate-generated passes, so that is confirmation of the desired go-bindata version running :)

jackfrancis:
Great, thanks! Can we get that deis/go-dev change included in this PR so it's even more signal (you'll need to rebase 😝 as well)

Then we can test/merge, and start enumerating through PRs that have this blockage and verify that this is a reliable fix.

Thanks again!

tariq1890:
Sure np :). It is done!

PR #3529, opened 2018-07-23
Kubernetes: enable docker health probe

jackfrancis:
**What this PR does / why we need it**: Adds a docker health probe to clusters. The health probe runs by systemd and probes docker health restart as needed. The health probe is designed to test the entire docker stack not just docker ps which has proven prone to false positives when healthy.

This is actually @khenidak's work :)

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Kubernetes: enable docker health probe
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -406,6 +406,7 @@ function ensureContainerd() {
 function ensureDocker() {
     wait_for_file 600 1 $DOCKER || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart docker
+    systemctlEnableAndStart docker-health-probe
 }
 function ensureKMS() {
     systemctlEnableAndStart kms
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -406,6 +406,7 @@ function ensureContainerd() {
 function ensureDocker() {
     wait_for_file 600 1 $DOCKER || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart docker
+    retrycmd_if_failure 6 1 10 docker pull busybox # pre-pull busybox, but don't exit if fail
     systemctlEnableAndStart docker-health-probe
 }
 function ensureKMS() {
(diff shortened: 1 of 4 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -406,6 +406,8 @@ function ensureContainerd() {
 function ensureDocker() {
     wait_for_file 600 1 $DOCKER || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart docker
+    retrycmd_if_failure 6 1 10 docker pull busybox # pre-pull busybox, but don't exit if fail
```
jackfrancis: @khenidak I added this

jackfrancis:
@CecileRobertMichon @mboersma FYI, let's consider merging this this week

parts/k8s/kubernetescustomscript.sh
khenidak: This is a nice shortcut. thanks

jackfrancis:
@mboersma to you for review :)

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -47,6 +47,19 @@ write_files:
          "max-file": "5"
       }
     }
+
+- path: "/etc/systemd/system/docker-health-probe.service"
```
mboersma: Could this systemd job definition somehow be shared between master and agent? Maybe using a `{{ template "docker-health-probe.yaml" }}` to include it?

Just wondering how to avoid duplication as the source of errors, but sometimes it's simpler to copy-and-paste anyway.

parts/k8s/kubernetesprovisionsource.sh
```diff
@@ -96,4 +96,36 @@ systemctl_restart() {
             sleep $wait_sleep
         fi
     done
+}
+docker_health_probe()
+{
+  # finds out if docker runtime is misbehaving
+  every=10 #check every n seconds
```
mboersma: Did these values come from experience with testing, or just seem like reasonable choices?

mboersma: I ran `shellcheck` on this script just for grins, and it mostly complained about variable quoting so :+1:.

PR #3536, opened 2018-07-24
WIP Golden image

seanknox:
Still a WIP.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


packer/extract-kubelet.sh
```diff
@@ -0,0 +1,19 @@
...
+img unpack $HYPERKUBE_URL
+
+path=$(find /home/packer/rootfs -name "hyperkube")
+cp "$path" "/usr/local/bin/kubelet"
+cp "$path" "/usr/local/bin/kubectl"
+
+chmod a+x /usr/local/bin/kubelet /usr/local/bin/kubectl
+rm -rf /home/packer/rootfs
+
+echo "Install complete successfully" > /var/log/azure/golden-image-install.complete
```
CecileRobertMichon: we might want to move this out to somewhere not specific to hyperkube

examples/kubernetes.json
```diff
@@ -1,34 +1,36 @@
...
+  "apiVersion": "vlabs",
+  "properties": {
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes"
+    },
+    "masterProfile": {
+      "count": 1,
+      "dnsPrefix": "",
+			"vmSize": "Standard_D2_v2",
+			"distro": "aks"
```
CecileRobertMichon: for now we shouldn't change the default apimodel, we should add a aksbase specific apimodel until we are ready to make it the default

packer/Makefile
```diff
@@ -0,0 +1,18 @@
+SHELL += -eu
+
+BLUE	:= \033[0;34m
+GREEN	:= \033[0;32m
+RED   := \033[0;31m
+NC    := \033[0m
+
+# ∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨∨
+
+export AZURE_LOCATION       		 			?= West US 2
```
CecileRobertMichon: not sure these should be hardcoded in the makefile (they should be set as env vars in CI)

packer/extract-kubelet.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -eux
+
+HYPERKUBE_VERSION="v1.10.5"
```
CecileRobertMichon: //TODO: we need to look into installing multiple k8s versions and selecting the right one during CSE

packer/install-core-packages.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -eux
```
CecileRobertMichon: We should discuss making all those install scripts used by both packer and kubernetescustomscript for CSE installs to avoid having duplicated code and different install approaches + be able to change/tune install code in one place as the code evolves. Also would make sure that clusters with and without golden images get the exact same stuff.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -560,10 +593,12 @@ if [ -f $CUSTOM_SEARCH_DOMAIN_SCRIPT ]; then
     $CUSTOM_SEARCH_DOMAIN_SCRIPT > /opt/azure/containers/setup-custom-search-domain.log 2>&1 || exit $ERR_CUSTOM_SEARCH_DOMAINS_FAIL
 fi
 
+holdWALinuxAgent
 installDeps
+ensureRpc
```
CecileRobertMichon: I don't see the `ensureRpc` function defined

parts/k8s/kubernetescustomscript.sh
```diff
@@ -572,23 +607,19 @@ configureK8s
...
 fi
 
 # containerd needs to be installed before extractHyperkube
 # so runc is present.
-echo `date`,`hostname`, installContainerdStart>>/opt/m
-installContainerd
-echo `date`,`hostname`, extractHyperkubeStart>>/opt/m
 extractHyperkube
-echo `date`,`hostname`, extractHyperkubeDone>>/opt/m
+fi
```
CecileRobertMichon: typo? there's no `if`

parts/k8s/kubernetescustomscript.sh
seanknox: typo, fixed.

parts/k8s/kubernetescustomscript.sh
seanknox: good catch! fixed

packer/extract-kubelet.sh
seanknox: Alternatively, we could download all of the hyperkube images as part of the golden image build, and then use CSE to **run** the correct one.

packer/Makefile
seanknox: agreed.

examples/kubernetes.json
seanknox: good catch, didn't mean to check this in.

packer/extract-kubelet.sh
CecileRobertMichon: Yes, exactly what I meant

packer/install-dependencies.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
CecileRobertMichon: @seanknox @jackfrancis: proposal is to replace all the scripts in the packer vhd json with this + the cleanup-vhd script. The script still needs work to handle versions and disabling/removing certain extra things in CSE (eg. etcd on agents) but this is to give you an idea. Main benefit is that we have one source of truth when it comes to what gets installed on clusters so we can evolve the code in one place.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -61,557 +63,77 @@ else
...
-else
-    echo "skipping master node provision operations, this is an agent node"
+    installDeps
+    installContainerRuntime
+    installNetworkPlugin
+    installContainerd
+    extractHyperkube
+    installFlexVolDrivers
+else 
+    echo "Golden image; skipping dependencies installation"
```
CecileRobertMichon: For now skipping all installs if using golden image for simplicity but we should discuss checking each dependency individually.

PatrickLang: I discussed this with Cecile - I think having some sort of check would be good for partial upgrades. Use cases:

1. in-place security updates
2. During test automation, we need to be able to test new versions of Kubernetes, Azure-CNI, moby, and so on. We shouldn't need to recreate the VHD in those scenarios

CecileRobertMichon:
TODO:

- [ ] install multiple hyperkube and etcd versions
- [ ] remove unused dependencies during CSE (eg. etcd on agent nodes)
- [ ] replace `FULL_INSTALL_REQUIRED` with more granular dependency checks: ie. each dependency should be installed iff it's missing
- [ ] automate pipeline azcopy
- [ ] automate pipeline SAS token creation
- [ ] automate pipeline publishing to Azure marketplace

PR #3543, opened 2018-07-24
Kubernetes: --max-pods=30 should be Azure CNI-only

jackfrancis:
**What this PR does / why we need it**: Instead of defaulting to --max-pods to 30 (the exceptional case) and then overriding to 110 (the more general case), it's easier to set to 110 by default and override to 30 when Azure CNI.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Kubernetes: --max-pods=30 should be Azure CNI-only
```

### DIFF
--- a/pkg/acsengine/defaults-kubelet.go
+++ b/pkg/acsengine/defaults-kubelet.go
@@ -36,7 +36,7 @@ func setKubeletConfig(cs *api.ContainerService) {
 		"--cluster-domain":                  "cluster.local",
 		"--network-plugin":                  "cni",
 		"--pod-infra-container-image":       cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase + KubeConfigs[o.OrchestratorVersion]["pause"],
-		"--max-pods":                        strconv.Itoa(DefaultKubernetesMaxPodsVNETIntegrated),
+		"--max-pods":                        strconv.Itoa(DefaultKubernetesMaxPods),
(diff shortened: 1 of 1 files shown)


donaldguy:
This logic should probably also be reflected in https://github.com/Azure/acs-engine/blob/5d2bedd8ff6c063f2ef98219c8861b0e8c81a122/pkg/api/convertertoagentpoolonlyapi.go#L489-L498 ?

jackfrancis:
Thank you @donaldguy, I agree!

@JunSun17 @weinong FYI

pkg/api/convertertoagentpoolonlyapi.go
```diff
@@ -487,11 +487,10 @@ func convertV20180331AgentPoolOnlyAgentPoolProfile(agentPoolProfile *v20180331.A
 	var maxPods string
 	// agentPoolProfile.MaxPods is 0 if maxPods field is not provided in API model
 	if agentPoolProfile.MaxPods == nil {
-		// default is kubenet
-		if networkProfile == nil || networkProfile.NetworkPlugin == v20180331.Kubenet {
-			maxPods = DefaultKubernetesMaxPodsKubenet
-		} else {
+		if networkProfile.NetworkPlugin == v20180331.Azure {
```
JunSun17: Can you change to networkProfile != nil && networkProfile.NetworkPlugin == v20180331.Azure, to avoid npe.

Issue #3546, opened 2018-07-25
Enabling Accelerated Networking leads to DNS troubles


sylr:
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE
**What version of acs-engine?**: [https://github.com/sylr/acs-engine/tree/v0.19.5+lectra.1](v0.19.5+lectra.1)
**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**: Kubernetes

---

**What happened**:

When I enable accelerated networking in my cluster I can observe **random** DNS resolving latencies (5 seconds timeouts) that I don't have when cluster does not have accelerated networking enabled

**How to reproduce it** (as minimally and precisely as possible):

acs-engine template:

```json
{
  "apiVersion": "vlabs",
  "properties": {
    "clusterName": "k8s4-euw-dev",
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorRelease": "1.10",
      "kubernetesConfig": {
        "kubernetesImageBase": "gcrio.azureedge.net/google_containers/",
        "clusterSubnet": "10.199.0.0/16",
        "dnsServiceIP": "10.200.0.10",
        "serviceCidr": "10.200.0.0/16",
        "dockerBridgeSubnet": "172.17.0.1/16",
        "enableRbac": true,
        "etcdVersion": "3.2.16",
        "etcdDiskSizeGB": "511",
        "networkPolicy": "none",
        "addons": [
          {
            "name": "tiller",
            "enabled": true,
            "config": {
              "max-history": "20"
            }
          }
        ],
        "apiServerConfig": {
          "--oidc-username-claim": "upn"
        }
      }
    },
    "masterProfile": {
      "count": 3,
      "dnsPrefix": "k8s4-EUW-dev",
      "vmSize": "Standard_F2s",
      "osDiskSizeGB": 255,
      "vnetSubnetID": "/subscriptions/xxxxx-xxxxx-xxxxxx-xxxx-xxxx/resourceGroups/network-EUW-dev-RG/providers/Microsoft.Network/virtualNetworks/Network-EUW-dev-VNet/subnets/k8s4-master-EUW-dev-subnet",
      "vnetCidr": "10.100.192.0/18",
      "firstConsecutiveStaticIP": "10.100.195.132",
      "storageProfile" : "ManagedDisks"
    },
    "agentPoolProfiles": [
      {
        "name": "agent",
        "count": 2,
[...]

PR #3548, opened 2018-07-25
Tune kube-dns deployment to fix problems with Accelerated Networking

sylr:
Surprinsingly enough, turning Accelerated Networking on provokes random DNS requests timeouts.

This patch improves the configuration of dnsmasq and turn on glibc mechanism that circumvent the
problem I have diagnosed (although I still don't know why Accelerated Networking triggers it).

```
  # Forward cluster.local to skydns
  --server=/cluster.local/127.0.0.1#10053
  # Forward all other zones to Azure DNS
  --server=168.63.129.16
  # Force dnsmasq to directly return with a NXDomain response for bogus
  # request due to the search domains defined in /etc/resolv.conf
  --address=/com.svc.cluster.local/net.svc.cluster.local/org.svc.cluster.local/com.cluster.local/net.cluster.local/org.cluster.local/
```

```yaml
  # Tune /etc/resolv.conf to resend requests if response grouped and lost
  dnsConfig:
    options:
    - name: attempts
      value: "2"
    - name: single-request-reopen
```

https://github.com/Azure/acs-engine/issues/3546

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
@@ -129,9 +129,11 @@ spec:
         - "-k"
         - "--cache-size=1000"
         - "--no-resolv"
-        - "--server=127.0.0.1#10053"
+        - "--server=/<kubernetesKubeletClusterDomain>/127.0.0.1#10053"
         - "--server=/in-addr.arpa/127.0.0.1#10053"
         - "--server=/ip6.arpa/127.0.0.1#10053"
+        - "--server=168.63.129.16"
+        - "--address=/com.svc.<kubernetesKubeletClusterDomain>/net.svc.<kubernetesKubeletClusterDomain>/org.svc.<kubernetesKubeletClusterDomain>/com.<kubernetesKubeletClusterDomain>/net.<kubernetesKubeletClusterDomain>/org.<kubernetesKubeletClusterDomain>/"
         - "--log-facility=-"
         image: <kubernetesDNSMasqSpec>
         name: dnsmasq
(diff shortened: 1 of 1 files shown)


jackfrancis:
@sylr Thanks for providing this, although it doesn't appear to work.

@feiskyer does the additional `dnsConfig` configuration look valid to you?

parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -173,6 +175,11 @@ spec:
             cpu: 10m
             memory: 50Mi
       dnsPolicy: Default
+      dnsConfig:
+        options:
+        - name: attempts
+          value: "2"
+        - name: single-request-reopen
```
feiskyer: Could we set this only if Accelerated Networking is enabled?

parts/k8s/addons/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -129,9 +129,11 @@ spec:
         - "-k"
         - "--cache-size=1000"
         - "--no-resolv"
-        - "--server=127.0.0.1#10053"
+        - "--server=/cluster.local/127.0.0.1#10053"
         - "--server=/in-addr.arpa/127.0.0.1#10053"
         - "--server=/ip6.arpa/127.0.0.1#10053"
+        - "--server=168.63.129.16"
```
feiskyer: Is this server general for all clouds? If not, we should read it from host

sylr:
Ok so after more testing this does not fix anything.

I thought it did but in reality I've set up the kube-dns replicas to 1 and the kube-dns pod was on the same host as the pod I tested from.

If the pod I test from and the kube-dns pod are on two different node I get to reproduce the problem.

sylr:
@jackfrancis I've opened an issue where I explain everything https://github.com/Azure/acs-engine/issues/3546

PR #3554, opened 2018-07-25
Bump ginkgo to a tagged release

carolynvs:
This updates our dependency on ginkgo to a tagged release (instead of a rando commit).  That unfortunately introduced some new dependencies, namely fsnotify, which dep has trouble with under certain conditions. So I applied the workaround from https://github.com/golang/dep/issues/1799.

I recommend reviewing this by commit, I separated all the vendor files into a separate commit.


carolynvs:
I recommend eating Popeye's fried chicken and then analyzing the crumbs. 🍗 😋

PR #3556, opened 2018-07-26
OSA external API

jim-minter:
Third iteration of OSA external API taking into account all discussion in #3434 and #3456.
@pweil- @kargakis @mjudeikis @qike-ms @amanohar @nicolerenee ptal

Notes:
- as I suppose we will have compute, infra and master pools, I have recombined InfraPoolProfile back into AgentPoolProfiles.
- I have removed more complex validation (`if infra.Count < 3` type stuff) - I think this belongs in the plugin, not here.  It is already present in azure-helm.
- I have added OSType as I think it's probable enough that we might want it some day and it's neater for it to be in from the start rather than defaulted later (explicit over implicit).
- I have implemented the reverse converter as it will be necessary to return output values.
- I believe we should include Name and Role in AgentPoolProfile for mergeability (explicit over implicit).
- I have removed the OpenShiftVersion field as @amanohar suggested.
- I have removed the k8s errors import, I don't think it's appropriate in an API package.

Checklist:
- [ ] Decide if pointer is wanted in `type AgentPoolProfiles []*AgentPoolProfile` (AI @amanohar).
- [ ] Discuss protocol around DNS fields.  What we have is basically workable but doesn't feel very elegant or future-proof (e.g. multiple routers, non-wildcard setups).
- [ ] Discuss and get consensus around AgentPoolProfiles.  Are we happy enough with it?
- [ ] Verify if any more than ServicePrincipalProfile is needed when MSI is enabled.
- [ ] Merge this thing already.


pkg/api/osa/vlabs/validate.go
```diff
@@ -0,0 +1,127 @@
...
+	if app.Count < 1 {
+		return fmt.Errorf("invalid count %q", app.Count)
+	}
+
+	if app.VMSize == "" {
+		return fmt.Errorf("vmSize must not be empty")
+	}
+
+	switch app.OSType {
+	case OSTypeLinux, OSTypeWindows:
```
kargakis: If we are not going to have Windows nodes working by the time this API gets branched, we should invalidate Windows nodes.

pweil-:
> The idea of being able to control the settings individually on different pool types still appeals to me. For now everything is shared and can be embedded, in the future we'll have to start validating against role and settings if we have unshared items.

discussed in scrum and resolved.  Keeping the single pool slice with specific config structs if/when necessary.

pkg/api/osa/vlabs/validate.go
jim-minter: maybe this is a bad approach, but I was separating basic validation (here) from specific implementation validation (in azure-helm).  The latter would disallow Windows nodes.

jim-minter:
@pweil- @kargakis I agree I don't like validation in multiple places.  It's already in too many.  I'll remove it from this PR and we'll implement it in azure-helm.

jim-minter:
@pweil- re: returning []error, completely agree with you, it's just not appropriate to embed the k8s errors package here.  We can do it in azure-helm.

jim-minter:
@amanohar I think this is ready for merge, what about you?

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,133 @@
...
+	Deleting ProvisioningState = "Deleting"
+	// Migrating means the OSA resource is being migrated from one subscription
+	// or resource group to another.
+	Migrating ProvisioningState = "Migrating"
+	// Upgrading means the existing OAS resource is being upgraded.
+	Upgrading ProvisioningState = "Upgrading"
+)
+
+// RouterProfile represents an OpenShift router.
+type RouterProfile struct {
```
pweil-: :+1:

pkg/api/osa/vlabs/types.go
```diff
@@ -0,0 +1,133 @@
...
+	Name string `json:"name,omitempty"`
```
amanohar: Is the name for each `RouterProfile ` for easy referencing?

jim-minter: yes; the name "default" might have slightly special powers as well.

jim-minter:
note, there are at least two additional items we may need to add to the external API, but I suggest those as a follow-up.

1- username(s) and password(s) for red hat services (e.g. registry.redhat.io, insights)

2- authentication configuration for openshift (is it worth adding an extensible struct now with pre-existing members for htpasswd (or not!) and aad authentication?)

kargakis:
> 2- authentication configuration for openshift (is it worth adding an extensible struct now with pre-existing members for htpasswd (or not!) and aad authentication?)

I remember we had related discussions a couple of months ago. Today, I would prefer if we could agree on a single option (aad) that is not configurable.

jim-minter:
> > 2- authentication configuration for openshift (is it worth adding an extensible struct now with pre-existing members for htpasswd (or not!) and aad authentication?)

> I remember we had related discussions a couple of months ago. Today, I would prefer if we could agree on a single option (aad) that is not configurable.

Yes, I'd like to get the barest scaffolding in if we can.

jim-minter:
@jackfrancis or @CecileRobertMichon please can you merge this?

Issue #3507, opened 2018-07-18
CIS benchmark gaps: k8s v1.8 master


jackfrancis:
```
$ kubectl get nodes -o wide
NAME                        STATUS    ROLES     AGE       VERSION   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION      CONTAINER-RUNTIME
k8s-agentpool1-12380904-0   Ready     agent     16m       v1.8.15   <none>        Ubuntu 16.04.4 LTS   4.15.0-1013-azure   docker://1.13.1
k8s-agentpool1-12380904-1   Ready     agent     14m       v1.8.15   <none>        Ubuntu 16.04.4 LTS   4.15.0-1013-azure   docker://1.13.1
k8s-master-12380904-0       Ready     master    19m       v1.8.15   <none>        Ubuntu 16.04.4 LTS   4.15.0-1013-azure   docker://1.13.1
```

Showing master failures from `$ ./kube-bench master --version 1.8`:

```
[INFO] 1.1 API Server
```
- [ ] `[FAIL] 1.1.6 Ensure that the --insecure-port argument is set to 0 (Scored)`
- [ ] `[FAIL] 1.1.12 Ensure that the admission control policy is set to DenyEscalatingExec (Scored)`
- [ ] `[FAIL] 1.1.13 Ensure that the admission control policy is set to SecurityContextDeny (Scored)`
- [ ] `[FAIL] 1.1.21 Ensure that the --kubelet-certificate-authority argument is set as appropriate (Scored)`
- [ ] `[FAIL] 1.1.24 Ensure that the admission control policy is set to PodSecurityPolicy (Scored)`
- [ ]  `[FAIL] 1.1.32 Ensure that the admission control policy is set to NodeRestriction (Scored)`
- [ ]  `[FAIL] 1.1.33 Ensure that the --experimental-encryption-provider-config argument is set as appropriate (Scored)`
- [ ] `[FAIL] 1.1.35 Ensure that the admission control policy is set to EventRateLimit (Scored)`

```
[INFO] 1.3 Controller Manager
```

- [ ] `[FAIL] 1.3.7  Ensure that the RotateKubeletServerCertificate argument is set to true (Scored)`

```
[INFO] 1.4 Configuration Files
```

- [ ] `[FAIL] 1.4.7 Ensure that the etcd pod specification file permissions are set to 644 or more restrictive (Scored)`
[...]

PR #3565, opened 2018-07-26
Kubernetes CIS: apiserver: DenyEscalatingExec

jackfrancis:
**What this PR does / why we need it**: See: #3507 

And see:

https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#denyescalatingexec

FYI this change is only for 1.8 and below clusters (as there are not yet CIS benchmarks for k8s 1.9 and above)

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Kubernetes CIS: apiserver: DenyEscalatingExec
```


jackfrancis:
@marrobi FYI

@weinong @nicolerenee Please audit for any negative AKS effect. The proposal is to set this config flag by default for CIS benchmark adherence

jackfrancis:
This works but we will not introduce by default:

https://github.com/Azure/acs-engine/pull/3423

PR #3570, opened 2018-07-26
change outbound networking check to use more portable constructs

PaulCharlton:
(1) DNS lookup which may be local to a custom VNET, and (2) external site with port 443 which is not blocked by most corporate "on-prem" firewalls.

reasons:  the previous test of google's DNS server on port 53 is blocked by many "on-prem" (and private Azure Tenant) corporate deployments, and will accordingly fail to provision the k8s cluster

issue filed subsequently -> https://github.com/Azure/acs-engine/issues/3571

Was not possible to deploy ACS clusters to generally closed VNETs.  Most corporate deployments __DO__ allow outbound traffic to port 443 via https protocol

```release-note
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -62,7 +62,7 @@ else
 fi
 
 function testOutboundConnection() {
-    retrycmd_if_failure 20 1 3 curl -I https://www.google.com || retrycmd_if_failure 20 1 3 curl -I https://www.1688.com || exit $ERR_OUTBOUND_CONN_FAIL
+    retrycmd_if_failure 20 1 3 nc -v www.google.com 443 || retrycmd_if_failure 20 1 3 nc -v www.1688.com 443 || exit $ERR_OUTBOUND_CONN_FAIL
 }
 
 function waitForCloudInit() {
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -62,7 +62,7 @@ else
 fi
 
 function testOutboundConnection() {
-    retrycmd_if_failure 20 1 3 nc -v 8.8.8.8 53 || retrycmd_if_failure 20 1 3 nc -v 8.8.4.4 53 || exit $ERR_OUTBOUND_CONN_FAIL
+    retrycmd_if_failure 20 1 3 curl -I https://www.google.com || retrycmd_if_failure 20 1 3 curl -I https://www.1688.com || exit $ERR_OUTBOUND_CONN_FAIL
 }
 
 function waitForCloudInit() {


jackfrancis:
@apeschel who is also interested in this surface area

I'll run some tests against this PR and report back after we have 10k deployments or so, to ensure that as a baseline the test changes aren't more subject to underlying cloud flakiness as compared to the google DNS test. Following that we can discuss any pros/cos for this approach as compared to the prior one.

PaulCharlton:
@jackfrancis additional context - already discussed with @apeschel.  Anything with port 443 is a more realistic test as I do not see many pods directly relying on an out-of-cluster DNS provider.  However, the entire infrastructure of K8s does need to be able to pull from remote docker registries, and pull from git repo such as GitHub by port 443 in lieu of port 22 which is often also blocked by corporate firewall and NSG rules.   Same applies to some national level firewalls such as GFOC.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -62,7 +62,7 @@ else
 fi
 
 function testOutboundConnection() {
-    retrycmd_if_failure 20 1 3 nc -v 8.8.8.8 53 || retrycmd_if_failure 20 1 3 nc -v 8.8.4.4 53 || exit $ERR_OUTBOUND_CONN_FAIL
+    retrycmd_if_failure 20 1 3 curl -I https://www.google.com || retrycmd_if_failure 20 1 3 curl -I https://www.1688.com || exit $ERR_OUTBOUND_CONN_FAIL
```
jackfrancis: A cleaner implementation of the intent would be `nc -v google.com 443` (same for the alibaba test)

PaulCharlton: @jackfrancis perhaps, however, the "curl https://" prefix also tests DNS name resolution, which has been an issue I have had to resolve for corporate clients before they could pull containers from remote repositories.

jackfrancis: `nc -v google.com` will test name resolution

PaulCharlton: @jackfrancis you are, of course, correct.

PaulCharlton:
@jackfrancis changes made as requested

jackfrancis:
Tests against this changeset check out (didn't introduce any add'l CSE flakiness).

parts/k8s/kubernetescustomscript.sh
```diff
@@ -62,7 +62,7 @@ else
 fi
 
 function testOutboundConnection() {
-    retrycmd_if_failure 20 1 3 nc -v 8.8.8.8 53 || retrycmd_if_failure 20 1 3 nc -v 8.8.4.4 53 || exit $ERR_OUTBOUND_CONN_FAIL
+    retrycmd_if_failure 20 1 3 nc -v www.google.com 443 || retrycmd_if_failure 20 1 3 nc -v www.1688.com 443 || exit $ERR_OUTBOUND_CONN_FAIL
```
jackfrancis: Noticed that 8.8.8.8 and 8.8.8.4 are listening on TCP 443, so let's make one final change and test the IP directly to avoid the DNS lookup (we're not testing for DNS resolution here).

PaulCharlton: Hey @jackfrancis from my point of view -- testing for name resolution is essential at some point early in the process, we have found customer sites (private Azure Cloud) which have custom DNS that was misconfigured in that ARR is required for name resolution within the cluster, and the customer AAD was required for their intranet which we inter-operate with.

What's the downside of performing name resolution at this point in the startup process?

PaulCharlton: also, I have not yet been able to find an official google doc that port 443 has formal long term support at that IP address

mboersma: > What's the downside of performing name resolution at this point in the startup process?

I think the issue is that this not only changes the port from 53 to the safer 443, but that it switches from IPs to hostnames. The DNS lookup arguably wasn't intended to be part of `testOutboundConnection`.

In general, it's probably just as reliable to contact `www.google.com:443` as `8.8.8.8:53`, and I agree with your point that `8.8.8.8:443` might not be officially supported. We were just wondering if there's a way to do this that doesn't make the test (slightly) more complicated by using a hostname--maybe there isn't.

eosfor: I guess this is why they are listening on 443 https://developers.google.com/speed/public-dns/docs/dns-over-https. but this still is not a best option to use, imho. google dns IPs may be blocked completely inside of an org network. I just tested 2 different org networks I have access to, and it works in one but does not work in another one. May I ask why do you need to test "internet connectivity" this way? I'd better try accessing MSFT endpoints, for instance, like login.microsoft.com or portal.azure.com over 443, or something like that. But definitely not DNS, as this service may be considered "insecure" inside of an org

eosfor: @jackfrancis @CecileRobertMichon please also keep in mind that there could be some kind of "proxy" related issues, when you may need to authenticate against proxy or kind of transparent proxy, before it will be able to forward your http(s) connection to somewhere. So sending simple TCP-SYN may not work.

eosfor: @jackfrancis got it. One thing, it maybe off topic but still. I'd think of a possibility to first download all required binaries into a storage count, aside of cluster provisioning. this step can be done by an admin or something. For instance I run a script on my laptop which downloads all needed binaries, and then I can upload them manually to a storage account. And then I just point acs-engine to that storage account to pick binaries from it. In this case we may not even need an internet connectivity. Service Endpoint will do the trick.

jackfrancis: @eosfor We're actually in the process of baking all of these dependencies onto an acs-engine-specific disk image, ETA for that is 4-8 weeks.

eosfor: @jackfrancis just came into my mind, there is a 168.63.129.16 address. Maybe you in MSFT can leverage it to get to the AKS related binaries. This one is required to be accessible for all the stuff in Azure, so, i'd say it is better choice.

jackfrancis: We're actually hard at work getting pre-baked OS images over the finish line, after which we'll remove this CSE check entirely (we will no longer need to call out to the internet to scaffold a k8s node).

eosfor: @jackfrancis, will you announce this somehow? Will it affect AKS?

PR #3585, opened 2018-07-30
Fix validation when using keyvaultSecretRef

ritazh:
Fix a regression introduced recently.
When using `keyvaultSecretRef` in apimodel, currently getting:
```
must either set useManagedIdentity in the kubernetes config or set --client-id and --client-secret
```


jackfrancis:
@cpuguy83 FYI, does this check out (I think your change to accommodate no sub id introduced this change)

PR #3586, opened 2018-07-30
Autoscale api model

JackQuincy:
Updates to the unversioned to support autoscaling in AKS.
None this is needed for work in AKS.
I didn't add parameter validation as for agentpools only it makes more sense to just happen in AKS for this. If we move this into standard acs-engine clusters when autoscalePool is set true we should make sure max and min counts are set and that Count is between them. And we should ensure Max and min aren't set when autoscalePool isn't set.
**If applicable**:
- [x] unit tests

**Release note**:

```release-note
NONE
```


JackQuincy:
This has passed E2E before, and none of this is called anywhere in ACS-Engine. So I'm not going to trigger the e2e pr tests again. @jackfrancis can you take another look?

PR #3587, opened 2018-07-30
Fix a DCOS upgrade test case when ~/.azure/config exists

mboersma:
`make test` failed for me locally. This was due to new behavior in #3519 that makes `acs-engine` look for a subscription ID in the local `az` CLI config. The one test spec that explicitly wants to use subscriptionID `00000000-0000-0000-0000-000000000000` to produce a "no subscription ID" error is no longer able to if `~/.azure/config` exists.

This change passed CI and made it into master presumably because no `~/.azure` directory exists on the CircleCI agents.

I added a hack to unset the `HOME` environment variable for this test case, which defeats the logic trying to look up the CLI's subscription ID so things work as before. This works but is ugly; I'd love to hear other suggestions.

### DIFF
--- a/cmd/dcos-upgrade_test.go
+++ b/cmd/dcos-upgrade_test.go
@@ -30,9 +30,11 @@ var _ = Describe("the upgrade command", func() {
 		privKey, err := ioutil.TempFile("", "id_rsa")
 		Expect(err).To(BeNil())
 		defer os.Remove(privKey.Name())
+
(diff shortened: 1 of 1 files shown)


mboersma:
@jackfrancis the lookup of ~/.azure/config is done in our own code, not using the `az` CLI, so unsetting `AZURE_CONFIG_DIR` won't work unfortunately.

PR #3588, opened 2018-07-31
Remove OCP 3.9 tests

mjudeikis:
Deprecate OCP 3.9 tests

cc: @kargakis @jim-minter


kargakis:
There is more config that needs to be removed in the `workflows` section.

PR #3598, opened 2018-08-01
Parallelizing all validate test cases

tariq1890:
**What this PR does / why we need it**: Finishing what I started in #3512 
I've run these unit tests multiple times just to make surer they are safe changes. The test outcomes have been consistent so far.

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20180331/validate_test.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/validate_test.go
@@ -2,6 +2,8 @@ package v20180331
 
 import (
 	"testing"
+
+	"github.com/Azure/acs-engine/pkg/helpers"
 )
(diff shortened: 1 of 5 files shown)


mboersma:
@tariq1890 this is great. Looks sane and I've run it many times locally, so I think it's safe.

It didn't seem to speed up `make test` for me though. Did you notice an improvement?

tariq1890:
It's actually an improvement of just 50ms.it was averaging on 320ms before
and now it's 270ms(I ran just the validate test cases here )

I thought it'd be good to have this change anyway, since this does indicate
that the unit tests run independent of each other and are unaffected by the
outcomes of the other.

PR #3610, opened 2018-08-03
Output VMSS cloud config settings.

JunSun17:
Output VMSS cloud config settings for AKS to consume.

### DIFF
--- a/parts/iaasoutputs.t
+++ b/parts/iaasoutputs.t
@@ -29,9 +29,5 @@
     "primaryScaleSetName": {
         "type": "string",
         "value": "[variables('primaryScaleSetName')]"
-    },
-    "vmType": {
-        "type": "string",
-        "value": "[variables('vmType')]"
     }
 
--- a/parts/iaasoutputs.t
+++ b/parts/iaasoutputs.t
@@ -25,4 +25,13 @@
     "primaryAvailabilitySetName": {
         "type": "string",
         "value": "[variables('primaryAvailabilitySetName')]"
+    },
+    "primaryScaleSetName": {
+        "type": "string",
+        "value": "[variables('primaryScaleSetName')]"
+    },
+    "vmType": {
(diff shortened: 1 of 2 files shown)


jackfrancis:
@JunSun17 What's the functional purpose of this change?

parts/iaasoutputs.t
```diff
@@ -25,4 +25,13 @@
     "primaryAvailabilitySetName": {
         "type": "string",
         "value": "[variables('primaryAvailabilitySetName')]"
+    },
+    "primaryScaleSetName": {
+        "type": "string",
+        "value": "[variables('primaryScaleSetName')]"
+    },
+    "vmType": {
```
weinong: we really dont need this

JunSun17: This change is originally target to pass the cloud config settings of vmType and primaryScaleSetName to AKS (in the way of exposing it to ARM output).

I will see if we can build these strings in AKS for its cloud provider config.

parts/k8s/kubernetesmastervars.t
```diff
@@ -359,11 +359,11 @@
     "nsgID": "[resourceId('Microsoft.Network/networkSecurityGroups',variables('nsgName'))]",
 {{if not AnyAgentUsesVirtualMachineScaleSets}}
     "primaryAvailabilitySetName": "[concat('{{ (index .AgentPoolProfiles 0).Name }}-availabilitySet-',variables('nameSuffix'))]",
-    "primaryScaleSetName": "''",
+    "primaryScaleSetName": "",
```
jackfrancis: What is the purpose of these changes (from literal `''` to empty string)?

JunSun17: If it is set to "''", it will not be an empty string in the output. the output will be ''. We need an empty string as the result.

PR #3611, opened 2018-08-03
update go-dev tools image

mboersma:
Adds `azcopy` and `unzip` for your convenience. See https://github.com/deis/docker-go-dev/releases/tag/v1.14.0


CecileRobertMichon:
thank you @mboersma !

PR #2862, opened 2018-05-04
get FD count for managed disk based on region

jackfrancis:
**What this PR does / why we need it**: Assigns fault domain count for managed disk based on region support. See:

https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/managed-disks-common-fault-domain-region-list.md

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1509

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
get FD count for managed disk based on region
```

PR #3617, opened 2018-08-06
get FD count for managed disk based on region

### LINKED ISSUE #1509: Enable Setting Fault & Update Domains for ACS
FEATURE REQUEST

Forked ACS-Engine (I believe from Current will update when confirmed)

k8s 1.7.2

**What happened**:
North Europe had a storage issue, this inevitably took out the Fault Domain with 2 out of the 3 masters.  The single remaining master did not pick up management as it could not form a quorum

There should be an option to set the number of Fault Domains, or provide warnings when a chosen FD count would present operational risks. E.g. with 2 FDs I assume you need 5 or 7 masters (if you can't have a quorum with 2 I am not a etcd expert so don't know!).  This also needs to be in place for agents (as in min number of agents to provide capacity).

Ideally by taking out a FD in an Azure Region, probably best not though please? :)

Need to think about Update Domains too.

mboersma:
**What this PR does / why we need it**: Assigns the fault domain count for managed disks based on region support. See:

https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/managed-disks-common-fault-domain-region-list.md

**Which issue this PR fixes**: 
Fixes #1509
Closes #2862

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:
```release-note
get FD count for managed disk based on region
```

### DIFF
--- a/parts/k8s/kubernetesagentresourcesvmas.t
+++ b/parts/k8s/kubernetesagentresourcesvmas.t
@@ -87,9 +87,9 @@
       "apiVersion": "[variables('apiVersionStorageManagedDisks')]",
       "properties":
         {
-            "platformFaultDomainCount": 2,
+            "platformFaultDomainCount": "[variables('platformFaultDomainCount')]",
             "platformUpdateDomainCount": 3,
-		"managed" : "true"
+            "managed": true
         },
 
       "type": "Microsoft.Compute/availabilitySets"
@@ -348,4 +348,4 @@
       }
     }
     {{end}}
-    
+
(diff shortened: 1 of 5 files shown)


mboersma:
@jackfrancis I refactored this so the ugly ARM template function exists in only one place and doesn't rely on other parameters.

By default, the user will get the max FDs available for the Azure region chosen for deployment. If she adds a non-zero value for `platformFaultDomainCount` to azure.parameters.json, that value is used instead.

(I still think there may be a more idiomatic, cleaner way to do this--I'm all ears for suggestions.)

PR #3623, opened 2018-08-07
Kubernetes: Remove unnecessary ARM vars

jackfrancis:
**What this PR does / why we need it**: Removed two unnecessary Kubernetes variables, and put addon-specific variables behind a "only create these variables if the addon is in use" guard.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
NONE
```

### DIFF
--- a/parts/k8s/kubernetesmastervars.t
+++ b/parts/k8s/kubernetesmastervars.t
@@ -341,9 +341,7 @@
     "kubeServiceCidr": "[parameters('kubeServiceCidr')]",
     "kubeClusterCidr": "[parameters('kubeClusterCidr')]",
     "dockerBridgeCidr": "[parameters('dockerBridgeCidr')]",
-{{if IsKubernetesVersionGe "1.6.0"}}
-    "registerWithTaints": "node-role.kubernetes.io/master=true:NoSchedule",
-{{else}}
+{{if not IsKubernetesVersionGe "1.6.0"}}
     {{if HasLinuxAgents}}
     "registerSchedulable": "false",
     {{else}}
--- a/parts/k8s/kubernetesmastervars.t
+++ b/parts/k8s/kubernetesmastervars.t
@@ -88,7 +88,7 @@
     "kubernetesCcmImageSpec": "[parameters('kubernetesCcmImageSpec')]",
     "kubernetesAddonManagerSpec": "[parameters('kubernetesAddonManagerSpec')]",
     "kubernetesAddonResizerSpec": "[parameters('kubernetesAddonResizerSpec')]",
-{{if DashboardAddonEnabled}}
+{{if .OrchestratorProfile.KubernetesConfig.IsDashboardEnabled}}
     "kubernetesDashboardSpec": "[parameters('kubernetesDashboardSpec')]",
     "kubernetesDashboardCPURequests": "[parameters('kubernetesDashboardCPURequests')]",
     "kubernetesDashboardMemoryRequests": "[parameters('kubernetesDashboardMemoryRequests')]",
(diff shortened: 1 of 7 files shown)


parts/k8s/kubernetesmastervars.t
```diff
@@ -342,15 +353,6 @@
     "kubeServiceCidr": "[parameters('kubeServiceCidr')]",
     "kubeClusterCidr": "[parameters('kubeClusterCidr')]",
     "dockerBridgeCidr": "[parameters('dockerBridgeCidr')]",
-{{if IsKubernetesVersionGe "1.6.0"}}
-    "registerWithTaints": "node-role.kubernetes.io/master=true:NoSchedule",
-{{else}}
```
jackfrancis: Removing pre-1.6 vintage configuration as these are no longer supported

pkg/api/types.go
```diff
@@ -888,9 +888,8 @@ func (o *OrchestratorProfile) IsMetricsServerEnabled() bool {
 }
 
 // IsContainerMonitoringEnabled checks if the container monitoring addon is enabled
-func (o *OrchestratorProfile) IsContainerMonitoringEnabled() bool {
+func (k *KubernetesConfig) IsContainerMonitoringEnabled() bool {
```
jackfrancis: Moved this to be a method of KubernetesConfig (this is more idiomatic with other "is this addon enabled?" functionality) as a part of this work

pkg/acsengine/template_generator.go
```diff
@@ -1145,6 +1145,21 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"EnablePodSecurityPolicy": func() bool {
 			return helpers.IsTrueBoolPointer(cs.Properties.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy)
 		},
+		"ACIConnectorAddonEnabled": func() bool {
```
CecileRobertMichon: I don't think we need those functions, I think we can do `{{if .OrchestratorProfile.KubernetesConfig.ACIConnectorAddonEnabled}}` directly in the template just like we use .MasterProfile for https://github.com/Azure/acs-engine/blob/master/parts/k8s/kubernetesmastervars.t#L257

parts/k8s/kubernetesmastervars.t
```diff
@@ -88,20 +88,24 @@
...
     "kubernetesDashboardCPURequests": "[parameters('kubernetesDashboardCPURequests')]",
     "kubernetesDashboardMemoryRequests": "[parameters('kubernetesDashboardMemoryRequests')]",
     "kubernetesDashboardCPULimit": "[parameters('kubernetesDashboardCPULimit')]",
     "kubernetesDashboardMemoryLimit": "[parameters('kubernetesDashboardMemoryLimit')]",
+{{end}}
     "kubernetesExecHealthzSpec": "[parameters('kubernetesExecHealthzSpec')]",
     "kubernetesDNSSidecarSpec": "[parameters('kubernetesDNSSidecarSpec')]",
     "kubernetesHeapsterSpec": "[parameters('kubernetesHeapsterSpec')]",
     "kubernetesMetricsServerSpec": "[parameters('kubernetesMetricsServerSpec')]",
+{{if IsNVIDIADevicePluginEnabled}}
```
CecileRobertMichon: is this one different?

jackfrancis: I actually don't know the order of precedence here. Both the method on `Properties` and the function inside the template funcMap have the same name. I'm kind of hesitant to poke around at present.

CecileRobertMichon: Oh I see that function was already there... It's the same thing as the others we don't need it but since it's already implemented it's already used in other places so makes sense to fix it separately (a bunch of other template funcMap functions are probably unnecessary as well so we could remove all of them in one PR later)

PR #3639, opened 2018-08-08
Allow kube-proxy to run regardless of any node taint

seanknox:

at the moment a user can taint a node and inadvertently prevent kube-proxy from running on that node. this allows kube-proxy to run on any node of the cluster regardless of the taint.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
@@ -15,9 +15,10 @@ spec:
         tier: node
     spec:
       tolerations:
-      - key: CriticalAddonsOnly
-        operator: Exists
-      - operator: Exists
+      - operator: "Exists"
+        effect: "NoExecute"
+      - operator: "Exists"
+        effect: "NoSchedule"
       containers:
       - command:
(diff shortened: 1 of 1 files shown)


parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
```diff
@@ -15,10 +15,9 @@ spec:
         tier: node
     spec:
       tolerations:
-      - key: node-role.kubernetes.io/master
-        operator: Equal
-        value: "true"
-        effect: NoSchedule
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - operator: Exists
```
CecileRobertMichon: is the double `operator: Exists ` intentional?

seanknox: this is cribbed from kubeadm, believe both are needed.

seanknox:
@lachie83 can I get a 👍 from you on this? This applies tolerations to kube-proxy so it'll run anywhere, or at least that is the intention. Grabbed from the official daemonset.

PR #3644, opened 2018-08-09
Scale down shouldn't depend on VM index

### LINKED ISSUE #2362: Scaling does not work on upgraded clusters with multiple agent nodes

No

Issue

v0.12.0, the earlier version might have the same problem

Kubernetes

**What happened**:

First upgrade a cluster with 3 agent nodes. When scale up or down, got the following errors:

```
./bin-v0.12.0/acs-engine scale --subscription-id xyz --deployment-dir ./_output/xyz  --location westus2 --resource-group xyz  --master-FQDN "xyz.westus2.cloudapp.azure.com"--node-pool "agentpool1" --new-node-count 1 --auth-method client_secret --client-id xyz  --client-secret xyz
INFO[0000] validating...
INFO[0003] Name suffix: 34851071
ERRO[0004] Failed to drain node , got error resource name may not be empty
ERRO[0004] Got error Node '' failed to drain with error: 'resource name may not be empty', while draining the nodes to be deleted
Error: Node '' failed to drain with error: 'resource name may not be empty'
Usage:
  acs-engine scale [flags]
Flags:
      --auth-method device        auth method (default:device, `client_secret`, `client_certificate`) (default "device")
      --azure-env string          the target Azure cloud (default "AzurePublicCloud")
      --certificate-path string   path to client certificate (used with --auth-method=client_certificate)
      --classic-mode              enable classic parameters and outputs
      --client-id string          client id (used with --auth-method=[client_secret|client_certificate])
      --client-secret string      client secret (used with --auth-mode=client_secret)
      --deployment-dir generate   the location of the output from generate
  -h, --help                      help for scale
      --language string           language to return error messages in (default "en-us")
      --location string           location the cluster is deployed in
      --master-FQDN string        FQDN for the master load balancer, Needed to scale down Kubernetes agent pools
      --new-node-count int        desired number of nodes
      --node-pool string          node pool to scale
      --private-key-path string   path to private key (used with --auth-method=client_certificate)
      --resource-group string     the resource group where the cluster is deployed
      --subscription-id string    azure subscription id
Global Flags:
      --debug   enable verbose debug logs
FATA[0004] Node '' failed to drain with error: 'resource name may not be empty'
```

```
./bin-v0.12.0/acs-engine scale --subscription-id xyz --deployment-dir ./_output/xyz  --location westus2 --resource-group xyz  --master-FQDN "xyz.westus2.cloudapp.azure.com"--node-pool "agentpool1" --new-node-count 5 --auth-method client_secret --client-id xyz  --client-secret xyz
INFO[0000] validating...
INFO[0004] Name suffix: 34851071
INFO[0004] Found no resources with type Microsoft.Network/routeTables in the template.  source="scaling command line"
INFO[0004] Starting ARM Deployment (xyz). This will take some time...
INFO[0164] Finished ARM Deployment (xyz).
FATA[0164] resources.DeploymentsClient#CreateOrUpdate: Failure sending request: StatusCode=200 -- Original Error: Long running operation terminated with status 'Failed': Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details."
```

Scaling should work.

For the case with 3 agent nodes, notice the numbering of agent nodes is changed from 0, 1, 2 to 0, 1, 3 after upgrading. this might be related to the scaling issue.

shanalily:
When upgrading a cluster, there can be gaps in the index in the VM names (e.g. there are 3 VMs with names k8s-agentpool1-00000000-0, k8s-agentpool1-00000000-1, k8s-agentpool1-00000000-3 but none ending in 2). This is a problem when scaling down a cluster with the `acs-engine scale` command because the index in the name is used for iterating over the VMs so there might be an attempt to drain a node with no name and the scale operation will fail. The index isn't needed for iterating over the VMs so they can be stored in a slice instead.

**Which issue this PR fixes**: fixes #2362

I want to know that I'm not breaking anything, though I've been using this change for a while without problems. This seems like a simple fix so I feel like I'm missing something.

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -264,7 +264,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 			}
 
 			vmsToDelete := make([]string, 0)
-			for i := len(vms) - 1; i >= sc.newDesiredAgentCount; i-- {
+			for i := currentNodeCount - 1; i >= sc.newDesiredAgentCount; i-- {
 				vmsToDelete = append(vmsToDelete, vms[i])
 			}
 
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -9,6 +9,7 @@ import (
 	"math/rand"
 	"os"
 	"path"
+	"sort"
 	// "sort"
 	"strings"
 	"time"
@@ -213,8 +214,6 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	orchestratorInfo := sc.containerService.Properties.OrchestratorProfile
 	var currentNodeCount, highestUsedIndex, index, winPoolIndex int
 	winPoolIndex = -1
-	// indexes := make([]int, 0)
-	// indexToVM := make(map[int]string)
 	vms := make([]string, 0)
 	if sc.agentPool.IsAvailabilitySets() {
 		for vmsListPage, err := sc.client.ListVirtualMachines(ctx, sc.resourceGroupName); vmsListPage.NotDone(); err = vmsListPage.Next() {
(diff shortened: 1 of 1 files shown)


jackfrancis:
Scale tests succeeded

jackfrancis:
@JackQuincy thanks for the feedback, agree that int sort is preferable to reduce the number of confused humans.

We have a long-term intention of doing just what you say, which is to define our own criteria for sorting nodes by "most appropriate for cordon/drain), and then enforce it for both scale and upgrade.

shanalily:
That makes sense, I'll fix it to the delete highest index first (I'll try to get to that today, if not then this weekend).

PR #3650, opened 2018-08-09
Add Packer

CecileRobertMichon:
**What this PR does / why we need it**: Add Packer for building custom vhd images https://www.packer.io/docs/builders/azure.html. Based on #3536

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
This is the second PR of a series of custom VHDs related PRs to breakdown #3536.
```

### DIFF
--- a/packer.mk
+++ b/packer.mk
@@ -1,5 +1,5 @@
 build-packer:
-	@packer build -var-file=packer/settings.json packer/vhd-image.json
+	@packer build -var-file=packer/settings.json packer/vhd-image-builder.json
 
 init-packer:
 	@./packer/init-variables
--- a/packer.mk
+++ b/packer.mk
@@ -14,4 +14,4 @@ az-copy: az-login
 	azcopy --source "${OS_DISK_SAS}" --destination "${CLASSIC_BLOB}/${VHD_NAME}" --dest-sas "${CLASSIC_SAS_TOKEN}"
 
 generate-sas: az-login
-	az storage container generate-sas --name vhds --permissions lr --connection-string "${CLASSIC_SA_CONNECTION_STRING}" --start ${START_DATE} --expiry ${EXPIRY_DATE} | tee vhd-sas)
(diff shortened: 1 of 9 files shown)


packer.mk
```diff
@@ -0,0 +1,17 @@
...
+	@./packer/init-variables
+
+az-login:
+	az login --service-principal -u ${CLIENT_ID} -p ${CLIENT_SECRET} --tenant ${TENANT_ID}
+
+run-packer:
+	@packer version && make az-login && make init-packer && (make build-packer | tee packer-output)
+
+az-copy:
+	@make az-login && (azcopy --source "${OS_DISK_SAS}" --destination "${CLASSIC_BLOB}/${VHD_NAME}" --dest-sas "${CLASSIC_SAS_TOKEN}")
```
mboersma: I think `az login` should be a make target dependency instead:

```Makefile
az-copy: az-login
        azcopy --source "${OS_DISK_SAS}" --destination "${CLASSIC_BLOB}/${VHD_NAME}" --dest-sas "${CLASSIC_SAS_TOKEN}"
```

Same for the `generate-sas:` target below.

CecileRobertMichon: Good point, I'll fix it. Thanks!

packer.mk
```diff
@@ -0,0 +1,17 @@
...
+	@packer build -var-file=packer/settings.json packer/vhd-image-builder.json
+
+init-packer:
+	@./packer/init-variables
+
+az-login:
+	az login --service-principal -u ${CLIENT_ID} -p ${CLIENT_SECRET} --tenant ${TENANT_ID}
+
+run-packer: az-login
+	@packer version && make init-packer && (make build-packer | tee packer-output)
```
mboersma: When you invoke `make` from inside a Makefile, it should be referred to as `$(MAKE)` according to the [GNU docs](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable).

CecileRobertMichon: fixed

PR #3665, opened 2018-08-13
Fix the issue that the VMSS upgrading doesn't work if the topology do…

chengliangli0918:
…esn't have master nodes

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -3,6 +3,8 @@ package kubernetesupgrade
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"math/rand"
 	"time"
 
 	"github.com/Azure/acs-engine/pkg/acsengine"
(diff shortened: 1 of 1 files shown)


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -383,6 +385,43 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
 }
 
 func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
+	if len(ku.ClusterTopology.AgentPoolScaleSetsToUpgrade) != 0 {
```
jackfrancis: This should probably be `> 0`, right?

chengliangli0918: thanks for the comments, I will change it to > 0 in next iteration.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -383,6 +385,43 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
...
+		}
+
+		if err := transformer.NormalizeForVMSSScaling(ku.logger, templateMap); err != nil {
+			ku.logger.Errorf("unable to update template, error: %v.", err)
+			return err
+		}
+
+		random := rand.New(rand.NewSource(time.Now().UnixNano()))
+		deploymentSuffix := random.Int31()
+		deploymentName := fmt.Sprintf("agentscaleset-%s-%d", time.Now().Format("06-01-02T15.04.05"), deploymentSuffix)
```
CecileRobertMichon: `time.Now().Format("06-01-02T15.04.05")` why is the timestamp hardcoded here?

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -383,6 +385,43 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
 }
 
 func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
+	if len(ku.ClusterTopology.AgentPoolScaleSetsToUpgrade) > 0 {
+		// need to apply the ARM template with target kubernets version to the VMSS first in order that the new VMSS instances
```
CecileRobertMichon: kubernets --> Kubernetes

chengliangli0918: Nice catch, thanks. Fix it in commit 3.

pkg/operations/kubernetesupgrade/upgrader.go
chengliangli0918: This is not hardcoded timestamp, it defines a format to change the time.Now() to string, e.g. 18-08-12T22.33.17. 
This new line just follows the existing pattern used on Line 59 in pkg\operations\kubernetesupgrade\upgradeagentnode.go

deploymentName := fmt.Sprintf("master-%s-%d", time.Now().Format("06-01-02T15.04.05"), deploymentSuffix)

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -383,6 +385,43 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
 }
 
 func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
+	if len(ku.ClusterTopology.AgentPoolScaleSetsToUpgrade) > 0 {
+		// need to apply the ARM template with target Kubernetes version to the VMSS first in order that the new VMSS instances
+		// created can get the expected Kubernetes version. Otherwise the new instances created still have old Kubernetes version
+		// if the topology dosn't have master nodes(so there is no ARM deployments in previous upgradeMasterNodes step)
```
CecileRobertMichon: the last line should be `if the topology doesn't have master nodes (so there are no ARM deployments in previous upgradeMasterNodes step) `

chengliangli0918: good catch, thank you. Changed in commit 4.

PR #3668, opened 2018-08-13
rationalize conditional ARM params

jackfrancis:
**What this PR does / why we need it**: We need to ensure that whenever we conditionally inject parameters values, we accordingly conditionally define parameters, otherwise ARM will prompt for values interactively during group deployment.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
NONE
```

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -238,6 +238,7 @@
       },
       "type": "string"
     },
+{{if .OrchestratorProfile.KubernetesConfig.IsDashboardEnabled}}
     "kubernetesDashboardSpec": {
       "metadata": {
         "description": "The container spec for kubernetes-dashboard-amd64."
@@ -268,6 +269,7 @@
       },
       "type": "string"
     },
+{{end}}
     "kubernetesExecHealthzSpec": {
       "metadata": {
         "description": "The container spec for exechealthz-amd64."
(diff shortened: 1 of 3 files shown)


parts/k8s/kubernetesparams.t
```diff
@@ -358,20 +365,23 @@
...
     "kubernetesAADPodIdentityEnabled": {
       "defaultValue": false,
       "metadata": {
         "description": "AAD Pod Identity status"
       },
       "type": "bool"
     },
+{{end}}
     "kubernetesACIConnectorEnabled": {
-      "defaultValue": false,
```
CecileRobertMichon: why remove the default here?

pkg/acsengine/template_generator.go
```diff
@@ -760,6 +760,13 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"EnablePodSecurityPolicy": func() bool {
 			return helpers.IsTrueBoolPointer(cs.Properties.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy)
 		},
+		"AddonEnabled": func(addon string) bool {
```
CecileRobertMichon: where is this used?

jackfrancis: cruft!

parts/k8s/kubernetesparams.t
jackfrancis: because it’s never relied upon in the updated implementation

pkg/acsengine/template_generator.go
jackfrancis: will fix

PR #3680, opened 2018-08-15
Golden image multiple k8s versions

CecileRobertMichon:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


CecileRobertMichon:
Running https://jenkins.azure-containers.io/view/acs-engine/job/k8s-deploy-linux-default-CSE-PR/ on this branch since 8/16 at 10:20am PST

parts/k8s/kubernetesinstalls.sh
```diff
@@ -121,7 +121,7 @@ function installContainerd() {
 	echo "Successfully installed cri-containerd..."
 }
 
-function extractHyperkube() {
+function pullHyperkube() {
     TMP_DIR=$(mktemp -d)
     retrycmd_if_failure 100 1 30 curl -sSL -o /usr/local/bin/img "https://acs-mirror.azureedge.net/img/img-linux-amd64-v0.4.6"
```
jackfrancis: Should we have a separate func to get the `img` binary dependency?

CecileRobertMichon: We could but it's only being used in this function for now so I would just be breaking the function down into two separate functions and always calling them one after the other. I also tried to limit the changes and since it was already in the same func I left it as is. If you think it's important I can do it, let me know.

PR #3688, opened 2018-08-16
Add dkms to manage nvidia gpu kmod compliation across different linux kernels

lachie83:
This PR adds DKMS to manage Nvidia GPU kmod compilation across different versions of the Linux kernel - https://wiki.archlinux.org/index.php/Dynamic_Kernel_Module_Support

Specifically it does the following
* Add dkms package
* Installs nividia drivers using dkms flag which registers the kmod in dkms
* Creates dkms.service in systemd that autoinstalls any missing registered kmods for the running version of the linux kernel
* This systemd unit is a run to complete job and only runs once when the VM boots

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -81,16 +81,18 @@ write_files:
 {{end}}
 
 {{if IsNSeriesSKU .}}
-- path: "/etc/systemd/system/dkms.service"
+- path: "/etc/systemd/system/nvidia-modprobe.service"
   permissions: "0644"
   owner: "root"
   content: |
     [Unit]
-    Description=Builds and install new kernel modules through DKMS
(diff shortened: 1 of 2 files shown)


jackfrancis:
@lachie83 my GPU tests passed. Ready to merge with your 👍

lachie83:
I'm still having issues with k8s <= 1.9 because this command isn't being run at boot `nvidia-modprobe -u -c0`. It only shows up after a reboot. Going to investigate. 

For reference - the device plugin is running this command in k8s >= 1.10 which is why it works there.

lachie83:
Updated the systemd unit and I'm seeing k8s <= 1.9 GPU kmod being mounted following a node reboot.

lachie83:
Ready for review @jackfrancis @sozercan @wbuchwalter

jackfrancis:
@lachie83 running another round of tests, /lgtm pending green

jackfrancis:
test signal is 💚

PR #3695, opened 2018-08-16
Remove security rules from NSG.

marrobi:
Please look at #3693 first, as this is a priority, although this PR could replace #3693 .

As per https://github.com/Azure/acs-engine/pull/3693 the allow_kube_tls rule may not be required. In addition there is no need to allow SSH or RDP through the NSG if there are no masters - they are annotated "Allow SSH traffic to master"/"Allow RDP traffic to master".

The case with masters is dealt with in k8s/kubernetesmasterresources.t as per:

```
 {{if not IsHostedMaster}}
      ,{{template "k8s/kubernetesmasterresources.t" .}}
    {{else}}
```

If a user needs to RDP or SSH onto nodes, then they should be aware they are opening the ports, and removing a layer of security.

In addition the SSH rule is flagged as a risk by Azure Security Centre.

Again I do not believe any adverse affect, although users may be relying on these rules to administer agent pool nodes.

**If applicable**:
- [ ] documentation
- [ ] unit tests

### DIFF
--- a/parts/k8s/kubernetesbase.t
+++ b/parts/k8s/kubernetesbase.t
@@ -126,52 +126,7 @@
       "location": "[variables('location')]",
       "name": "[variables('nsgName')]",
       "properties": {
-        "securityRules": [
-{{if .HasWindows}}
-          {
-            "name": "allow_rdp",
-            "properties": {
-              "access": "Allow",
-              "description": "Allow RDP traffic to master",
-              "destinationAddressPrefix": "*",
-              "destinationPortRange": "3389-3389",
-              "direction": "Inbound",
-              "priority": 102,
-              "protocol": "Tcp",
-              "sourceAddressPrefix": "*",
-              "sourcePortRange": "*"
-            }
-          },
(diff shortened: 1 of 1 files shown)


jackfrancis:
@marrobi Would changing the `priority` of these built-in rules allow additional rules that collide with 443/22 to work?

marrobi:
I don't believe so as all the rules in question (existing and additional) are allow rules. Priority will only work if need to overide a deny or vice vera?

I guess they could possibly be restricted to the node IP, rather than Any (hence wouldn't affect the load balancer). Actually, as the nodes don't have public IPs (is there an instance where they do, unless added manually? ) these rules are redundant anyway...?

In my opinion all traffic inbound into the VNet from outside should be denied by default anyway if not needed for normal operation.

Get Outlook for Android<https://aka.ms/ghei36>

CecileRobertMichon:
Closing since this was covered by https://github.com/Azure/acs-engine/pull/4164

PR #3709, opened 2018-08-20
[WIP] Using dep 0.5.0 and adding dep check to the CI/CD cycle

### LINKED ISSUE #3702: Add dep check to CI/CD cycle
Per the latest go-dev docker image, we are using the latest release of dep which includes the check feature.

dep check essentially ensures that vendor directories, source imports, and the toml files are in sync. It works fast and it is much better than calling dep ensure and performing a diff.

Possible scenarios when this could be useful

I) When a developer adds a new dependency but forgets to check in the vendor files.
ii) When a dev adds new imports and it only compiles in their local machine (because of their local gopath) but forgets to add the corresponding the dep stuff. dep check will detect this loss of sync immediately.
iii) Inconsistencies caused due to conflicting transitive dependencies.

If you are okay with this, I can proceed with the required changes, :). More details over here

https://github.com/golang/dep/pull/1932

tariq1890:
**What this PR does / why we need it**: The go-dev docker image now uses dep 0.5.0. The acs-engine docker uses this version as well. dep check is a useful command to use in our CI/CD cycle.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3702 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


tariq1890:
Looks like there are multiple ways to resolve this. Would need to reach out to the golang/dep maintainers for more insight.

PR #3721, opened 2018-08-21
Kubernetes: Use AKS curated OS images by default

jackfrancis:
**What this PR does / why we need it**: Uses AKS-curated OS images by default.

**If applicable**:
- [x] documentation
- [ ] unit tests

**Release note**:

```
Kubernetes: Use AKS curated OS images by default
```


CecileRobertMichon:
/hold

will update image version to 0.10.0 once it's published

neurot1cal:
Is there any documentation around what this new Distribution contains? Is this a fork of Ubuntu or an entirely new Distribution? Will this apply to acs-engine nodes as well, or only AKS?

CecileRobertMichon:
@neurot1cal not yet, we are still in testing phase. The image is based on Ubuntu 16.04-LTS  (same as previous default) and contains pre-installed software needed for deployment that was previously fetched at provisioning time. The objective is to reduce deployment time (significantly) and eliminate transient errors due to general download flakiness. The plan is to enable this for both acs-engine Linux nodes (at first) and AKS nodes (after baking in acs-engine for a bit).

docs/clusterdefinition.md
```diff
@@ -479,7 +479,7 @@ We consider `kubeletConfig`, `controllerManagerConfig`, `apiServerConfig`, and `
...
-| distro                       | no                                        | Select Master(s) Operating System (Linux only). Currently supported values are: `ubuntu` and `coreos` (CoreOS support is currently experimental). Defaults to `ubuntu` if undefined. Currently supported OS and orchestrator configurations -- `ubuntu`: DCOS, Docker Swarm, Kubernetes; `RHEL`: OpenShift; `coreos`: Kubernetes. [Example of CoreOS Master with CoreOS Agents](../examples/coreos/kubernetes-coreos.json) |
+| distro                       | no                                        | Select Master(s) Operating System (Linux only). Currently supported values are: `ubuntu, `aks` and `coreos` (CoreOS support is currently experimental). Defaults to `aks` if undefined. `aks` is a custom image based on `ubuntu` that comes with pre-installed software necessary for Kubernetes deployments. Currently supported OS and orchestrator configurations -- `ubuntu` and `aks`: DCOS, Docker Swarm, Kubernetes; `RHEL`: OpenShift; `coreos`: Kubernetes. [Example of CoreOS Master with CoreOS Agents](../examples/coreos/kubernetes-coreos.json) |
```
mboersma: A backtick was lost here: should be

...values are: \`ubuntu\`, \`aks\`...

CecileRobertMichon: ah thanks will fix :)

pkg/acsengine/defaults.go
```diff
@@ -590,7 +590,7 @@ func setAgentNetworkDefaults(a *api.Properties, isUpgrade, isScale bool) {
 		if !a.OrchestratorProfile.IsOpenShift() {
 			// Set default Distro to Ubuntu
```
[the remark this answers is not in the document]
CecileRobertMichon: I just removed them, I don't think it was a very useful comment

CecileRobertMichon:
@neurot1cal this is in master if you want to give it a try :)

CecileRobertMichon:
@ablekh this is already on in AKS worldwide as of mid-October. There are still more improvements coming as this is a continual process to our commitment to deliver fast, reliable deployments. Unfortunately AKS cluster deletion is not affected by this change and not related to the acs-engine project which only does cluster provisioning. AKS handles on the service side (some improvements might be coming there too).

ablekh:
@CecileRobertMichon Thank you so much for your update. Hmm ... then, based on the time frame that you mentioned, unfortunately, it means that this change was definitely not nearly enough to get cluster creation speed down to feasible numbers (please see details in my issue referenced above). Having said that, I definitely appreciate the team's commitment and hard work. I hope that the goal of achieving fast and reliable AKS deployments will be reached sooner, rather than later. :-) And, for that, perhaps, some changes beyond `acs-engine` are needed. Re: cluster deletion - Understood. I hope that relevant team(s) at MS will address this issue as well.

CecileRobertMichon:
@ablekh out of curiosity what range of durations are you seeing in AKS deployments lately? Have you noticed any improvements recently? Agreed that more work is needed to get there, this effort was just one part of it. After this change, we've observed average cluster deployments in acs-engine drop to ~4 minutes for 3 nodes and ~7 minutes for 55 nodes. AKS create is built on top off acs-engine but there are more operations that happen in AKS provisioning than in acs-engine so it needs additional time.

ablekh:
@CecileRobertMichon Well, the number that I shared in my issue 711 referenced above, unfortunately, apply largely to end of October - beginning of November time frame. Hence, my original comment here. Almost always, I create a 3-node cluster on AKS (and then scale as needed), so my numbers refer to that size. Based on my recollection, an average time for creating a 3-node AKS cluster that I experienced was ~20 min. IMO, 80% of the total time spent on other (non-`acs-engine`) operations seems like way too much to me ...

ablekh:
@CecileRobertMichon You're very welcome. And thank you very much, too. :-)

PR #3724, opened 2018-08-22
Handle multiple CNI versions in golden image

CecileRobertMichon:
**What this PR does / why we need it**: Allow installing multiple cni versions on custom image and keeping one at provisioning

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/packer/vhd-image-builder.json
+++ b/packer/vhd-image-builder.json
@@ -61,7 +61,7 @@
         {
             "type": "shell",
             "inline": [
-                "sudo /bin/bash -eux /home/packer/install-dependencies.sh",
+                "sudo /bin/bash -ux /home/packer/install-dependencies.sh",
                 "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
                 "rm /home/packer/*.sh"
             ]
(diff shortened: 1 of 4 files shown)


packer/vhd-image-builder.json
```diff
@@ -61,7 +61,7 @@
         {
             "type": "shell",
             "inline": [
-                "sudo /bin/bash -eux /home/packer/install-dependencies.sh",
+                "sudo /bin/bash -ux /home/packer/install-dependencies.sh",
```
CecileRobertMichon: we don't want the script to exit at the first error since there are retries

parts/k8s/kubernetesinstalls.sh
```diff
@@ -99,36 +98,45 @@ function installClearContainersRuntime() {
...
-        fi
+    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
+        installAzureCNI
     fi
+    installCNI
+}
+
+function downloadCNI() {
+    mkdir -p $CNI_DOWNLOADS_DIR
+    CNI_TGZ_TMP=$(echo ${CNI_PLUGINS_URL} | cut -d "/" -f 5)
```
jackfrancis: lol

CecileRobertMichon: is that bad? 😬

jackfrancis: Laughter is the best medicine, so very very good.

PR #3727, opened 2018-08-22
Adding use cases and example for WinRM extension

PatrickLang:

After discussion with @jsturtevant , I thought it would be good to update the doc to show how to use this extension to connect from the Linux master nodes to a Windows agent node.

I also noticed that an explicit firewall rule wasn't set. I fixed that so that this extension won't break when the firewall is reenabled soon.

### DIFF
--- a/extensions/winrm/README.md
+++ b/extensions/winrm/README.md
@@ -2,6 +2,29 @@
 
 This extension enabled WinRM with HTTPS and a self-signed certificate on each Windows node. This makes it possible to troubleshoot Windows nodes from the master over ssh, without needing Remote Desktop.
 
+
+## Use Cases
+
+### Remote Shells
+
+PowerShell can be used to connect to the Windows node from the same Azure vNet. If you want to connect from the Linux master, then it's easiest to run PowerShell as a container:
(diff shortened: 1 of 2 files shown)


extensions/winrm/README.md
```diff
@@ -2,6 +2,29 @@
...
+# This will prompt for the Windows node's username & password
+$credential = Get-Credential
+
+# Now connect
+Enter-PsSession <hostname> -Credential $credential -Authentication Basic -UseSSL
+```
+
+### Log Gathering
+
 If you want to pull all the logs off Windows nodes quickly, deploy with this extension, then SSH to the master node and run [logslurp](https://github.com/PatrickLang/logslurp) to gather them all at once.
```
jsturtevant: Can you add an example of how to to use this from inside the master?  Is there a docker image we can use?  The logslurp docs require you to build your own image...

CecileRobertMichon:
@PatrickLang closing this as part of the acs-engine PR queue cleanup, please reopen in https://github.com/Azure/acs-engine/pull if it's still relevant!

PR #3732, opened 2018-08-23
Return deployment result when available in an error condition.

JunSun17:
Currently when an error happens in ARM deployment, the future is not checked and an empty deployment result is returned. This PR tries to retrieve deployment result future and returns it when available.

### DIFF
--- a/pkg/armhelpers/deployments.go
+++ b/pkg/armhelpers/deployments.go
@@ -24,14 +24,14 @@ func (az *AzureClient) DeployTemplate(ctx context.Context, resourceGroupName, de
 
 	// wait till future completes
 	e1 := future.WaitForCompletion(ctx, az.deploymentsClient.Client)
-	if err != nil && e1 != nil {
+	if e1 != nil {
 		log.Errorf("Wait for future to complete failed: %+v.", e1)
 		return resources.DeploymentExtended{}, err
 	}
 
 	// get future results
 	deploymentResult, e2 := future.Result(az.deploymentsClient)
-	if err != nil && e2 != nil {
+	if e2 != nil {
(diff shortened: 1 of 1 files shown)


pkg/armhelpers/deployments.go
```diff
@@ -21,25 +21,26 @@ func (az *AzureClient) DeployTemplate(ctx context.Context, resourceGroupName, de
 
 	log.Infof("Starting ARM Deployment (%s). This will take some time...", deploymentName)
 	future, err := az.deploymentsClient.CreateOrUpdate(ctx, resourceGroupName, deploymentName, deployment)
-	if err != nil {
-		return de, err
+
+	// wait till future completes
+	e1 := future.WaitForCompletion(ctx, az.deploymentsClient.Client)
+	if err != nil && e1 != nil {
+		return resources.DeploymentExtended{}, err
```
mboersma: Is there never a case where `az.deploymentsClient.CreateOrUpdate` succeeds but an error is returned from `future.WaitForCompletion`? 

We never return the `e1` or `e2` error so I worried that might lose some information. Should we at least log the error we don't return?

mboersma: If an `err` is returned from `az.deploymentsClient.CreateOrUpdate`, is it safe to keep using the `future` instance as we do here?

mboersma: That makes sense. If we're sure calling `future.WaitForCompletion` will be safe even when `err` was returned (I tried to follow the code but couldn't quite locate the implementation in autorest-go), and if `e1` and `e2` are logged, then :+1:.

JunSun17:
The original code behavior matches the sample provided by Azure Go SDK. Will close this PR for now.

PR #3744, opened 2018-08-24
Creating kube manifest dirs on node provision.

### LINKED ISSUE #3506: create /etc/kubernetes/manifests on worker nodes
Lots of spurious kubelet messages:

```
Jul 18 18:13:05 k8s-agentmd-34767551-2 kubelet[8431]: E0718 18:13:05.032451    8431 file_linux.go:61] Unable to read config path "/etc/kubernetes/manifests": path does not exist, ignoring```

### PR DESCRIPTION
**Pull request by tariq1890**
**What this PR does / why we need it**: Fixes #3506 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -67,6 +67,7 @@ installNetworkPlugin
 installContainerd
 extractHyperkube
 ensureRPC
+createKubeManifestDir
 
 if [[ ! -z "${MASTER_NODE}" ]]; then
     configureEtcd
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -263,6 +263,11 @@ function ensureEtcd() {
     retrycmd_if_failure 120 5 10 curl --cacert /etc/kubernetes/certs/ca.crt --cert /etc/kubernetes/certs/etcdclient.crt --key /etc/kubernetes/certs/etcdclient.key ${ETCD_CLIENT_URL}/v2/machines || exit $ERR_ETCD_RUNNING_TIMEOUT
 }
 
+function createKubeManifestDir() {
+    KUBEMANIFESTDIR=/etc/kubernetes/manifests
+    mkdir -p $KUBEMANIFESTDIR
+}
+
 function writeKubeConfig() {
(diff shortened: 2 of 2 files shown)


jackfrancis:
@CecileRobertMichon your thoughts on if there's somewhere more elegant to put this? It seems like this directory is being created on master implicitly by cloud-init because it has files that are in there (i.e., cloud-init rationalizes a filepath that doesn't yet exist by creating directories if needed)

CecileRobertMichon:
I think we should put this in custom data (cloud-init), it could be part of the runcmd (https://github.com/Azure/acs-engine/blob/master/parts/k8s/kubernetesagentcustomdata.yml#L219) or in "write_files" if it's possible to create a directory

jackfrancis:
I read through this and it suggests we write_files isn't able to create an empty dir:

https://cloudinit.readthedocs.io/en/latest/topics/examples.html#

jackfrancis:
`runcmd` is too non-deterministic as far as I'm concerned

parts/k8s/kubernetescustomscript.sh
```diff
@@ -67,6 +67,7 @@ installNetworkPlugin
 installContainerd
 extractHyperkube
 ensureRPC
+createKubeManifestDir
```
jackfrancis: Note: if we *do* decide to keep this in CSE, let's move this fun invocation into the `if $FULL_INSTALL_REQUIRED; then` block. We should create this directory during VHD image creation process.

CecileRobertMichon: `if $FULL_INSTALL_REQUIRED;` is for deployment not using the AKS image so I don't think we should move it, we want this to run on every deployment

CecileRobertMichon:
Keeping it in CSE is probably fine then

CecileRobertMichon:
If we want to do this during VHD creation we'll need to call the function from https://github.com/Azure/acs-engine/blob/master/packer/install-dependencies.sh

CecileRobertMichon:
@jackfrancis thinking about this more, I don't think we need to do this in VHD since mkdir is a primitive operation and we don't expect any delay/flakiness from it. I'm open to being convinced of the opposite though.

PR #3745, opened 2018-08-24
Add new Azure client from a tenant

sozercan:
We are using this in the RP for getting an azure client from a specific tenant.
I see other clients are used in the CLI, I can add this there as well if needed

### DIFF
--- a/pkg/armhelpers/azureclient.go
+++ b/pkg/armhelpers/azureclient.go
@@ -157,6 +157,26 @@ func NewAzureClientWithClientSecret(env azure.Environment, subscriptionID, clien
 	return getClient(env, subscriptionID, tenantID, armSpt, graphSpt), nil
 }
 
+// NewAzureClientWithClientSecretExternalTenant returns an AzureClient via client_id and client_secret from a tenant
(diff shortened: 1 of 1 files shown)


pkg/armhelpers/azureclient.go
```diff
@@ -157,6 +157,26 @@ func NewAzureClientWithClientSecret(env azure.Environment, subscriptionID, clien
 	return getClient(env, subscriptionID, tenantID, armSpt, graphSpt), nil
 }
 
+// NewAzureClientWithClientSecretExternalTenant returns an AzureClient via client_id and client_secret from a tenant
+func NewAzureClientWithClientSecretExternalTenant(env azure.Environment, subscriptionID, tenantID, clientID, clientSecret string) (*AzureClient, error) {
```
CecileRobertMichon: when is this new function going to get called?

sozercan: It is called from RP, nothing from acs-engine side directly

amanohar: @CecileRobertMichon this function is going to be used by the RP. We add it here as all the Azure client initialization code and functions are in ACS E so it makes sense keep this consolidated in one place with the added advantage of being able to reuse common functions. Can we get a sign off?

PR #3749, opened 2018-08-27
AzureCNI/Custom VNET - Remove ip address network configuration race condition

f4tq:
Case handled:  AzureCNI/Custom Vnet

[There is a documented race condition](https://github.com/Azure/acs-engine/blob/master/docs/kubernetes/features.md) whereby it is impossible to ensure that master ip addresses are allocated serially as REQUIRED by k8 bootstrap scripts.  I hit this problem every time i.e. the cluster fails to start. 

This PR guarantees the 'MasterCount' ip addresses are allocated serially beginning with *firstConsecutiveStaticIP* **before** dynamic ip allocation begins. 

The generated dynamic ips stanza are made dependent on the static ip (masterCount) stanzas.
  
To implement this, both the static and dynamic ips are in separate arm stanzas.
The static network interfaces reference the lb and intlb(for masterCount> 1) as before.

It passes all tests and works predictably with automation onto existing subnets.

Tested with v0.20.7 and v0.21.1

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ x] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

Strike 
```
If possible, assign to the firstConsecutiveStaticIP configuration property an IP address that is near the "end" of the available IP address space in the desired subnet.
For example, if the desired subnet is a /24, choose the "239" address in that network space
```
And change to: 
```Ensure that firstConsecutiveStaticIP + (masterCount-1) ip addresses are free within the master subnet as required by k8 installation scripts.```

### DIFF
--- a/parts/k8s/kubernetesmasterresources.t
+++ b/parts/k8s/kubernetesmasterresources.t
@@ -1,3 +1,4 @@
+{{ $Context := . }}
 {{if .MasterProfile.IsManagedDisks}}
     {
       "apiVersion": "[variables('apiVersionStorageManagedDisks')]",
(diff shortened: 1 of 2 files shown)


f4tq:
Oh well, Adobe won't let me give CircleC access to all my private repositories.  But who would anyway?

Maybe someone who's forked over all their private repos can take this PR and submit it.

See ya.  I tried...

PR #3756, opened 2018-08-28
Make cmd more testable by abstracting authArgs and AzureClient

serbrech:
Improve testability of the cmd package by abstracting the authArgs dependency.  
This allows tests to stub the `authArgs.getClient()` call and return a fake `ACSEngineClient` instead of the concrete `AzureClient` instance

If this is ok, the same pattern can be applied for other command that depends on the `AzureClient`.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -44,8 +44,7 @@ const (
 )
 
 type deployCmd struct {
-	authArgs
-
+	authProvider
 	apimodelPath      string
 	dnsPrefix         string
 	autoSuffix        bool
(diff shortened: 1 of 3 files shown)


jackfrancis:
@serbrech FYI, this looks fine, but it occured to me that our E2E doesn't traverse `bin/acs-engine deploy` and instead uses `bin/acs-engine generate` + `az group deployment create`.

Let's use this as an excuse to build a 1st class `bin/acs-engine deploy` E2E test. Stand by... :)

serbrech:
added #3776

Issue #3632, opened 2018-08-08
Improve Windows node e2e tests for K8


jsturtevant:
**Is this a request for help?**: No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): FEATURE REQUEST

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes

Improve e2e tests for Windows containers.  After talking with @PatrickLang around DNS issues, I did a review of the e2e tests for windows containers and notice a quite a few missing scenarios.  Opening this PR to start the work of adding more e2e test and get feedback on tests we should plan on including.

List of improvements and additions planned:

- [X] Include a linux agent node in end to end tests.  This will enable testing of other scenarios down the line.  
- [x] Enable DNS checks that are [disabled during windows](https://github.com/Azure/acs-engine/blob/e053cb099487fdd91fbdfb14413c37d5186f6ee7/test/e2e/kubernetes/kubernetes_test.go#L84) only tests.  
- [X] Port above tests that make sense to a windows node 
- [X] Test outbound Connectivity [already implemented](https://github.com/Azure/acs-engine/blob/e053cb099487fdd91fbdfb14413c37d5186f6ee7/test/e2e/kubernetes/kubernetes_test.go#L788)
- [X] Windows to Windows network connectivity via DNS
- [X] Windows to Linux network connectivity via DNS
- [X] Linux to Windows network connectivity via DNS
- [x] ~~After pod DNS is confirmed working kill the pod and try DNS again~~ by using stability iterations for scenarios above this should be covered
- [x] Enable [after awhile DNS check](https://github.com/Azure/acs-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L694) when have windows nodes.  
- [ ] Add [after awhile DNS check](https://github.com/Azure/acs-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L694) tests for dns lookups for windows containers.
- [ ] Add HPA auto tests ([currently linux only](https://github.com/Azure/acs-engine/blob/e053cb099487fdd91fbdfb14413c37d5186f6ee7/test/e2e/kubernetes/kubernetes_test.go#L546))

**Anything else we need to know**:
Will implement iteratively and submit a pull request as each test comes online to merge.

jsturtevant:
@PatrickLang could you assign me to this issue

PR #3760, opened 2018-08-28
[WIP] Windows dns connectivity - e2e tests

jsturtevant:
adds unit tests for e2e scenarios in windows containers.  From issue #3632, specifically:

-  Windows to Windows network connectivity via DNS
-  Windows to Linux network connectivity via DNS
-  Linux to Windows network connectivity via DNS

I am currently getting error:

```
 Error trying to deploy iis-dns-kubernetes-eastus-55793-65288 [microsoft/iis:windowsservercore-1803] in namespace default:error: v1beta1.Deployment is not suitable for converting to "apps/v1beta1"
```

@PatrickLang  This appears to be related to 1.11.  Is it ok to move from `apiVersion": "extensions/v1beta1` in the windows deploy script to a newer apiversion?

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


jsturtevant:
The line of code that has the older version set is: https://github.com/Azure/acs-engine/blob/f019949ff5c333719376f3bf88000d7a5cd8de58/test/e2e/kubernetes/deployment/deployment.go#L104

I found that the `kubectl` version affects if the test will pass.  I was running 1.11 yesterday when I got the error above, on my other dev box I was running `kubectl` 1.9 and do not have the error.  

I assume at some point we will run into this during the e2e tests as we move up versions.   I found that I didn't need to pass the schema when creating a pod when writing tests for this pr here: https://github.com/jsturtevant/acs-engine/blob/1b5dd2353ef94350e7cc41fea887e08a0f8074b0/test/e2e/kubernetes/pod/pod.go#L141.  

Removing the schema might be easy way to make the tests work across versions.

@jackfrancis @CecileRobertMichon    any thoughts on the version for the deployments?

jsturtevant:
A few notes for this PR:

## Stability iterations
STABILITY_ITERATIONS does not seem to be set on the CI builds.  Looking at the last few builds in Circle CI I am seeing this set to 0.    When I enabled it locally I found a bug on the DNS checks recently added by @jackfrancis. I've fixed and used the same pattern for checking the DNS on windows.  During that I refactored some common code (the `RunCommandMultipleTimes`) as it was being used in 4+ places.

@jackfrancis I can add the STABILITY_ITERATIONS  to the CircleCI config.yml file.  What is the target you would like?

## Api versions removed
I've also removed apiversion checks on in the `deployment.go` file to fix kubectl errors mentioned above.  This should resolve flaky tests based on `kubectl` version as mentioned above.  

## Windows DNS notes
@PatrickLang  During testing I ran into two interesting things in windows DNS:

1. `nslookup <svc-name>` does not resolve but   `nslookup <svc-name>.default.svc.cluster.local` does.  I ended up using `powershell iwr` as it was used elsewhere and i figured it more likely covers developer use case of calling HTTP call on service.
2. Using `curl` in 1803 image as a command would fail.  `powershell iwr` works.  For instance:

```
# doesn't work
kubectl run windows-1803-curl -n default --image microsoft/windowsservercore:1803 --image-pull-policy=IfNotPresent --restart=Never --overrides '{ "spec": {"nodeSelector":{"beta.kubernetes.io/os":"windows"}}}' --command -- curl nginx-dns-kubernetes-eastus-21376-93545

# works
kubectl run windows-1803-ps -n default --image microsoft/windowsservercore:1803 --image-pull-policy=IfNotPresent --restart=Never --overrides '{ "spec": {"nodeSelector":{"beta.kubernetes.io/os":"windows"}}}' --command -- powershell iwr -UseBasicParsing -TimeoutSec 60 nginx-dns-kubernetes-eastus-21376-93545
```

jackfrancis:
@jsturtevant thanks for this!

How many iterations did you need to fall into the bug you mentioned? It was set to zero intentionally for PR E2E tests so as not to suddenly introduce already existing transient flakiness that would block PR signal. But if the number is low (10?) we can consider turning that on in the near term.

jsturtevant:
@jackfrancis The bug was related to having windows nodes in the cluster. The linux only tests where being scheduled onto the windows node and failing.  

If STABILITY_ITERATIONS is disabled for PRs that sounds reasonable.  Though, I didn't see the STABILITY_ITERATIONS being set for `build_and_test_master` test runs either.  Is the expectation it is  being run there?

I think a low number (5? or 10?) would be fine to start with and we could adjust as needed.

jsturtevant:
@PatrickLang Could you give a review?  Thanks!

test/e2e/kubernetes/pod/pod.go
```diff
@@ -158,6 +159,70 @@ func RunLinuxPod(image, name, namespace, command string, printOutput bool) (*Pod
...
+		if succeeded {
+			successes++
+		}
+		err = p.Delete()
+		if err != nil {
+			return successes, err
+		}
+	}
+
+	log.Printf("Ran command successfully on %d of %d test attempts\n\n", successes, attempts)
```
PatrickLang: I think this would be good to `defer` logging the summary to consistently log how many iterations were attempted, desired, and succeeded even on failure. It would be good to know if there's a pattern that leads to failing at N of M attempts. On a failed run it won't say how many attempts were made, just that one of them failed

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -962,6 +881,79 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+
+				By("Exposing a internal service for the windows iis deployment")
+				err = windowsIISDeployment.Expose("ClusterIP", 80, 80)
+				Expect(err).NotTo(HaveOccurred())
+				windowsService, err := service.Get(windowsDeploymentName, "default")
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Connecting to Windows from another Windows deployment")
+				name := fmt.Sprintf("windows-2-windows-%s", cfg.Name)
+				command := fmt.Sprintf("iwr -UseBasicParsing -TimeoutSec 60 %s", windowsService.Metadata.Name)
```
PatrickLang: agree on using `Invoke-WebRequest`. It would be good to know if the network path fails too, which nslookup won't catch.

There are some known problems with DNS search suffices that were caught in Kubernetes conformance testing. We're still fixing those.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -962,6 +881,79 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				name = fmt.Sprintf("windows-2-linux-%s", cfg.Name)
+				command = fmt.Sprintf("iwr -UseBasicParsing -TimeoutSec 60 %s", linuxService.Metadata.Name)
+				successes, err = pod.RunCommandMultipleTimes(pod.RunWindowsPod, windowsServerImage, name, command, cfg.StabilityIterations)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(successes).To(Equal(cfg.StabilityIterations))
+
+				By("Connecting to Windows from Linux deployment")
+				name = fmt.Sprintf("linux-2-windows-%s", cfg.Name)
+				command = fmt.Sprintf("curl %s", windowsService.Metadata.Name)
+				successes, err = pod.RunCommandMultipleTimes(pod.RunLinuxPod, "radial/busyboxplus:curl", name, command, cfg.StabilityIterations)
```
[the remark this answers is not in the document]
jsturtevant: I've updated back to using the alpine images the rest of tests use with wget.

test/e2e/kubernetes/pod/pod.go
```diff
@@ -158,6 +159,70 @@ func RunLinuxPod(image, name, namespace, command string, printOutput bool) (*Pod
...
+			p, err = podRunnerCmd(image, podName, "default", command, true)
+		} else {
+			p, err = podRunnerCmd(image, podName, "default", command, false)
+		}
+		if err != nil {
+			return successes, err
+		}
+		succeeded, _ := p.WaitOnSucceeded(1*time.Second, 2*time.Minute)
+		cmd := exec.Command("kubectl", "logs", podName, "-n", "default")
+		out, err := util.RunAndLogCommand(cmd)
```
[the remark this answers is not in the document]
jsturtevant: This was from the original implementation. I could go either way.  Seems like we might want to see details if it fails and grabbing logs each round keeps the code simpler.  I'll leave it for now until we find it is to much

test/e2e/kubernetes/pod/pod.go
jsturtevant: Great idea.  I have updated it to defer and log the successful attempts, actual attempts, and desired attempts

jsturtevant:
@PatrickLang  I did a rebase to pull in the changes and addressed your questions/comments.  I think this is ready for final review.  

One final question I have is the setting for STABILITY_ITERATIONS.  It does not appear to be set for any of the builds.  If we don't get that set then none of this will be called.  Starting out with a small number seems to make sense to see how often we run into flaky tests.  I have been using a setting of 5 for my tests and haven't had any trouble.

PatrickLang:
Sorry for the delay here - have been behind the last few days working on plans for kubernetes v1.13 + Windows. Looks good, launching last tests

jsturtevant:
@PatrickLang it looks like it failed on the should autoscale test. I think it is a flaky test?

PatrickLang:
/LGTM

I restarted the failed test. I think it's flaky

PatrickLang:
@jackfrancis - is this a known test flake? 

```
Azure Container Cluster using the Kubernetes Orchestrator
/go/src/github.com/Azure/acs-engine/test/e2e/kubernetes/kubernetes_test.go:63
  with a linux agent pool
  /go/src/github.com/Azure/acs-engine/test/e2e/kubernetes/kubernetes_test.go:581
    should be able to autoscale [It]
    /go/src/github.com/Azure/acs-engine/test/e2e/kubernetes/kubernetes_test.go:624

    Expected error:
        <*errors.fundamental | 0xc0000fc5a0>: {
            msg: "Timeout exceeded (30m0s) while waiting for 2 Pod replicas from Deployment php-apache-kubernetes-southeastasia-31965-41004",
            stack: [0x70c22b, 0x45da91],
        }
        Timeout exceeded (30m0s) while waiting for 2 Pod replicas from Deployment php-apache-kubernetes-southeastasia-31965-41004
    not to have occurred
```

jackfrancis:
@PatrickLang we do see HPA being the highest rate of flake. I assume there's something there, just haven't root caused it yet.

jsturtevant:
@PatrickLang did the test run again or is that the failure from before?

PatrickLang:
/assign @tariq1890 

Can you merge? I restarted the failed test but don't think its related.

tariq1890:
@PatrickLang Can you check the latest failures? Pretty sure they are flakes, but i'd like to confirm. Also deferring this one to @jackfrancis as e2e's are not my forte :)

jsturtevant:
The failure on 1.9 is the cse timeout.  I believe that is a flake issue.

The failure on 1.11 is a timeout on provisioning external the IP addresses.  I haven't seen that one before. It looks like it failed for both the windows and linux tests.  Are there any known issues in 1.11 for provisioning external IP's?

jackfrancis:
Both are possible flakes, re-running.

jsturtevant:
Looks like they were flakes and this is ready to go :+1:

jackfrancis:
Thanks @jsturtevant!

PR #3763, opened 2018-08-29
Update to go 1.11 and fix new linting errors

mboersma:

Updates the go compiler used in the go-dev tools image and CircleCI to the [newest release](https://golang.org/doc/go1.11) and fixes the new `go vet` errrors it found.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Update to go 1.11 and fix new linting errors
```


mboersma:
Apparently `gofmt -s` and `goimports` now disagree about the level of indentation for some struct fields, so linting won't pass because they can't both be satisfied. I'll work it out or disable one of the linters.

pkg/helpers/helpers_test.go
```diff
@@ -174,12 +120,12 @@ EPDesL0rH+3s1CKpgkhYdbJ675GFoGoq+X21QaqsdvoXmmuJF9qq9Tq+JaWloUNq
 	pemBuffer := bytes.Buffer{}
 	pem.Encode(&pemBuffer, pemBlock)
 
-	if pemBuffer.String() != expectedPrivateKeyString {
-		t.Fatalf("Private Key did not match expected format/value")
+	if !strings.HasPrefix(pemBuffer.String(), "-----BEGIN RSA PRIVATE KEY-----") {
```
tariq1890: Can we add lines that test the following for the Private Key?

1. privateKey.N.BitLen() equals the SSHKey bit size (which is set at 4096 for us) 
2. privatKey.Validate() does not throw an err

mboersma: Good call--I added that.

mboersma:
This passed all CircleCI jobs previously, and the recent commit just added two unit test clauses, so I think this is good to merge.

PR #3764, opened 2018-08-29
Add build metadata to VHD

CecileRobertMichon:
**What this PR does / why we need it**: Add info for debugging custom images, including commit ID, build ID and build number.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -43,4 +43,6 @@ done
 
 df -h
 
-echo "Install completed successfully on " `date` > /var/log/azure/golden-image-install.complete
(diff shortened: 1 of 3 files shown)


packer/install-dependencies.sh
```diff
@@ -3,6 +3,10 @@
 source /home/packer/provision_installs.sh
 source /home/packer/provision_source.sh
 
+echo "\nVSTS Build NUMBER: ${1}" >> /var/log/azure/golden-image-install.complete
```
jackfrancis: Can we make these named env variables instead of ordinal arguments?

PR #3627, opened 2018-08-07
Kubernetes: variables --> parameters

jackfrancis:
**What this PR does / why we need it**: This PR proves out two patterns:

- converge flat params structure into leaner, object representation
  - done for `cloudProviderBackoffConfig`
- refer explicitly to params instead of vars (especially vars that are nothing more than thinly-wrapped params)
  - demonstrated here w/ `etcdVersion`
  - introduces a new `WrapAsParameter` template function for direct access to params from template processing
  

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [x] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
NONE
```

PR #3774, opened 2018-08-31
bool ARM params need toLower

jackfrancis:
**What this PR does / why we need it**: Simply referencing a boolean ARM parameter like this:

`parameters('cloudproviderConfig').cloudProviderBackoff`

Will yield a `True` string value (note the uppercase "T").

Recently a change was made in how we dispatch dynamic properties to ARM template JSON and many vars were converted to params as a way of optimizing overhead. This on made the change on the backoff vars specifically:

https://github.com/Azure/acs-engine/pull/3627

That change seems to have induced a change from "true" to "True" in the rendered value @ `/etc/kubernetes/azure.json` on node vms, effectively turning off both rate limiting and backoff (cloudprovider is apparently not resilient against an uppercase "True").

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
bool ARM params need toLower
```


jackfrancis:
See here for the `toLower` solution used:

https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-string#tolower

jackfrancis:
validated:

```
azureuser@k8s-master-27894273-0:~$ sudo cat /etc/kubernetes/azure.json | grep cloudProviderBackoff
    "cloudProviderBackoff": true,
    "cloudProviderBackoffRetries": 6,
    "cloudProviderBackoffExponent": 1.5,
    "cloudProviderBackoffDuration": 5,
    "cloudProviderBackoffJitter": 1,
azureuser@k8s-master-27894273-0:~$ sudo cat /etc/kubernetes/azure.json | grep cloudProviderRatelimit
    "cloudProviderRatelimit": true,
```

Issue #3776, opened 2018-08-31
Missing acs-engine deploy e2e test


serbrech:
e2e tests always set the cluster up using `acs-engine generate`, followed by a `az deployment create` on the generated arm template. https://github.com/Azure/acs-engine/blob/708c688ba25c89fbf605428d46739ee39c44cfc8/test/e2e/runner/cli_provisioner.go#L133

Q: Should all functional tests be run on both `generate` and `deploy` setup? 

- [ ] add Deploy cmd to [cli.go](https://github.com/Azure/acs-engine/blob/master/test/e2e/engine/cli.go)
- [ ] parameterize provisioner to use deploy or generate setup
- [ ] refactor cli_provisioner to allow using the deploy command
- [ ] add deploy e2e test to ci

PR #3777, opened 2018-08-31
Add acs-engine deploy e2e test

serbrech:
see #3776 and #3756
e2e tests always set the cluster up using `acs-engine generate`, followed by a `az deployment create` on the generated arm template. https://github.com/Azure/acs-engine/blob/708c688ba25c89fbf605428d46739ee39c44cfc8/test/e2e/runner/cli_provisioner.go#L133

Q: Should all functional tests be run on both `generate` and `deploy` setup? 

- [x] add Deploy cmd to [cli.go](https://github.com/Azure/acs-engine/blob/master/test/e2e/engine/cli.go)
- [x] parameterize provisioner to use deploy or generate setup
- [x] refactor cli_provisioner to allow using the deploy command
- [ ] add deploy e2e test to ci

### DIFF
--- a/test/e2e/config/config.go
+++ b/test/e2e/config/config.go
@@ -85,6 +85,9 @@ func (c *Config) SetKubeConfig() {
 
 // GetSSHKeyPath will return the absolute path to the ssh private key
 func (c *Config) GetSSHKeyPath() string {
+	if c.UseDeployCommand {
+		return filepath.Join(c.CurrentWorkingDir, "_output", c.Name, "azureuser_rsa")
+	}
 	return filepath.Join(c.CurrentWorkingDir, "_output", c.Name+"-ssh")
 }
 
--- a/test/e2e/config/config.go
+++ b/test/e2e/config/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	CurrentWorkingDir   string
 	SoakClusterName     string `envconfig:"SOAK_CLUSTER_NAME"`
 	ForceDeploy         bool   `envconfig:"FORCE_DEPLOY"`
+	UseDeployCommand    bool   `envconfig:"USE_DEPLOY_COMMAND"`
 }
 
 const (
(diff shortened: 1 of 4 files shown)


serbrech:
There is currently an annoying difference between the `deploy` and `generate` path.

The existing e2e tests use generate. The test code takes care of generating the ssh keys and inject the public key in the `apimodel.json`

The deploy command on the other hand, includes the generation of the ssh key if there is none in the base apimodel. the name and location of that key is not parameterizable from the command line. it needs to be in the original apimodel.

Since both test path use the same apimodel examples, I added a couple `if/else` to take care of it [in this commit](https://github.com/Azure/acs-engine/pull/3777/commits/6d54d7e7e5c0ebe9047232fbae1db19f92602117)
A better way would be to add a command line parameter to provide the ssh key, and generate it the same way for both test path. 

This quirk might also just disappear if we refactor the `deploy` cmd to reuse the `generate` cmd logic.

To run the e2e tests using the deploy command, the CI needs to add `USE_DEPLOY_COMMAND=true` as environment variable parameter : 

i.e : 

```console
CLUSTER_DEFINITION=examples/kubernetes.json `
CLEANUP_ON_EXIT=false `
SUBSCRIPTION_ID=95a2efa9-87b0-4337-a4fe-6ac6e325b0c3 `
CLIENT_ID=85f44b3c-61c2-4d9d-8d1f-edf073067f4e `
CLIENT_SECRET=514287d6-64fd-45c5-a10f-65af0b085006 `
TENANT_ID=72f988bf-86f1-41af-91ab-2d7cd011db47 `
LOCATION=westeurope `
NAME="" ` #this needs to be defined and empty for the tests to create a new cluster. Most machine have name defined as their HostName :)
USE_DEPLOY_COMMAND=true `
make test-kubernetes
```

jackfrancis:
And btw, rationalizing `generate` and `deploy` execution paths so that they are 100% functionally equivalent is the eventual goal.

PR #3778, opened 2018-08-31
only test outbound connection if not using VHD image

jackfrancis:
**What this PR does / why we need it**: // TODO audit VHD path to ensure no remaining internet-dependent I/O is required.

Exception is for GPU nodes, for which we will address separately by adding that cluster configuration to the packer CI implementation.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
only test outbound connection if not using VHD image
```


jackfrancis:
Moved here: https://github.com/Azure/acs-engine/pull/3781

PR #3780, opened 2018-08-31
Remove exit code 50

jackfrancis:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
NONE
```

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -41,6 +41,86 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     pullHyperkube
 done
 
+DASHBOARD_VERSIONS="1.8.3 1.6.3"
+for DASHBOARD_VERSION in ${DASHBOARD_VERSIONS}
+    pullImg "k8s.gcr.io/kubernetes-dashboard-amd64:v${DASHBOARD_VERSION}"
+done
+
+EXECHEALTHZ_VERSIONS="1.2"
+for EXECHEALTHZ_VERSION in ${EXECHEALTHZ_VERSIONS}
+    pullImg "k8s.gcr.io/exechealthz-amd64:${EXECHEALTHZ_VERSION}"
+done
+
+ADDON_RESIZER_VERSIONS="1.8.1 1.7"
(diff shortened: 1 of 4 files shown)


parts/k8s/kubernetesconfigs.sh
```diff
@@ -242,7 +242,6 @@ function ensureDocker() {
     DOCKER_JSON_FILE=/etc/docker/daemon.json
     wait_for_file 1200 1 $DOCKER_JSON_FILE || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart docker
-    retrycmd_if_failure 6 1 10 docker pull busybox # pre-pull busybox, but don't exit if fail
```
CecileRobertMichon: Should we run this if golden image isn't being used?

CecileRobertMichon: (in the `if $FULL_INSTALL_REQUIRED; then ` block)

jackfrancis: It was just a convenience (no error thrown). We'll assume use-cases that don't use the VHD image will have outbound internet access, and so the probe script's usage of `docker run busybox` will simply download the image first.

packer/install-dependencies.sh
```diff
@@ -41,6 +41,88 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     pullHyperkube
 done
 
+DASHBOARD_VERSIONS="1.8.3 1.6.3"
+for DASHBOARD_VERSION in ${DASHBOARD_VERSIONS}
+    pullImg "k8s.gcr.io/kubernetes-dashboard-amd64:v${DASHBOARD_VERSION}"
```
CecileRobertMichon: where is the "not golden image" equivalent for this?

jackfrancis: There is none. This is additional foo for VHD images: when the k8s cluster comes online its component images will already be local. For non-VHD use-cases the functionality will be as it is today, where the initial bootstrapping process will include fetching docker images.

PR #3781, opened 2018-08-31
CSE: Remove outbound internet access requirement for VHD installs

jackfrancis:
**What this PR does / why we need it**: // TODO audit VHD path to ensure no remaining internet-dependent I/O is required.

Exception is for GPU nodes, for which we will address separately by adding that cluster configuration to the packer CI implementation.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
CSE: Remove outbound internet access requirement for VHD installs
```


packer/install-dependencies.sh
```diff
@@ -33,6 +33,88 @@ installContainerd
...
+for DASHBOARD_VERSION in ${DASHBOARD_VERSIONS}; do
+    pullImg "k8s.gcr.io/kubernetes-dashboard-amd64:v${DASHBOARD_VERSION}"
+done
+
+EXECHEALTHZ_VERSIONS="1.2"
+for EXECHEALTHZ_VERSION in ${EXECHEALTHZ_VERSIONS}; do
+    pullImg "k8s.gcr.io/exechealthz-amd64:${EXECHEALTHZ_VERSION}"
+done
+
+ADDON_RESIZER_VERSIONS="1.8.1 1.7"
```
CecileRobertMichon: why is one with a patch version and the other without? Is that a thing?

packer/install-dependencies.sh
```diff
@@ -33,6 +33,88 @@ installContainerd
...
+for TILLER_VERSION in ${TILLER_VERSIONS}; do
+    pullImg "gcr.io/kubernetes-helm/tiller:v${TILLER_VERSION}"
+done
+
+RESCHEDULER_VERSIONS="0.4.0 0.3.1"
+for RESCHEDULER_VERSION in ${RESCHEDULER_VERSIONS}; do
+    pullImg "k8s.gcr.io/rescheduler:v${RESCHEDULER_VERSION}"
+done
+
+VIRTUAL_KUBELET_VERSIONS="latest"
```
CecileRobertMichon: are we not pinning to a specific version on purpose here?

packer/install-dependencies.sh
jackfrancis: cf. https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/k8s_versions.go#L15
cf. https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/k8s_versions.go#L142

packer/install-dependencies.sh
jackfrancis: cf. https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/k8s_versions.go#L24

packer/install-dependencies.sh
CecileRobertMichon: oh weird

PR #3784, opened 2018-09-01
[WIP] Install GPU drivers in VHD

CecileRobertMichon:
**What this PR does / why we need it**: Pre-install GPU drivers in VHD.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -113,6 +113,7 @@ fi
 
 if [[ "${GPU_NODE}" = true ]]; then
     installGPUDrivers
+    ensureGPUDrivers
 fi
 
 if $FULL_INSTALL_REQUIRED; then
--- a/parts/k8s/kubernetesinstalls.sh
+++ b/parts/k8s/kubernetesinstalls.sh
@@ -26,43 +26,48 @@ function installDeps() {
 }
 
 function installGPUDrivers() {
-    # latest version of the drivers. Later this parameter could be bubbled up so that users can choose specific driver versions.
-    DV=396.26
-    DEST=/usr/local/nvidia
-	NVIDIA_DOCKER_VERSION=2.0.3
-	DOCKER_VERSION=1.13.1-1
-	NVIDIA_CONTAINER_RUNTIME_VERSION=2.0.0
-    # First we remove the nouveau drivers, which are the open source drivers for NVIDIA cards. Nouveau is installed on NV Series VMs by default.
-	# We also installed needed dependencies.
-    rmmod nouveau
-    echo blacklist nouveau >> /etc/modprobe.d/blacklist.conf
-    update-initramfs -u
-    mkdir -p $DEST
-    cd $DEST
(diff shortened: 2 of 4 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -113,6 +113,7 @@ fi
 
 if [[ "${GPU_NODE}" = true ]]; then
     installGPUDrivers
+    ensureGPUDrivers
```
tariq1890: Do we need another "gpu driver installed succesfully check" before we invoke ensureGPUDrivers?

CecileRobertMichon: @tariq1890 not sure what you mean, installGPUDrivers is called right before and that should check that the driver is installed and install it if not.

tariq1890: Is there a possibility of installGPUDrivers failing?

CecileRobertMichon: yes in which case the script will exit non-zero at ensureGPUDrivers. What this PR is doing is separating the install part of the GPU drivers install from what needs to be done at runtime so that the VHD builder can call the install part only. The function itself is unchanged from what was previously there, it was just split into two sub functions which are called one after the other in CSE to preserve behavior.

However, looking through it it does look like installGPUDrivers is missing some exit codes so I look into adding those.

CecileRobertMichon:
https://jenkins.azure-containers.io/view/acs-engine/job/k8s-deploy-gpu-1.11/1466/

parts/k8s/kubernetesinstalls.sh
```diff
@@ -26,45 +26,48 @@ function installDeps() {
...
+        apt_get_update
+        retrycmd_if_failure 30 5 300 apt-get install -y linux-headers-$(uname -r) gcc make dkms || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+        retrycmd_if_failure 30 5 300 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-docker2=$NVIDIA_DOCKER_VERSION+docker$DOCKER_VERSION nvidia-container-runtime=$NVIDIA_CONTAINER_RUNTIME_VERSION+docker$DOCKER_VERSION || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+        pkill -SIGHUP dockerd
+        mkdir -p $DEST
+        cd $DEST
+        # Download the .run file from NVIDIA.
+        # Nvidia libraries are always install in /usr/lib/x86_64-linux-gnu, and there is no option in the run file to change this.
+        # Instead we use Overlayfs to move the newly installed libraries under /usr/local/nvidia/lib64
+        retrycmd_if_failure 5 10 60 curl -fLS https://us.download.nvidia.com/tesla/$DV/NVIDIA-Linux-x86_64-$DV.run -o nvidia-drivers-$DV || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
```
CecileRobertMichon: @tariq1890 added a specific exit code for GPU install

PR #3790, opened 2018-09-04
E2E: enable "skip collect logs after test"

jackfrancis:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
NONE
```

### DIFF
--- a/test/e2e/runner.go
+++ b/test/e2e/runner.go
@@ -191,7 +191,7 @@ func teardown() {
 		log.Printf("cannot create directory for logs: %s", err)
 	}
 
-	if cliProvisioner.Config.IsKubernetes() && cfg.SoakClusterName == "" {
+	if cliProvisioner.Config.IsKubernetes() && cfg.SoakClusterName == "" && !cfg.SkipLogsCollection {
 		err = cliProvisioner.FetchProvisioningMetrics(logsPath, cfg, acct)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


test/e2e/runner.go
```diff
@@ -191,7 +191,7 @@ func teardown() {
 		log.Printf("cannot create directory for logs: %s", err)
 	}
 
-	if cliProvisioner.Config.IsKubernetes() && cfg.SoakClusterName == "" {
+	if cliProvisioner.Config.IsKubernetes() && cfg.SoakClusterName == "" && !cfg.SkipLogsCollection {
```
CecileRobertMichon: As a follow up we could remove `&& cfg.SoakClusterName == ""` and set SkipLogsCollection to true for soak tests what do you think?

jackfrancis: As-is it still just doesn't make sense to collect the logs in the way we currently do on a soak cluster, because we don't have any easy way to tag the logs (e.g., timestamp) to differentiate one log collection from the next, which soak clusters would need to do.

Basically, in practice what we're collecting at present are provisioning logs, and we *do* want the soak cluster tests to do that the first time they run, and the current implementation (though hacky) does do that.

CecileRobertMichon: Oh true I forgot it does collect logs the first time around

PR #3792, opened 2018-09-04
Ensuring etcd is not installed twice on VHD images

tariq1890:
Currently, we seem to be installing etcd before we make the check for golden image. This potentially leads to redundant etcd installs. This PR should hopefully fix that


CecileRobertMichon:
@tariq1890 the installEtcd function first checks if that version of etcd is installed and skips the install if it is https://github.com/Azure/acs-engine/blob/master/parts/k8s/kubernetesinstalls.sh#L11

PR #3794, opened 2018-09-05
Design doc

tariq1890:
Design doc draft for acs-engine


docs/acs-engine-design-doc.md
```diff
@@ -0,0 +1,284 @@
...
+
+**Strongly-typed schema for the ARM templates**
+
+As of now, we have no standard/formal representation of the ARM templates. They are declared as loosely typed in the Azure Go SDK.
+
+_**Pros**_
+
+- A formal representation would help us create a more direct mapping between the api model inputs and their corresponding ARM template files.
+
+- This also widens the scope for customizations in cluster definitions in the future.
```
jackfrancis: I would re-phrase this to be something like "This will allow us to accommodate future ARM template customization  more effectively, because we can express and maintain the variety of inter-dependent outputs natively, as first class data representations."

jackfrancis:
One more change! Let's rename `docs/images/acs-engine-1.jpg` to `docs/images/proposed-template-refactor.jpg` so that it is more easily associated with this markdown file.

tariq1890:
@jackfrancis  Done!

PR #3804, opened 2018-09-05
store Azure CNI version in api model

jackfrancis:
**What this PR does / why we need it**: Stores the statically assigned Azure CNI version in the api model so that it is more easily determined which version of Azure CNI a cluster was built with.

Note: this does not enable configurable versioning of Azure CNI at deployment time.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
store Azure CNI version in api model
```

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -422,6 +422,14 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
 			a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs = true
 		}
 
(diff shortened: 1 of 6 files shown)


jackfrancis:
@weinong @sharmasushant FYI

CecileRobertMichon:
What about making it a variable in the ARM template? Only reason I'm saying that is it might be confusing to have a field in the apimodel / in their cluster configuration that gets ignored. From what I understand we're not allowing this value to be configurable for now.

jackfrancis:
@CecileRobertMichon I'm open to rationalizing this whole thing (including making the version configurable) and reflecting that in the ARM template. For now this incremental change is useful.

PR #3808, opened 2018-09-06
adding ablitiy to login to an external tenant in azure client stuff

JackQuincy:
What this PR does / why we need it:
We are using this in the RP for getting an azure client from a specific tenant.
I see other clients are used in the CLI, I can add this there as well if needed

related to an older pr only this one uses a cert to login

### DIFF
--- a/pkg/armhelpers/azureclient.go
+++ b/pkg/armhelpers/azureclient.go
@@ -211,25 +211,7 @@ func NewAzureClientWithClientCertificate(env azure.Environment, subscriptionID, 
 		return nil, err
 	}
 
-	if certificate == nil {
-		return nil, errors.New("certificate should not be nil")
-	}
-
-	if privateKey == nil {
-		return nil, errors.New("privateKey should not be nil")
-	}
-
-	armSpt, err := adal.NewServicePrincipalTokenFromCertificate(*oauthConfig, clientID, certificate, privateKey, env.ServiceManagementEndpoint)
-	if err != nil {
-		return nil, err
-	}
-	graphSpt, err := adal.NewServicePrincipalTokenFromCertificate(*oauthConfig, clientID, certificate, privateKey, env.GraphEndpoint)
-	if err != nil {
-		return nil, err
-	}
-	graphSpt.Refresh()
-
-	return getClient(env, subscriptionID, tenantID, armSpt, graphSpt), nil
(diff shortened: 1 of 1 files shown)


pkg/armhelpers/azureclient.go
```diff
@@ -232,6 +232,34 @@ func NewAzureClientWithClientCertificate(env azure.Environment, subscriptionID,
 	return getClient(env, subscriptionID, tenantID, armSpt, graphSpt), nil
 }
 
+// NewAzureClientWithClientCertificateExternalTenant returns an AzureClient via client_id and jwt certificate assertion against a 3rd party tenant
+func NewAzureClientWithClientCertificateExternalTenant(env azure.Environment, subscriptionID, tenantID, clientID string, certificate *x509.Certificate, privateKey *rsa.PrivateKey) (*AzureClient, error) {
+	oauthConfig, err := adal.NewOAuthConfig(env.ActiveDirectoryEndpoint, tenantID)
```
tariq1890: do we need to call this from getAuthArgs as well? Perhaps add a new auth method named "external-tenant".

JackQuincy: I'm confused. What are you suggesting?

tariq1890: Referring to this : https://github.com/Azure/acs-engine/blob/e93da3756bc9b8a647b5771b08524915bcf34d70/cmd/root.go#L193

pkg/armhelpers/azureclient.go
```diff
@@ -211,6 +211,20 @@ func NewAzureClientWithClientCertificate(env azure.Environment, subscriptionID,
...
+	if err != nil {
+		return nil, err
+	}
+
+	return internalnewAzClientWithCertificate(env, oauthConfig, subscriptionID, clientID, tenantID, certificate, privateKey)
+}
+
+func internalnewAzClientWithCertificate(env azure.Environment, oauthConfig *adal.OAuthConfig, subscriptionID, clientID, tenantID string, certificate *x509.Certificate, privateKey *rsa.PrivateKey) (*AzureClient, error) {
```
tariq1890: making this "newAzClientWithCertificate" would be idiomatic and self-explanatory.

CecileRobertMichon: Let's change it to `newAzureClientWithCertificate` or `internalNewAzureClientWithCertificate` (not sure how important the internal part it, I'll defer to @JackQuincy for that) for consistency with other function names in this file

PR #3813, opened 2018-09-07
only enable aggregated api if rbac != disabled

jackfrancis:
**What this PR does / why we need it**: Prior to this change, default configuration for RBAC and enable aggregated APIs was distinct. These default enforcements need to happen in tandem, as aggregated API functionality depends upon RBAC. So, if RBAC is disabled, we need to ensure that aggregated API functionality is not by-default enabled.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
only enable aggregated api if rbac != disabled
```

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -59,6 +59,14 @@ func IsTrueBoolPointer(b *bool) bool {
 	return false
 }
 
+// IsFalseBoolPointer is a simple boolean helper function for boolean pointers
+func IsFalseBoolPointer(b *bool) bool {
+	if b != nil && !*b {
+		return true
+	}
+	return false
+}
+
 // PointerToBool returns a pointer to a bool
 func PointerToBool(b bool) *bool {
 	p := b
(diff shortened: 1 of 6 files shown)


pkg/helpers/helpers.go
```diff
@@ -59,6 +59,14 @@ func IsTrueBoolPointer(b *bool) bool {
 	return false
 }
 
+// IsFalseBoolPointer is a simple boolean helper function for boolean pointers
+func IsFalseBoolPointer(b *bool) bool {
```
tariq1890: Question: Do you think IsFalse would suffice as there being a bool pointer is already suggestive?

jackfrancis: Because we already have lots of places using `IsTrueBoolPointer` it makes the most sense to keep the same naming pattern.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -703,7 +703,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 		})
 
 		It("should be able to autoscale", func() {
-			if eng.HasLinuxAgents() {
+			if eng.HasLinuxAgents() && eng.ExpandedDefinition.Properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
```
jackfrancis: HPA implementation uses metrics-server, which requires aggregated api

PR #3820, opened 2018-09-07
AKS doesn't use VHD image

jackfrancis:
**What this PR does / why we need it**: Ensure AKS doesn't use VHD image

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
AKS doesn't use VHD image
```

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -488,7 +488,6 @@ func setExtensionDefaults(a *api.Properties) {
 	}
 }
 
-// SetMasterNetworkDefaults for masters
 func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
 	// don't default Distro for OpenShift
 	if !a.OrchestratorProfile.IsOpenShift() {
@@ -585,7 +584,6 @@ func setVMSSDefaults(a *api.Properties) {
 	}
 }
 
-// setAgentProfileDefaults sets defaults for agent profiles
 func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
 	// configure the subnets if not in custom VNET
 	if a.MasterProfile != nil && !a.MasterProfile.IsCustomVNET() {


pkg/acsengine/defaults.go
```diff
@@ -675,6 +668,19 @@ func setStorageDefaults(a *api.Properties) {
 	}
 }
 
+func setHostedMasterOverrides(a *api.Properties) {
+	a.HostedMasterProfile.Subnet = DefaultKubernetesMasterSubnet
+
+	for _, profile := range a.AgentPoolProfiles {
+		// don't default Distro for OpenShift
+		if !a.OrchestratorProfile.IsOpenShift() {
+			if profile.Distro == "" {
```
CecileRobertMichon: since `setAgentNetworkDefaults` happens before this, profile.Distro won't be empty by the time we reach this function

pkg/acsengine/defaults.go
```diff
@@ -675,6 +668,16 @@ func setStorageDefaults(a *api.Properties) {
 	}
 }
 
+func setHostedMasterOverrides(a *api.Properties) {
+	a.HostedMasterProfile.Subnet = DefaultKubernetesMasterSubnet
+
+	for _, profile := range a.AgentPoolProfiles {
+		if !a.OrchestratorProfile.IsOpenShift() {
+			profile.Distro = api.Ubuntu
```
CecileRobertMichon: I think we might want to do this in `setAgentNetworkDefaults` instead like so: 
```
			if profile.Distro == "" {
                                    if cs.Properties.HostedMasterProfile != nil { 
                                                profile.Distro = api.Ubuntu
                                    } else {
				                profile.Distro = api.AKS
                                    }
			}
		}
```

CecileRobertMichon: I'm worried about overriding the value no matter what because this might override the value for a scale or an upgrade if we do enable the AKS distro progressively

CecileRobertMichon: Also it's more readable to have both default values in one place IMO

pkg/acsengine/defaults.go
```diff
@@ -233,16 +233,22 @@ func setPropertiesDefaults(cs *api.ContainerService, isUpgrade, isScale bool) (b
...
-	setMasterNetworkDefaults(properties, isUpgrade)
-
-	setHostedMasterNetworkDefaults(properties)
+	// Set master profile defaults if this cluster configuration includes master node(s)
+	if cs.Properties.MasterProfile != nil {
+		setMasterProfileDefaults(properties, isUpgrade)
+	}
 
-	setAgentNetworkDefaults(properties, isUpgrade, isScale)
+	setAgentProfileDefaults(properties, isUpgrade, isScale)
```
jackfrancis: Renamed this func because it's not network-specific, it's agent profile-specific.

pkg/acsengine/defaults.go
```diff
@@ -233,16 +233,22 @@ func setPropertiesDefaults(cs *api.ContainerService, isUpgrade, isScale bool) (b
 
 	setOrchestratorDefaults(cs, isUpgrade || isScale)
 
-	setMasterNetworkDefaults(properties, isUpgrade)
-
-	setHostedMasterNetworkDefaults(properties)
+	// Set master profile defaults if this cluster configuration includes master node(s)
+	if cs.Properties.MasterProfile != nil {
+		setMasterProfileDefaults(properties, isUpgrade)
```
jackfrancis: Renamed this func because it's not network-specific, it's master profile-specific.

CecileRobertMichon: 👍 💯

pkg/acsengine/defaults.go
```diff
@@ -590,8 +585,8 @@ func setVMSSDefaults(a *api.Properties) {
 	}
 }
 
-// SetAgentNetworkDefaults for agents
-func setAgentNetworkDefaults(a *api.Properties, isUpgrade, isScale bool) {
+// setAgentProfileDefaults sets defaults for agent profiles
```
jackfrancis: Renamed this func because it's not network-specific, it's agent profile-specific.

pkg/acsengine/defaults.go
CecileRobertMichon: 👍 💯

pkg/acsengine/defaults.go
```diff
@@ -590,8 +584,7 @@ func setVMSSDefaults(a *api.Properties) {
 	}
 }
 
-// SetAgentNetworkDefaults for agents
```
CecileRobertMichon: but this comment was so useful 😜

PR #3821, opened 2018-09-07
Change av profile default logic based on new version support matrix

tariq1890:


jackfrancis:
I'm okay with this change, but I agree w/ @CecileRobertMichon that we should do >= 1.10.2 to reflect the real requirement. We don't have to document the boring detail about why imds didn't work until 1.10.2, but the tl;dr is that vmss + Azure cloudprovider is functional as of k8s 1.10.2.

PR #3823, opened 2018-09-10
add support for Kubernetes 1.11.3

mboersma:

See https://github.com/kubernetes/kubernetes/releases/tag/v1.11.3

TODO:

- [x] upload Windows binary

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:
```release-note
add support for Kubernetes 1.11.3
```


pkg/api/common/versions.go
```diff
@@ -75,6 +75,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.11.0":         false,
 	"1.11.1":         true,
 	"1.11.2":         true,
+	"1.11.3":         true,
 	"1.12.0-alpha.1": true,
```
CecileRobertMichon: Should we set 1.11.1 support to false since it is no longer one of the two latest patches?

mboersma: Yes we should--I'll add that.

PR #3826, opened 2018-09-10
AKS uses VHD + outbound inet test

jackfrancis:
**What this PR does / why we need it**: Enable VHD image for AKS node provisioning. Restores CSE 50 exit code for VHDs to accommodate AKS req's until all external i/o dependencies are implemented as local to VHD image.

Also ensures that OS disk size configurations < 30 GB will not attempt to use AKD VHD image, which requires 100 GB.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
AKS uses VHD + outbound inet test
```

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -41,12 +41,13 @@ function holdWALinuxAgent() {
     fi
 }
 
+testOutboundConnection
+
 if [[ ! -z "${MASTER_NODE}" ]]; then
     installEtcd
 fi
 
 if $FULL_INSTALL_REQUIRED; then
-    testOutboundConnection
     holdWALinuxAgent
     installDeps
 else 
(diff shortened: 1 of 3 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -41,12 +41,13 @@ function holdWALinuxAgent() {
     fi
 }
 
+testOutboundConnection
```
jackfrancis: Until we prove that we can bootstrap a k8s control plane w/ no outbound internet access we should put this back.

pkg/acsengine/defaults.go
```diff
@@ -618,7 +618,7 @@ func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
 		// don't default Distro for OpenShift
 		if !a.OrchestratorProfile.IsOpenShift() {
 			if profile.Distro == "" {
-				if a.HostedMasterProfile != nil {
+				if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < api.VHDDiskSizeAKS {
```
jackfrancis: - `profile.OSDiskSizeGB == 0` we interpret as "no user configuration provided, i.e., zero value"; we can use the 100GB VHD image in this condition
- for explicit OS disk sizes < 100 GB we need to use the base Ubuntu OS image, as we can't use a VHD image reference whose OS disk size is greater than the configured disk size in the ARM template

PR #3830, opened 2018-09-11
Add support to add a list of auxiliary tokens in request header

amanohar:
Adds support to add a list of auxiliary tokens in request header. Aux tokens enables ARM to support requests which may require authentication across multiple tenants.

Primary token will always be extracted from the “Authorization” header. Auxiliary tokens will be extracted from the header “Authorization-Auxiliary”

Example in Azure Python CLI: https://github.com/Azure/azure-cli/pull/6423/commits/0217c4b8c8737940c0c92226347c423f0f8ec413

### DIFF
--- a/pkg/armhelpers/azureclient.go
+++ b/pkg/armhelpers/azureclient.go
@@ -46,6 +46,7 @@ var (
 // This client is backed by real Azure clients talking to an ARM endpoint.
 type AzureClient struct {
 	acceptLanguages []string
+	auxiliaryTokens []string
 	environment     azure.Environment
 	subscriptionID  string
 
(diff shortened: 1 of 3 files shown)


pkg/armhelpers/azureclient.go
```diff
@@ -462,3 +463,45 @@ func (az *AzureClient) addAcceptLanguages() autorest.PrepareDecorator {
...
+					r.Header.Set("x-ms-authorization-auxiliary", fmt.Sprintf("Bearer %s", token))
+				}
+			}
+			return r, nil
+		})
+	}
+}
+
+// AddAuxillaryTokens sets the list of aux tokens to accept on this request
+func (az *AzureClient) AddAuxillaryTokens(tokens []string) {
```
tariq1890: I guess it should be AddAuxiliaryTokens

pkg/armhelpers/azureclient.go
```diff
@@ -462,3 +463,45 @@ func (az *AzureClient) addAcceptLanguages() autorest.PrepareDecorator {
...
+	az.auxiliaryTokens = tokens
+	az.authorizationClient.Client.RequestInspector = az.addAuxiliaryTokens()
```
tariq1890: We seem to be calling az.addAuxiliaryTokens() multiple times only to get the same instance of autorest.PrepareDecorator. We could assign this to a variable instead and re-use that.

pkg/armhelpers/azureclient.go
```diff
@@ -462,3 +463,45 @@ func (az *AzureClient) addAcceptLanguages() autorest.PrepareDecorator {
 		})
 	}
 }
+
+func (az *AzureClient) addAuxiliaryTokens() autorest.PrepareDecorator {
```
CecileRobertMichon: would it be possible to rename this function to something more specific so we don't have two functions with the same name?

pkg/armhelpers/azureclient.go
CecileRobertMichon: Is this function going to be called in the RP? is it possible to add unit tests for it?

pkg/armhelpers/azureclient.go
amanohar: Aux token is being assigned to `az.auxiliaryTokens` and then set in the Header `addAuxiliaryTokens`. I am following the coding convention in the file used for `addAcceptLanguages`

pkg/armhelpers/azureclient.go
amanohar: Happy to update it. Any recommendations? I was following the `AddAcceptLanguages` and `addAcceptLanguages` example in the file.

jackfrancis: how about `setAuxiliaryTokens`?

Also, is there a good reason why we're using encapsulation like this? (`AddAuxiliaryTokens` is a public method and the one we're discussing is a private method).

@tariq1890 do you have any thoughts?

tariq1890: There is definitely redundancy. In the case of passing acceptLanguages and auxiliary tokens, there is actually an overwrrite of the requestPreparer happening.

amanohar: @jackfrancis I will update it `setAuxiliaryTokens`

>Also, is there a good reason why we're using encapsulation like this?

I was following the coding convention in the file that's already being used by `addAcceptLanguages`. I am not aware of the initial reason to follow this encapsulation. But it sure introduces a lot of redudancy and superflous for a seemingly simple task. I am happy to update this to call `setAuxiliaryTokens` only once.

amanohar: Updated function name + removed redundant setting of header. PTAL.

pkg/armhelpers/azureclient.go
```diff
@@ -462,3 +463,47 @@ func (az *AzureClient) addAcceptLanguages() autorest.PrepareDecorator {
...
+
+// AddAuxiliaryTokens sets the list of aux tokens to accept on this request
+func (az *AzureClient) AddAuxiliaryTokens(tokens []string) {
+	az.auxiliaryTokens = tokens
+	requestWithTokens := az.setAuxiliaryTokens()
+
+	az.authorizationClient.Client.RequestInspector = requestWithTokens
+	az.deploymentOperationsClient.Client.RequestInspector = requestWithTokens
+	az.deploymentsClient.Client.RequestInspector = requestWithTokens
+	az.deploymentsClient.Client.RequestInspector = requestWithTokens
```
CecileRobertMichon: Line 496 and 495 are duplicates

amanohar: Thanks. Removed duplicates in both functions.

amanohar:
@CecileRobertMichon @JackQuincy, add unit test. PTAL.

amanohar:
@jackfrancis can this be merged please?

jackfrancis:
beat ya

PR #3840, opened 2018-09-12
Adding rawData script input for k8s addons

tariq1890:
fixes #3811

PR #3841, opened 2018-09-13
update keyvault-flexvol image and default to enabled

ritazh:


pkg/api/vlabs/validate_test.go
```diff
@@ -284,7 +284,7 @@ func Test_OrchestratorProfile_Validate(t *testing.T) {
 
 	for testName, test := range tests {
 		t.Run(testName, func(t *testing.T) {
-			t.Parallel()
+			//t.Parallel()
```
tariq1890: Is this required? I tested this test case change and it seems to work consistently with t.Parallel() enabled.

tariq1890:
Looks good to me save for that one review comment. @jackfrancis This would need to be merged before #3840 is merged.

PR #3843, opened 2018-09-13
chore(deps): upgrade go-autorest to v10.15.4

ultimateboy:

Updates the version of `go-autorest` to the latest release, which includes several bug fixes.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -22,7 +22,7 @@
   version = "v19.0.0"
 
 [[projects]]
-  digest = "1:935cf7044794f9149498fb4ecea4393cfee9bf3083760fefb835dc63307d11ce"
+  digest = "1:aafd4229d26448b58665f4067d3fa5e945a2c90a30a639346767c21e4be9beb7"
   name = "github.com/Azure/go-autorest"
   packages = [
     "autorest",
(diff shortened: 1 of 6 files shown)


tariq1890:
@weinong @jackfrancis @mboersma I see the possibility of some incongruity here as we are updating the go-autorest version and not the Azure Go SDK (which uses the go-autorest library heavily). What are your thoughts about updating the Azure Go SDK as well?

Gopkg.lock
```diff
@@ -22,7 +22,7 @@
   version = "v19.0.0"
```
tariq1890: Let's upgrade to v20.1.0

mboersma: I'd like to do that as well--many services (including containerservices itself) have been updated in the SDK since v19.0.0. I think it would be safest to use the go-autorest version that was actually used to generate the Go SDK, if we can find that out.

mboersma: Apparently that was [10.5.3](https://github.com/Azure/azure-sdk-for-go/blob/v20.1.0/Gopkg.toml#L25). Close enough.

tariq1890: I checked with Martin Strobel earlier, the go sdk maintainer. He says using 10.15.4 of go-autorest along with v20.1.0 of go sdk should be fine.

mboersma: OTOH, upgrading sdk-for-go has a chance of breaking some API acs-engine is already using (and this has a cascade effect when vendored into AKS). If this is passing, maybe that should be a separate PR?

jackfrancis:
@tariq1890 @mboersma what's the // TODO at this point according to your recommendations?

Gopkg.lock
tariq1890: Sure!

tariq1890:
I guess the TODO would be to vendor in an Azure SDK Go release which contains the go-autorest referenced in this PR.

jackfrancis:
@tariq1890 can we do that as a follow-up? Or should we do it in this PR?

tariq1890:
We can do that in another PR. Hopefully, the next release of AzureGo SDK will have the deps set straight.

Issue #3773, opened 2018-08-30
UseManagedIdentity validations and e2e tests


ritazh:
User assigned msi feature was added to k8s v1.12.0-alpha.1. Let's make sure we only support this feature here for k8s version 1.12+

cc @kkmsft

kkmsft:
Adding two more issues to be covered in this bug:
- Validation for the empty user assigned identity
- Add e2e test flow

Issue #3523, opened 2018-07-20
Add validation for aad-pod-identity addon to ensure only AvailabilitySet is used


ritazh:

ritazh:
add configurable resource limits and e2e test

PR #3769, opened 2018-08-30
[WIP] VMSS support for k8s masters

ritazh:
**What this PR does / why we need it**:
* Adds support for virtual machine scale set infrastructure support for masters

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3609 

**Special notes for your reviewer**:
Work in progress - do not merge

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
Add Virtual Machine Scale Sets support for Kubernetes v1.10+ master availability profile
```

//cc @kkmsft @khenidak

PR #3848, opened 2018-09-14
Validations for user assigned identity and aad-pod-identity

kkmsft:
- Ensure that user assigned identity is used only for K8s versions > v1.12.0-alpha.1. #3773
- Ensure that add-pod-identity is used only with the availability set (supported as of now). #3523

Validation to ensure that

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -271,6 +271,17 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
 							minVersion.String(), o.OrchestratorVersion)
 					}
 				}
+				// If the user assigned identity is enabled, then make sure the version is higher than v1.12.0-alpha.1
+				if o.KubernetesConfig.UseManagedIdentity && o.KubernetesConfig.UserAssignedID != "" {
+					minVersion, err := semver.Make("1.12.0-alpha.1")
+					if err != nil {
+						return errors.Errorf("could not validate the version")
+					}
+					if sv.LT(minVersion) {
+						return errors.Errorf("user assigned msi is only available in kubernetes version %s or greater. current version used is %s",
+							minVersion.String(), o.OrchestratorVersion)
+					}
+				}
 			}
 		case OpenShift:
 			// TODO: add appropriate additional validation logic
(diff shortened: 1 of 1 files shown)


pkg/api/vlabs/validate.go
```diff
@@ -503,6 +514,10 @@ func (a *Properties) validateAddons() error {
 						return errors.New("NVIDIA Device Plugin add-on can only be used Kubernetes 1.10 or above. Please specify \"orchestratorRelease\": \"1.10\"")
 					}
 				}
+			case "aad-pod-identity":
+				if helpers.IsTrueBoolPointer(addon.Enabled) && !isAvailabilitySets {
+					return errors.Errorf("aad-pod-identity supports only availablity sets as of now. Please specify \"availabilityProfile\": \"%s\"", AvailabilitySet)
```
tariq1890: nit: typo in availablity

pkg/api/vlabs/validate.go
```diff
@@ -271,6 +271,17 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
 							minVersion.String(), o.OrchestratorVersion)
 					}
 				}
+				// If the user assigned identity is enabled, then make sure the version is higher than v1.12.0-alpha.1
+				if o.KubernetesConfig.UseManagedIdentity && o.KubernetesConfig.UserAssignedID != "" {
```
tariq1890: Could you add unit tests for these new validations?

tariq1890:
Thanks for the PR :)

ritazh:
@kkmsft thank for looking at this. can you pls take a look at this: https://github.com/Azure/acs-engine/pull/3769/files#diff-f45af0b14c127459636440a5981a6934R634 I have added validations for k8s versions as well as conditions that require user assigned identity to work with vmss master. I think what you have here is a subset of the conditions I have added in this function. I have also added unit tests in the other PR. Let me know if you think that function already covers this case, if not, either we can add it to that other PR or you can add it here. LMK!

pkg/api/vlabs/validate.go
```diff
@@ -271,6 +271,17 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
...
+					minVersion, err := semver.Make("1.12.0-alpha.1")
+					if err != nil {
+						return errors.Errorf("could not validate the version")
+					}
+					if sv.LT(minVersion) {
+						return errors.Errorf("user assigned msi is only available in kubernetes version %s or greater. current version used is %s",
```
CecileRobertMichon: nit: let's capitalize "User", "Kubernetes", and "Current"

pkg/api/vlabs/validate.go
ritazh: This validation is now here: https://github.com/Azure/acs-engine/blob/master/pkg/api/vlabs/validate.go#L672-L674.

PR #3859, opened 2018-09-17
remove privileged on specific addons where not required

jessfraz:
adds cap NET_ADMIN which should be all that is needed

Signed-off-by: Jess Frazelle <acidburn@microsoft.com>

### DIFF
--- a/parts/k8s/addons/azure-cni-networkmonitor.yaml
+++ b/parts/k8s/addons/azure-cni-networkmonitor.yaml
@@ -31,7 +31,8 @@ spec:
         - name: azure-cnms
           image: <azureCNINetworkMonitorImage>
           securityContext:
-            privileged: true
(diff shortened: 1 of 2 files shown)


jessfraz:
i was going to test those seperately... I dont know what they all need.

jackfrancis:
@jessfraz I have two tests in a row that suggests the Azure NetworkPolicy addon implementation w/ this change breaks cluster DNS.

Specifically this:

https://github.com/Azure/acs-engine/blob/master/test/e2e/kubernetes/workloads/validate-dns.yaml

CecileRobertMichon:
@jessfraz closing this since we moved to https://github.com/Azure/aks-engine/! What is the status of this? Is it still relevant and should we reopen an equivalent PR for aks-engine? Thanks :)

PR #3861, opened 2018-09-17
add function CreateContainer and SaveBlockBlob

mick0601:
We need functions to create Container and save Blob to ServiceAccount for ManagedOpenShift SP
We also realized that the CreateContainer and SaveBlob should be public functions so that other SPs can reuse that.

**If applicable**:
- [ ] documentation
- [ ] unit tests
Passed by running makedev.ps1 & make test
=======================
Ginkgo ran 28 suites in 12m56.9024074s
Test Suite Passed
==========================

**Release note**:

```release-note
```


pkg/armhelpers/storage.go
```diff
@@ -49,3 +50,21 @@ func (as *AzureStorageClient) DeleteBlob(vhdContainer, vhdBlob string) error {
 
 	return blobRef.Delete(&azStorage.DeleteBlobOptions{})
 }
+
+// CreateContainer creates the CloudBlobContainer if it does not exist
+func (as *AzureStorageClient) CreateContainer(containerName string) (bool, error) {
+	bs := as.client.GetBlobService()
```
tariq1890: Lines 56 & 57 are duplicated. Can we move them to a separate method?

pkg/armhelpers/storage.go
```diff
@@ -49,3 +50,21 @@ func (as *AzureStorageClient) DeleteBlob(vhdContainer, vhdBlob string) error {
...
+	containerRef := bs.GetContainerReference(containerName)
+	created, err := containerRef.CreateIfNotExists(nil)
+
+	return created, err
+}
+
+// SaveBlockBlob initializes a block blob by taking the byte
+func (as *AzureStorageClient) SaveBlockBlob(containerName, blobName string, b []byte, options *azStorage.PutBlobOptions) error {
+	bs := as.client.GetBlobService()
```
tariq1890: I like the variable names!

pkg/armhelpers/storage.go
mick0601: Thanks for the reviewing, I removed the duplicate lines

mick0601:
@CecileRobertMichon We think those two functions need to be public because the CreateContainer and SaveBlob are very common functions that other RP may also need them. 
Unit test is just added.

pkg/armhelpers/interfaces.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/azure-sdk-for-go/services/preview/msi/mgmt/2015-08-31-preview/msi"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	azStorage "github.com/Azure/azure-sdk-for-go/storage"
```
tariq1890: Can we not add this?

pkg/armhelpers/mockclients.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/Azure/azure-sdk-for-go/services/preview/msi/mgmt/2015-08-31-preview/msi"
 	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	azStorage "github.com/Azure/azure-sdk-for-go/storage"
```
tariq1890: Here as well

pkg/armhelpers/storage.go
```diff
@@ -43,9 +44,29 @@ func (az *AzureClient) getStorageKeys(ctx context.Context, resourceGroup, accoun
...
+// CreateContainer creates the CloudBlobContainer if it does not exist
+func (as *AzureStorageClient) CreateContainer(containerName string) (bool, error) {
+	containerRef := getContainerRef(as, containerName)
+	created, err := containerRef.CreateIfNotExists(nil)
+
+	return created, err
+}
+
+// SaveBlockBlob initializes a block blob by taking the byte
+func (as *AzureStorageClient) SaveBlockBlob(containerName, blobName string, b []byte, options *azStorage.PutBlobOptions) error {
```
tariq1890: Instead of passing an azStorage options as a parameter and necessitating an import in other packages, we can add an options field in the AzureStorageClient and then construct the PutBlobOptions from that. Better SOC that way IMO.

tariq1890:
@mick0601 Thanks for the fixes. Looks good so far. Added a few more comments :)

pkg/armhelpers/storage.go
tariq1890: I see that PutBlobOptions is composed of headers. So the new options field in AzureStorageClient can be a Map.

mick0601: If we put the option to the AzureStorageClient field, the values of the option will be predefined and we lose the agility of changing the option when saving a specific blob with a specific option.
We may set this field to a public but that also require the caller to setthe option values everytime it calls the SaveBlockBlob.

tariq1890: I see your point. Then we should not be using an AzureStorageClient wrapper in the first place.

pkg/armhelpers/storage.go
```diff
@@ -43,9 +44,29 @@ func (az *AzureClient) getStorageKeys(ctx context.Context, resourceGroup, accoun
...
+	containerRef := getContainerRef(as, vhdContainer)
 	blobRef := containerRef.GetBlobReference(vhdBlob)
 
 	return blobRef.Delete(&azStorage.DeleteBlobOptions{})
 }
+
+// CreateContainer creates the CloudBlobContainer if it does not exist
+func (as *AzureStorageClient) CreateContainer(containerName string) (bool, error) {
+	containerRef := getContainerRef(as, containerName)
+	created, err := containerRef.CreateIfNotExists(nil)
```
tariq1890: This needs to have ContainerOptions as a parameter. Let's add flexibility and agility here

pkg/armhelpers/interfaces.go
```diff
@@ -131,6 +132,10 @@ type ACSEngineClient interface {
 type ACSStorageClient interface {
 	// DeleteBlob deletes the specified blob in the specified container.
 	DeleteBlob(container, blob string) error
```
tariq1890: add BlobOption parameter here please

pkg/armhelpers/interfaces.go
```diff
@@ -131,6 +132,10 @@ type ACSEngineClient interface {
...
+	// CreateContainer creates the CloudBlobContainer if it does not exist
+	CreateContainer(containerName string) (bool, error)
```
tariq1890: Ditto

pkg/operations/deletevm.go
```diff
@@ -75,7 +75,7 @@ func CleanDeleteVirtualMachine(az armhelpers.ACSEngineClient, logger *log.Entry,
 		}
 
 		logger.Infof("deleting blob: %s/%s", vhdContainer, vhdBlob)
-		if err = as.DeleteBlob(vhdContainer, vhdBlob); err != nil {
+		if err = as.DeleteBlob(vhdContainer, vhdBlob, nil); err != nil {
```
tariq1890: Thanks a lot @mick0601 . Last request :Can we pass an empty DeleteOptionBlob struct pointer instead of nil here?

mick0601: Yes, I will make that change in a real quick. Thanks again for the help, Tariq

tariq1890:
Thanks a lot @mick0601 :)

Issue #1919, opened 2017-12-12
Multizone support?


omerlh:
**Is this a request for help?**:

No

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

Feature Request

**What version of acs-engine?**:

```
Version: v0.10.0
GitCommit: 7f3e25c2
GitTreeState: clean
```

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes, 1.7

Kubernetes already have support for [multizone](https://kubernetes.io/docs/admin/multiple-zones/) - e.g. deploying cluster over multiple regions in Azure, for high availability. How can that be done using ACS engine?
Currently, I create 2 different clusters (in eastus and westus region), with traffic manager to spread the traffic between the clusters. This is not ideal, but working. Ideally, this would be managed by kubernetes multizone feature - which will also be more efficient on resources.

hmarcelodn:
I experienced same issue and I did exactly the same than you @omerlh

dstrebel:
Kubernetes multizone feature is for single region zones and not multiregion. Azure currently is in preview for Availability Zones, which also utilize VM Scale Sets. Azure VM Scale Set support is slated for [1.10](https://github.com/kubernetes/features/issues/513) . So support for K8 multizone will come sometime after that release. If you want cross region you would be looking at utilizing K8 Federation.

xukai286:
we need this feature. thank you

feiskyer:
We are working on upstream kubernetes for this feature. See https://github.com/kubernetes/features/issues/586.

aheumaier:
@feiskyer Is this still valid ? or merged now ?

feiskyer:
@aheumaier yep, k8s part has been merge now. #3453 is working on acs-engine part.

ritazh:
Note: https://github.com/Azure/acs-engine/pull/3453 adds zones support for agents. We will add zones support for masters in a different PR.

aluncoppack:
>We will add zones support for masters in a different PR.

@ritazh, this is fantastic to hear. Do you have a link to the master PR?

ritazh:
@aluncoppack AZ for master PR: #3864

Issue #3835, opened 2018-09-12
Update default apiversion


CecileRobertMichon:
Current default apiversion used in ARM templates is `2016-03-30`. We should update it to a more recent version. `2018-04-01` is already used for VMs with accelerated networking enabled (default is true so most VMs).

PR #3864, opened 2018-09-18
Add availability zone support for masters

ritazh:
Adds availability zones support for masters

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/Azure/acs-engine/issues/1919

**If applicable**:
- [x] documentation
- [x] unit tests

**Release note**:

```release-note
Add availability zones support for k8s masters
```
cc @khenidak @feiskyer


pkg/acsengine/defaults.go
```diff
@@ -242,7 +242,7 @@ func setPropertiesDefaults(cs *api.ContainerService, isUpgrade, isScale bool) (b
 
 	setStorageDefaults(properties)
 	setExtensionDefaults(properties)
-	setVMSSDefaults(properties)
+	setVMSSDefaultsForAgents(properties)
```
jackfrancis: Do we have a `HasVMSSAgentPool`-type method? It'd be more expressive to do:

```
if properties.HasVMSSAgentPool() {
    setVMSSDefaultsForAgents(properties)
}
```

pkg/acsengine/defaults.go
```diff
@@ -582,10 +582,26 @@ func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
 	if a.MasterProfile.HTTPSourceAddressPrefix == "" {
 		a.MasterProfile.HTTPSourceAddressPrefix = "*"
 	}
+	// Set VMSS Defaults for Masters
```
jackfrancis: Is there a reason we're not doing this in a func like the vmss for agents defaults flow above?

pkg/api/types.go
```diff
@@ -715,6 +717,17 @@ func (p *Properties) HasVirtualMachineScaleSets() bool {
 	return false
 }
 
+// HasAllZonesAgentPools will return true if all of the agent pools have zones
```
jackfrancis: Let's go full java and change this func name to `HasZonesForAllAgentPools`

tariq1890: :'(

jackfrancis:
FYI got an E2E flake on 1.8:

https://circleci.com/gh/Azure/acs-engine/43044?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

(just in case we get another flake or two we can start to debug more deeply)

ritazh:
Also added this: https://jenkins.azure-containers.io/view/acs-engine/job/k8s-availability-zones-all/

ritazh:
@jackfrancis @tariq1890 ready for another round of review

pkg/api/vlabs/validate_test.go
```diff
@@ -1647,12 +1647,73 @@ func TestProperties_ValidateZones(t *testing.T) {
 	tests := []struct {
 		name                string
 		orchestratorVersion string
+		masterProfile       *MasterProfile
 		agentProfiles       []*AgentPoolProfile
 		expectedErr         string
 	}{
+		{
+			name:                "Master profile with zones version",
+			orchestratorVersion: "1.11.3",
```
CecileRobertMichon: we should use `orchestratorRelease` in unit tests whenever possible as individual patches will get deprecated as we add new ones (we now support the two latest patches of each minor version)

pkg/acsengine/defaults.go
```diff
@@ -584,8 +591,25 @@ func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
...
+// setVMSSDefaultsForMasters
+func setVMSSDefaultsForMasters(a *api.Properties) {
+	if a.MasterProfile.Count > 100 {
+		a.MasterProfile.SinglePlacementGroup = helpers.PointerToBool(false)
+	}
+	if a.MasterProfile.SinglePlacementGroup == nil {
+		a.MasterProfile.SinglePlacementGroup = helpers.PointerToBool(api.DefaultSinglePlacementGroup)
+	}
+	if a.MasterProfile.SinglePlacementGroup == helpers.PointerToBool(false) {
+		a.MasterProfile.StorageProfile = api.ManagedDisks
```
CecileRobertMichon: Is this going to silently override the user's choice if they select storage account? Should we validate + error if the user selects storageAccount + singlePlacement `false`

pkg/acsengine/defaults.go
```diff
@@ -584,8 +591,25 @@ func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
...
+	}
+	if a.MasterProfile.HasAvailabilityZones() {
```
CecileRobertMichon: same here for overriding LB defaults

parts/k8s/kubernetesmasterresourcesvmss.t
```diff
@@ -186,22 +186,29 @@
 },
 {{end}}
 {
-  "apiVersion": "[variables('apiVersionDefault')]",
+  "apiVersion": "2017-08-01",
```
CecileRobertMichon: Any reason for using `2017-08-01` and not a more recent version like `2018-04-01` that we already use for accelerated networking? 

In any case, we have a backlog item to rationalize the default version https://github.com/Azure/acs-engine/issues/3835 so the version will be bumped soon. If the version needs to be `2017-08-01` and is reused in multiple places I would recommend creating a variable for it.

pkg/acsengine/defaults.go
ritazh: I personally think this is too much detail for end users. so we should set for them but I am ok with either. What would you recommend?

pkg/acsengine/defaults.go
ritazh: Same here.

pkg/acsengine/defaults.go
CecileRobertMichon: The foundation of acs-engine is that it is highly customizable so I think we should assume that anything that is exposed to the user can and will be customized by users. If we want to set a value for the users we shouldn't expose it to them in the apimodel. I think a simple validation check should be enough in this case. That's my recommendation but I'll let you make the call.

ritazh: made it a validation instead of default

pkg/acsengine/defaults.go
ritazh: now it's a validation instead of default

CecileRobertMichon:
As discussed offline, let's re-add the default settings with an additional nil check to only set the default if the value is nil + only error on validation if the value isn't nil && isn't what we want

pkg/acsengine/defaults.go
```diff
@@ -584,8 +591,22 @@ func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
...
+// setVMSSDefaultsForMasters
+func setVMSSDefaultsForMasters(a *api.Properties) {
+	if a.MasterProfile.Count > 100 {
+		a.MasterProfile.SinglePlacementGroup = helpers.PointerToBool(false)
+	}
+	if a.MasterProfile.SinglePlacementGroup == nil {
+		a.MasterProfile.SinglePlacementGroup = helpers.PointerToBool(api.DefaultSinglePlacementGroup)
+	}
+	if a.MasterProfile.HasAvailabilityZones() && (a.OrchestratorProfile.KubernetesConfig == nil || a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "") {
+		a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = "Standard"
```
CecileRobertMichon: if `a.OrchestratorProfile.KubernetesConfig == nil` is true it would cause a nil point dereference.

At this point this should never be true (we should be setting `KubernetesConfig` before we reach this line) but the logic of the if statement is contradictory.

pkg/acsengine/defaults.go
```diff
@@ -594,10 +615,7 @@ func setVMSSDefaults(a *api.Properties) {
 			if profile.SinglePlacementGroup == nil {
 				profile.SinglePlacementGroup = helpers.PointerToBool(api.DefaultSinglePlacementGroup)
 			}
-			if profile.SinglePlacementGroup == helpers.PointerToBool(false) {
-				profile.StorageProfile = api.ManagedDisks
-			}
-			if profile.HasAvailabilityZones() {
+			if profile.HasAvailabilityZones() && (a.OrchestratorProfile.KubernetesConfig == nil || a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "") {
```
CecileRobertMichon: ditto

pkg/api/vlabs/validate.go
```diff
@@ -274,6 +277,9 @@ func (a *Properties) validateOrchestratorProfile(isUpdate bool) error {
 						return errors.Errorf("loadBalancerSku is only available in Kubernetes version %s or greater; unable to validate for Kubernetes version %s",
 							minVersion.String(), o.OrchestratorVersion)
 					}
+					if *a.OrchestratorProfile.KubernetesConfig.ExcludeMasterFromStandardLB == false {
```
CecileRobertMichon: let's use the helper function `helpers.IsFalseBoolPointer` here instead of dereferencing the pointer which might be nil

pkg/api/vlabs/validate.go
```diff
@@ -470,6 +466,38 @@ func (a *Properties) validateAgentPoolProfiles(isUpdate bool) error {
...
+				// agent pool profiles
+				for _, agentPoolProfile := range a.AgentPoolProfiles {
+					if agentPoolProfile.AvailabilityProfile == AvailabilitySet {
+						return errors.New("Availability Zones are not supported with an AvailabilitySet. Please either remove availabilityProfile or set availabilityProfile to VirtualMachineScaleSets")
+					}
+					if agentPoolProfile.Count < len(agentPoolProfile.AvailabilityZones)*2 {
+						return errors.New("the node count and the number of availability zones provided can result in zone imbalance. To achieve zone balance, each zone should have at least 2 nodes or more")
+					}
+				}
+				if a.OrchestratorProfile.KubernetesConfig != nil && a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku != "Standard" {
```
CecileRobertMichon: `&& a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku != nil` 


Defined KubernetesConfig but no LoadBalancerSku defined should not fail

pkg/acsengine/defaults.go
```diff
@@ -422,11 +429,11 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
 			a.OrchestratorProfile.KubernetesConfig.UseInstanceMetadata = helpers.PointerToBool(api.DefaultUseInstanceMetadata)
 		}
 
-		if a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
+		if !a.HasAvailabilityZones() && a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
```
CecileRobertMichon: Is `LoadBalancerSku` not used if the cluster doesn't have Availability Zones?

ritazh: if we are not using availability zones, then `LoadBalancerSku` should be set to `Basic` if not set.

pkg/acsengine/defaults.go
```diff
@@ -584,8 +591,22 @@ func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
 	}
 }
 
-// setVMSSDefaults
-func setVMSSDefaults(a *api.Properties) {
+// setVMSSDefaultsForMasters
+func setVMSSDefaultsForMasters(a *api.Properties) {
+	if a.MasterProfile.Count > 100 {
```
CecileRobertMichon: The only supported values for `MasterProfile.Count` are 1, 3 and 5 so this is unreachable code

pkg/acsengine/defaults.go
CecileRobertMichon: Oh I misread that, missed the `!`, makes sense 👍

pkg/api/types.go
```diff
@@ -715,6 +717,31 @@ func (p *Properties) HasVirtualMachineScaleSets() bool {
 	return false
 }
 
+// HasZonesForAllAgentPools returns true if all of the agent pools have zones
+func (p *Properties) HasZonesForAllAgentPools() bool {
```
CecileRobertMichon: nit: this function could break as soon as it finds an agent pool without AZ similar to what  https://github.com/Azure/acs-engine/pull/3864/files#diff-0226d5bd49c7195fda1784e5dd4274c9R573 does
```
for _, agentPoolProfile := range p.AgentPoolProfiles { 
			if !agentPoolProfile.HasAvailabilityZones() { 
				return false
			} 
return true
```

pkg/api/vlabs/types.go
```diff
@@ -555,12 +561,31 @@ func (m *MasterProfile) IsVirtualMachineScaleSets() bool {
...
-	isAll := p.MasterProfile.IsVirtualMachineScaleSets()
-	if isAll {
+// HasAvailabilityZones returns true if the master profile has availability zones
+func (m *MasterProfile) HasAvailabilityZones() bool {
+	return m.AvailabilityZones != nil && len(m.AvailabilityZones) > 0
+}
+
+// IsClusterAllAvailabilityZones returns true if the cluster contains AZs for all agents and masters profiles
+func (p *Properties) IsClusterAllAvailabilityZones() bool {
+	isAll := p.MasterProfile != nil && p.MasterProfile.HasAvailabilityZones()
```
CecileRobertMichon: does this mean that this function should return false if MasterProfile is nil but all agents have AZs? Maybe it should be an || instead

What do you think about making this a single line? :
return  (p.MasterProfile == nil || p.MasterProfile.HasAvailabilityZones()) && p.HasZonesForAllAgentPools()

ritazh: this function should return true if the cluster contains all profiles with AZs. so in this case, if MasterProfile is nil then it should return false.

ritazh: > What do you think about making this a single line? :
return (p.MasterProfile == nil || p.MasterProfile.HasAvailabilityZones()) && p.HasZonesForAllAgentPools()

how about 
```
return (p.MasterProfile != nil && p.MasterProfile.HasAvailabilityZones()) && p.HasZonesForAllAgentPools())
```

CecileRobertMichon: So that means it won't be possible to use AZ for agent pool only (AKS) clusters? The way I see this if a cluster only has agents and all its agents use AZ then all the cluster nodes use AZ. If that's not the logic we are trying to achieve let's rename the function to something like `MastersAndAgentsUseAvailabilityZones`

ritazh: > AZ for agent pool only (AKS) clusters

That is the next thing we want to focus on after we have enabled AZ for masters and agents as it is done currently. 

> The way I see this if a cluster only has agents and all its agents use AZ then all the cluster nodes use AZ

This PR currently has validations to ensure all profiles (masters and agents) use AZ in the cluster. We do not allow a mix of AZ and non-AZ when creating a new cluster. Zones need to be set explicitly configured hence the requirement to ask the user to configure AZ for both masters and agents. Let me know if you have other questions or recommendations. 

----

To avoid confusion, I have added the following for the `availabilityZones` feature as documented in `clusterdefinition.md`. Please let me know if any of this is not clear or can be improved.

To protect your cluster from datacenter-level failures, you can enable the Availability Zones feature for your cluster by configuring `"availabilityZones"` for the master profile and all of the agentPool profiles in the cluster definition. This feature only applies to Kubernetes clusters version 1.12+. Supported values are arrays of strings, each representing a supported availability zone in a region for your subscription. e.g. `"availabilityZones": ["1","2"]` represents zone 1 and zone 2 can be used. To get supported zones for a region in your subscription, run `az vm list-skus --location centralus --query "[?name=='Standard_DS2_v2'].[locationInfo, restrictions"] -o table`. You should see values like `'zones': ['2', '3', '1']` appear in the first column. If `NotAvailableForSubscription` appears in the output, then create an Azure support ticket to enable zones for that region. NOTE: To ensure high availability, each profile must define at least two nodes per zone. e.g. An agent pool profile with 2 zones: `"availabilityZones": ["1","2"]` must have at least 4 nodes total with `"count": 4`. When `"availabilityZones"` is configured, the `"loadBalancerSku"` will default to `Standard` as Standard LoadBalancer is required for availability zones.

PR #3869, opened 2018-09-19
Adding DeleteApp func to AzureClient and returning appObjectID in CreateApp

shrutir25:
This PR adds functionality to delete an AAD App. An AAD App can be deleted by providing in the application Object ID. Currently, the `CreateApp` function does not retrieve or return this parameter. 

**If applicable**:
- [ ] documentation
- [x] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -365,7 +365,7 @@ func autofillApimodel(dc *deployCmd) error {
 					},
 				}
 			}
-			applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
+			_, applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
 			if err != nil {
 				return errors.Wrap(err, "apimodel invalid: ServicePrincipalProfile was empty, and we failed to create valid credentials")
 			}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -365,10 +365,11 @@ func autofillApimodel(dc *deployCmd) error {
 					},
 				}
 			}
-			_, applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
+			applicationResp, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
 			if err != nil {
 				return errors.Wrap(err, "apimodel invalid: ServicePrincipalProfile was empty, and we failed to create valid credentials")
 			}
+			applicationID := to.String(applicationResp.AppID)
 			log.Warnf("created application with applicationID (%s) and servicePrincipalObjectID (%s).", applicationID, servicePrincipalObjectID)
 
 			log.Warnln("apimodel: ServicePrincipalProfile was empty, assigning role to application...")
--- a/pkg/armhelpers/graph.go
+++ b/pkg/armhelpers/graph.go
@@ -29,7 +29,7 @@ func (az *AzureClient) CreateGraphApplication(ctx context.Context, applicationCr
 	return az.applicationsClient.Create(ctx, applicationCreateParameters)
 }
 
-// DeleteGraphApplication creates an application via the graphrbac client
+// DeleteGraphApplication deletes an application via the graphrbac client
 func (az *AzureClient) DeleteGraphApplication(ctx context.Context, applicationObjectID string) (result autorest.Response, err error) {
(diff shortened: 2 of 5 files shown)


pkg/armhelpers/graph.go
```diff
@@ -28,6 +29,11 @@ func (az *AzureClient) CreateGraphApplication(ctx context.Context, applicationCr
 	return az.applicationsClient.Create(ctx, applicationCreateParameters)
 }
 
+// DeleteGraphApplication creates an application via the graphrbac client
```
mboersma: This comment should say "deletes an application".

pkg/armhelpers/interfaces.go
```diff
@@ -102,7 +103,8 @@ type ACSEngineClient interface {
 
 	// CreateGraphPrincipal creates a service principal via the graphrbac client
 	CreateGraphPrincipal(ctx context.Context, servicePrincipalCreateParameters graphrbac.ServicePrincipalCreateParameters) (graphrbac.ServicePrincipal, error)
-	CreateApp(ctx context.Context, applicationName, applicationURL string, replyURLs *[]string, requiredResourceAccess *[]graphrbac.RequiredResourceAccess) (applicationID, servicePrincipalObjectID, secret string, err error)
+	CreateApp(ctx context.Context, applicationName, applicationURL string, replyURLs *[]string, requiredResourceAccess *[]graphrbac.RequiredResourceAccess) (applicationObjectID, applicationID, servicePrincipalObjectID, secret string, err error)
+	DeleteApp(ctx context.Context, appName, applicationObjectID string) (autorest.Response, error)
```
mboersma: Can we make this parameter `applicationName` for consistency with `CreateApp` and with other parameters?

pkg/armhelpers/graph.go
```diff
@@ -90,12 +97,22 @@ func (az *AzureClient) CreateApp(ctx context.Context, appName, appURL string, re
...
 	}
 
 	servicePrincipalObjectID = to.String(servicePrincipalResp.ObjectID)
 
-	return applicationID, servicePrincipalObjectID, servicePrincipalClientSecret, nil
+	return applicationObjectID, applicationID, servicePrincipalObjectID, servicePrincipalClientSecret, nil
+}
+
+// DeleteApp is a simpler method for deleting an application and the associated spn
+func (az *AzureClient) DeleteApp(ctx context.Context, appName, applicationObjectID string) (autorest.Response, error) {
```
mboersma: Although it's verbose, I think `applicationName` would be a more consistent parameter name.

pkg/armhelpers/interfaces.go
shrutir25: sure makes sense !

pkg/armhelpers/graph.go
shrutir25: yup ! I ll change it now

cmd/deploy.go
```diff
@@ -365,7 +365,7 @@ func autofillApimodel(dc *deployCmd) error {
 					},
 				}
 			}
-			applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
+			_, applicationID, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)
```
tariq1890: @shrutir25 I think it would be good to avoid underscore placeholders since it goes against the tenets of idiomatic Go. We could change the return type of CreateApp to autorest.Response instead.

I see two benefits

1. We address the concern above.

2. It also becomes consistent with DeleteApplication.

pkg/armhelpers/graph.go
```diff
@@ -50,7 +56,7 @@ func (az *AzureClient) ListRoleAssignmentsForPrincipal(ctx context.Context, scop
 }
 
 // CreateApp is a simpler method for creating an application
-func (az *AzureClient) CreateApp(ctx context.Context, appName, appURL string, replyURLs *[]string, requiredResourceAccess *[]graphrbac.RequiredResourceAccess) (applicationID, servicePrincipalObjectID, servicePrincipalClientSecret string, err error) {
+func (az *AzureClient) CreateApp(ctx context.Context, appName, appURL string, replyURLs *[]string, requiredResourceAccess *[]graphrbac.RequiredResourceAccess) (applicationRes graphrbac.Application, servicePrincipalObjectID, servicePrincipalClientSecret string, err error) {
```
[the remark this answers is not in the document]
shrutir25: Thanks for the review ! Just made the change.

shrutir25:
> whoops a nit : should be "PointerToString returns a pointer to a string"

I hadnt saved the file with the change :)

tariq1890:
Thanks once again @shrutir25

Issue #3850, opened 2018-09-14
move ARM template outputs into data model


jackfrancis:
Move these ARM template output properties into the data model so that we can get access to them without having to submit an operation to ARM:

```
    "vnetResourceGroup": {
        "type": "string",
        "value": "[variables('virtualNetworkResourceGroupName')]"
    },
    "subnetName": {
        "type": "string",
        "value": "[variables('subnetName')]"
    },
    "securityGroupName": {
        "type": "string",
        "value": "[variables('nsgName')]"
    },
    "virtualNetworkName": {
        "type": "string",
        "value": "[variables('virtualNetworkName')]"
    },
    "routeTableName": {
        "type": "string",
        "value": "[variables('routeTableName')]"
    },
    "primaryAvailabilitySetName": {
        "type": "string",
        "value": "[variables('primaryAvailabilitySetName')]"
    },
    "primaryScaleSetName": {
        "type": "string",
        "value": "[variables('primaryScaleSetName')]"
    }
```

tariq1890:
Closing since #3877 has been merged

PR #3877, opened 2018-09-19
Extracting property values to make ARM output variables accessible

tariq1890:
**What this PR does / why we need it**: This PR extracts values in code which are normally using template params and ARM variables. This is so done for easier programmatic access to the ARM output params

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: Working towards #3850 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


pkg/acsengine/template_generator.go
```diff
@@ -802,12 +802,24 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			imageRef := cs.Properties.MasterProfile.ImageRef
 			return imageRef != nil && len(imageRef.Name) > 0 && len(imageRef.ResourceGroup) > 0
 		},
+		"GetRouteTableName": func() string {
+			return cs.Properties.GetRouteTableName()
```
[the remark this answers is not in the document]
tariq1890: I think it should be tied to the Properties struct. What are thoughts on moving cs.Properties to a variable?

tariq1890:
@jackfrancis @CecileRobertMichon ready for review.

pkg/api/types.go
```diff
@@ -715,6 +719,113 @@ func (p *Properties) HasVirtualMachineScaleSets() bool {
 	return false
 }
 
+// K8sOrchestratorName returns the 3 character orchestrator code for kubernetes-based clusters.
+func (p *Properties) K8sOrchestratorName() string {
+	if p.OrchestratorProfile.IsKubernetes() ||
+		p.OrchestratorProfile.IsOpenShift() {
+		if p.HostedMasterProfile != nil {
+			return "aks"
```
CecileRobertMichon: this should be a const as well

is it a new value introduced?

CecileRobertMichon: also should `(p.OrchestratorProfile.IsOpenShift() && p.HostedMasterProfile != nil)` return `aks`?

CecileRobertMichon: also should `(p.OrchestratorProfile.IsOpenShift() && p.HostedMasterProfile != nil)` return `aks`?

tariq1890: no. You'll find this is params_k8s.go

pkg/api/types.go
```diff
@@ -715,6 +719,113 @@ func (p *Properties) HasVirtualMachineScaleSets() bool {
 	return false
 }
 
+// K8sOrchestratorName returns the 3 character orchestrator code for kubernetes-based clusters.
+func (p *Properties) K8sOrchestratorName() string {
```
CecileRobertMichon: unit test for this new function?

tariq1890: Sure. Missed that one

pkg/api/const.go
```diff
@@ -72,6 +76,8 @@ const (
 	// DefaultKubernetesFirstConsecutiveStaticIPOffsetVMSS specifies the IP address offset of master 0 in VMSS
 	// when VNET integration is enabled.
 	DefaultKubernetesFirstConsecutiveStaticIPOffsetVMSS = 4
+	// DefaultSubnetNameResourceSegmentIndex specifies the default subnet name resource segment index.
+	DefaultSubnetNameResourceSegmentIndex = 10
```
CecileRobertMichon: where is this used?

tariq1890: https://github.com/tariq1890/acs-engine/blob/extract_output_params/pkg/api/types.go#L784

CecileRobertMichon: ah my search failed me

pkg/api/types.go
```diff
@@ -717,6 +721,113 @@ func (p *Properties) HasVMSSAgentPool() bool {
...
+	return p.K8sOrchestratorName() + "-" + p.AgentPoolProfiles[0].Name + "-" + p.GenerateClusterID() + "-vmss"
+}
+
+// IsHostedMasterProfile returns true if the cluster has a hosted master
+func (p *Properties) IsHostedMasterProfile() bool {
+	return p.HostedMasterProfile != nil
+}
+
+// GetSubnetName returns the subnet name of the cluster based on its current configuration.
+func (p *Properties) GetSubnetName() string {
```
CecileRobertMichon: I don't see most of these functions being used in the var file yet, is it intentional?

tariq1890: yes. This is intended to be non-breaking change. If everyone is okay with what we have so far. We can start using these methods to extract the output parameters in the templates.

CecileRobertMichon:
I really like this 💯

pkg/api/types.go
CecileRobertMichon: > also should (p.OrchestratorProfile.IsOpenShift() && p.HostedMasterProfile != nil) return aks?
This is the current behavior so we should leave it that way

pkg/api/types.go
```diff
@@ -65,6 +69,17 @@ type Properties struct {
 	AzProfile               *AzProfile               `json:"azProfile,omitempty"`
 }
 
+// ClusterMetadata represents the metadata of the ACS cluster.
```
jackfrancis: s/ACS/AKS

jackfrancis: or, if it's not AKS-specific (i.e., if it also applies to acs-engine built clusters), then just say "Kubernetes cluster" in the comment

jackfrancis:
this /lgtm with a minor comment

pkg/acsengine/defaults.go
```diff
@@ -720,7 +720,7 @@ func setHostedMasterProfileDefaults(a *api.Properties) {
 
 func setDefaultCerts(a *api.Properties) (bool, error) {
 	if a.MasterProfile != nil && a.OrchestratorProfile.OrchestratorType == api.OpenShift {
-		return certgen.OpenShiftSetDefaultCerts(a, DefaultOpenshiftOrchestratorName, GenerateClusterID(a))
+		return certgen.OpenShiftSetDefaultCerts(a, api.DefaultOpenshiftOrchestratorName, a.GenerateClusterID())
```
weinong: nit: `GenerateClusterID` seems to suggest to generate every time it is called... `GetClusterID` seems better?

tariq1890: Certainly!

weinong: ok.. maybe it's generated after all...

jackfrancis: Just to clarify, the existing `GenerateClusterID` (before this change) also generates w/ every invocation as well, right?

weinong: Talked to @tariq1890 , it's seeded using the hash. so clusterID is always the same regardless how many times it is called

weinong: In this case, I'd like to get `GetClusterID` :)

tariq1890: @jackfrancis no, I've added a new field in Properties to store the ClusterID after it's generated during the first time.

PR #3879, opened 2018-09-20
Adding test case for Generate Cluster ID

tariq1890:
Added a Testcase to confirm the GenerateClusterID behavior. Committing since it adds to our coverage :)

### DIFF
--- a/pkg/acsengine/engine_test.go
+++ b/pkg/acsengine/engine_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


pkg/acsengine/engine_test.go
```diff
@@ -591,3 +591,18 @@ func TestGenerateKubeConfig(t *testing.T) {
...
+	p := &api.Properties{
+		AgentPoolProfiles: []*api.AgentPoolProfile{
+			{
+				Name: "foo_agent",
+			},
+		},
+	}
+
+	clusterID := GenerateClusterID(p)
+	if clusterID == "" {
```
CecileRobertMichon: Are there any additional checks we could do like the length of the string or its content with a regexp?

tariq1890: Sure. Makes sense. I can do that.

tariq1890:
@CecileRobertMichon done!

PR #3881, opened 2018-09-20
Replace deprecated Azure SDK method calls

tariq1890:
Replacing WaitForCompletion method calls with WaitForCompletion as the former is deprecated as per the Azure Go SDK godoc

### DIFF
--- a/pkg/armhelpers/compute.go
+++ b/pkg/armhelpers/compute.go
@@ -24,7 +24,7 @@ func (az *AzureClient) DeleteVirtualMachine(ctx context.Context, resourceGroup, 
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


jackfrancis:
@tariq1890 could you link to docs?

@weinong @amanohar for review

thanks!

tariq1890:
https://godoc.org/github.com/Azure/go-autorest/autorest/azure#Future.WaitForCompletion

tariq1890:
This is a non-breaking change since the deprecated method just redirects to the recommended method (i.e. WaitForCompletionRef).

jackfrancis:
@weinong @amanohar docs are very clear, so we'll merge this change to improve go-autorest usage hygiene

PR #3882, opened 2018-09-20
api server validation test on nodes during cse

jackfrancis:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
api server validation test on nodes during cse
```


jackfrancis:
Confirmed:

```
Executed "nc -vz 10.255.255.5 443" 66 times
+ echo 'Custom script finished successfully'
```

This may be delaying the ARM deployment time a bit, we'll want to watch.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -123,6 +127,10 @@ if $FULL_INSTALL_REQUIRED; then
     fi
 fi
 
+if [[ -z "${MASTER_NODE}" ]]; then
+    connectToAPIServer
```
CecileRobertMichon: does this need to be at the end of the script because the apiserver won't be reachable earlier?

jackfrancis: More importantly: none of the steps that occur before this (i.e., the end of CSE) depend on API server connectivity. So putting it earlier in the script would de-optimize the overall implementation of cluster deployment as we'd be blocking operations that should otherwise be performed concurrently.

CecileRobertMichon: This is the last step in the script so no steps occurring after this depend on it either. What is this test tryin to catch in that case? Won't it just add and additional step and make deployments longer? If we expect this to catch potential connectivity issues that were previously caught with the exit 50 test then all the dependency installs that happen before this do depend on it (if anything needs to be installed because of a missing version or if not using VHD).

PR #3890, opened 2018-09-21
fix(defaults): Checks for set backoff and rate limit values

thomastaylor312:
The defaults file was always setting rate limits and backoff to true for the azure cloud config. This checks if the value exists before setting defaults

**Release note**:
```release-note
The defaults file was always setting rate limits and backoff to true for the azure cloud config. This checks if the value exists before setting defaults
```

### DIFF
--- a/pkg/acsengine/defaults.go
+++ b/pkg/acsengine/defaults.go
@@ -351,30 +351,33 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
 			o.KubernetesConfig.ServiceCIDR = DefaultKubernetesServiceCIDR
 		}
 		// Enforce sane cloudprovider backoff defaults, if CloudProviderBackoff is true in KubernetesConfig
-		o.KubernetesConfig.CloudProviderBackoff = true
-		if o.KubernetesConfig.CloudProviderBackoffDuration == 0 {
-			o.KubernetesConfig.CloudProviderBackoffDuration = DefaultKubernetesCloudProviderBackoffDuration
+		if o.KubernetesConfig.CloudProviderBackoff {
+			if o.KubernetesConfig.CloudProviderBackoffDuration == 0 {
+				o.KubernetesConfig.CloudProviderBackoffDuration = DefaultKubernetesCloudProviderBackoffDuration
+			}
+			if o.KubernetesConfig.CloudProviderBackoffExponent == 0 {
+				o.KubernetesConfig.CloudProviderBackoffExponent = DefaultKubernetesCloudProviderBackoffExponent
+			}
+			if o.KubernetesConfig.CloudProviderBackoffJitter == 0 {
+				o.KubernetesConfig.CloudProviderBackoffJitter = DefaultKubernetesCloudProviderBackoffJitter
+			}
+			if o.KubernetesConfig.CloudProviderBackoffRetries == 0 {
+				o.KubernetesConfig.CloudProviderBackoffRetries = DefaultKubernetesCloudProviderBackoffRetries
+			}
 		}
-		if o.KubernetesConfig.CloudProviderBackoffExponent == 0 {
-			o.KubernetesConfig.CloudProviderBackoffExponent = DefaultKubernetesCloudProviderBackoffExponent
-		}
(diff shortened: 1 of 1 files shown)


pkg/acsengine/defaults.go
```diff
@@ -351,30 +351,33 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
...
-		}
-		if o.KubernetesConfig.CloudProviderBackoffExponent == 0 {
-			o.KubernetesConfig.CloudProviderBackoffExponent = DefaultKubernetesCloudProviderBackoffExponent
-		}
-		if o.KubernetesConfig.CloudProviderBackoffJitter == 0 {
-			o.KubernetesConfig.CloudProviderBackoffJitter = DefaultKubernetesCloudProviderBackoffJitter
-		}
-		if o.KubernetesConfig.CloudProviderBackoffRetries == 0 {
-			o.KubernetesConfig.CloudProviderBackoffRetries = DefaultKubernetesCloudProviderBackoffRetries
+		if o.KubernetesConfig.CloudProviderBackoff {
```
CecileRobertMichon: where is the default for CloudProviderBackoff being set if it's nil?

pkg/acsengine/defaults.go
```diff
@@ -351,30 +351,33 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
...
 		// Enforce sane cloudprovider rate limit defaults, if CloudProviderRateLimit is true in KubernetesConfig
 		// For k8s version greater or equal to 1.6.6, we will set the default CloudProviderRate* settings
-		o.KubernetesConfig.CloudProviderRateLimit = true
 		if o.KubernetesConfig.CloudProviderRateLimit && k8sSemVer.GTE(minVersion) {
-			if o.KubernetesConfig.CloudProviderRateLimitQPS == 0 {
-				o.KubernetesConfig.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
-			}
-			if o.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
-				o.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
+			if o.KubernetesConfig.CloudProviderRateLimit && k8sSemVer.GTE(minVersion) {
```
CecileRobertMichon: same uestion with `o.KubernetesConfig.CloudProviderRateLimit`

pkg/acsengine/defaults.go
thomastaylor312: It is a bool, so the zero value is `false`

pkg/acsengine/defaults.go
```diff
@@ -351,30 +351,33 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
...
+				if o.KubernetesConfig.CloudProviderRateLimitQPS == 0 {
+					o.KubernetesConfig.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
+				}
+				if o.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
```
tariq1890: Could we have a helper method to get the values for these. The helper method would return a default value is the said field has nil or its equivalent as its value.

pkg/acsengine/defaults.go
CecileRobertMichon: IMO we should still set it to the default constant https://github.com/Azure/acs-engine/blob/master/pkg/acsengine/const.go#L82 in case we ever decide to change that default

pkg/acsengine/defaults.go
thomastaylor312: These were pre-existing. I do think that would be a good idea though.

pkg/acsengine/defaults.go
thomastaylor312: I agree, but that probably shouldn't be set here. It could be the user has set it to `false` or that it defaulted to `false`

thomastaylor312:
@jackfrancis put together a more fully featured PR in #3891 that addresses this. Closing.

PR #3891, opened 2018-09-21
Actually allow cloudprovider rate limit / backoff disabling

jackfrancis:
**What this PR does / why we need it**: Cleaned up some tech debt: made these options `*bool` so it's easier to distinguish between the following user opinions:

1. I don't care
2. Enable
3. Disable

For user scenario #1 we will continue to enable by default.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Actually allow cloudprovider rate limit / backoff disabling
```


pkg/acsengine/defaults.go
```diff
@@ -350,26 +348,31 @@ func setOrchestratorDefaults(cs *api.ContainerService, isUpdate bool) {
...
-			o.KubernetesConfig.CloudProviderBackoffExponent = DefaultKubernetesCloudProviderBackoffExponent
+
+		if o.KubernetesConfig.CloudProviderBackoff == nil {
+			o.KubernetesConfig.CloudProviderBackoff = helpers.PointerToBool(DefaultKubernetesCloudProviderBackoff)
 		}
-		if o.KubernetesConfig.CloudProviderBackoffJitter == 0 {
-			o.KubernetesConfig.CloudProviderBackoffJitter = DefaultKubernetesCloudProviderBackoffJitter
+		// Enforce sane cloudprovider backoff defaults, if CloudProviderBackoff is true in KubernetesConfig
+		if helpers.IsTrueBoolPointer(o.KubernetesConfig.CloudProviderBackoff) {
+			if o.KubernetesConfig.CloudProviderBackoffDuration == 0 {
```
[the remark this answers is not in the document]
jackfrancis: Correct. That would be the equivalent of disabling the functionality altogether. (represented in the *bool property `CloudProviderBackoff`)

tariq1890: Can we move L351-L372 to another method?

pkg/acsengine/const.go
```diff
@@ -78,22 +78,6 @@ const (
 	DefaultKubernetesCtrlMgrTerminatedPodGcThreshold = "5000"
 	// DefaultKubernetesCtrlMgrUseSvcAccountCreds is "true", see --use-service-account-credentials at https://kubernetes.io/docs/admin/kube-controller-manager/
 	DefaultKubernetesCtrlMgrUseSvcAccountCreds = "false"
-	// DefaultKubernetesCloudProviderBackoff is false to disable cloudprovider backoff implementation for API calls
```
tariq1890: Any particular reason to move this to the api package?

jackfrancis: nope, not in the final implementation!

jackfrancis: (addressed in latest commit)

pkg/api/const.go
```diff
@@ -187,3 +187,18 @@ const (
...
+const (
+	// DefaultKubernetesCloudProviderBackoffRetries is 6, takes effect if DefaultKubernetesCloudProviderBackoff is true
+	DefaultKubernetesCloudProviderBackoffRetries = 6
+	// DefaultKubernetesCloudProviderBackoffJitter is 1, takes effect if DefaultKubernetesCloudProviderBackoff is true
+	DefaultKubernetesCloudProviderBackoffJitter = 1.0
+	// DefaultKubernetesCloudProviderBackoffDuration is 5, takes effect if DefaultKubernetesCloudProviderBackoff is true
+	DefaultKubernetesCloudProviderBackoffDuration = 5
+	// DefaultKubernetesCloudProviderBackoffExponent is 1.5, takes effect if DefaultKubernetesCloudProviderBackoff is true
+	DefaultKubernetesCloudProviderBackoffExponent = 1.5
+	// DefaultKubernetesCloudProviderRateLimitQPS is 3, takes effect if DefaultKubernetesCloudProviderRateLimit is true
```
andyzhangx: A few customers are complaining about set this issue too small (current is 3), are we going to increase this value?

rocketraman: @jackfrancis At least in my newly installed 1.12.1 acs-engine cluster the defaults led to many errors in the kube controller like this:

```
E1028 21:32:55.584474       1 operation_generator.go:187] VolumesAreAttached failed for checking on node "k8s-agentpool2-36147952-vmss000000" with: azure - cloud provider rate limited(read) for operation:VMSSGetInstanceView
```

Upping the limits manually in `/etc/kubernetes/azure.json` and bouncing the controller pod solved the problem, but I wouldn't expect to have to do or know about this at all.

jackfrancis: @rocketraman how many nodes are in your cluster? what limits are working for you?

rocketraman: @jackfrancis On this 1.12.1 cluster I have only 6 agent nodes. I used a bucket of 70 (instead of 10) and a qps of 6 instead of 3.

jackfrancis:
@feiskyer Does @rocketraman's error look like a rate limit error? Or is it just wrapped inside the rate limiter stdout?

rocketraman:
> @feiskyer Does @rocketraman's error look like a rate limit error? Or is it just wrapped inside the rate limiter stdout?

@jackfrancis  For what its worth, the errors disappeared after changing the buckets and qps settings, so it does seem to be a rate limit error.

feiskyer:
@jackfrancis @rocketraman Yep, the error message is from cloud provider's rate limits.

PR #3892, opened 2018-09-21
Re-enable outbound internet check

jackfrancis:
**What this PR does / why we need it**: The merry-go-round continues.

Following recent incidents where CSE 50 aided in identifying persistent network i/o failures on vms, let's put this back. I'm advocating we increase the tolerance, so there are 40 attempts each to www.google.com and www.1688.com before declaring a lack of outbound internet connectivity.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
Re-enable outbound internet check
```


tariq1890:
+100

PR #3894, opened 2018-09-24
Handle iterated subtest execution correctly

tariq1890:
Please see : https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20180331/validate_test.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/validate_test.go
@@ -646,7 +646,7 @@ func TestValidateAADProfile(t *testing.T) {
 				Properties: &Properties{
 					EnableRBAC: helpers.PointerToBool(true),
 					AADProfile: &AADProfile{
-						ServerAppID: "ccbfaea3-7312-497e-81d9-9ad9b8a99853",
+						ServerAppSecret: "ccbfaea3-7312-497e-81d9-9ad9b8a99853",
 					},
 				},
 			},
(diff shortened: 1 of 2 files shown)


pkg/api/agentPoolOnlyApi/v20180331/validate_test.go
```diff
@@ -646,7 +646,7 @@ func TestValidateAADProfile(t *testing.T) {
 				Properties: &Properties{
 					EnableRBAC: helpers.PointerToBool(true),
 					AADProfile: &AADProfile{
-						ServerAppID: "ccbfaea3-7312-497e-81d9-9ad9b8a99853",
+						ServerAppSecret: "ccbfaea3-7312-497e-81d9-9ad9b8a99853",
```
jackfrancis: How did this test ever succeed? Is it because before this fix these tests were never being run?

tariq1890: yes :'(

pkg/api/vlabs/validate.go
```diff
@@ -371,7 +371,7 @@ func (a *Properties) validateMasterProfile() error {
 			return e
 		}
 		if !a.IsClusterAllVirtualMachineScaleSets() {
-			return errors.New("VirtualMachineScaleSets for master profile must be used together with virtualMachineScaleSets for agent profiles. Set \"availabilityProfile\" to \"VirtualMachineScaleSets\" for agent profiles")
+			return errors.New("VirtualMachineScaleSets for master profile must be used together with virtualMachineScaleSets for agent profiles. See \"availabilityProfile\" to \"VirtualMachineScaleSets\" for agent profiles")
 		}
```
ritazh: Is there a reason we don't want `Set \"availabilityProfile\" to \"VirtualMachineScaleSets\" for agent profiles` here?

tariq1890: whoops. Major fail. Will correct .

tariq1890: @ritazh Corrected. Thanks!

PR #3908, opened 2018-09-26
Accelerated networking for Windows

### LINKED ISSUE #3678: Support AcceleratedNetworkingEnabled for Windows-based Agent pools

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of acs-engine?**: latest (0.20.9)

---

As part of accelerated networking support (PR https://github.com/Azure/acs-engine/pull/3011), enableAcceleratedNetworking setting does not take any effect if the agent pool is of osType=Windows. It seems that the files kuberneteswinagentresourcesvmss.t and kuberneteswinagentresourcesvmas.t do not have the enableAcceleratedNetworking setting whereas the corresponding linux files have. 
Azure supports Accelerated Networking on Windows VMs. See https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-powershell

Kubernetes

**What happened**:
Specifying AcceleratedNetworkingEnabled for agent pool profile when OsType=Windows does not take any effect. The generated ARM template (azuredeploy.json) does not have the setting for EnableAcceleratedNetworking on both VMSS based or AS based VMs. The ApiModel.json still has the setting to true

azuredeploy.json has setting for enableAcceleratedNetworking when the agent pool profile has osType=Windows and AcceleratedNetworkingEnabled set to true.

Take the file examples/kubernetes-config/kubernetes-accelerated-network.json and in the agent pool, add osType=Windows (keep the AcceleratedNetworkingEnabled to true). Generate the ARM template. The ARM template does not have the enableAcceleratedNetworking enabled on the network interface configuration.

CecileRobertMichon:
**What this PR does / why we need it**: Adds accelerated networking support for Windows nodes. The reason for using a separate var for Windows is that using the same variable as Linux would break back compat because the var could have previously set to true even though the property wasn't applied on windows nodes and so it is not an accurate representation of the current state of the cluster.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3678 

**If applicable**:
- [x] documentation
- [ ] unit tests

**Release note**:

```release-note
```


parts/k8s/kuberneteswinagentresourcesvmas.t
```diff
@@ -16,7 +16,11 @@
     },
 {{end}}
     {
+      {{if .AcceleratedNetworkingEnabled}}
+      "apiVersion": "2018-04-01",
```
CecileRobertMichon: will need to merge this with #3909

CecileRobertMichon:
back compat: https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/151/console

CecileRobertMichon:
new back compat: https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/155/console

CecileRobertMichon:
@PatrickLang this is ready for review

PatrickLang:
@dineshgovindasamy @daschott - any concerns with allowing accelerated networking for Kubernetes nodes? It's off by default but customers can opt-in on compatible VM SKUs

daschott:
@jmesser81 any concerns with this?

CecileRobertMichon:
@PatrickLang as is the PR is actually enabling it by default (same as for Linux) for supported SKUs. I will change it to disabled by default for now since it seems to be what we want.

JMesser81:
@daschott @CecileRobertMichon - no concerns with having it be as an opt-in. I don't think we've tested this but there isn't any reason why it shouldn't just work.

pkg/acsengine/defaults.go
```diff
@@ -647,6 +647,10 @@ func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabled = helpers.PointerToBool(!isUpgrade && !isScale && helpers.AcceleratedNetworkingSupported(profile.VMSize))
 		}
 
+		if profile.AcceleratedNetworkingEnabledWindows == nil {
+			profile.AcceleratedNetworkingEnabledWindows = helpers.PointerToBool(false)
```
jackfrancis: bool literal `false` should be a named const

romina2001:
@CecileRobertMichon is this still possible? Having accelerated networking enabled for windows node?

CecileRobertMichon:
@romina2001 this issue has the latest status on this: https://github.com/Azure/aks-engine/issues/2281

PR #3909, opened 2018-09-26
Update apiversion to make it consistent in k8s templates

### LINKED ISSUE #3835: Update default apiversion

CecileRobertMichon:
**What this PR does / why we need it**: Update the default api version to 2018-06-01 (the most recent one used in the code, for managed identity) and use the default in k8s templates instead of having different hardcoded versions for each feature that needs a more recent api version. This change will need to be tested carefully to make sure that we don't regress any features. However, I expect that it won't be as big of an api version jump as it seems since most resources were already using a 2018 api version (2018-04-01) for accelerated networking (on by default).

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3835 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -138,7 +138,7 @@ The following is an example of the template.json file.
 				"description": "Storage API Version"
 			}
 		},
-		"apiVersionDefault": {
+		"apiVersionCompute": {
 			"type": "string",
 			"minLength": 1,
 			"metadata": {
@@ -183,7 +183,7 @@ The following is an example of the template.json file.
       }, 
       "type": "Microsoft.Storage/storageAccounts"	
 	}, {
-      "apiVersion": "[parameters('apiVersionDefault')]",
+      "apiVersion": "[parameters('apiVersionCompute')]",
       "dependsOn": [],
       "location": "[resourceGroup().location]",
       "type": "Microsoft.Compute/virtualMachines/extensions",
@@ -222,7 +222,7 @@ Replace "**EXTENSION-NAME**" with the name of the extension.
 {
     "name": "EXTENSION-NAME",
     "type": "Microsoft.Resources/deployments",
-    "apiVersion": "[variables('apiVersionDefault')]",
+    "apiVersion": "[variables('apiVersionCompute')]",
     "dependsOn": [
         "vmLoopNode"
     ],
@@ -233,8 +233,8 @@ Replace "**EXTENSION-NAME**" with the name of the extension.
             "contentVersion": "1.0.0.0"
         },
         "parameters": {
-            "apiVersionDefault": {
-                "value": "[variables('apiVersionDefault')]"
+            "apiVersionCompute": {
+                "value": "[variables('apiVersionCompute')]"
             },
             "username": {
                 "value": "[parameters('linuxAdminUsername')]"
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -222,7 +222,7 @@ Replace "**EXTENSION-NAME**" with the name of the extension.
 {
     "name": "EXTENSION-NAME",
     "type": "Microsoft.Resources/deployments",
-    "apiVersion": "[variables('apiVersionLinkDefault')]",
+    "apiVersion": "[variables('apiVersionDefault')]",
     "dependsOn": [
         "vmLoopNode"
     ],
@@ -233,9 +233,6 @@ Replace "**EXTENSION-NAME**" with the name of the extension.
             "contentVersion": "1.0.0.0"
         },
         "parameters": {
-            "apiVersionStorage": {
-                "value": "[variables('apiVersionStorage')]"
-            },
             "apiVersionDefault": {
                 "value": "[variables('apiVersionDefault')]"
             },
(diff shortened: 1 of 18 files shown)


docs/extensions.md
```diff
@@ -222,7 +222,7 @@ Replace "**EXTENSION-NAME**" with the name of the extension.
 {
     "name": "EXTENSION-NAME",
     "type": "Microsoft.Resources/deployments",
-    "apiVersion": "[variables('apiVersionLinkDefault')]",
+    "apiVersion": "[variables('apiVersionDefault')]"
```
tariq1890: why remove the "," here?

CecileRobertMichon: whoops that was accidental, will fix thanks for catching

tariq1890:
I've re-run the linux e2e test

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -84,19 +80,20 @@
    {
       "location": "[variables('location')]",
       "name": "[variables('{{.Name}}AvailabilitySet')]",
-      "apiVersion": "[variables('apiVersionStorageManagedDisks')]",
+      "apiVersion": "[variables('apiVersionDefault')]",
```
jackfrancis: move the `apiVersion` change out of the `{{if .IsManagedDisks}}` block now that it's the same for both conditions

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -257,7 +246,7 @@
...
+              "uri": "[concat(reference(concat('Microsoft.Storage/storageAccounts/',variables('storageAccountPrefixes')[mod(add(div(copyIndex(variables('{{.Name}}Offset')),variables('maxVMsPerStorageAccount')),variables('{{.Name}}StorageAccountOffset')),variables('storageAccountPrefixesCount'))],variables('storageAccountPrefixes')[div(add(div(copyIndex(variables('{{.Name}}Offset')),variables('maxVMsPerStorageAccount')),variables('{{.Name}}StorageAccountOffset')),variables('storageAccountPrefixesCount'))],variables('{{.Name}}AccountName')),variables('apiVersionDefault')).primaryEndpoints.blob,'osdisk/', variables('{{.Name}}VMNamePrefix'), copyIndex(variables('{{.Name}}Offset')), '-osdisk.vhd')]"
```
jackfrancis: Gah. What's being changed here?

jackfrancis:
@weinong @nicolerenee FYI, the AKS changes here are:

- The `"type": "Microsoft.Network/networkInterfaces"` resource type is upgraded to use the `"2018-06-01"` API
  - *was* using version `"2018-04-01"`
- The `"type": "Microsoft.Storage/storageAccounts"` resource type is upgraded to use the `"2018-06-01"` API
  - *was* using version `"2016-04-30-preview"`
- The `"type": "Microsoft.Compute/virtualMachines"` resource type is upgraded to use the `"2018-06-01"` API
  - *was* using version `"2016-04-30-preview"`

parts/k8s/kubernetesagentresourcesvmas.t
CecileRobertMichon: s/`variables('apiVersionStorage')`/`variables('apiVersionDefault')`/

tariq1890: such noob diffing by GitHub

parts/k8s/kubernetesmasterresources.t
```diff
@@ -1,14 +1,16 @@
...
-            "platformFaultDomainCount": 2,
-            "platformUpdateDomainCount": 3,
-		        "managed" : true
-        },
+      {
+        "platformFaultDomainCount": 2,
+        "platformUpdateDomainCount": 3
+      },
+      "sku": {
+        "name": "Aligned"
```
CecileRobertMichon: The only significant change in templates in this PR: `managed` is now deprecated in favor of `Aligned` sku for availability sets with managed disks 

https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/availabilitysets

jackfrancis: This change is simply to fulfill the new interface defined by the new API version, right? (i.e., no functional change there)

CecileRobertMichon: It should be from what I understand. That's something we should verify and be 100% confident about before we commit this change.

CecileRobertMichon:
Interestingly `2018-06-01` isn't a valid apiversion for storage accounts but `2018-07-01` is. I'll need to make some changes (that's why all the tests are failing).

weinong:
LGTM. In general, we will tell whether it breaks or not when we vendor it in

parts/k8s/kubernetesagentresourcesvmas.t
CecileRobertMichon: changed the `apiVersion` back to `apiVersionStorage` for storage account so this isn't true anymore

CecileRobertMichon:
Summary of changes:

CecileRobertMichon:
upgrade back-compat: https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/174/console

PR #3911, opened 2018-09-27
Search domain improvements

### LINKED ISSUE #3737: Is Custom Search Domain functionality broken?

YES

ISSUE

v0.20.9

Kubernetes 1.11.2

**What happened**:

We have a custom vnet k8s installation. The vnet uses a non-standard DNS server (corporate Active Directory), so we need to make use of CSD functionality in order to get machines able to resolve each other by DNS.

I have the following configuration block to my apimodel.json

```
"properties": {
    "linuxProfile": {
      "customSearchDomain": {
        "name": "mydomain.com",
        "realmUser": "SVC_blah",
        "realmPassword": "biglongpassword"
      }
    }
}
```

On deployment, first thing I noticed was exit code 80.

This is because realm,sssd etc packages failed to install because no prior `apt-get update` before attempting to install the packages, and thus 404'ing on locating those packages.

After fixing that, then I realised the `realmUser`, `realmPassword`, and `domain` variables were not attempted to be replaced as per https://github.com/Azure/acs-engine/blob/a79c2d7f5464b8a42748de69ca4481ca7dde4a9c/parts/k8s/kubernetesagentcustomdata.yml#L174-L178 

So I manually substituted those variables into the script myself just to see if I could get that working, and found that the `realm` command is missing the `--computer-ou` switch, which I need to join the machine to a specific OU.

ie.
```
echo "biglongpassword" | realm join -U SVC_blah@`echo "mydomain.com" | tr /a-z/ /A-Z/` `echo "mydomain.com" | tr /a-z/ /A-Z/` --computer-ou "OU=bleh,OU=blah,DC=MYDOMAIN,DC=COM"
```

I appreciate maybe the --computer-ou use case hasn't arisen for anyone yet, and I don't mind helping to produce that feature.

After adding the `--computer-ou` manually I was able to get the machine to join the domain and an `/etc/krb5.keytab` file generated. 

Then I set about trying to use a variance of the `register-dns.sh` script found at https://github.com/tesharp/acs-engine/blob/register-dns-extension/extensions/register-dns/v1/register-dns.sh 

I think I need to add the `-g` switch to the `nsupdate` command to make it use established kerberos creds, but it doesn't seem to find them, and also complains it can't find a default realm anywhere (as there is no /etc/krb5.conf file to find it).

I believe you can put such a thing in the `/etc/sssd/sssd.conf` file, so maybe thats enough, but at this point I was hoping for a bit of advice to see whether I'm going down the right road here before going deeper! 

Machine registration, DNS working and general happiness

Try it yourself as per above.

grenzr:

As noted in #3737, I was unable to get this feature working out of the box without:

- FEATURE: addition of computerOU switch to the realm command so that machines can join into a specific OU path in Active Directory
- BUGFIX: addition of `apt-get update` before custom search domain packages are installed
- BUGFIX: fix custom search domain sed replacement in acs-engine/parts/k8s/kubernetesagentcustomdata.yml (not yet working)

fixes #3737

**If applicable**:
- [x] documentation
- [ ] unit tests

**Release note**:

```release-note
- FEATURE: addition of computerOU switch to the realm command so that machines can join into a specific OU path in Active Directory
- BUGFIX: addition of `apt-get update` before custom search domain packages are installed
- BUGFIX: fix custom search domain sed replacement in acs-engine/parts/k8s/kubernetesagentcustomdata.yml (not yet working)
```

### DIFF
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -380,6 +380,7 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
     sed -i "s|<searchDomainName>|{{WrapAsParameter "searchDomainName"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmUser>|{{WrapAsParameter "searchDomainRealmUser"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmPassword>|{{WrapAsParameter "searchDomainRealmPassword"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
+    sed -i "s|<searchDomainComputerOU>|{{WrapAsParameter "searchDomainComputerOU"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"    
 {{end}}
 {{if .OrchestratorProfile.KubernetesConfig.IsContainerMonitoringEnabled}}
     sed -i "s|<omsAgentVersion>|{{WrapAsParameter "omsAgentVersion"}}|g" "/etc/kubernetes/addons/omsagent-daemonset.yaml"
--- a/parts/k8s/kubernetesmastercustomdata.yml
+++ b/parts/k8s/kubernetesmastercustomdata.yml
@@ -380,7 +380,7 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
     sed -i "s|<searchDomainName>|{{WrapAsParameter "searchDomainName"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmUser>|{{WrapAsParameter "searchDomainRealmUser"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmPassword>|{{WrapAsParameter "searchDomainRealmPassword"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
-    sed -i "s|<searchDomainComputerOU>|{{WrapAsParameter "searchDomainComputerOU"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"    
+    sed -i "s|<searchDomainComputerOU>|{{WrapAsParameter "searchDomainComputerOU"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
 {{end}}
 {{if .OrchestratorProfile.KubernetesConfig.IsContainerMonitoringEnabled}}
     sed -i "s|<omsAgentVersion>|{{WrapAsParameter "omsAgentVersion"}}|g" "/etc/kubernetes/addons/omsagent-daemonset.yaml"
--- a/docs/clusterdefinition.md
+++ b/docs/clusterdefinition.md
@@ -540,6 +540,7 @@ A cluster can have 0 to 12 agent pool profiles. Agent Pool Profiles are used for
 | customSearchDomain.name          | no       | describes the search domain to be used on all linux clusters                                                                                                                     |
 | customSearchDomain.realmUser     | no       | describes the realm user with permissions to update dns registries on Windows Server DNS                                                                                         |
 | customSearchDomain.realmPassword | no       | describes the realm user password to update dns registries on Windows Server DNS                                                                                                 |
+| customSearchDomain.computerOU | no       | optional OU path to register acs-engine machines into Windows Server DNS                                                                                                 |
 | customNodesDNS.dnsServer         | no       | describes the IP address of the DNS Server                                                                                                                                       |
 
 #### secrets
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -207,6 +207,7 @@ AGENT_ARTIFACTS_CONFIG_PLACEHOLDER
     sed -i "s|<searchDomainName>|{{WrapAsParameter "searchDomainName"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmUser>|{{WrapAsParameter "searchDomainRealmUser"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmPassword>|{{WrapAsParameter "searchDomainRealmPassword"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
+    sed -i "s|<searchDomainComputerOU>|{{WrapAsParameter "searchDomainComputerOU"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"    
 {{end}}
 
 - path: "/opt/azure/containers/provision.sh"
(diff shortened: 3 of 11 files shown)


jackfrancis:
This /lgtm @CecileRobertMichon could you take a pass as well?

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -356,6 +356,7 @@ MASTER_ARTIFACTS_CONFIG_PLACEHOLDER
     sed -i "s|<searchDomainName>|{{WrapAsParameter "searchDomainName"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmUser>|{{WrapAsParameter "searchDomainRealmUser"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
     sed -i "s|<searchDomainRealmPassword>|{{WrapAsParameter "searchDomainRealmPassword"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
+    sed -i "s|<searchDomainComputerOU>|{{WrapAsParameter "searchDomainComputerOU"}}|g" "/opt/azure/containers/setup-custom-search-domains.sh"
```
CecileRobertMichon: you will need to add this line to https://github.com/Azure/acs-engine/blob/fix-multiple-agentpools/parts/k8s/kubernetesmastercustomdatavmss.yml#L384 as well

grenzr: ok thanks, will do - however my issue is that the sed substitutions do not occur during bootstrap of each machine.
see comment in #3737  for further info: 
https://github.com/Azure/acs-engine/issues/3737#issuecomment-418442112 

I'm not sure the kubelet service (which I believe is supposed to trigger this script) is working properly. Any idea why that might be?

grenzr:
freshly rebased and added searchDomainComputerOU as requested. Just need some support on why the sed substitution script (supposedly triggered by kubelet.service) doesn't occur during machine bootstrap if possible please?

jackfrancis:
@grenzr what does the shell script look like after kubelet starts? Specifically the value of `sdComputerOU=`?

(speaking to your statement that the sed replacement isn't working)

grenzr:
@jackfrancis soo... I just rebased this branch locally with v0.22.4 and built a new acs-engine binary and tested a deployment with it (including my changes above).

It failed again with exit code 80 on the CSE.

The contents of `/opt/azure/containers/setup-custom-search-domains.sh` looks like this :

```
# cat /opt/azure/containers/setup-custom-search-domains.sh
#!/bin/bash
set -x
source /opt/azure/containers/provision_source.sh

sdName="<searchDomainName>"
sdRealmUser=$"<searchDomainRealmUser>"
sdRealmPassword=$"<searchDomainRealmPassword>"
sdComputerOU=$"<searchDomainComputerOU>"
ucDomainName=$(echo "${sdName}" | tr /a-z/ /A-Z/)

computerOUSwitch=""
if [[ ! -z "${sdComputerOU}" ]]; then
  computerOUSwitch="--computer-ou='${sdComputerOU}'"
fi

echo "  dns-search ${sdName}" >> /etc/network/interfaces.d/50-cloud-init.cfg
systemctl_restart 20 5 10 restart networking

retrycmd_if_failure 10 5 120 apt-get update
retrycmd_if_failure 10 5 120 apt-get -y install \
  realmd \
  sssd \
  sssd-tools \
  samba-common \
  samba \
  python2.7 \
  samba-libs \
  packagekit

echo "${sdRealmPassword}" | \
  realm join -U "${sdRealmUser}@${ucDomainName}" ${ucDomainName} ${computerOUSwitch}
```

ie. none of the substitute occurred at all.

`/etc/systemd/system/kubelet.service` contains:

```
[Unit]
Description=Kubelet
ConditionPathExists=/usr/local/bin/kubelet
# Required

[Service]
Restart=always
EnvironmentFile=/etc/default/kubelet
SuccessExitStatus=143
ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
ExecStartPre=/bin/mkdir -p /var/lib/cni
ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
# This is a partial workaround to this upstream Kubernetes issue:
#  https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731
ExecStartPre=/sbin/sysctl -w net.ipv4.tcp_retries2=8
ExecStartPre=-/sbin/ebtables -t nat --list
ExecStartPre=-/sbin/iptables -t nat --list
ExecStart=/usr/local/bin/kubelet \
        --enable-server \
        --node-labels="${KUBELET_NODE_LABELS}" \
        --v=2 \
        --volume-plugin-dir=/etc/kubernetes/volumeplugins \
        $KUBELET_CONFIG $KUBELET_OPTS \
        $KUBELET_REGISTER_NODE $KUBELET_REGISTER_WITH_TAINTS

[Install]
WantedBy=multi-user.target
```

So when kubelet starts its supposed to call this first: `/opt/azure/containers/kubelet.sh` (which contains the sed lines we care about at the bottom).

`systemctl list-unit-files` shows  `kubelet.service                           disabled` so I'm guessing this service never starts in order to do the substitutions.

However, the custom-search-domains.sh is called, but fails (as expected) with :

```
Processing triggers for libc-bin (2.23-0ubuntu10) ...
Processing triggers for systemd (229-4ubuntu21.4) ...
Processing triggers for ureadahead (0.100.0-19) ...
Processing triggers for ufw (0.35-0ubuntu2) ...
Processing triggers for dbus (1.10.6-1ubuntu3.3) ...
+ '[' 0 -eq 0 ']'
+ break
+ echo Executed '"apt-get' -y install realmd sssd sssd-tools samba-common samba python2.7 samba-libs 'packagekit"' 1 times
Executed "apt-get -y install realmd sssd sssd-tools samba-common samba python2.7 samba-libs packagekit" 1 times
+ echo '<searchDomainRealmPassword>'
+ realm join -U '<searchDomainRealmUser>@<SEARCHDOMAINNAME>' '<SEARCHDOMAINNAME>' '--computer-ou='\''<searchDomainComputerOU>'\'''
realm: No such realm found
```

Isn't the kubelet service supposed to be enabled or if not, when is it? If not at boot then it looks like its happening too late for the custom search domain script.

grenzr:
heh... it can't start on boot if the service isn't installed yet, eh? :)
looks like its hanging onto the `ensureKubelet()` function, which is called by `/opt/azure/containers/provision.sh`..

Yeah looks like ensureKubelet is called too late as it happens well after attempts to run `$CUSTOM_SEARCH_DOMAIN_SCRIPT` .. I think here lies the problem..

grenzr:
freshly rebased from master, and have now provided a PoC solution in the latest commit, which appears to bring up a working cluster now.

jackfrancis:
@grenzr let's undo the last commit, that's not the way we want to move this forward (it has wide ranging consequences)

Question: is the new computerOU configuration optional? Or a new requirement for this functionality?

grenzr:
@jackfrancis sure, I'll remove it shortly. Was only there temporarily for demo. Do you have any thoughts on a better alternative?

Yes, the computerOU I believe is optional currently, as I wanted to maintain backwards compatibility with the existing implementation (although I'm not sure it was working properly to begin with?)

This piece only adds the option to the realm command if its non-empty:
```
computerOUSwitch=""
if [[ ! -z "${sdComputerOU}" ]]; then
  computerOUSwitch="--computer-ou='${sdComputerOU}'"
fi
```
and the defaultValue for it is an empty string in `kubernetesparams.t`

Anything else I've missed?

I haven't run any ginkgo tests across it locally yet.

grenzr:
PoC commit removed

CecileRobertMichon:
@grenzr thank you for your contribution! We've relocated to https://github.com/Azure/aks-engine/. What is the status of this PR? I am closing this one as we are no longer accepting PRs for acs-engine but love to see this go through in aks-engine. Please let me know if you are willing to reopen the PR at https://github.com/Azure/aks-engine/pulls. Thanks!

PR #3917, opened 2018-09-27
[WIP]Helper method to escape special characters in bash

tariq1890:
This PR adds a utility method to escape special chars in bash. Normally we'd get around this by wrapping the string in single quotes. The utility method handles the edge cases where we need to single-quote-wrap strings with single quotes.

### DIFF
--- a/pkg/helpers/helpers_test.go
+++ b/pkg/helpers/helpers_test.go
@@ -5,6 +5,8 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"math/rand"
+	"os/exec"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/helpers/helpers_test.go
```diff
@@ -231,3 +233,29 @@ func TestEqualError(t *testing.T) {
...
+	testcases := []string{
+		"Hel'lo'p\"la`ygr'ound'",
+		`"PwEV@QG7/PYt"re9`,
+	}
+
+	for _, test := range testcases {
+		r := BashEscapeString(test)
+
+		if runtime.GOOS != "windows" {
+			out, err := exec.Command("/bin/sh", "-c", "testvar="+r+"; echo $testvar").Output()
```
jackfrancis: Just to be clear let's do `/bin/bash`

tariq1890: Ok

pkg/helpers/helpers_test.go
```diff
@@ -231,3 +233,29 @@ func TestEqualError(t *testing.T) {
 		}
 	}
 }
+
+func TestBashEscapeString(t *testing.T) {
+
+	testcases := []string{
+		"Hel'lo'p\"la`ygr'ound'",
+		`"PwEV@QG7/PYt"re9`,
```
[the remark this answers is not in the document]
tariq1890: Resolved

pkg/helpers/helpers_test.go
```diff
@@ -231,3 +233,109 @@ func TestEqualError(t *testing.T) {
...
+		`conse''cutive`,
+		"conse\\\\cutive",
+		"consec\"\"utive",
+		`PwEV@QG7/PYt"re9`,
+		"Lnsr@191",
+		"Jach#321",
+		"Bgmo%219",
+	}
+
+	for _, test := range tests {
```
CecileRobertMichon: why not merge the two test functions so we only call `BashEscapeString` once for each test case?

tariq1890: Sure thing.

PR #3919, opened 2018-09-28
Allow service principals with double-quote (") and backslash (\)

### LINKED ISSUE #3038: deal with backtick (`) chars in sp keys
Exact details TBD, but from a high level we should either fail validation or escape backticks. Observation suggests that currently we're not dealing with it at all which yields non-deterministic side-effects during deployment.

tariq1890:
This fixes an issue when passwords contain JSON special chars like " and \

Fixes #3038

### DIFF
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -134,6 +134,9 @@ function configureK8s() {
     set +x
     echo "${KUBELET_PRIVATE_KEY}" | base64 --decode > "${KUBELET_PRIVATE_KEY_PATH}"
     echo "${APISERVER_PUBLIC_KEY}" | base64 --decode > "${APISERVER_PUBLIC_KEY_PATH}"
+    # Perform the required JSON escaping for special characters " and \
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesconfigs.sh
```diff
@@ -131,6 +131,10 @@ function configureK8s() {
     chmod 0600 "${AZURE_JSON_PATH}"
     chown root:root "${AZURE_JSON_PATH}"
 
+    # Perform the required JSON escaping for special characters " an \
```
jackfrancis: s/an/and

tariq1890: lol

parts/k8s/kubernetesconfigs.sh
```diff
@@ -131,6 +131,10 @@ function configureK8s() {
     chmod 0600 "${AZURE_JSON_PATH}"
     chown root:root "${AZURE_JSON_PATH}"
 
+    # Perform the required JSON escaping for special characters " and \
```
CecileRobertMichon: let's move this below `set +x` on line 138

tariq1890:
@CecileRobertMichon Fixed :)

CecileRobertMichon:
/lgtm

oops

PR #3922, opened 2018-09-28
Stop ginkgo tests after first failure

mboersma:
In some environments, each e2e spec can wait for a timeout before failing, which can make the overall test suite take a long time before it returns an overall failure. This tells `ginkgo` to bail after the first red flag instead.

The tradeoff here is speed versus potentially gaining more knowledge about the overall set of failed test cases. That's usually been the right tradeoff IMHO, but I'm curious what others think.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Stop ginkgo tests after first failure
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ ifneq ($(GIT_BASEDIR),)
 endif
 
 test: generate
-	ginkgo -skipPackage test/e2e/dcos,test/e2e/kubernetes,test/e2e/openshift -r .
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
would this change impact our log gathering?

jackfrancis:
@CecileRobertMichon if it does then our log gathering is probably not properly implemented

mboersma:
> would this change impact our log gathering?

I don't think so. It will just stop executing any new test cases after one fails, but cleanup / teardown and the rest of `ginkgo` machinery should proceed as before.

PR #3925, opened 2018-09-28
Refactor VM prefix to template functions

CecileRobertMichon:
**What this PR does / why we need it**: instead of hardcoding VM name prefixes in several places, centralize in one common function. 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesagentvars.t
+++ b/parts/k8s/kubernetesagentvars.t
@@ -4,6 +4,9 @@
 {{end}}
     "{{.Name}}Count": "[parameters('{{.Name}}Count')]",
     "{{.Name}}VMNamePrefix": "{{GetAgentVMPrefix .}}",
+{{if .IsWindows}}
+    "winResourceNamePrefix" : "[substring(parameters('nameSuffix'), 0, 5)]",
+{{end}}
 {{if .IsAvailabilitySets}}
     "{{.Name}}Offset": "[parameters('{{.Name}}Offset')]",
     "{{.Name}}AvailabilitySet": "[concat('{{.Name}}-availabilitySet-', parameters('nameSuffix'))]",
--- a/parts/k8s/kubernetesagentvars.t
+++ b/parts/k8s/kubernetesagentvars.t
@@ -3,16 +3,15 @@
     "{{.Name}}StorageAccountsCount": "[add(div(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')), mod(add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),2), add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),1)))]",
 {{end}}
     "{{.Name}}Count": "[parameters('{{.Name}}Count')]",
+    "{{.Name}}VMNamePrefix": "{{GetAgentVMPrefix .}}",
 {{if .IsAvailabilitySets}}
     "{{.Name}}Offset": "[parameters('{{.Name}}Offset')]",
     "{{.Name}}AvailabilitySet": "[concat('{{.Name}}-availabilitySet-', parameters('nameSuffix'))]",
-{{end}}
-    "{{.Name}}VMNamePrefix": "{{GetAgentVMPrefix variables('{{.Name}}Index')}}",
+{{else}}
     {{if .IsLowPriorityScaleSet}}
     "{{.Name}}ScaleSetPriority": "[parameters('{{.Name}}ScaleSetPriority')]",
     "{{.Name}}ScaleSetEvictionPolicy": "[parameters('{{.Name}}ScaleSetEvictionPolicy')]",
     {{end}}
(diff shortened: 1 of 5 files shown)


pkg/api/types_test.go
```diff
@@ -2043,3 +2043,106 @@ func TestProperties_GetClusterMetadata(t *testing.T) {
...
+				},
+			},
+			expectedVMPrefix: "44196k8s900",
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
```
tariq1890: Parallel! <3

pkg/api/types.go
```diff
@@ -748,30 +748,56 @@ func (p *Properties) K8sOrchestratorName() string {
 	return ""
 }
 
-func (p *Properties) getAgentVMPrefix() string {
-	return p.K8sOrchestratorName() + "-agentpool-" + p.GetClusterID() + "-"
+// GetAgentPoolIndex returns the prefix of agentpool VMs given the pool index1
```
tariq1890: This needs to return -1 if not found.

CecileRobertMichon: good point, fixed

pkg/api/types.go
```diff
@@ -748,30 +748,56 @@ func (p *Properties) K8sOrchestratorName() string {
...
+func (p *Properties) GetAgentPoolIndex(a *AgentPoolProfile) int {
```
CecileRobertMichon: //TODO: add unit test for this new function

tariq1890: I think a better approach would be to just pass the name instead of the entire struct pointer. Lets pass in .Name from the template file. Also, you need to fix the godoc of this method. Personally, I would just make this a private method and not have a godoc string. But If you see a utility for this as a public method, let's make sure we fix the godoc.

tariq1890: So based on that decision, I would rename it either getAgentPoolIndexByName/GetAgentPoolIndexByName.

pkg/api/types.go
```diff
@@ -748,30 +748,56 @@ func (p *Properties) K8sOrchestratorName() string {
...
+	for i, profile := range p.AgentPoolProfiles {
+		if profile.Name == a.Name {
+			index = i
+			break
+		}
+	}
+	return index
+}
+
+// GetAgentVMPrefix returns the prefix of agentpool VMs given the pool index
```
tariq1890: Based on my previous comment, Let the godoc be " GetAgentVMPrefix returns the prefix of agentpool VMs given the agent pool profile name"

parts/k8s/kubernetesagentvars.t
```diff
@@ -3,23 +3,15 @@
     "{{.Name}}StorageAccountsCount": "[add(div(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')), mod(add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),2), add(mod(variables('{{.Name}}Count'), variables('maxVMsPerStorageAccount')),1)))]",
 {{end}}
     "{{.Name}}Count": "[parameters('{{.Name}}Count')]",
+    "{{.Name}}VMNamePrefix": "{{GetAgentVMPrefix .}}",
```
tariq1890: I am guessing this needs to be

`"{{.Name}}VMNamePrefix": "{{GetAgentVMPrefix .Name}}"`

pkg/api/types.go
```diff
@@ -748,30 +748,55 @@ func (p *Properties) K8sOrchestratorName() string {
...
+		if profile.Name == name {
+			index = i
+			break
+		}
+	}
+	return index
+}
+
+// GetAgentVMPrefix returns the VM prefix for an agentpool
+func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile) string {
```
CecileRobertMichon: @tariq1890 the VM prefix depends on more than just the name so we need to pass in the whole agent pool profile. I've fixed the comment to better reflect what the function is doing (the previous one was a leftover from an iteration where I passed in the index)

parts/k8s/kubernetesagentvars.t
CecileRobertMichon: No, the function still takes an agent profile

CecileRobertMichon: `getAgentPoolIndexByName` I've changed to just take in the name though

tariq1890: Ah yes. I see that now. Makes sense.

pkg/api/types.go
tariq1890: Sounds good. Thanks!

tariq1890: Let's address the case of -1 index in this method. Return an empty string maybe ? We can add a test to reflect this as well.

PR #3929, opened 2018-10-01
Implementing complete shellQuote of service principal secrets.

tariq1890:


pkg/acsengine/template_generator.go
```diff
@@ -919,5 +919,8 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		},
 		"quote":      strconv.Quote,
 		"shellQuote": helpers.ShellQuote,
+		"GetServicePrincipalClientSecret": func() string {
```
jackfrancis: This func doesn't appear to handle the `if properties.ServicePrincipalProfile.KeyvaultSecretRef != nil {` case in `pkg/acsengine/params_k8s.go`.

tariq1890: I checked the code and the KeyVaultSecretRef part of SPs seems to be used only in the e2e and they don't rely on these template helper methods. So any workflow that involves KeyVault will be unaffected by this change from what I've seen.

parts/k8s/kuberneteswinagentresourcesvmas.t
```diff
@@ -276,7 +276,7 @@
...
+          "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('kubernetesAPIServerIP'),' -KubeDnsServiceIp ',parameters('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -AgentKey ',parameters('clientPrivateKey'),' -AADClientId ',variables('servicePrincipalClientId'),' -AADClientSecret ',{{GetServicePrincipalClientSecret | windowsEscape}},variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]"
```
PatrickLang: `variables('singleQuote')` in ` ',{{GetServicePrincipalClientSecret | windowsEscape}},variables('singleQuote'), '` is either extraneous or unbalanced

parts/k8s/kuberneteswinagentresourcesvmss.t
```diff
@@ -115,7 +115,7 @@
...
+                    "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('kubernetesAPIServerIP'),' -KubeDnsServiceIp ',parameters('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -AgentKey ',parameters('clientPrivateKey'),' -AADClientId ',variables('servicePrincipalClientId'),' -AADClientSecret ',{{GetServicePrincipalClientSecret | windowsEscape}},variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]"
```
PatrickLang: same issue as in parts/k8s/kuberneteswinagentresourcesvmas.t

pkg/helpers/helpers.go
```diff
@@ -164,6 +164,11 @@ func ShellQuote(s string) string {
 	return `'` + strings.Replace(s, `'`, `'\''`, -1) + `'`
 }
 
+// WindowsEscape safely escapes strings with single quotes which need to be used as Powershell arguments
+func WindowsEscape(s string) string {
+	return strings.Replace(s, `'`, `''`, -1)
```
PatrickLang: I don't think Windows actually needs this escaped if I'm reading it correctly. However, arm templates can't handle a single quote. You could surround the whole string in double quotes and PowerShell will handle it ok.

PatrickLang: after discussion, I think this will be cleanest

`return "@\" + s + "\"@"`

PatrickLang: see https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-6#here-strings

PatrickLang:
Ok - I dug deeper into this, and there's a fundamental problem with how this was done on Windows.

Putting arguments into the ARM template on the commandToExecute should only be done if that value is in an ARM parameter. That implies that it's something that will be different for each node. Putting stuff here is really hard because it needs to be triple escaped. Once because it's in an ARM template, once because it's on a Windows cmd.exe invocation, and a third time because it's being passed to powershell.exe. Some of the escapes needed in PowerShell.exe break if they're passed from cmd.exe.

If it's something consistent for all nodes - it should be written into kuberneteswindowssetup.ps1. There you only need double escaping:
level 1 - arm. since the whole customData is encased in single quotes, you can treat everything inside it as a literal. 
level 2 - PowerShell. Use `@"literal"@` and it will work without any extra escapes needed for single or double quotes

PatrickLang:
alright - another option. If this may need to change per-machine later, we can just use a better encoding to base64 and unencode it in the powershell script itself

PatrickLang:
Can you try this instead for Windows? It will be less painful than the triple escape https://github.com/PatrickLang/acs-engine/commit/93edde75e5838f307bc2688628f8a088b2fe8144

CecileRobertMichon:
@tariq1890 what is the status of this PR? Can you please make sure it's transferred to aks-engine if still needed?

PR #3948, opened 2018-10-03
Move utility methods to the helper package

tariq1890:


pkg/acsengine/defaults.go
```diff
@@ -705,98 +705,98 @@ func setHostedMasterProfileDefaults(a *api.Properties) {
...
 	if provided["ca"] {
-		caPair = &PkiKeyCertPair{CertificatePem: a.CertificateProfile.CaCertificate, PrivateKeyPem: a.CertificateProfile.CaPrivateKey}
+		caPair = &helpers.PkiKeyCertPair{CertificatePem: p.CertificateProfile.CaCertificate, PrivateKeyPem: p.CertificateProfile.CaPrivateKey}
 	} else {
-		caCertificate, caPrivateKey, err := createCertificate("ca", nil, nil, false, false, nil, nil, nil)
+		var err error
+		caPair, err = helpers.CreatePkiKeyCertPair("ca")
 		if err != nil {
-			return false, err
+			return false, nil
```
mboersma: Shouldn't we continue to return the `err` here?

pkg/helpers/pki_test.go
```diff
@@ -103,10 +102,59 @@ func TestSubjectAltNameInCert(t *testing.T) {
 
 	// Test SAN is not empty
 	SubjectAltNames := []string{"santest.mydomain.com", "santest2.testdomain.net"}
-	masterExtraFQDNs := append(formatAzureProdFQDNs(dnsPrefix), SubjectAltNames...)
+	formattedDNSPrefixes := []string{
```
mboersma: It's too bad this data structure is a copy of the one in pkg/acsengine/engine_test.go, but since they're in different packages and it's just test code, I guess that's ok.

pkg/acsengine/defaults.go
tariq1890: Thanks for this.

tariq1890:
@mboersma Fixed :)

PR #3951, opened 2018-10-04
AKS distro is for Kubernetes only

jackfrancis:
**What this PR does / why we need it**: Set default distro value to AKS (for VHD delivery) only if Kubernetes

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
AKS distro is for Kubernetes only
```


pkg/acsengine/defaults.go
```diff
@@ -492,7 +492,7 @@ func setExtensionDefaults(a *api.Properties) {
 
 func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
 	// don't default Distro for OpenShift
```
tariq1890: Cam we update the comments?

pkg/acsengine/defaults.go
```diff
@@ -647,7 +647,7 @@ func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
 		}
 
 		// don't default Distro for OpenShift
-		if !a.OrchestratorProfile.IsOpenShift() {
+		if a.OrchestratorProfile.IsKubernetes() {
```
tariq1890: comments here too?

pkg/acsengine/defaults.go
```diff
@@ -491,10 +491,11 @@ func setExtensionDefaults(a *api.Properties) {
 }
 
 func setMasterProfileDefaults(a *api.Properties, isUpgrade bool) {
-	// don't default Distro for OpenShift
-	if !a.OrchestratorProfile.IsOpenShift() {
-		if a.MasterProfile.Distro == "" {
+	if a.MasterProfile.Distro == "" {
+		if a.OrchestratorProfile.IsKubernetes() {
 			a.MasterProfile.Distro = api.AKS
+		} else {
```
CecileRobertMichon: we don't want to set the distro value for openshift

pkg/acsengine/defaults.go
```diff
@@ -646,14 +647,15 @@ func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabled = helpers.PointerToBool(!isUpgrade && !isScale && helpers.AcceleratedNetworkingSupported(profile.VMSize))
 		}
 
-		// don't default Distro for OpenShift
-		if !a.OrchestratorProfile.IsOpenShift() {
-			if profile.Distro == "" {
+		if profile.Distro == "" {
```
tariq1890:
```
if profile.Distro == "" {
			if a.OrchestratorProfile.IsKubernetes() && profile.OSDiskSizeGB >= api.VHDDiskSizeAKS {
				profile.Distro = api.AKS
			} else {
				profile.Distro = api.Ubuntu
			}
		}
```

tariq1890: I've optimized it to that. Does it look okay?

jackfrancis: I have a slight preference for the current implementation as it emphasizes the k8s-specific conditions (which could grow over time based on more k8s-specific distros.

CecileRobertMichon: The condition `profile.OSDiskSizeGB >= api.VHDDiskSizeAKS` is not equivalent to `profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < api.VHDDiskSizeAKS`

tariq1890: I see. profile.OSDiskSizeGB=0 represents a nil state I am assuming?

CecileRobertMichon: Also again we don't want to set the distro for openshift

jackfrancis: Also what @CecileRobertMichon said. It's bit clumsy, but we want to include the zero value (0) when evaluating what to do. zero value means AKS distro, in this case.

pkg/acsengine/defaults.go
```diff
@@ -646,14 +647,15 @@ func setAgentProfileDefaults(a *api.Properties, isUpgrade, isScale bool) {
...
+			if a.OrchestratorProfile.IsKubernetes() {
 				if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < api.VHDDiskSizeAKS {
 					profile.Distro = api.Ubuntu
 				} else {
 					profile.Distro = api.AKS
 				}
+			} else if !a.OrchestratorProfile.IsOpenShift() {
```
jackfrancis: @CecileRobertMichon we're covering the OpenShift condition here

CecileRobertMichon: I know, my comment was regarding Tariq's proposed implementation

tariq1890: Cool then. Thanks for the clarifications

PR #3750, opened 2018-08-27
Reduce the Kubernetes version support matrix

CecileRobertMichon:
**What this PR does / why we need it**: Deprecates Kubernetes versions except for the two latest patches of each minor release while maintaining support for scale and upgrade from these versions. Exception is version 1.6.9 which we need to keep support for because of API v20160930.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [x] unit tests
- [x] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
```

PR #3952, opened 2018-10-04
Remove Kubernetes 1.7.x from supported versions 

mboersma:

Following on to #3750, this further reduces our Kubernetes support matrix by removing 1.7 as a supported release. (Note that 1.6.9 remains supported in acs-engine for historical reasons.)

**If applicable**:
- [ ] documentation
- [X] unit tests

**Release note**:

```release-note
Remove Kubernetes 1.7.x from supported versions
```


mboersma:
/hold

This works, but I think there are further opportunities here to remove code. I'd still like to take another pass at that and to test the vendoring of this into AKS.

mboersma:
Thanks for fixing the mocks in a more durable way @jackfrancis. This looks good to me.

PR #3953, opened 2018-10-04
Migrating cloud spec config to api package

tariq1890:


jackfrancis:
Can you explain more about why moving this to `api` is an improvement?

tariq1890:
Sure. So my argument is that that the set Properties defaults logic be contained in the same package as that of Properties struct itself. The PR here is done to enable that.

jackfrancis:
Got it.

pkg/api/types.go
```diff
@@ -1281,3 +1281,36 @@ func (k *KubernetesConfig) SetCloudProviderRateLimitDefaults() {
...
+		return AzureUSGovernmentCloud
+	default:
+		return AzureCloudSpec
+	}
+}
+
+// getCloudTargetEnv determines and returns whether the region is a sovereign cloud which
+// have their own data compliance regulations (China/Germany/USGov) or standard
+//  Azure public cloud
+func getCloudTargetEnv(location string) string {
```
jackfrancis: Should this be a method of `ContainerService`?

tariq1890: Oh yes. That would be more consistent

tariq1890: Fixed

jackfrancis:
Looks like we need to delete the original `getCloudTargetEnv` function definition under the `acsengine` package.

tariq1890:
Yes. Thanks for pointing that out. I am working on it right now.

PR #3959, opened 2018-10-05
Don't require VM tags

mboersma:

Removes our absolute reliance on VM tags in the `upgrade` and `scale` commands.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Don't require VM tags when upgrading or scaling
```

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -24,7 +24,6 @@ import (
 	"github.com/Azure/acs-engine/pkg/operations"
 	"github.com/leonelquinteros/gotext"
 	"github.com/pkg/errors"
-
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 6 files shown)


cmd/scale.go
```diff
@@ -451,6 +446,19 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	return f.SaveFile(sc.deploymentDirectory, "apimodel.json", b)
 }
 
+func vmInAgentPool(vmName string, sc *scaleCmd) bool {
+	if !strings.Contains(vmName, sc.agentPoolToScale) {
+		log.Warnf("VM %s is missing prefix \"%s\"", vmName, sc.agentPoolToScale)
+		return false
+	}
+	// Use strings.Contains for the nameSuffix as the Windows Agent Pools use only a substring of the first 5 characters of the entire nameSuffix
+	if !strings.Contains(vmName, sc.nameSuffix) {
```
CecileRobertMichon: shouldn't this only check if the vmName contains the substring of the name suffix instead of the entire name suffix?

PR #3968, opened 2018-10-05
Enable upgrade to next supported Kubernetes version

CecileRobertMichon:
**What this PR does / why we need it**: Remove constraint of only being able to upgrade to the next minor version if the next minor version is deprecated. Instead, allow upgrading to the next minor version OR the next available minor version. So if I have 1.7 cluster and 1.7 and 1.8 are deprecated, I can upgrade to 1.9 but not 1.10+. If I have a 1.9 cluster I can upgrade to 1.9 or 1.10.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/orchestrators_test.go
+++ b/pkg/api/orchestrators_test.go
@@ -56,7 +56,8 @@ func TestOrchestratorUpgradeInfo(t *testing.T) {
 			OrchestratorType:    Kubernetes,
 			OrchestratorVersion: deployedVersion,
 		}
-		v := getKubernetesAvailableUpgradeVersions(deployedVersion, common.GetAllSupportedKubernetesVersions(false, false))
+		v, e := getKubernetesAvailableUpgradeVersions(deployedVersion, common.GetAllSupportedKubernetesVersions(false, false))
+		Expect(e).To(BeNil())
 		orch, e := GetOrchestratorVersionProfile(csOrch, false)
 		Expect(e).To(BeNil())
 		Expect(len(orch.Upgrades)).To(Equal(len(v)))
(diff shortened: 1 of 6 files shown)


CecileRobertMichon:
https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/176/console

pkg/api/orchestrators_test.go
```diff
@@ -298,3 +238,44 @@ func TestDockerceInfoInfo(t *testing.T) {
 	_, e := dockerceInfo(csOrch, false)
 	Expect(e).To(BeNil())
 }
+
+func TestGetKubernetesAvailableUpgradeVersions(t *testing.T) {
```
CecileRobertMichon: This is one of those "the tests are just as important if not more important than the code" PRs

CecileRobertMichon:
/hold 

TODO: retest upgrade and back compat

CecileRobertMichon:
https://jenkins.azure-containers.io/view/acs-engine%20ad-hoc/job/k8s-upgrade/177/console

CecileRobertMichon:
/unhold

back compat upgrade passed

tariq1890:
@CecileRobertMichon I think we should leverage the capabilities of semver library. I have this link for reference - https://github.com/Masterminds/semver/blob/master/README.md#sorting-semantic-versions.

My suggestion is that we use sort and then use the sorted array to get the Max and Min versions. This will simplify the logic by a great deal and do away with the ifs and for

tariq1890:
Based on the `pre-release` bool, we can choose to include the pre-release versions when in the array to be sorted.

PR #3974, opened 2018-10-07
Refactor: Moving set defaults logic from package acsengine to package api

tariq1890:
This PR moves the entire set defaults code to the api package. This is done so that the SetDefaults method may be moved out of the templateGenerate method safely.


pkg/acsengine/template_generator.go
```diff
@@ -52,25 +52,22 @@ func (t *TemplateGenerator) GenerateTemplate(containerService *api.ContainerServ
 	defer func() {
 		properties.OrchestratorProfile.OrchestratorVersion = orchVersion
 	}()
-	if certsGenerated, err = setPropertiesDefaults(containerService, isUpgrade, isScale); err != nil {
```
jackfrancis: So to be clear, we've decomposed defaults enforcement from the main `GenerateTemplate` func? So going forward `GenerateTemplate` will no assume a fully rationalized api model?

tariq1890: Yes. The PR goes with that assumption.

pkg/helpers/helpers_test.go
```diff
@@ -73,7 +73,7 @@ func TestPointerToBool(t *testing.T) {
 		t.Fatalf("expected PointerToBool(true) to return *true, instead returned %#v", ret)
```
jackfrancis: nit: these print statements should print %d, boolVar instead of true/false strings

tariq1890: Sure

pkg/acsengine/const.go
```diff
@@ -1,44 +1,8 @@
...
-	DefaultDCOSMasterSubnet = "192.168.255.0/24"
-	// DefaultDCOSFirstConsecutiveStaticIP  specifies the static IP address on master 0 for a DCOS cluster
-	DefaultDCOSFirstConsecutiveStaticIP = "192.168.255.5"
-	// DefaultDCOSBootstrapStaticIP specifies the static IP address on bootstrap for a DCOS cluster
-	DefaultDCOSBootstrapStaticIP = "192.168.255.240"
-	// DefaultKubernetesMasterSubnet specifies the default subnet for masters and agents.
-	// Except when master VMSS is used, this specifies the default subnet for masters.
-	DefaultKubernetesMasterSubnet = "10.240.0.0/16"
-	// DefaultKubernetesClusterSubnet specifies the default subnet for pods.
-	DefaultKubernetesClusterSubnet = "10.244.0.0/16"
```
jackfrancis: Did this const get moved or just deleted?

tariq1890: These have been moved to package api.

pkg/acsengine/const.go
```diff
@@ -1,44 +1,8 @@
...
-	// DefaultDockerBridgeSubnet specifies the default subnet for the docker bridge network for masters and agents.
-	DefaultDockerBridgeSubnet = "172.17.0.1/16"
```
jackfrancis: ditto

tariq1890: These constants should have been moved to the right place(in api/const/go). Thanks for pointing it out. Will make the required changes.

pkg/acsengine/const.go
```diff
@@ -128,28 +64,8 @@ const (
 	DefaultKeyVaultFlexVolumeAddonName = "keyvault-flexvolume"
 	// DefaultELBSVCAddonName is the name of the elb service addon deployment
 	DefaultELBSVCAddonName = "elb-svc"
-	// DefaultKubernetesDNSServiceIP specifies the IP address that kube-dns
-	// listens on by default. must by in the default Service CIDR range.
-	DefaultKubernetesDNSServiceIP = "10.0.0.10"
```
jackfrancis: ditto

pkg/acsengine/const.go
```diff
@@ -128,28 +64,8 @@ const (
...
-	// DefaultKubernetesServiceCIDR specifies the IP subnet that kubernetes will
-	// create Service IPs within.
-	DefaultKubernetesServiceCIDR = "10.0.0.0/16"
```
jackfrancis: ditto

tariq1890: acknowledged

pkg/acsengine/const.go
```diff
@@ -170,31 +86,10 @@ const (
...
-	DefaultJumpboxUsername = "azureuser"
-	// DefaultKubeletPodMaxPIDs specifies the default max pid authorized by pods
-	DefaultKubeletPodMaxPIDs = 100
-	// DefaultKubernetesAgentSubnetVMSS specifies the default subnet for agents when master is VMSS
-	DefaultKubernetesAgentSubnetVMSS = "10.248.0.0/13"
 	// DefaultUserAssignedID specifies the default name for the user assigned identity
 	DefaultUserAssignedID = "acsenginetestid"
 )
 
-const (
```
jackfrancis: ditto these consts

tariq1890:
@jackfrancis review comments addressed

pkg/acsengine/const.go
```diff
@@ -170,31 +86,10 @@ const (
...
-	// DCOSMaster represents the master node type
```
jackfrancis: So the idea behind deleting these consts and not moving them elsewhere is that they are unused?

Should we delete the `DCOSNodeType` type struct as well?

pkg/acsengine/params_k8s.go
```diff
@@ -19,36 +19,36 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 
 	if orchestratorProfile.IsKubernetes() ||
 		orchestratorProfile.IsOpenShift() {
-		k8sVersion := orchestratorProfile.OrchestratorVersion
+		kubeConfig := api.KubeConfigs[orchestratorProfile.OrchestratorVersion]
```
jackfrancis: This is a good optimization that avoids unnecessary lookups. 👍 

While we're here, let's do this, if you don't mind :)

1) rename `KubeConfigs` to `K8sComponentsByVersionDict` (or something)
2) generate the `K8sComponentsByVersionDict` map only once, instead of calling something like `getKubeConfigs()` every time it is referenced

Does that make sense?

pkg/acsengine/params_k8s.go
```diff
@@ -19,36 +19,36 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 
 	if orchestratorProfile.IsKubernetes() ||
 		orchestratorProfile.IsOpenShift() {
-		k8sVersion := orchestratorProfile.OrchestratorVersion
+		kubeConfig := api.K8sComponentsByVersionMap[orchestratorProfile.OrchestratorVersion]
```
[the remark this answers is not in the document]
tariq1890: ah yes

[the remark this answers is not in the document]
tariq1890: Done

PR #3978, opened 2018-10-08
Using local rand object to generate cluster ids

tariq1890:
The current rand object that we use is globally instantiated and it has shown to cause inconsistent outputs in race conditions. We will now use a locally instantiated rand object instead and perform rand operations with mute locks.

This PR should hopefully eliminate the current flakiness observed in our unit test suite.


CecileRobertMichon:
ignore my PR description comment, race condition :)

PR #3982, opened 2018-10-09
E2E: actually fail when no InternalIP, ssh master tweaks, delete retries

jackfrancis:
**What this PR does / why we need it**: Prior usage was not failing when a node had no InternalIP. Additionally, simplify "the ssh to master" implementations, and retry all deletes.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
E2E: actually fail when no InternalIP
```

### DIFF
--- a/test/e2e/kubernetes/deployment/deployment.go
+++ b/test/e2e/kubernetes/deployment/deployment.go
@@ -163,11 +163,7 @@ func (d *Deployment) Delete(retries int) error {
 		}
 	}
 
-	if kubectlError != nil {
-		return kubectlError
-	}
-
-	return nil
+	return kubectlError
 }
 
 // Expose will create a load balancer and expose the deployment on a given port
--- a/test/e2e/kubernetes/deployment/deployment.go
+++ b/test/e2e/kubernetes/deployment/deployment.go
@@ -134,22 +134,39 @@ func Get(name, namespace string) (*Deployment, error) {
 }
 
 // Delete will delete a deployment in a given namespace
-func (d *Deployment) Delete() error {
-	cmd := exec.Command("kubectl", "delete", "deploy", "-n", d.Metadata.Namespace, d.Metadata.Name)
-	out, err := util.RunAndLogCommand(cmd)
-	if err != nil {
-		log.Printf("Error while trying to delete deployment %s in namespace %s:%s\n", d.Metadata.Namespace, d.Metadata.Name, string(out))
-		return err
+func (d *Deployment) Delete(retries int) error {
+	var kubectlOutput []byte
+	var kubectlError error
+	for i := 0; i < retries; i++ {
+		cmd := exec.Command("kubectl", "delete", "deploy", "-n", d.Metadata.Namespace, d.Metadata.Name)
+		kubectlOutput, kubectlError = util.RunAndLogCommand(cmd)
+		if kubectlError != nil {
+			log.Printf("Error while trying to delete deployment %s in namespace %s:%s\n", d.Metadata.Namespace, d.Metadata.Name, string(kubectlOutput))
+			continue
+		}
+		break
 	}
-	// Delete any associated HPAs
+
+	if kubectlError != nil {
+		return kubectlError
+	}
+
 	if d.Metadata.HasHPA {
-		cmd := exec.Command("kubectl", "delete", "hpa", "-n", d.Metadata.Namespace, d.Metadata.Name)
-		out, err := util.RunAndLogCommand(cmd)
-		if err != nil {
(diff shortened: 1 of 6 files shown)


test/e2e/kubernetes/deployment/deployment.go
```diff
@@ -134,22 +134,39 @@ func Get(name, namespace string) (*Deployment, error) {
...
+			}
+			break
 		}
 	}
+
+	if kubectlError != nil {
+		return kubectlError
+	}
+
 	return nil
```
CecileRobertMichon: can we just return kubectlError?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -389,16 +347,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 							address := node.Status.GetAddressByType("InternalIP")
 							if address == nil {
 								log.Printf("One of our nodes does not have an InternalIP value!: %s\n", node.Metadata.Name)
+								Fail("All nodes should have an InternalIP address")
 							}
-							Expect(address).NotTo(Equal(nil))
```
CecileRobertMichon: was this not failing when address was nil?

test/e2e/kubernetes/deployment/deployment.go
jackfrancis: yep :)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -30,14 +30,16 @@ import (
 )
 
 const (
-	WorkloadDir = "workloads"
-	PolicyDir   = "workloads/policies"
-	SSHPort     = "50001"
+	WorkloadDir           = "workloads"
+	PolicyDir             = "workloads/policies"
+	deleteResourceRetries = 10
```
CecileRobertMichon: if we are going to use a const for number of retries and it's the same for every call, why pass it as a parameter to the function and not just use it in the delete function directly?

test/e2e/kubernetes/persistentvolumeclaims/persistentvolumeclaims.go
```diff
@@ -85,14 +85,23 @@ func Describe(pvcName, namespace string) error {
...
+			continue
+		}
+		break
+	}
+
+	if kubectlError != nil {
+		return kubectlError
 	}
+
 	return nil
```
CecileRobertMichon: same comment about returning kubectlError here

test/e2e/kubernetes/service/service.go
```diff
@@ -74,14 +74,23 @@ func Get(name, namespace string) (*Service, error) {
...
+			continue
+		}
+		break
 	}
+
+	if kubectlError != nil {
+		return kubectlError
+	}
+
 	return nil
```
CecileRobertMichon: ditto

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Good question. Just an iterative placeholder until we have stronger opinions about particular use cases of delete that should be retried fewer or more times (or until we make it user-configurable)

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: I think the way ginkgo implements the equality check doesn't work in the way it was being used.

test/e2e/kubernetes/kubernetes_test.go
tariq1890: Thoughts on having a generic returnable function that takes in function/cmd parameters?

test/e2e/kubernetes/kubernetes_test.go
CecileRobertMichon: looking at gingko docs it looks like `Fail()`  won't run the `AfterSuite()` stuff so maybe not ideal

http://onsi.github.io/ginkgo/#getting-started-writing-your-first-test

CecileRobertMichon: could we maybe add an `else{}` to protect against panic instead and let the assertion do its thing?

jackfrancis: Good catch. What do you think about this `getAddressError` implementation?

CecileRobertMichon: I don't think that will help. The succeeding line will still be run without an else statement to protect it. I think `Expect(address).NotTo(Equal(nil))` <==> `Expect(getAddressError).NotTo(HaveOccurred())`

CecileRobertMichon: Should be `Expect(address).NotTo(BeNil())`

Issue #3851, opened 2018-09-14
Cleanup Windows installations


PatrickLang:
This is list to track what needs to be done to make Windows work better with upstream builds.

First Step - support upstream binaries
- [x] Add new parameter `$global:KubeBinariesReleaseUrl` to parts\k8s\kuberneteswindowssetup.ps1 to download [node binaries] #4008 (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#node-binaries-1) and [client binaries](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#client-binaries-1). Install those instead of binaries from `$global:KubeBinariesSASURL` 
- [x] Same for Azure CNI #4007 
- [ ] Add ContainerD support #4006 
- [ ] Remove support for 1.9 in k8s_versions.go so we can delete NSSM
- [ ] Remove dependency on NSSM. Use new flags added in https://github.com/kubernetes/kubernetes/pull/60144

Second Step - don't break kubetest Azure provider
- [ ] Update https://github.com/kubernetes/test-infra/pull/7625 or issue new PR if it's not merged. Remove the step to build the zip, and instead set the path to Kubernetes binaries

Third Step - clean up

- [ ] Delete code in parts\k8s\kuberneteswindowssetup.ps1 related to NSSM and zip file
- [ ] Delete acs-engine\scripts\build-windows-k8s.sh from this tree

--- 

**History on acs-engine\scripts\build-windows-k8s.sh**

This script was written before Kubernetes had upstream builds for Windows (kubelet.exe, kube-proxy.exe, kubectl.exe). It was designed to pull sources from git, apply some cherry-picks, build the binaries, download NSSM, and package everything up into a ZIP.

PR #3994, opened 2018-10-11
Split Windows setup scripts, prepare for cleanup and multiple CRI

PatrickLang:
The Windows setup scripts were difficult to manage for a few reasons:
- Although the files looked like PowerShell scripts, they were subject to some limitations:
  - Golang was used to double-escape some characters so it could be put into ARM customData, and some like '`' couldn't be used
- Testing was difficult
  - The single script was interpreted as golang templates, which means you can't copy a file from the repo to a machine to test it
  - The scripts were not idempotent

This changes things so that only `kuberneteswindowssetup.ps1` is used as a golang template. All changes that are done by acs-engine are in that single file. They are validated and passed as parameters to powershell functions in separate files now. To avoid the escaping problems - all of those files are enclosed in a ZIP, base64 encoded, and extracted at runtime.

If you need to test changes, you can now:
- go change or replace c:\azuredata\k8s\*.ps1
- re-run c:\azuredata\customdatasetupscript.ps1 with the same parameters

This is the first PR towards https://github.com/Azure/acs-engine/issues/3851


PatrickLang:
/hold
I'm still doing more testing on this, but want to take advantage of automation and start reviews.

PatrickLang:
scheduled extra tests using kubenet/wincni. will remove hold after those pass

PatrickLang:
/remove hold
kubenet tests passed 8/10. Since those are low priority and not blocking, I think that's good enough.

pkg/acsengine/engine.go
```diff
@@ -650,8 +650,8 @@ func getSecurityRules(ports []int) string {
 	return buf.String()
 }
 
-// getSingleLineForTemplate returns the file as a single line for embedding in an arm template
-func (t *TemplateGenerator) getSingleLineForTemplate(textFilename string, cs *api.ContainerService, profile interface{}) (string, error) {
+// getSingleLineForTemplate returns the file as a single line
```
jackfrancis: method comment copy/paste cruft (should be `// getSingleLine ...`)

jackfrancis:
we'll want to fix the comment nit to avoid lint failures in the future, otherwise lgtm

Issue #4029, opened 2018-10-15
Feature Request: Support Docker CE 18.03


MarcPow:
**Is this a request for help?**:

No
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

Feature Request
---

**What version of acs-engine?**:

0.23.1
---

    "dockerEngineVersion": {
      "allowedValues": [
        "17.05.*",
        "17.04.*",
        "17.03.*",
        "1.13.*",
        "1.12.*",
        "1.11.*"
      ],
      "metadata": {
        "description": "The docker engine version to install."
      },
      "type": "string"
    },

CecileRobertMichon:
Sadly, because of unresolved license limitations we are currently unable to move to new Docker CE versions (see https://github.com/Azure/AKS/issues/63 for context). There has been a push to move to Moby instead, you can track the effort at https://github.com/Azure/acs-engine/pull/3896

PatrickLang:
Docker EE-basic 18.03 is available for all licensed Windows Server customers. I'm tracking this upgrade in #3852 

Linux users would need their own Docker EE license.

PR #4030, opened 2018-10-15
Optionally support Docker 18.03 and 18.06

MarcPow:

Allows users to optionally select Docker 18.03 or 18.06 

https://github.com/Azure/acs-engine/issues/4029


MarcPow:
Ah - I may need to abandon this.  Per @CecileRobertMichon in the tracking issue, there are some licensing restrictions that prevent me from making this change.

jackfrancis:
Can't do this due to licensing issues. moby support is in-progress

Issue #4027, opened 2018-10-15
Can't provision Hybrid cluster issues with cse provisioning


sylus:
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): Bug

---

**What version of acs-engine?**: v0.23.1

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**

Kubernetes 1.11.3

**What happened**:

Running the following:

```
az group deployment create --name "k8s-acs-stc-mgmt" `
>>                            --resource-group "k8s-management-rg" `
>>                            --template-file "./_output/k8s-acs-stc-mgmt/azuredeploy.json" `
>>                            --parameters "./_output/k8s-acs-stc-mgmt/azuredeploy.parameters.json"
```

Where the initial cluster spec is:

```sh
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorRelease": "1.11",
      "kubernetesConfig": {
        "kubernetesImageBase": "k8s-gcrio.azureedge.net/",
        "useManagedIdentity": false,
        "privateCluster": {
          "enabled": true
        },
        "addons": [
          {
            "name": "tiller",
            "enabled" : true
          },
          {
            "name": "kubernetes-dashboard",
            "enabled" : true
          },
          {
            "name": "rescheduler",
            "enabled" : true
          }
        ]
      }
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "k8s-acs-stc-mgmt",
      "vmSize": "Standard_D4s_v3",
      "OSDiskSizeGB": 200,
      "vnetSubnetId": "/subscriptions/XXX/resourceGroups/network-management-rg/providers/Microsoft.Network/virtualNetworks/k8s-mgmt-vnet/subnets/k8sMasterSubnet",
      "firstConsecutiveStaticIP": "172.20.58.4",
      "vnetCidr": "172.20.58.0/23"
    },
    "agentPoolProfiles": [
      {
        "name": "linuxpool1",
        "count": 3,
        "customNodeLabels": {
          "os": "linux"
        },
        "vmSize": "Standard_D8s_v3",
        "OSDiskSizeGB": 200,
        "storageProfile" : "ManagedDisks",
        "availabilityProfile": "AvailabilitySet",
[...]

PR #4038, opened 2018-10-16
Fix regression in Hybrid clusters using custom Vnets

tariq1890:
Fixes issue #4027


jackfrancis:
Added param-->var replacement for vmss

Approach makes sense @tariq1890!

parts/k8s/kubernetesmastervars.t
```diff
@@ -155,25 +155,29 @@
...
     "virtualNetworkName": "[split(parameters('masterVnetSubnetID'), '/')[variables('vnetNameResourceSegmentIndex')]]",
     "virtualNetworkResourceGroupName": "[split(parameters('masterVnetSubnetID'), '/')[variables('vnetResourceGroupNameResourceSegmentIndex')]]",
+    "masterSubnet": "",
   {{else}}
     "subnetName": "[concat(parameters('orchestratorName'), '-subnet')]",
     "vnetID": "[resourceId('Microsoft.Network/virtualNetworks',variables('virtualNetworkName'))]",
     "vnetSubnetID": "[concat(variables('vnetID'),'/subnets/',variables('subnetName'))]",
     "virtualNetworkName": "[concat(parameters('orchestratorName'), '-vnet-', parameters('nameSuffix'))]",
     "virtualNetworkResourceGroupName": "''",
+    "masterSubnet": "[parameters('masterSubnet')]",
```
CecileRobertMichon: Do we need to translate this to a variable? Why not add a default value of `""` for the parameter?

tariq1890: I like the idea. I think it's much better and cleaner. I followed this because of my limited ARM knowledge and variables to params delegation is a pattern that has already being followed. Thanks for the suggestion!

@jackfrancis FYI

PR #4094, opened 2018-10-22
update azure-npm to v1.0.13

saiyan86:

This PR changes the order of Azure-NPM for applying network policies from FIFO to LIFO.
Kubernetes administrators/network operators usually apply a deny-all policy then whitelist certain traffic. Without this PR to achieve this, one would have to whitelist all traffic first then apply the deny-all policy. 

Related PR:https://github.com/Azure/azure-container-networking/pull/258


saiyan86:
@jackfrancis Please review. Thanks!

Issue #4007, opened 2018-10-12
Make Azure CNI URL configurable


PatrickLang:

PatrickLang:
Here's example downloads that we want to make work: 
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#downloads-for-v1121

PR #4097, opened 2018-10-22
enable user-configurable Azure CNI URL

jackfrancis:
**What this PR does / why we need it**: Enables user-configurable Azure CNI URL. Linux implementation is code-sane, but doesn't actually do anything, as that'll require more work integrating w/ VHD. Because Windows is the desired initial use-case, I'm advocating we take this as-is (once we prove Windows functionality) and leave remaining Linux implementation to be done when desired.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
enable user-configurable Azure CNI URL
```


jsturtevant:
fixes: https://github.com/Azure/acs-engine/issues/4007

pkg/api/types.go
```diff
@@ -360,6 +360,8 @@ type KubernetesConfig struct {
 	LoadBalancerSku                  string            `json:"loadBalancerSku,omitempty"`
 	ExcludeMasterFromStandardLB      *bool             `json:"excludeMasterFromStandardLB,omitempty"`
 	AzureCNIVersion                  string            `json:"azureCNIVersion,omitempty"`
+	AzureCNIURLLinux                 string            `json:"azureCNIURLLinux,omitempty"`
```
tariq1890: Can we have a struct (AzureCNIConfig perhaps? )that wraps AzureCNIVersion, AzureCIURLLinux, AzureCNIURLWindows, and AzureCNIPluginURL ? This way we make all the four user configurable and struct can be reused in KubernetesConfig as well as KubernetesSpecConfig structs.

KubernetesConfig is getting really bloated right now :(

jsturtevant:
reference back to epic #3851 for tracking

PR #4103, opened 2018-10-23
Make windows binary url configurable

### LINKED ISSUE #4008: Make Kubernetes node binary URL configurable
Today, testing a Kubernetes release requires doing a new build of ACS-Engine. We need to be able to test fixes from Windows Server, ContainerD / Docker, Kubernetes, and Azure-CNI together to improve Kubernetes conformance test results. If we have to rebuild acs-engine, then we cannot move fast enough.

The epic tracking Windows setup cleanup is #3851, be sure to check this off the list when this item is done.

jsturtevant:
Today, testing a Kubernetes release requires doing a new build of ACS-Engine. We need to be able to test fixes from Windows Server, ContainerD / Docker, Kubernetes, and Azure-CNI together to improve Kubernetes conformance test results. If we have to rebuild acs-engine, then we cannot move fast enough.

The epic tracking Windows setup cleanup is #3851, be sure to check this off the list when this item is done.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #4008 

This is step one in removing the use of package built by `build-windows-k8s.sh`.  There will be follow up PR as defined in #3851.  

cc: @PatrickLang 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Make windows binary url configurable
```


jsturtevant:
@PatrickLang still need to run a few test deployments but could you review to make sure this was along the lines of what your thinking?

jsturtevant:
@jackfrancis @PatrickLang ready for a review

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -139,17 +139,27 @@ try
...
-        Write-Log "download kubelet binaries and unzip"
-        Get-KubeBinaries -KubeBinariesSASURL $global:KubeBinariesSASURL
+        Write-Log "Download kubelet binaries and unzip"
+        Get-KubePackage -KubeBinariesSASURL $global:KubeBinariesPackageSASURL
+
+        # this overwrite the binaries that are download from the custom packge with binaries 
+        # The custom package has a few files that are nessary for future steps (nssm.exe)
+        # this is a temporary work around to get the binaries until we depreciate 
+        # custom package and nssm.exe as defined in #3851.
+        if ($global:WindowsKubeBinariesURL){
```
jackfrancis: So before we were always calling `Get-KubeBinaries` and passing in `KubeBinariesSASURL`. Now we're only calling `Get-KubeBinaries` if `WindowsKubeBinariesURL` is non-empty. Is that the correct interpretation?

Is that change in implementation intentional?

jsturtevant: I renamed the old `Get-KubeBinaries` to `Get-KubePackage` to be more explicit in its functionality and created the new function with the old name.  I did this to make it more descriptive in what each function does.  `Get-KubePackage` will be removed in future commits as described in  #3851.

jackfrancis:
@PatrickLang for final review/merge

docs/design/proposed-template-refactor.md
```diff
@@ -35,13 +35,13 @@ Once the input is validated, the template generator is invoked which will conver
     {{if .HasWindows}}
       "kubeBinariesSASURL": {
         "metadata": {
-          "description": "The download url for kubernetes windows binaries."
+          "description": "The download url for kubernetes windows binaries pacakge that is created by scripts/build-windows-k8s.sh"
```
PatrickLang: nit: `pacakge `

parts/k8s/kubernetesbase.t
```diff
@@ -6,13 +6,13 @@
     {{if .HasWindows}}
       "kubeBinariesSASURL": {
         "metadata": {
-          "description": "The download url for kubernetes windows binaries."
+          "description": "The download url for kubernetes windows binaries pacakge that is created by scripts/build-windows-k8s.sh"
```
PatrickLang: nit: `pacakge `

parts/k8s/kuberneteswindowssetup.ps1
PatrickLang: This approach looks good to me

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -183,6 +183,40 @@ Get-KubeBinaries
...
+Get-KubeBinaries
+{
+    Param(
+        [Parameter(Mandatory=$true)][string]
+        $KubeBinariesURL
+    )
+
+    if ($computerInfo.WindowsVersion -eq "1709") 
+    {        
+        Write-Log "Server version 1709 does not support using kubernetes binaries in tar file."
```
PatrickLang: Why not? It's not a top priority but I think it would still work.

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -183,6 +183,40 @@ Get-KubeBinaries
...
+        $KubeBinariesURL
+    )
+
+    if ($computerInfo.WindowsVersion -eq "1709") 
+    {        
+        Write-Log "Server version 1709 does not support using kubernetes binaries in tar file."    
+        return
+    }
+    
+    $binaryPackage = "c:\k.tar.gz"
```
PatrickLang: I think we should avoid putting more stuff in non-ACL'd and root folders. Can you move this to a temp dir? https://stackoverflow.com/a/34560029

PatrickLang: the other case of this in `Get-KubernetesPackage` can be removed later. It depends on being extracted from `c:\k.zip`

parts/k8s/windowskubeletfunc.ps1
PatrickLang: oh yeah, no tar. disregard

parts/k8s/windowskubeletfunc.ps1
jsturtevant: Made this update.

pkg/acsengine/params.go
```diff
@@ -231,13 +231,19 @@ func getParameters(cs *api.ContainerService, generatorCode string, acsengineVers
...
 			if kubeBinariesSASURL == "" {
 				kubeBinariesSASURL = cloudSpecConfig.KubernetesSpecConfig.KubeBinariesSASURLBase + api.K8sComponentsByVersionMap[k8sVersion]["windowszip"]
 			}
-
 			addValue(parametersMap, "kubeBinariesSASURL", kubeBinariesSASURL)
-			addValue(parametersMap, "windowsPackageSASURLBase", cloudSpecConfig.KubernetesSpecConfig.WindowsPackageSASURLBase)
+
+			// Kubernetes node binaries as packaged by upstream kubernetes
+			// example at https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#node-binaries-1
+			addValue(parametersMap, "windowsKubeBinariesURL", properties.OrchestratorProfile.KubernetesConfig.WindowsNodeBinariesURL)
```
tariq1890: can we write a unit test to assert the presence of this key and its value. We have written similar uni tests in `params_k8s_test.go`

pkg/api/converterfromapi.go
```diff
@@ -735,6 +735,7 @@ func convertKubernetesConfigToVLabs(api *KubernetesConfig, vlabs *vlabs.Kubernet
 	vlabs.CustomCcmImage = api.CustomCcmImage
 	vlabs.UseCloudControllerManager = api.UseCloudControllerManager
 	vlabs.CustomWindowsPackageURL = api.CustomWindowsPackageURL
+	vlabs.WindowsNodeBinariesURL = api.WindowsNodeBinariesURL
```
tariq1890: Let's also update the corresponding unit tests to reflect these new fields.

pkg/api/convertertoapi.go
```diff
@@ -695,6 +695,7 @@ func convertVLabsKubernetesConfig(vlabs *vlabs.KubernetesConfig, api *Kubernetes
 	api.CustomCcmImage = vlabs.CustomCcmImage
 	api.UseCloudControllerManager = vlabs.UseCloudControllerManager
 	api.CustomWindowsPackageURL = vlabs.CustomWindowsPackageURL
+	api.WindowsNodeBinariesURL = vlabs.WindowsNodeBinariesURL
```
tariq1890: Same as previous comment

tariq1890:
@jsturtevant Thanks for the PR. Looks good save for a few comments.

pkg/acsengine/params.go
jsturtevant: The tests in `params_k8s_test.go` call to `params_k8s.go` by calling `assignKubernetesParameters()`. This is is in `params.go` under `getParameters()` which doesn't seem to have any unit tests.  

Should I try creating a test function like `TestAssignKubernetesParameters` in `params_k8s_test.go`? Or should i try to mimic the `TestKubernetesParamsAddons` but for the `kubernetesConfig` object?

tariq1890: @jsturtevant Since this is a k8s-specific param. We should be populating this in `param_k8s.go`.

jsturtevant: In this case there are some kubernetes config settings that are where already set in the `params.go`  and since it directly correlated to the parameter above I would like to keep it next to it (`kubeBinariesSASURL`).  

We could move all these k8 specific params to the `params_k8s.go` but that should probably be done in a separate PR.

tariq1890: Yes. I just realized. Sure. That can be done in a separate PR. As for your previous question, We can go with the former

pkg/api/converterfromapi.go
jsturtevant: There do not appear to be any unit tests for `convertfromapi.go`

pkg/api/convertertoapi.go
jsturtevant: There is not a precedent for testing this particular set of settings.  I gave it a go but it feels a little strange as it's just copying values from one place to another.

I tried two different ways to test.  I reverted one of the tries and went a different route.  Can rebase those commits away unless you perfer the first try.

pkg/acsengine/params.go
jsturtevant: added the tests for the `params.go` file in a similar fashion as `params_k8s_test.go`

PR #4107, opened 2018-10-24
Stop kubelet before update and reboot faster

PatrickLang:

If you're creating a cluster as part of a test pass - the nodes appear to be ready before any patches are installed. This change will stop the kubelet so the node is not "ready", and also reboot faster. Once the reboot happens - the kubelet will be running and the testing can commence.

cc @adelina-t


PatrickLang:
This was failing until #4119 & #4141 merged. Retesting now

PatrickLang:
@adelina-t @dineshgovindasamy - I added the workaround to remove the Azure CNI lock after reboot. Can you review?

PatrickLang:
/hold

The scheduled task still doesn't run after reboot. If you run `Get-ScheduledTask -TaskName "RemoveLock" | Start-ScheduledTask` it will complete, but it never starts automatically. Need to dig more into this.

CecileRobertMichon:
@PatrickLang can you please move this to https://github.com/Azure/aks-engine?

PR #4114, opened 2018-10-24
Make --profiling user configurable

### LINKED ISSUE #3500: apiServerConfig --profiling not configurable

---

---

**What version of acs-engine?**: v0.19.3

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)** Kubernetes v1.11.0

**What happened**:

According to the acs-engine documentation [1], the ``apiServerConfig``'s ``--profiling`` is set by default to ``false`` and it is not user configurable (and cannot be set to true through acs-engine).

Because of this, the Kubernetes metrics API will not handle DELETE requests properly [2][3], treating them as GET instead. Some e2e tests (e.g.: ``[sig-scalability] Load capacity    [Feature:ManualPerformance] should be able to handle 3 pods per node { ReplicationController} with 0 secrets, 0 configmaps and 0 daemons``) requires the metrics to be reset / deleted before starting, but they fail because of this issue.

[1]https://github.com/Azure/acs-engine/blob/0acbf7c4c9a74f60888862e679298fd62e875526/docs/clusterdefinition.md#apiserverconfig
[2] https://github.com/kubernetes/kubernetes/blob/501ce61224be3f1fe4d707b4246de2d59c43b26f/staging/src/k8s.io/apiserver/pkg/server/config.go#L576
[3] https://github.com/kubernetes/kubernetes/blob/501ce61224be3f1fe4d707b4246de2d59c43b26f/staging/src/k8s.io/apiserver/pkg/server/routes/metrics.go#L48

Profiling should be configurable.

If profiling is enabled, the following commands should work as expected:

    kubectl proxy --port=8888 &
    curl -X DELETE http://localhost:8888/metrics  # this should output "metrics reset"

    
If profiling is disabled, the curl above would be treated as a GET instead, and all the metrics will returned.

Create a ``kubernetes.json`` file with ``--profiling=true`` [1].
Generate files:

    acs-engine generate kubernetes.json
    
Checking the generated files, the ``azuredeploy.json`` file will contain ``--profiling=false`` for: kubernetesControllerManagerConfig, kubernetesAPIServerConfig, kubernetesSchedulerConfig.

Continuing with the deployment will result in having a Kubernetes deployment in which its API server is configured with --profiling=false and the metrics not being deletable.

[1] https://paste.ubuntu.com/p/rgPV5hmxCZ/

ritazh:
**What this PR does / why we need it**: Make `--profiling` user configurable

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #3500

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```
cc @PatrickLang

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -365,6 +365,12 @@ const (
 	DefaultKubernetesMaxPodsKubenet = "110"
 	// DefaultKubernetesMaxPodsAzureCNI is the maximum number of pods to run on a node for Azure CNI.
 	DefaultKubernetesMaxPodsAzureCNI = "30"
+	// DefaultKubernetesAPIServerEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
+	DefaultKubernetesAPIServerEnableProfiling = "false"
+	// DefaultKubernetesCtrMgrEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
+	DefaultKubernetesCtrMgrEnableProfiling = "false"
+	// DefaultKubernetesSchedulerEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
+	DefaultKubernetesSchedulerEnableProfiling = "false"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


jackfrancis:
This looks good. Can you add some simple unit tests for each that (1) validates the default if no user-config and (2) validates that user-configurable values are processed?

`TestKubeletMaxPods` does this for kubelet `--max-pods` config (actually that test is more tricky becauset there are multiple default paths, but you get the idea)

PR #4119, opened 2018-10-24
Configure Docker Version on Windows Nodes

jsturtevant:
**What this PR does / why we need it**:
Allows configuration of Docker version on windows nodes and sets a default that fixes deployments with server 1803.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #4118  and address making dockerEngineVersion configurable for #3852

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [x] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
Make docker version configurable on Windows nodes with default of Docker EE 17.06
```

PR #4141, opened 2018-10-29
Adding DOCKER_API_VERSION workaround

### LINKED ISSUE #4118: Windows failed to create kubelet: failed to get docker version
no
---

issue
---

any using windows server 1803 image 
---

Kubernetes

**What happened**:
deployed 1803 windows container images.  Windows nodes never join cluster.  get the error as described here: https://github.com/kubernetes/kubernetes/issues/69996

```
server.go:262] failed to run Kubelet: failed to create kubelet: failed to get docker version: Error response from daemon: client version 1.38 is too new. Maximum supported API version is 1.37
```

Kubelet to start and and connect nodes to the cluster

deploy a k8s cluster with 1803 node as of today

Because the Docker version is installed by the image the version got auto upgraded.  Should pin the Docker version to a known working value so that Windows Server moving to new Docker version upgrades in Azure don't cause deployment errors.

PatrickLang:

The October windows image `MicrosoftWindowsServer:WindowsServerSemiAnnual:Datacenter-Core-1803-with-Containers-smalldisk:1803.0.20181017` brought in Docker EE-basic 18.03, which has a problem with Kubernetes v1.12 (https://github.com/kubernetes/kubernetes/issues/69996)

This fix pins the `DOCKER_API_VERSION` environment variable system-wide to match the API supported by Docker for versions 17.06 and 18.03. This fixes the problem with kubelet startup. 

For newer versions, the key should be cleared.

**Which issue this PR fixes** 
fixes #4118 
depends on #4119

### DIFF
--- a/parts/k8s/windowsconfigfunc.ps1
+++ b/parts/k8s/windowsconfigfunc.ps1
@@ -58,7 +58,7 @@ function Install-Docker
         }
 
         default {
-            Write-Log "Docker version $DockerVersion found, clearing DOCKER_API_VERSION")
+            Write-Log "Docker version $DockerVersion found, clearing DOCKER_API_VERSION"
             [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', $null, [System.EnvironmentVariableTarget]::Machine)
         }
     }
--- a/parts/k8s/windowsconfigfunc.ps1
+++ b/parts/k8s/windowsconfigfunc.ps1
@@ -41,6 +41,28 @@ function Install-Docker
         $DockerVersion
     )
 
+    # Note: to get a list of all versions, use this snippet
+    # $versions = (curl.exe -L "https://go.microsoft.com/fwlink/?LinkID=825636&clcid=0x409" | ConvertFrom-Json).Versions | Get-Member -Type NoteProperty | Select-Object Name
+    # Docker version to API version decoder: https://docs.docker.com/develop/sdk/#api-version-matrix
+
+    switch ($DockerVersion.Substring(0,5))
+    {
+        "17.06" {
+            Write-Log "Docker 17.06 found, setting DOCKER_API_VERSION to 1.30"
+            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.30', [System.EnvironmentVariableTarget]::Machine)
+        }
+
+        "18.03" {
+            Write-Log "Docker 18.03 found, setting DOCKER_API_VERSION to 1.37"
+            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.37', [System.EnvironmentVariableTarget]::Machine)
+        }
+
+        default {
+            Write-Log "Docker version $DockerVersion found, clearing DOCKER_API_VERSION")
+            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', $null, [System.EnvironmentVariableTarget]::Machine)
+        }
+    }
+
     try {
         Find-Package -Name Docker -ProviderName DockerMsftProvider -RequiredVersion $DockerVersion -ErrorAction Stop
         Write-Log "Found version $DockerVersion. Installing..."


PatrickLang:
@jsturtevant can you review? I'm testing now

PatrickLang:
18.03 with Windows Server version 1803 works:
```
 OS Image:                   Windows Server Datacenter
 Operating System:           windows
 Architecture:               amd64
 Container Runtime Version:  docker://18.3.1
 Kubelet Version:            v1.12.2
 Kube-Proxy Version:         v1.12.2
ProviderID:                  azure:///subscriptions/34229103-1113-415f-ba20-80f86dd700ce/resourceGroups/plang1029upgarg/
providers/Microsoft.Compute/virtualMachines/21418k8s9000
Non-terminated Pods:         (0 in total)
  Namespace                  Name    CPU Requests  CPU Limits  Memory Requests  Memory Limits
  ---------                  ----    ------------  ----------  ---------------  -------------
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                       Requests  Limits
  --------                       --------  ------
  cpu                            0 (0%)    0 (0%)
  memory                         0 (0%)    0 (0%)
  attachable-volumes-azure-disk  0         0
Events:
  Type    Reason                 Age              From                   Message
  ----    ------                 ----             ----                   -------
  Normal  Starting               5m               kubelet, 21418k8s9000  Starting kubelet.
  Normal  NodeHasSufficientDisk  5m (x2 over 5m)  kubelet, 21418k8s9000  Node 21418k8s9000 status is now: NodeHasSuffici
entPID
```

Still waiting on my 17.06 test pass

parts/k8s/windowsconfigfunc.ps1
```diff
@@ -41,6 +41,28 @@ function Install-Docker
...
 
+    # Note: to get a list of all versions, use this snippet
+    # $versions = (curl.exe -L "https://go.microsoft.com/fwlink/?LinkID=825636&clcid=0x409" | ConvertFrom-Json).Versions | Get-Member -Type NoteProperty | Select-Object Name
+    # Docker version to API version decoder: https://docs.docker.com/develop/sdk/#api-version-matrix
+
+    switch ($DockerVersion.Substring(0,5))
+    {
+        "17.06" {
+            Write-Log "Docker 17.06 found, setting DOCKER_API_VERSION to 1.30"
+            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.30', [System.EnvironmentVariableTarget]::Machine)
```
jsturtevant: In 17.06 1.30 is the default api version.  Is this check necessary?

PatrickLang: This is actually a key read by the kubelet. The kubelet doesn't set the version used in the client today, so this is the way to force it.

jsturtevant:
Does it make sense to have the docker API version configurable from the stand point of acs-engine model?  Along side of the configuration of the docker version, you could also set the api version you want to use at the model level:

```
 "windowsProfile": {
      "adminUsername": "azureuser",
      "adminPassword": "replacepassword1234$",
      "windowsDockerVersion": "17.06.2-ee-16"
      "dockerapiversion": "1.30"
    },
```

parts/k8s/windowsconfigfunc.ps1
jsturtevant: It is working in 17.06 via https://github.com/Azure/acs-engine/pull/4119.  Not sure I understand why the extra environment variable is needed in this case?

PatrickLang:
In this case I really want acs-engine to "do the right thing". I don't want the cluster admin to have to decide on what Docker API version.

I would argue this should be automatic as part of the Docker API, but I don't know the background for why it's like this.

PatrickLang:
17.06 tests:
```
OS Image:                   Windows Server Datacenter
 Operating System:           windows
 Architecture:               amd64
 Container Runtime Version:  docker://17.6.2
 Kubelet Version:            v1.12.2
 Kube-Proxy Version:         v1.12.2
ProviderID:                  azure:///subscriptions/34229103-1113-415f-ba20-80f86dd700ce/resourceGroups/plang1029defarg/
providers/Microsoft.Compute/virtualMachines/87970k8s9000
Non-terminated Pods:         (0 in total)
  Namespace                  Name    CPU Requests  CPU Limits  Memory Requests  Memory Limits
  ---------                  ----    ------------  ----------  ---------------  -------------
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                       Requests  Limits
  --------                       --------  ------
  cpu                            0 (0%)    0 (0%)
  memory                         0 (0%)    0 (0%)
  attachable-volumes-azure-disk  0         0
Events:                          <none>
```

PR #4154, opened 2018-10-31
Calico support for azure-vnet-ipam

caseydavenport:

This PR does a few things:
- Updates Calico to v3.3.0 (latest as of this moment)
- Introduces support for using Calico network policy with Azure vnet networking.

There are a couple areas where I could use some acs-engine help. I'll call them out explicitly in comments on the areas in question.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Support networkPolicy=calico with networkPlugin=azure
```


parts/k8s/kubernetesconfigs.sh
```diff
@@ -185,8 +185,10 @@ function configureCNI() {
     retrycmd_if_failure 30 6 10 modprobe br_netfilter || exit $ERR_MODPROBE_FAIL
     echo -n "br_netfilter" > /etc/modules-load.d/br_netfilter.conf
     if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
-        mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
-        chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
+	# Casey TODO: When using Calico with Azure networking, we don't want this CNI
```
caseydavenport: This is one area where I'm not sure what the best approach is - when using Calico for network policy with the Azure plugin, Calico will install its own config via the DaemonSet, so we don't want to install this one.

I don't see a NETWORK_POLICY environment variable set in this script, though I could be missing something?

parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml
```diff
@@ -137,8 +137,7 @@ data:
           "nodename": "__KUBERNETES_NODE_NAME__",
           "mtu": 1500,
           "ipam": {
-              "type": "host-local",
```
caseydavenport: This is the other area that needs some design input - right now I've adjusted the config to be hardcoded to use the azure-vnet-ipam plugin. However, this means the old "Calico + kubenet" approach won't work anymore.

I think we need a way to modify this config based on the networking type specified, but not sure what the best way to do that is. Any thoughts?

caseydavenport:
@jackfrancis would appreciate your thoughts on this one :)

parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml
tariq1890: Hi @caseydavenport. To make certain parts of the calico YAML configurable, we add placeholders in the YAML file. For the user to provide their desired config, they specify it as part of the Add-on Config in their apimodel. You can check [this](https://github.com/Azure/acs-engine/blob/1f1dda675dc168ee46a47ab3a572f8871496272c/pkg/api/types.go#L90). After, you will find that the subsitution logic for these placeholders are found [here](https://github.com/Azure/acs-engine/blob/1f1dda675dc168ee46a47ab3a572f8871496272c/parts/k8s/kubernetesmastercustomdata.yml#L361).

tariq1890: Also check L345 of this file, you will see that we use the Calico IPv4 as the placeholder here

jackfrancis:
This is looking good @caseydavenport. Just to clarify that your intent matches the implementation: the template files under `parts/k8s/addons/<version>` will be rendered only for k8s major.minor versions (e.g., v1.10.*n*) that match `<version>`. I see you have included changes to the 1.10 version of the calico manifest. Given the current implementation, not including your changes to the < 1.10 versions and the 1.11 version means that the calico + Azure CNI implementation will only be delivered w/ k8s 1.10 or k8s 1.12+.

The backstory there is that the tolerations spec is slightly different for 1.12 and greater, other than that the manifests are identical.

So this is just a reminder that once we get a working manifest for one version of k8s, we'll want to copy/paste the changes across all the 1.11 through 1.7 templates (tedious work, I know). Lemme know if that doesn't make sense.

caseydavenport:
@jackfrancis yep, I'm happy to copy to older (and newer) versions once we agree this change set makes sense! TBH, I did v1.10 because it seemed to be what was used by default and was easy to test my changes - suspected I'd need to port them elsewhere as well, though.

I've pushed a change which configures the CNI config based on the specified plugin and tested it works locally. The last bit is around not installing the Azure CNI config when Calico is enabled, still need to find a way to do that.

parts/k8s/kubernetesconfigs.sh
caseydavenport: @jackfrancis I'd appreciate any thoughts on how to handle this bit of code.

We need skip installation of this file when using NetworkPolicy=Calico with NetworkPlugin=Azure.

But right now, this always gets installed when NetworkPlugin=Azure. Is there recommended way for me to plumb the right information to this stage?

feiskyer: @caseydavenport NETWORK_POLICY could be added in kubernetesmastervars.t, e.g. `' NETWORK_POLICY=',parameters('networkPolicy')`

feiskyer:
The following test is failing because of adding calico with azure plugin:

```
--- FAIL: Test_Properties_ValidateNetworkPluginPlusPolicy (0.00s)
    validate_test.go:704: should error on networkPolicy="calico" + networkPlugin="azure"
```

jackfrancis:
@caseydavenport, I made two changes to move testing forward:

- now passing networkPolicy (i.e., calico) directly to the provision scripts so that they may conditionally do foo based on whether or not calico is present
- fixed validation test

We're testing this continuously starting now, so far looks good!

caseydavenport:
@jackfrancis awesome, thanks for that!

This looks good to me - I'll get started porting the manifest changes to other versions.

caseydavenport:
@jackfrancis I've updated the other addon manifests for the other k8s versions as well to match what I had done for 1.10.

What's the next step for this guy?

parts/k8s/addons/1.10/kubernetesmasteraddons-calico-daemonset.yaml
```diff
@@ -395,7 +390,7 @@ spec:
         # This container installs the Calico CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: quay.io/calico/cni:v3.1.3
+          image: quay.io/calico/cni:v3.4.0-0.dev-13-gcda66d6
```
jackfrancis: Is there a final version for the cni image, say the equivalent of `quay.io/calico/typha:v3.3.0`?

jackfrancis:
@caseydavenport this /lgtm for merge once we validate we have the `install-cni` image reference we want

jackfrancis:
@caseydavenport have you tested this change against a 1.12 cluster buildout? Initial test signal was non-passing, wil retry.

caseydavenport:
@jackfrancis ok, I got to the bottom of it.

For whatever reason 1.12 has a different timing and exposed a race in the new Azure code I wrote in the CNI plugin. I've got a patch working here: https://github.com/projectcalico/cni-plugin/pull/634

Will update this PR to include it once merged.

jackfrancis:
@caseydavenport awesome, thx, I'll be ready!

caseydavenport:
@jackfrancis I've updated to the new version.

An reason the circle tests don't seem to be reporting?

jackfrancis:
@caseydavenport just takes a while :)

These tests don't validate calico + Azure CNI, I'll run another round of tests and report back. What are we going to do about these dev-versioned images? Do we want to ship with those references?

caseydavenport:
> What are we going to do about these dev-versioned images? Do we want to ship with those references?

@jackfrancis If you'd like I can re-tag as something else, but these tags are stable references and IIRC the plan was to ship this for now and update to use Calico v3.4.0 as soon as it is available, which is currently looking like early December.

I'm happy to tag as something else if that's desirable, but ultimately the code is the same.

caseydavenport:
Thanks everyone for the help getting this merged!

PR #4085, opened 2018-10-19
remove unused nsg for AKS

weinong:
**What this PR does / why we need it**:
AKS doesn't need these NSG rules.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #
#4086

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
```
Remove unused NSG rules for AKS

PR #4164, opened 2018-11-01
Remove unreachable NSG code

CecileRobertMichon:
**What this PR does / why we need it**: PR #4085 added an if statement to remove NSG for AKS but that code is effectively unreachable because it is inside an "if hosted master statement"

Non-hosted master clusters have their own NSG definition in master resources:
https://github.com/Azure/acs-engine/blob/fix-nsg/parts/k8s/kubernetesmasterresources.t#L87

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/parts/k8s/kubernetesbase.t
+++ b/parts/k8s/kubernetesbase.t
@@ -146,54 +146,7 @@
       "location": "[variables('location')]",
       "name": "[variables('nsgName')]",
       "properties": {
-        "securityRules": [
-{{if .HasWindows}}
-          {
-            "name": "allow_rdp",
-            "properties": {
-              "access": "Allow",
-              "description": "Allow RDP traffic to master",
-              "destinationAddressPrefix": "*",
-              "destinationPortRange": "3389-3389",
-              "direction": "Inbound",
-              "priority": 102,
-              "protocol": "Tcp",
-              "sourceAddressPrefix": "*",
-              "sourcePortRange": "*"
-            }
-          },
-{{end}}
-{{if not IsHostedMaster}}
-          {
-            "name": "allow_ssh",
-            "properties": {
-              "access": "Allow",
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesbase.t
```diff
@@ -146,54 +146,7 @@
       "location": "[variables('location')]",
       "name": "[variables('nsgName')]",
       "properties": {
-        "securityRules": [
-{{if .HasWindows}}
```
CecileRobertMichon: AKS doesn't have Windows, so this never applies

parts/k8s/kubernetesbase.t
```diff
@@ -146,54 +146,7 @@
...
-              "destinationPortRange": "3389-3389",
-              "direction": "Inbound",
-              "priority": 102,
-              "protocol": "Tcp",
-              "sourceAddressPrefix": "*",
-              "sourcePortRange": "*"
-            }
-          },
-{{end}}
-{{if not IsHostedMaster}}
```
CecileRobertMichon: this whole thing is inside a `{{if IsHostedMaster}}` block https://github.com/Azure/acs-engine/pull/4164/files#diff-66965012f8a909902930dd7fa6804d3eR96 so this is dead code

CecileRobertMichon:
@weinong FYI

parts/k8s/kubernetesbase.t
jackfrancis: Can we indent all the conditional blocks so that it is easier to follow?

CecileRobertMichon: Done. https://github.com/Azure/acs-engine/pull/4164/files?w=1 to see the diff without whitespaces.

PR #4169, opened 2018-11-02
distinct outbound test for mooncake clusters

jackfrancis:
**What this PR does / why we need it**: Also, allow a little longer for failure

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
distinct outbound test for mooncake
```

### DIFF
--- a/parts/k8s/kubernetesmasterresources.t
+++ b/parts/k8s/kubernetesmasterresources.t
@@ -929,7 +929,7 @@
         {{if IsOpenShift}}
           "script": "{{ Base64 OpenShiftGetMasterSh }}"
         {{else}}
-          "commandToExecute": "[concat('retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $retries); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; ERR_OUTBOUND_CONN_FAIL=50; retrycmd_if_failure 40 1 3 nc -vz k8s.gcr.io 443 || retrycmd_if_failure 40 1 3 nc -vz gcr.azk8s.cn 80 || exit $ERR_OUTBOUND_CONN_FAIL; for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),' ',variables('provisionScriptParametersMaster'), ' /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]"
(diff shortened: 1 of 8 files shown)


parts/k8s/kubernetesmasterresources.t
```diff
@@ -929,7 +929,7 @@
...
+          "commandToExecute": "[concat('retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $retries); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; ERR_OUTBOUND_CONN_FAIL=50; retrycmd_if_failure 75 1 3 nc -vz {{if IsMooncake}}gcr.azk8s.cn 80{{else}}k8s.gcr.io 443 || retrycmd_if_failure 75 1 3 nc -vz gcr.io 443{{end}} || exit $ERR_OUTBOUND_CONN_FAIL; for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),' ',variables('provisionScriptParametersMaster'), ' /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]"
```
CecileRobertMichon: Should the secondary check not be `gcr.io` in case `gcr.io` is down since the first one is already `gcr.io`? Maybe we could test `azure.com` instead?

jackfrancis: I'd like to get away from "random" internet destinations, if possible.

jackfrancis: I mean, that's my thinking, open to other options.

pkg/api/const.go
```diff
@@ -382,8 +382,10 @@ const (
 )
 
 const (
-	azurePublicCloud       = "AzurePublicCloud"
-	azureChinaCloud        = "AzureChinaCloud"
+	// AzurePublicCloud is a const string reference identifier for public cloud
+	AzurePublicCloud = "AzurePublicCloud"
+	// AzureChinaCloud is a const string reference identifier for china cloud
+	AzureChinaCloud        = "AzureChinaCloud"
 	azureGermanCloud       = "AzureGermanCloud"
```
CecileRobertMichon: why not change all the Azure cloud constant names for consistency?

jackfrancis: Because it's kind of golang bad form to Capitalize something that isn't exported

parts/k8s/kubernetesmasterresources.t
CecileRobertMichon: I don't think `azure.com` is random. We can think of something else too, I just don't think it makes sense to test `gcr.io` twice.

pkg/acsengine/template_generator.go
```diff
@@ -371,7 +371,11 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		},
 		"UseAksExtension": func() bool {
 			cloudSpecConfig := cs.GetCloudSpecConfig()
-			return cloudSpecConfig.CloudName == azurePublicCloud
+			return cloudSpecConfig.CloudName == api.AzurePublicCloud
+		},
+		"IsMooncake": func() bool {
+			cloudSpecConfig := cs.GetCloudSpecConfig()
+			return cloudSpecConfig.CloudName == api.AzureChinaCloud
```
CecileRobertMichon: I would add a nil check here for `cloudSpecConfig`

jackfrancis: I don't see how `cloudSpecConfig` could ever be `nil`. The function returns a type struct (not a pointer to one), so even if the map lookup key didn't exist it would return a zero value `AzureEnvironmentSpecConfig` (which is not `nil`)

parts/k8s/kubernetesmasterresources.t
jackfrancis: s/random/internet resource that isn't actually used during provisioning

If we want a fallback "generic internet destination" I'd vote for 8.8.8.8 as a fallback.

CecileRobertMichon: 8.8.8.8 works too, as long as the fallback is not the same destination we already checked.

feiskyer:
LGTM

andyzhangx:
That's a great improvement, thanks.

PR #4185, opened 2018-11-06
Add ContainerInventory(Docker based workflow) test 

rashmichandrashekar:
Update omsagent tag to use a test agent with changes to add logs for containerinventory (Docker based workflow)
Add test for ContainerInventory workflow


rashmichandrashekar:
Will send out an update PR

PR #4194, opened 2018-11-06
add AKS container images to VHD build script

jackfrancis:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
add AKS container images to VHD build script
```

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -131,9 +131,9 @@ for TUNNELFRONT_VERSION in ${TUNNELFRONT_VERSIONS}; do
     pullContainerImage "docker" "docker.io/deis/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
 done
 
-KUBE_SVC_REDIRECT_VERSIONS="v1.0.2"
+KUBE_SVC_REDIRECT_VERSIONS="1.0.2"
 for KUBE_SVC_REDIRECT_VERSION in ${KUBE_SVC_REDIRECT_VERSIONS}; do
-    pullContainerImage "docker" "docker.io/deis/kube-svc-redirect:${KUBE_SVC_REDIRECT_VERSION}"
+    pullContainerImage "docker" "docker.io/deis/kube-svc-redirect:v${KUBE_SVC_REDIRECT_VERSION}"
 done
 
 NGINX_VERSIONS="1.13.12-alpine"
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -126,6 +126,21 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
     pullContainerImage "docker" "nvidia/k8s-device-plugin:${NVIDIA_DEVICE_PLUGIN_VERSION}"
 done
 
+TUNNELFRONT_VERSIONS="v1.9.2-v4.0.4"
+for TUNNELFRONT_VERSION in ${TUNNELFRONT_VERSIONS}; do
+    pullContainerImage "docker" "docker.io/deis/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
+done
+
+KUBE_SVC_REDIRECT_VERSIONS="v1.0.2"
+for KUBE_SVC_REDIRECT_VERSION in ${KUBE_SVC_REDIRECT_VERSIONS}; do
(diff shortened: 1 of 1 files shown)


packer/install-dependencies.sh
```diff
@@ -126,6 +126,21 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
     pullContainerImage "docker" "nvidia/k8s-device-plugin:${NVIDIA_DEVICE_PLUGIN_VERSION}"
 done
 
+TUNNELFRONT_VERSIONS="v1.9.2-v4.0.4"
```
CecileRobertMichon: // TODO: cleanup images not needed at runtime in https://github.com/Azure/acs-engine/blob/cleanup-images/parts/k8s/kubernetesinstalls.sh#L220

packer/install-dependencies.sh
```diff
@@ -126,6 +126,21 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
     pullContainerImage "docker" "nvidia/k8s-device-plugin:${NVIDIA_DEVICE_PLUGIN_VERSION}"
```
CecileRobertMichon: is this missing a "v" for version? (I know it's not from this PR but just noticed)

packer/install-dependencies.sh
```diff
@@ -126,6 +126,21 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
...
 done
 
+TUNNELFRONT_VERSIONS="v1.9.2-v4.0.4"
+for TUNNELFRONT_VERSION in ${TUNNELFRONT_VERSIONS}; do
+    pullContainerImage "docker" "docker.io/deis/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
+done
+
+KUBE_SVC_REDIRECT_VERSIONS="v1.0.2"
```
CecileRobertMichon: let's make the version var "1.0.2" and put the "v" in the image string below to be consistent with the rest of the file and avoid human error

packer/install-dependencies.sh
```diff
@@ -126,6 +126,21 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
...
+for KUBE_SVC_REDIRECT_VERSION in ${KUBE_SVC_REDIRECT_VERSIONS}; do
+    pullContainerImage "docker" "docker.io/deis/kube-svc-redirect:${KUBE_SVC_REDIRECT_VERSION}"
+done
+
+NGINX_VERSIONS="1.13.12-alpine"
+for NGINX_VERSION in ${NGINX_VERSIONS}; do
+    pullContainerImage "docker" "nginx:${NGINX_VERSION}"
```
CecileRobertMichon: this one doesn't have the "v" for version?

jackfrancis: correct

packer/install-dependencies.sh
CecileRobertMichon: what does the dash mean in "v1.9.2-v4.0.4"? is it one version? or should it be a space?

packer/install-dependencies.sh
jackfrancis: there is no "v" prefix for this image

packer/install-dependencies.sh
jackfrancis: Talk to the publisher. :) It's just the way it's versioned.

jackfrancis:
```
Containers:
  tunnel-front:
    Container ID:   docker://bbf2a60b072c724b88132537a0caf961c467c73790c8e44f12a9467ba402e85e
    Image:          docker.io/deis/hcp-tunnel-front:v1.9.2-v4.0.4
```

jackfrancis: Actually, given that this is already oddly versioned as you noticed, let's skip the whole "v" removal thing and just keep this as an oddball.

CecileRobertMichon: sounds good

CecileRobertMichon:
https://msazure.visualstudio.com/One/_build/results?buildId=18402607&_a=summary&view=logs

PR #4208, opened 2018-11-07
setting default Images in addon defaults  instead of params_k8s.go

tariq1890:
Setting default Image names in the right place :)


pkg/api/addons.go
```diff
@@ -169,11 +179,11 @@ func (cs *ContainerService) setAddonsConfig() {
 		Containers: []KubernetesContainerSpec{
 			{
 				Name:           "omsagent",
-				Image:          "microsoft/oms:ciprod10162018-2",
 				CPURequests:    "50m",
 				MemoryRequests: "200Mi",
 				CPULimits:      "150m",
 				MemoryLimits:   "750Mi",
+				Image:          "microsoft/oms:ciprod10162018-2",
```
CecileRobertMichon: can this be changed to not be hardcoded anymore using `ContainerMonitoringAddonName` in k8sComponents? I see you already did half the work by changing the name to match the image

pkg/api/defaults_test.go
```diff
@@ -181,6 +181,60 @@ func TestAddonsIndexByName(t *testing.T) {
 	}
 }
 
+func TestAssignDefaultAddonImages(t *testing.T) {
```
CecileRobertMichon: 👏

pkg/api/addons.go
jackfrancis: punt on this while omsagent team is developing it

PR #4209, opened 2018-11-08
Fix validation for firstConsecutiveStaticIP

ritazh:
**What this PR does / why we need it**: Fix validation for firstConsecutiveStaticIP

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -383,6 +383,9 @@ func (dc *deployCmd) validateApimodel() (*api.ContainerService, string, error) {
 				return nil, "", errors.New("when using the kubernetes orchestrator, must either set useManagedIdentity in the kubernetes config or set --client-id and --client-secret or KeyvaultSecretRef of secret (also available in the API model)")
 			}
 		}
+		if p.MasterProfile != nil && p.MasterProfile.IsVirtualMachineScaleSets() && p.MasterProfile.VnetSubnetID != "" && p.MasterProfile.FirstConsecutiveStaticIP != "" {
+			return nil, "", errors.Errorf("when masterProfile's availabilityProfile is VirtualMachineScaleSets and a vnetSubnetID is specified, the firstConsecutiveStaticIP should be empty and will be determined by an offset from the first IP in the vnetCidr %s", p.MasterProfile.FirstConsecutiveStaticIP)
+		}
 	}
 
 	// This isn't terribly elegant, but it's the easiest way to go for now w/o duplicating a bunch of code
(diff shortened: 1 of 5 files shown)


cmd/deploy.go
```diff
@@ -383,6 +383,9 @@ func (dc *deployCmd) validateApimodel() (*api.ContainerService, string, error) {
 				return nil, "", errors.New("when using the kubernetes orchestrator, must either set useManagedIdentity in the kubernetes config or set --client-id and --client-secret or KeyvaultSecretRef of secret (also available in the API model)")
 			}
 		}
+		if p.MasterProfile != nil && p.MasterProfile.IsVirtualMachineScaleSets() && p.MasterProfile.VnetSubnetID != "" && p.MasterProfile.FirstConsecutiveStaticIP != "" {
```
jackfrancis: I think this should go in the `validate.go` area, as this validation is only a convenience for `acs-engine deploy` CLI usage.

ritazh: This logic was originally in `validate.go`. But as we have seen in e2e tests, what we want to do here is to validate the user-provided apimodel, not the generated apimodel. Since we want to notify the user that we are generating this `firstConsecutiveStaticIP` for them if they are using custom vnet and VMSS master profile, instead of asking them to provide a value.

When we run e2e tests, `cli_provisioner.go` calls `engine.ParseOutput` in `template.go` which calls `apiloader.LoadContainerServiceFromFile(path, true, false, nil)` that triggers `validate.go`on the generated apimodel.
https://github.com/Azure/acs-engine/blob/29ad61683b3c1944b1684f09ce11908ecace9307/test/e2e/runner/cli_provisioner.go#L223

This causes the e2e tests to fail since the generated apimodel now includes the`firstConsecutiveStaticIP` we generated based on the vnetCidr provided by the user.  
https://github.com/Azure/acs-engine/blob/29ad61683b3c1944b1684f09ce11908ecace9307/test/e2e/engine/template.go#L283-L294

When we run `acs-engine deploy`, `validateApimodel` in deploy.go validates the user-provided apimodel before `dc.containerService.SetPropertiesDefaults` is called. https://github.com/Azure/acs-engine/blob/29ad61683b3c1944b1684f09ce11908ecace9307/cmd/deploy.go#L88-L91

jackfrancis: Hmm, `vlabs/validate.go` should *only* validate user input, not the generated data...

CecileRobertMichon:
Closing since we've moved to https://github.com/Azure/aks-engine. @ritazh is this PR still relevant?

ritazh:
@CecileRobertMichon I think this PR was created to address a failed e2e test. We can revisit this if the issue comes up again. Thanks!

PR #4210, opened 2018-11-08
mount xtables lock in proxy

khenidak:
Mounts iptables lock from host into kubeproxy to allow `iptables .. -w` to work correctly


serbrech:
FYI @marrobi

Issue #4064, opened 2018-10-18
CSE error: exit code 84


MirzaSikander:
**Is this a request for help?**: yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): issue

---

**What version of acs-engine?**:
Version: v0.23.1
GitCommit: 94d7cf6fb
GitTreeState: clean

---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
kubernetes v1.10.8

**What happened**:
Deployment failed.

Here is the error
```
{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.","details":[{"code":"Conflict","message":"{\r\n \"status\": \"Failed\",\r\n \"error\": {\r\n \"code\": \"ResourceDeploymentFailure\",\r\n \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n \"details\": [\r\n {\r\n \"code\": \"VMExtensionProvisioningError\",\r\n \"message\": \"VM has reported a failure when processing extension 'vmssCSE'. Error message: \\\"Enable failed: failed to execute command: command terminated with exit status=84\\n[stdout]\\n\\n[stderr]\\n\\\".\"\r\n }\r\n ]\r\n }\r\n}"}]}
```
The VMSS deployment had a status of `Conflict`
![image](https://user-images.githubusercontent.com/2823474/47133153-474f3e80-d25c-11e8-91e6-aa4e8036f465.png)

**What you expected to happen**:
Deployment should succeed.

**How to reproduce it** (as minimally and precisely as possible):
Here is the cluster definition:
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "kubernetesConfig": {
        "gcHighThreshold" : 60,
        "gcLowThreshold" : 40,
        "addons": [
          {
            "name": "cluster-autoscaler",
            "enabled": true,
            "config": {
              "minNodes": "1",
              "maxNodes": "10"
            }
          }
        ]
      }
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "k8s-athens-internal-01",
      "vmSize": "Standard_D2_v3"
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 1,
        "vmSize": "Standard_D2_v3",
[...]

PR #4221, opened 2018-11-09
ensure N series clusters get aks-docker-engine

jackfrancis:
**What this PR does / why we need it**: This fixes is an upgrade scenario for N series agent pool nodes, which will have `"aks"` as the distro value in their existing api model, which is no longer a valid value for N series-backed agent pools.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: #4064

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
ensure N series clusters get aks-docker-engine
```

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -438,16 +438,20 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabledWindows = helpers.PointerToBool(DefaultAcceleratedNetworkingWindowsEnabled)
 		}
 
+		// Ensure distro is set properly for N Series SKUs no matter what, because
+		// (1) At present, "aks-docker-engine" is the only working distro base for running GPU workloads on N Series SKUs
+		// (2) Previous versions of acs-engine had working implementations using a different distro value,
+		//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
+		if IsNSeriesSKU(p) {
+			profile.Distro = AKSDockerEngine
+		}
+
 		if profile.Distro == "" && profile.OSType != Windows {
 			if p.OrchestratorProfile.IsKubernetes() {
 				if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
 					profile.Distro = Ubuntu
 				} else {
-					if IsNSeriesSKU(p) {
-						profile.Distro = AKSDockerEngine
-					} else {
-						profile.Distro = AKS
-					}
+					profile.Distro = AKS
 				}
 			} else if !p.OrchestratorProfile.IsOpenShift() {
 				profile.Distro = Ubuntu
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -458,7 +458,9 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 					profile.Distro = Ubuntu
 				}
 			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) && profile.IsNSeriesSKU() {
-				profile.Distro = AKSDockerEngine
+				if profile.Distro != Ubuntu {
+					profile.Distro = AKSDockerEngine
+				}
 			}
 		}
 
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -438,23 +438,27 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabledWindows = helpers.PointerToBool(DefaultAcceleratedNetworkingWindowsEnabled)
 		}
 
-		// Ensure distro is set properly for N Series SKUs no matter what, because
-		// (1) At present, "aks-docker-engine" is the only working distro base for running GPU workloads on N Series SKUs
-		// (2) Previous versions of acs-engine had working implementations using a different distro value,
-		//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
-		if IsNSeriesSKU(p) {
-			profile.Distro = AKSDockerEngine
-		}
-
-		if profile.Distro == "" && profile.OSType != Windows {
-			if p.OrchestratorProfile.IsKubernetes() {
-				if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
+		if profile.OSType != Windows {
+			if profile.Distro == "" {
+				if p.OrchestratorProfile.IsKubernetes() {
+					if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
+						profile.Distro = Ubuntu
+					} else {
+						// Ensure distro is set properly for N Series SKUs, because
+						// (1) At present, "aks-docker-engine" and "ubuntu" are the only working distro base for running GPU workloads on N Series SKUs
+						// (2) Previous versions of acs-engine had working implementations using the "aks" distro value,
+						//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
+						if profile.IsNSeriesSKU() {
+							profile.Distro = AKSDockerEngine
+						} else {
+							profile.Distro = AKS
+						}
+					}
+				} else if !p.OrchestratorProfile.IsOpenShift() {
 					profile.Distro = Ubuntu
-				} else {
-					profile.Distro = AKS
 				}
-			} else if !p.OrchestratorProfile.IsOpenShift() {
-				profile.Distro = Ubuntu
+			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) && profile.IsNSeriesSKU() {
+				profile.Distro = AKSDockerEngine
 			}
 		}
 
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -444,10 +444,6 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 					if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
 						profile.Distro = Ubuntu
 					} else {
-						// Ensure distro is set properly for N Series SKUs, because
-						// (1) At present, "aks-docker-engine" and "ubuntu" are the only working distro base for running GPU workloads on N Series SKUs
-						// (2) Previous versions of acs-engine had working implementations using the "aks" distro value,
-						//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
 						if profile.IsNSeriesSKU() {
 							profile.Distro = AKSDockerEngine
 						} else {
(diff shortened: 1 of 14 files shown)


pkg/api/defaults.go
```diff
@@ -438,16 +438,20 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabledWindows = helpers.PointerToBool(DefaultAcceleratedNetworkingWindowsEnabled)
 		}
 
+		// Ensure distro is set properly for N Series SKUs no matter what, because
+		// (1) At present, "aks-docker-engine" is the only working distro base for running GPU workloads on N Series SKUs
+		// (2) Previous versions of acs-engine had working implementations using a different distro value,
+		//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
+		if IsNSeriesSKU(p) {
+			profile.Distro = AKSDockerEngine
```
CecileRobertMichon: do we currently support N series VMs in Windows agent pools? If so (or maybe it's a good idea no matter what in case that changes later) we should also check `profile.OSType != Windows`

pkg/api/defaults.go
```diff
@@ -438,16 +438,20 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			profile.AcceleratedNetworkingEnabledWindows = helpers.PointerToBool(DefaultAcceleratedNetworkingWindowsEnabled)
 		}
 
+		// Ensure distro is set properly for N Series SKUs no matter what, because
+		// (1) At present, "aks-docker-engine" is the only working distro base for running GPU workloads on N Series SKUs
```
CecileRobertMichon: that's not technically true, "Ubuntu" still works. Does this mean agent pool profiles with the Ubuntu distro will be forced changed to AKSDockerEngine? What if I want to request Ubuntu as a user?

CecileRobertMichon:
```
if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
					profile.Distro = Ubuntu
				}
```
should still apply as well

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
 }
 
 func TestAKSDockerEngineDistro(t *testing.T) {
```
jackfrancis: @CecileRobertMichon start by reading these tests to get a sense of what this PR delivers

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
+	properties.setAgentProfileDefaults(false, false)
+
+	if properties.AgentPoolProfiles[0].Distro != "" {
+		t.Fatalf("Expected %s distro for N-series Windows VM, got %s instead", "", properties.AgentPoolProfiles[0].Distro)
+	}
+	if properties.AgentPoolProfiles[1].Distro != AKSDockerEngine {
+		t.Fatalf("Expected %s distro for N-series pool, got %s instead", AKSDockerEngine, properties.AgentPoolProfiles[1].Distro)
+	}
+
+	// Non-k8s context
```
CecileRobertMichon: what does non-k8s context mean?

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
+	mockCS = getMockBaseContainerService("1.10.9")
+	properties = mockCS.Properties
+	properties.MasterProfile.Count = 1
+	properties.setAgentProfileDefaults(false, false)
+
+	if properties.AgentPoolProfiles[0].Distro != Ubuntu {
+		t.Fatalf("Expected %s distro for N-series Windows VM, got %s instead", Ubuntu, properties.AgentPoolProfiles[0].Distro)
```
CecileRobertMichon: is this error a bad copy paste? The definition doesn't involve windows

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
+	properties = mockCS.Properties
+	properties.OrchestratorProfile.OrchestratorType = "Kubernetes"
+	properties.MasterProfile.Count = 1
+	properties.AgentPoolProfiles[0].VMSize = "Standard_NC6"
+	properties.AgentPoolProfiles[0].OSType = Windows
+	properties.AgentPoolProfiles[1].VMSize = "Standard_NC6"
+	properties.setAgentProfileDefaults(false, false)
+
+	if properties.AgentPoolProfiles[0].Distro != "" {
+		t.Fatalf("Expected %s distro for N-series Windows VM, got %s instead", "", properties.AgentPoolProfiles[0].Distro)
```
CecileRobertMichon: `Expected no distro value` would be a more elegant error message

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
-	if properties.AgentPoolProfiles[0].Distro != AKS {
-		t.Fatalf("Expected %s distro for N-series VM got %s instead", AKS, properties.AgentPoolProfiles[0].Distro)
+	if properties.AgentPoolProfiles[0].Distro != Ubuntu {
+		t.Fatalf("Expected %s distro for N-series pool with small disk, got %s instead", Ubuntu, properties.AgentPoolProfiles[0].Distro)
+	}
+	if properties.AgentPoolProfiles[1].Distro != Ubuntu {
+		t.Fatalf("Expected %s distro for D-series pool with small disk, got %s instead", Ubuntu, properties.AgentPoolProfiles[1].Distro)
+	}
+
+	// N Series agent pools should always get the "aks-docker-engine" distro for upgrade flows
```
CecileRobertMichon: what if the previous value was "Ubuntu"?

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
+		t.Fatalf("Expected %s distro for D-series pool, got %s instead", AKS, properties.AgentPoolProfiles[1].Distro)
+	}
+	if properties.AgentPoolProfiles[2].Distro != AKSDockerEngine {
+		t.Fatalf("Expected %s distro for D-series pool, got %s instead", AKSDockerEngine, properties.AgentPoolProfiles[1].Distro)
+	}
+	if properties.AgentPoolProfiles[3].Distro != Ubuntu {
+		t.Fatalf("Expected %s distro for D-series pool, got %s instead", Ubuntu, properties.AgentPoolProfiles[1].Distro)
+	}
+
+	// N Series agent pools should always get the "aks-docker-engine" distro for scale flows
```
CecileRobertMichon: same question, what if the previous value was "Ubuntu"?

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,125 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
 }
 
 func TestAKSDockerEngineDistro(t *testing.T) {
+	// N Series agent pools should always get the "aks-docker-engine" distro for default create flows
```
CecileRobertMichon: What about the case where if I specifically request "ubuntu" as a user?

pkg/api/defaults_test.go
jackfrancis: Adding coverage for the existing code logic that is `!IsKubernetes()` and `!IsOpenShift()`. Somewhat artificial but the logic is there and I don't want to remove it.

pkg/api/defaults_test.go
```diff
@@ -914,26 +914,135 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
...
+	}
+
+	// Non-k8s context
+	mockCS = getMockBaseContainerService("1.10.9")
+	properties = mockCS.Properties
+	properties.MasterProfile.Count = 1
+	properties.setAgentProfileDefaults(false, false)
+
+	if properties.AgentPoolProfiles[0].Distro != Ubuntu {
+		t.Fatalf("Expected no distro value for non k8s context, got %s instead", properties.AgentPoolProfiles[0].Distro)
```
CecileRobertMichon: this error message doesn't match the if check

CecileRobertMichon: I think that was meant for line 1035

pkg/api/types.go
```diff
@@ -1268,8 +1268,23 @@ func (k *KubernetesConfig) IsIPMasqAgentEnabled() bool {
...
 
-// IsNSeriesSKU returns whether or not the agent pool has Standard_N SKU VMs
-func IsNSeriesSKU(p *Properties) bool {
+// IsNSeriesSKU returns true if the agent pool contains an N-series (NVIDIA GPU) VM
+func (a *AgentPoolProfile) IsNSeriesSKU() bool {
+	return common.IsNvidiaEnabledSKU(a.VMSize)
+}
+
+// HasNSeriesSKU returns whether or not there is an N series SKU agent pool
+func HasNSeriesSKU(p *Properties) bool {
```
CecileRobertMichon: is this function a duplicate of the below one?

pkg/api/defaults.go
```diff
@@ -438,19 +438,29 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
+							profile.Distro = AKSDockerEngine
+						} else {
+							profile.Distro = AKS
+						}
 					}
+				} else if !p.OrchestratorProfile.IsOpenShift() {
+					profile.Distro = Ubuntu
+				}
+			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) && profile.IsNSeriesSKU() {
+				if profile.Distro != Ubuntu {
```
CecileRobertMichon: I have a slight preference for `if profile.Distro == AKS` for now in case we add another distro in the future that we don't want to override in upgrade/scale

pkg/api/defaults.go
```diff
@@ -438,19 +438,29 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
+			if profile.Distro == "" {
+				if p.OrchestratorProfile.IsKubernetes() {
+					if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
+						profile.Distro = Ubuntu
 					} else {
-						profile.Distro = AKS
+						// Ensure distro is set properly for N Series SKUs, because
+						// (1) At present, "aks-docker-engine" and "ubuntu" are the only working distro base for running GPU workloads on N Series SKUs
+						// (2) Previous versions of acs-engine had working implementations using the "aks" distro value,
+						//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
```
CecileRobertMichon: this comment doesn't really fit on this line anymore since the upgrade/scale case is below

test/e2e/engine/template.go
```diff
@@ -225,7 +225,7 @@ func (e *Engine) HasWindowsAgents() bool {
 // HasGPUNodes will return true if the VM SKU is GPU-enabled
 func (e *Engine) HasGPUNodes() bool {
 	for _, ap := range e.ExpandedDefinition.Properties.AgentPoolProfiles {
-		if strings.Contains(ap.VMSize, "Standard_N") {
+		if ap.IsNSeriesSKU() {
```
CecileRobertMichon: instead of having a for loop here you could reuse the `HasNSeriesSKU` function

pkg/api/types.go
jackfrancis: it's dead code, will remove

PR #4222, opened 2018-11-09
Add no outbound internet feature flag

### LINKED ISSUE #4216: Enable k8s deployments with no outbound internet
This issue is to track the progress of deployments with no outbound (except allowing outbound to cluster vnet). 

Sub items:

- [ ] Pre-download all dependencies in the OS image to avoid downloads during deployment

- [ ] Pre-pull all container addon images and do not attempt to pull when pods start if already present 

- [ ] Use internal load balancer so nodes can talk to the apiserver (right now all `kubectl` commands running on nodes timeout because they try to reach a public IP) - also the reason CSE fails with exit code 30

- [ ] Make sure all addons start and go to running state, each with an assigned IP in the vnet

- [ ] Make outbound connection check optional

CecileRobertMichon:
**What this PR does / why we need it**: Add a feature flag for no outbound internet for testing.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: #4216 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


pkg/api/const.go
```diff
@@ -240,7 +240,7 @@ const (
 	// AzureCniPluginVerLinux specifies version of Azure CNI plugin, which has been mirrored from
 	// https://github.com/Azure/azure-container-networking/releases/download/${AZURE_PLUGIN_VER}/azure-vnet-cni-linux-amd64-${AZURE_PLUGIN_VER}.tgz
 	// to https://acs-mirror.azureedge.net/cni
-	AzureCniPluginVerLinux = "v1.0.13"
+	AzureCniPluginVerLinux = "v1.0.12"
```
CecileRobertMichon: TODO: revert this once the newest VHD is published

PR #4223, opened 2018-11-09
Update to Docker 18.09 for Windows

PatrickLang:
[Docker EE 18.09](https://docs.docker.com/engine/release-notes/) is out, which has a newer API version. It's the version recommended by Docker Inc to work with Kubernetes 1.12+ on Windows. This PR sets it as the default version so we can get more test coverage on it.

Previously, we needed some workarounds to make k8s 1.12+ work (see #4118). That PR was designed so that when a new version came out, the workarounds would no longer be applied. That code will be kept in place so we can still test and use older versions if needed.

The same issue is also tracked in https://github.com/kubernetes/kubernetes/issues/69996

**Release note**:
```release-note
Use [Docker EE basic version 18.09.0](https://docs.docker.com/engine/release-notes/) by default for Windows nodes.
```

### DIFF
--- a/examples/windows/kubernetes-windows-docker-version.json
+++ b/examples/windows/kubernetes-windows-docker-version.json
@@ -21,7 +21,7 @@
     "windowsProfile": {
       "adminUsername": "azureuser",
       "adminPassword": "replacepassword1234$",
-      "windowsDockerVersion": "17.06.2-ee-16"
+      "windowsDockerVersion": "18.09.0"
     },
     "linuxProfile": {
       "adminUsername": "azureuser",
--- a/parts/windowsparams.t
+++ b/parts/windowsparams.t
@@ -78,7 +78,7 @@
       "type": "string"
     },
     "windowsDockerVersion": {
-      "defaultValue": "17.06.2-ee-16",
+      "defaultValue": "18.09.0",
       "metadata": {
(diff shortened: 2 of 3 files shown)


jsturtevant:
Does the api version need to be forced like in previous versions  (see #4141)?  Otherwise looks good to me

PatrickLang:
No API version force should be needed. K8s 1.12/1.13 are requesting API version 1.38. Per the release notes "Update API version to 1.39 moby/moby#37640"

PatrickLang:
In fact, the code in #4141 should delete the API override key. That was intentional so that if Kubernetes moves to API version 1.39, it will still work

PatrickLang:
Doing a lot of relaunches. Infra seems flaky right now - I had 3 tests fail to create VMs, and 3 others where DNS was broken on Linux. None of that seems related to this change so I'm rescheduling them in CircleCI

jackfrancis:
@PatrickLang see this code:

https://github.com/Azure/acs-engine/blob/master/pkg/api/types.go#L1094

We've been discussing methods of continuing to "bless forward" newer versions of various configurations over the lifecycle of long-lived clusters via upgrade. As that code is written, the next time this value is updated, an upgrade won't get it (because the value is only `""` on create).

Is that okay?

jsturtevant:
What's the proper way to get the value on upgrades?

PR #4224, opened 2018-11-09
ensure addon image is overwritten during upgrade

jackfrancis:
**What this PR does / why we need it**: Because we are currently persisting cluster state via the generated api model; AND we are responsible for curating a k8s version-specific, known-working set of component versions (implemented as addons); we need to be able to move forward (i.e., change) the key addons configuration as clusters are upgraded. The image reference is a key configuration, and we are not currently doing that, instead only assigning the known-working value when the `Image` key is an empty string, which will never be true in the upgrade case.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
ensure addon image is overwritten during upgrade
```


docs/clusterdefinition.md
```diff
@@ -180,7 +180,7 @@ Above you see custom configuration for both tiller and kubernetes-dashboard. Bot
 
 See https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/ for more on Kubernetes resource limits.
 
-Additionally above, we specified a custom docker image for tiller, let's say we want to build a cluster and test an alpha version of tiller in it.
+Additionally above, we specified a custom docker image for tiller, let's say we want to build a cluster and test an alpha version of tiller in it. Important note! customizing the image is not sticky across upgrade/scale, to ensure that acs-engine always delivers a version-curated, known-working addon when moving a cluster to a new version. Considering all that, providing a custom image reference for an addon configuration should be considered for testing/development, but not for a production cluster.
```
CecileRobertMichon: let's reference the "data" alternative here

(commenting so we don't forget)

jackfrancis: added

PR #4226, opened 2018-11-10
Adding containermonitoring to the list of addons that need tested

tariq1890:
This will be useful when oms agent team adds new versions of the oms agent


tariq1890:
So we don't have any means of validating new images of omsagent addons then?

PR #4240, opened 2018-11-13
Adding c:\tmp as needed to pass Kubernetes tests

PatrickLang:

These two test cases will fail on Windows due to a missing path. Since people may want to run conformance or other tests on a cluster, we should create the path to avoid failure.

- Kubernetes e2e suite.[sig-storage] HostPath should support r/w [NodeConformance]
- Kubernetes e2e suite.[sig-storage] HostPath should support subPath [NodeConformance]

```
Error: Error response from daemon: invalid volume specification: 'c:/tmp:c:/test-volume': invalid mount config for type "bind": bind source path does not exist: c:\tmp
```

cc @jsturtevant @adelina-t


adelina-t:
Yeah, this is useful. I usually add this as part of a preprovision extension, but if it's in acs-engine, that's even better.

PR #4248, opened 2018-11-14
Using the conventional azure environment variable names

tariq1890:
Making sure the azure environment variables are name consistently across acs-engine

### DIFF
--- a/scripts/azure-const.sh
+++ b/scripts/azure-const.sh
@@ -1,31 +1,31 @@
 #!/bin/bash
-if [ -z "$CLIENT_ID" ]; then
+if [ -z "$AZURE_CLIENT_ID" ]; then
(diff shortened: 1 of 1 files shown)


scripts/azure-const.sh
```diff
@@ -1,31 +1,31 @@
 #!/bin/bash
-if [ -z "$CLIENT_ID" ]; then
+if [ -z "$AZURE_CLIENT_ID" ]; then
```
CecileRobertMichon: what's this script used for? is it going to break any jobs that rely on it to change the env vars?

tariq1890: This script is used to generate `azureconst.go`, it won't break anything I am just doing this so that we remain consistent across acs-engine. Packer uses env vars prefixed with `AZURE_` for instance

PR #4262, opened 2018-11-19
Add DockerEngine feature flag

### LINKED ISSUE #4261: aks-docker-engine distro with v0.25.2 deployments fails on Azure US Gov

No

Issue

0.25.2

1.10.18

**What happened**:

Deploying the following template fails on Azure US Gov

```
{
  "apiVersion": "vlabs", 
  "location": "usgovvirginia", 
  "properties": {
    "agentPoolProfiles": [
      {
        "OSDiskSizeGB": 0, 
        "availabilityProfile": "VirtualMachineScaleSets", 
        "count": 1, 
        "distro": "aks-docker-engine", 
        "ipAddressCount": 128, 
        "kubernetesConfig": {
          "controllerManagerConfig": {
            "--terminated-pod-gc-threshold": "50"
          }, 
          "kubeletConfig": {
            "--image-pull-progress-deadline": "30m", 
            "--max-pods": "110"
          }
        }, 
        "name": "agentpool1", 
        "storageProfile": "ManagedDisks", 
        "vmSize": "Standard_DS13_v2"
      }
    ], 
    "linuxProfile": {
      "adminUsername": "user", 
      "ssh": {
        "publicKeys": [
          {
            "keyData": "********"
          }
        ]
      }
    }, 
    "masterProfile": {
      "OSDiskSizeGB": 0, 
      "count": 1, 
      "distro": "aks-docker-engine", 
      "dnsPrefix": "test1", 
      "vmSize": "Standard_D8_v3"
    }, 
    "orchestratorProfile": {
      "kubernetesConfig": {
        "addons": [
          {
            "enabled": false, 
            "name": "blobfuse-flexvolume"
          }, 
          {
            "enabled": false, 
            "name": "smb-flexvolume"
          }, 
          {
            "enabled": false, 
            "name": "keyvault-flexvolume"
          }, 
          {
            "enabled": true, 
            "name": "nvidia-device-plugin"
          }
        ], 
        "apiServerConfig": {
          "--enable-admission-plugins": "Priority,Initializers", 
          "--feature-gates": "PodPriority=true", 
          "--runtime-config": "admissionregistration.k8s.io/v1alpha1,scheduling.k8s.io/v1alpha1=true"
        }, 
        "etcdDiskSizeGB": "2048", 
        "kubeletConfig": {
          "--feature-gates": "PodPriority=true"
        }, 
        "schedulerConfig": {
          "--feature-gates": "PodPriority=true"
        }
      }, 
      "orchestratorVersion": "1.10.8"
    }, 
    "servicePrincipalProfile": {
      "clientId": "**************************", 
      "secret": "********"
    }
  }
}
```

We get the following error:

```
ERROR: Deployment failed. Correlation ID: fd54cb6b-7551-47e4-9be8-dafcc177273c. {
  "error": {
    "code": "InvalidParameter",
    "message": "The value of parameter imageReference.publisher is invalid.",
    "target": "imageReference.publisher"
  }
}
```

Deployment to succeed. 

Use the the above template to deploy to Azure US Gov cloud.

CecileRobertMichon:
**What this PR does / why we need it**: Deployments w/ GPU nodes in sovereign clouds need to install docker-engine without VHD as Moby is not compatible w/ GPU and AKS VHD is not yet available in sovereign clouds.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #4261 

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


pkg/api/types.go
```diff
@@ -1363,6 +1364,23 @@ func (k *KubernetesConfig) GetAzureCNIURLWindows(cloudSpecConfig AzureEnvironmen
 	return cloudSpecConfig.KubernetesSpecConfig.VnetCNIWindowsPluginsDownloadURL
 }
 
+// IsFeatureEnabled returns true if a feature flag is on for the provided feature
+func (f *FeatureFlags) IsFeatureEnabled(feature string) bool {
+	if f != nil {
```
jackfrancis: @tariq1890 thoughts on nil checking a pointer self instance reference?

pkg/api/vlabs/types.go
```diff
@@ -58,6 +58,7 @@ type AzProfile struct {
 type FeatureFlags struct {
 	EnableCSERunInBackground bool `json:"enableCSERunInBackground,omitempty"`
 	BlockOutboundInternet    bool `json:"blockOutboundInternet,omitempty"`
+	DockerEngine             bool `json:"dockerEngine,omitempty"`
```
tariq1890: Requesting a more descriptive FeatureFlag name here

pkg/api/types.go
tariq1890: I prefer nil check before the method call.

Also, if we ensure that a FeatureFlag struct is always initialized while instantiating the properties struct, then we can do away with the nil check.

jackfrancis: Me too. I'm not convinced we all think this is an improvement in quality.

pkg/api/vlabs/types.go
jackfrancis: Preference for just `DockerEngine`. The other two have verbs because they change the way various workflows operate. Whereas the verb for `DockerEngine` is just something superfluous like "use docker-engine".

pkg/api/types.go
tariq1890: @CecileRobertMichon Correct me if I am wrong, but nil check within the function ensures that the field of a nil pointer is never accessed  but the method call still happens and in this case, the first if condition fails and goes straight to false.

I am suggesting that we do the nil check outside of that method  and before the method is invoked. This pattern would stay consistent when we try to safely access both sub-fields and methods from a struct field.

CecileRobertMichon: > nil check within the function ensures that the field of a nil pointer is never accessed but the method call still happens and in this case, the first if condition fails and goes straight to false.

This is exactly right, I've updated my previous comment to be more clear.

My argument was that checking for nil outside the function means 1) duplicated nil checks if we call the same function multiple times and 2) prone to bugs if that function is ever called without a nil check 3) we can't unit test that the right behavior happens when the struct is nil (ie. that we return false)

CecileRobertMichon: This restructuring of the code by taking out the function of template_generator and unit testing it was a prototype, I was thinking of changing the other functions to follow the same pattern in a later PR so let's get it right and exclude  consistency from reasoning.

jackfrancis: I'm happy (1) not setting the bar as high as getting it right for all time and (2) not being 100% consistent.

/lgtm

PR #4271, opened 2018-11-21
Do not check agent & master pool size when using availability zones

sylr:

You might want to start little and grow over time. You should not be
forced to use at least 2 times the amount of availability zones.

```release-note
Remove master & agent pools' size checks for availability zones
```

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -490,17 +490,10 @@ func (a *Properties) validateZones() error {
 		// all zones or no zones should be defined for the cluster
 		if a.HasAvailabilityZones() {
 			if a.MastersAndAgentsUseAvailabilityZones() {
-				// master profile
-				if a.MasterProfile.Count < len(a.MasterProfile.AvailabilityZones)*2 {
-					return errors.New("the node count and the number of availability zones provided can result in zone imbalance. To achieve zone balance, each zone should have at least 2 nodes or more")
-				}
 				// agent pool profiles
 				for _, agentPoolProfile := range a.AgentPoolProfiles {
 					if agentPoolProfile.AvailabilityProfile == AvailabilitySet {
 						return errors.New("Availability Zones are not supported with an AvailabilitySet. Please either remove availabilityProfile or set availabilityProfile to VirtualMachineScaleSets")
-					}
-					if agentPoolProfile.Count < len(agentPoolProfile.AvailabilityZones)*2 {
-						return errors.New("the node count and the number of availability zones provided can result in zone imbalance. To achieve zone balance, each zone should have at least 2 nodes or more")
 					}
 				}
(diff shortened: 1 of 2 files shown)


jackfrancis:
@ritazh for comment, who did the original implementation

ritazh:
Hi @sylr, the Availability Zones feature was added to ensure high availability of user's applications or clusters in this case. To fully take advantage of AZ in a cluster and to ensure a balanced scale set, we decided to distribute n nodes across at least n/2 AZ zones. This is especially important as you scale out the cluster over n/2 zones as zones cannot be added post cluster creation. For example, if you have a 4-node agent pool with 1 zone, you are not really taking advantage of AZ.

Ref: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones

We are definitely open to suggestions to make this both more resilient and easy to use. Perhaps to make it less burdensome to start using this feature, instead of requiring `num of nodes >= num of zones *2`, we validate for `zones > 1`. WDYT?

cc @khenidak

sylr:
Hi @ritazh I think the use of Availability Zones should not be conditioned by the number of nodes at deployment time since the main purpose of using a VMSS is to be able to increase or decrease that number over time.

ritazh:
@sylr For AZ support, we wanted to make sure that the minimum number of nodes per zones is at least 2 as you have seen if a pod is bound to a PVC in a zone and one of the nodes in that zone goes down, we need to ensure at least we have another node to reschedule the pod with the PVC in the same zone.

sylr:
@ritazh even if you have another node in the same zone you can't be sure that it will have enough resources for your pod to be scheduled.

ritazh:
@sylr Thanks for your contribution! Looks like few unit tests need to be updated as well since the logic in validate.go has been updated. https://github.com/Azure/acs-engine/blob/master/pkg/api/vlabs/validate_test.go You can also run `make coverage` to see which tests are failing and need to be updated.

ritazh:
@sylr can you pls rebase and resolve the conflicts? Thanks!

tariq1890:
LGTM. @sylr Appreciate the contribution :). Could we have this PR from you in aks-engine. We are holding off on PRs in this repo at the moment

sylr:
@tariq1890 I believe it's not really clear how the members of the project want to do the transition from acs-engine to aks-engine so I think it's better to not commit the same code in the two repo and wait for a cutover and then merge.

tariq1890:
@sylr Good idea! Let's wait to hear from the members :)

jackfrancis:
@sylr this needs to land in aks-engine: it will receive all forward-looking additions. That's annoying, I know. We're happy to do the gruntwork for folks (i.e., manually move acs-engine PRs as appropriate to aks-engine) when we have some cycles (i.e., after the transition stabilizes). Or, feel free to move it over yourself sooner!

Again, thanks for your patience/understanding!

CecileRobertMichon:
see #206 in aks-engine

Issue #4277, opened 2018-11-22
Update kubletwin/pause to pipe ping output to null


zentron:
The "built-in" pause container that is created generates a _lot_ of log entries since it pings once per second. 
https://github.com/Azure/acs-engine/blob/400dfbb476b6eedbda7c763df57ff7ab6d9347b2/parts/k8s/windowskubeletfunc.ps1#L158

This causes headaches with any attempt to perform container logging using something like FileBeat since it (and the kubernetes recommended approach) involves tailing the logs generated on disk.  When there are many pods running this has an impact on disk IO (writing logs), CPU (logging tools needing to parse) and networking (pinging & logging tools sending messages around)

The default linux pause container simply [sits there and does nothing](https://github.com/kubernetes/kubernetes/blob/master/build/pause/pause.c#L65) and the alternative windows pause container `apprenda/pause` also [does nothing](https://github.com/apprenda/kubernetes-pause/blob/master/pause.ps1)

At a minimum this involves just changing the command to pipe to nul
> "CMD cmd /c ping -t localhost  > nul" | Out-File -encoding ascii -FilePath Dockerfile -Append

Or ideally change to also use sleep since i don't think a constant ping provides any benefit (and is not how any other pause containers work).

PR #4278, opened 2018-11-22
Update pause container to pipe ping result to nul

zentron:
As per https://github.com/Azure/acs-engine/issues/4277 pinging creates a few necessary issues down the line. This change is in line with how other pause containers run

### DIFF
--- a/parts/k8s/windowskubeletfunc.ps1
+++ b/parts/k8s/windowskubeletfunc.ps1
@@ -155,7 +155,7 @@ New-InfraContainer
     }
 
     "FROM $($windowsBase)" | Out-File -encoding ascii -FilePath Dockerfile
-    "CMD cmd /c ping -t localhost" | Out-File -encoding ascii -FilePath Dockerfile -Append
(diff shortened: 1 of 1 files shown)


tariq1890:
Thanks for the PR @zentron . I've requested Patrick to review this.

PatrickLang:
Can you elaborate a bit on what problem it causes? I’m wondering if we should change it to a loop that does nothing instead of ping.

Tests scheduled :)

zentron:
Noted on the original issue. We have over 50 pods on a node and each then gets a pause container. The problem comes when we try and store the log output of these containers. While additional 50 entries per second isn't the end of the world it does add unnecessary uverhead. 

Yep if it was a proper pause like the Linux one or sleep like the alternative windows 'apprenda/pause' one then it would be ideal.

CecileRobertMichon:
@zentron thank you for your contribution! We've relocated to https://github.com/Azure/aks-engine, if this PR is still relevant please reopen it in the new repo.

zentron:
Thanks fine, we have decided to move away from ACS for this (and several other reasons). Given that the PR was ready to merge a few months back it would be nice if it could still just be added to aks-engine rather than just closed & ignored

PR #4286, opened 2018-11-22
Make AvailabilitySet profile for master use Availability Zones

sylr:
Currently Azure `Availability Sets` and `Availability Zones` are exclusive (this might change in the future).

This patch allows to replace the master Availability Set with the use of Availability Zones without creating a new availabilityProfile which would be costly and obsolete if Azure Availability Sets becomes compatible with Availability Zones.


parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -77,7 +77,8 @@
       "type": "Microsoft.Network/networkInterfaces"
     },
 {{if .IsManagedDisks}}
-   {
+    {{if not (HasAvailabilityZones .)}}
```
ritazh: As discussed on slack, please remove AZ support for VMAS agent profile.

parts/k8s/kubernetesmasterresources.t
```diff
@@ -787,6 +802,9 @@
       },
       "location": "[variables('location')]",
       "name": "[concat(variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')))]",
+      {{if .MasterProfile.HasAvailabilityZones}}
+      "zones": "[split(string(parameters('availabilityZones')[mod(copyIndex(variables('masterOffset')), length(parameters('availabilityZones')))]), ',')]",
```
ritazh: How does this handle copyIndex 0, for example:
```
        "AvailabilityZones": [
            "1",
            "2"
        ]
```
and cases where the AvailabilityZones array is not contiguous, which can happen.
```
        "AvailabilityZones": [
            "1",
            "3"
        ]
```

PR #4290, opened 2018-11-26
enable typha and add horizontal autoscaler

briansan:
enables Typha for Calico and also adds a horizontal autoscaler for it

None

None

**Release note**:
NONE


jackfrancis:
@feiskyer FYI

jackfrancis:
Hi @briansan as implemented, this only enables Typha for k8s 1.12 and above. Which versions of k8s should this implementation of Typha work with?

Thanks!

feiskyer:
LGTM.

@caseydavenport I think this would work since v1.7?

briansan:
hi @jackfrancis, what makes you say so? Typha should be supported on all version of Kubernetes

jackfrancis:
@briansan We have version-specific implementations of the calico addon (see the version subdirs in the `parts/k8s/adddons` directory. I'm happy to do the work, we just need precise clarification of which versions of calico support the manifest definition additions as-is in the PR.

jackfrancis:
@briansan I noticed that the `calico-typha` Deployment still has `0` replicas. Is that intentional?

briansan:
@jackfrancis one of the goals of this PR is to set up an autoscaler which handles setting the replica count based on the number of k8s nodes

jackfrancis:
@briansan Got it, thanks!

PR #4294, opened 2018-11-27
feat(perf): Invoke-WebRequest much slower then browser download

sylus:
**What this PR does / why we need it**: Invoke-WebRequest much slower than a browser download when the progress bar is shown by an order of magnitude. Saw this when trying out the windows patches extension and a rather larger MSU.

http://download.windowsupdate.com/c/msdownload/update/software/updt/2018/08/windows10.0-kb4346783-x64_145555faf4e3e494c20f22508f9c28fe12af387a.msu

Before Patch: About an hour
After Patch: Less then a minute

> Note: The msu file was placed in a storage account using the premium service.

https://stackoverflow.com/questions/28682642/powershell-why-is-using-invoke-webrequest-much-slower-than-a-browser-download

**Special notes for your reviewer**: I am not sure if this applies when is running as part of the CustomScriptExtension but did for me when running manually. Feel free to close if is the case :)

**Release note**:

```release-note
Invoke-WebRequest much slower than a browser download 
```

### DIFF
--- a/extensions/windows-patches/v1/installPatches.ps1
+++ b/extensions/windows-patches/v1/installPatches.ps1
@@ -13,6 +13,7 @@ function DownloadFile([string] $URI, [string] $fullName)
 {
     try {
         Write-Host "Downloading $URI"
+        $ProgressPreference = 'SilentlyContinue'
         Invoke-WebRequest -UseBasicParsing $URI -OutFile $fullName
     } catch {
         Write-Error $_


PatrickLang:
Thanks! it looks like turning off the progress bar should help.

I tried other methods before (`start-bitstransfer`, system.net.webclient) but those required a logged-in user and would fail.

One problem though - can you check into why it's showing all lines changed? Did you change line endings or something? I want to make sure this merges just the changed lines cleanly.

sylus:
Here you @PatrickLang all updated without whitespace correction. :D

PR #4324, opened 2018-12-01
Add copyright headers to source files

mboersma:

Adds copyright headers to Go source files and enforces a `make validate-headers` target in CI. 

Better late than never. All credit goes to @CecileRobertMichon. See also Azure/aks-engine#23.

**Release note**:

```release-note
Add copyright headers to source files
```


tariq1890:
can we rename `validate-headers` to `validate-license-headers` ? I think headers is too generic and not necessarily suggestive of what it's actually doing.

tariq1890:
lgtm otherwise :)

PR #4332, opened 2018-12-03
Disable accelerated networking for PB series VMs

tariq1890:
This is needed in tandem with #4265 . Currently, it is known that PB series VMs have no accelerated Networking support.

We are forced to use/edit these static lists as we have no api from Azure to retrieve VM Skus and their properties :(


CecileRobertMichon:
Description? why is this needed/true? is there a relevant doc?

tariq1890:
@CecileRobertMichon I've updated the PR

CecileRobertMichon:
@tariq1890 did this make it in aks-engine? Should we close this one?

tariq1890:
Yes. Closing.

PR #4344, opened 2018-12-05
update Calico network policy examples docs

briansan:
- updates documentation around how to configure Calico policy on the acs-engine deploy


CecileRobertMichon:
@briansan can you please close this PR and reopen it in aks-engine? We are in the process of transitioning. There will be no further acs-engine releases, the next on will be in the aks-engine repo. See https://github.com/Azure/aks-engine/blob/master/docs/faq.md for details.

PR #4364, opened 2019-01-10
chore: add config for "stale" bot service

mboersma:
This should help in cleaning up our large backlog of issues here. We still need to move all relevant issues over to aks-engine by hand.

### DIFF
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -1,0 +1,61 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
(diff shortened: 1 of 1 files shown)


.github/stale.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+# only: issues
+
+# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
+pulls:
+  daysUntilStale: 30
```
CecileRobertMichon: maybe we can be more strict on PR staleness?

mboersma: Sure, since we don't intend to have any actual PRs here now. I'll change it to a week.

